PR #8, opened 2016-09-09
Serializes CensusContextImpl as a Protocol Buffer.

sebright:
This PR adds census-proto as a git submodule and then uses it for the `CensusContext` serialization format.

PR #10, opened 2016-09-16
Makes census-java compatible with Java 6

sebright:
This PR includes #2, because it uses the Maven build files to build with Java 6 and 7 in .travis.yml.

### DIFF
--- a/core/javatests/com/google/census/CensusScopeTest.java
+++ b/core/javatests/com/google/census/CensusScopeTest.java
@@ -41,58 +41,76 @@ public class CensusScopeTest {
 
   @Test
   public void testScope() {
-    try (CensusScope scope = new CensusScope(Census.getCurrent().with(K1, V1))) {
+    CensusScope scope = new CensusScope(Census.getCurrent().with(K1, V1));
+    try {
       assertEquals(
           Census.getDefault().with(K1, V1),
           Census.getCurrent());
+    } finally {
+      scope.close();
     }
     assertEquals(Census.getDefault(), Census.getCurrent());
   }
 
   @Test
   public void testNestedScope() {
-    try (CensusScope s1 = new CensusScope(Census.getCurrent().with(K1, V1))) {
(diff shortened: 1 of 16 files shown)


core/javatests/com/google/census/CensusScopeTest.java
```diff
@@ -41,58 +41,76 @@ public void testNoScope() {
 
   @Test
   public void testScope() {
-    try (CensusScope scope = new CensusScope(Census.getCurrent().with(K1, V1))) {
+    CensusScope scope = new CensusScope(Census.getCurrent().with(K1, V1));
```
adriancole: not sure if bazel does this, but not really required to limit tests to java6 is it?

core_native/java/com/google/census/CensusSerializer.java
```diff
@@ -36,14 +34,23 @@ static ByteBuffer serialize(CensusContextImpl context) {
           .append(TAG_DELIM)
           .append(tag.getValue());
     }
-    return ByteBuffer.wrap(builder.toString().getBytes(UTF_8));
+    try {
+      return ByteBuffer.wrap(builder.toString().getBytes("UTF-8"));
+    } catch (UnsupportedEncodingException e) {
+      throw new AssertionError("UTF-8 not supported.");
```
adriancole: worthwhile making this a static constant.

```
  public static final Charset UTF_8 = Charset.forName("UTF-8");
```

core/javatests/com/google/census/CensusScopeTest.java
sebright: If Census supports Java 6, then I think we should also run the tests with it.  I'd like to add the tests to pom.xml soon.

sebright:
I rebased and updated the Maven build files to work with the Protocol Buffers that were introduced in #8.  #8 also removed the use of `StandardCharsets.UTF_8`.

PR #15, opened 2016-09-21
Removes dependency on guava from census-core.

### LINKED ISSUE #13: Remove Guava and jsr305 dependencies from the API artifact
They are only lightly used and can be easily swapped out. A dependency-free library is more appealing to users. In particular, users are concerned about Guava dependency because they may also be using Guava, and the two Guava versions may conflict.

sebright:
This fixes #13. Another way to document methods that can return null is to add a new Nullable class to census-core.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -15,10 +15,7 @@
 java_library(
     name = "census-core",
     srcs = glob(["core/java/com/google/census/*.java"]),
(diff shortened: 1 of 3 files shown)


sebright:
I split this into two pull requests.  #18 removes jsr305.

PR #18, opened 2016-09-26
Removes dependency on jsr305 from census-core.

sebright:
I split this off from #15 so that we can decide whether to remove the two dependencies independently.


sebright:
I'll close this for now.  See #13.

PR #22, opened 2016-09-27
Adds grpc-based thread-local support

dinooliva:


sebright:
It looks like this branch continues from dinooliva/thread-local, but I think the merge will still work.

examples/java/com/google/census/CensusRunner.java
```diff
@@ -36,11 +39,42 @@
 
   public static void main(String args[]) {
     System.out.println("Hello Census World");
-    System.out.println("Default Tags: " + Census.getCensusContextFactory().getDefault());
```
adriancole: Nit: hard to see the changes due to extracting methods. Presumably biggest change here is catch blocks

grpc/java/com/google/census/CensusGrpcContext.java
```diff
@@ -0,0 +1,42 @@
...
+
+package com.google.census;
+
+import io.grpc.Context;
+
+/**
+ * Utils for embedding {@link CensusContext} in gRPC {@link Context} to support thread-local
+ * implicit propagation of the {@link CensusContext}.
+ */
+public abstract class CensusGrpcContext {
```
adriancole: Did you want this ctor public? Or is it intended as package private?

grpc_native/java/com/google/census/CensusGrpcContextImpl.java
```diff
@@ -0,0 +1,35 @@
...
+
+package com.google.census;
+
+import io.grpc.Context;
+
+/**
+ * Utils for embedding {@link CensusContext} in gRPC {@link Context} to support thread-local
+ * implicit propagation of the {@link CensusContext}.
+ */
+public class CensusGrpcContextImpl extends CensusGrpcContext {
```
adriancole: Intentionally not final?

examples/java/com/google/census/CensusRunner.java
sebright: Do you mean that the PR should have been broken into multiple commits to make the diffs clearer?

The main difference is that the new code uses the gRPC Context to keep track of the Census tags.  We could add back the `CensusScope` class, implemented using the gRPC context, to make this code clearer.

grpc/java/com/google/census/CensusGrpcContext.java
sebright: This class is supposed to act as an interface that can have methods added in Java 6 and 7, so the constructor is intentionally left public.

grpc_native/java/com/google/census/CensusGrpcContextImpl.java
sebright: No.  I fixed it in #27.

PR #27, opened 2016-09-29
Makes class modifiers more consistent.

sebright:


sebright:
This change was included in #54.

PR #38, opened 2016-09-30
Runs Maven Checkstyle plugin to enforce Google style.

sebright:
This PR also fixes the existing style violations.


sebright:
I rebased this onto #40.  #40 introduces a build profile for Java 6, and this PR uses it to avoid running maven-checkstyle-plugin 2.16 with Java 6.  Only version 2.16 supports Google style directly.

PR #48, opened 2016-11-17
Stats api

dinooliva:
Renames MetricName to MeasurementDescriptor and expands associated information.


core/java/com/google/instrumentation/stats/MeasurementDescriptor.java
```diff
@@ -0,0 +1,146 @@
...
+   */
+  public final String description;
+
+  /**
+   * The units in which {@link MeasurementDescriptor} values are measured.
+   */
+  public final MeasurementUnit unit;
+
+  /**
+   * Construtor.
```
a-veitch: typo (need a 'c')

core/java/com/google/instrumentation/stats/MeasurementDescriptor.java
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  /**
+   * Fundamental units of measurement.
+   */
+  public enum BasicUnit {
+    UNKNOWN,
+    BITS,
+    BYTES,
+    SCALAR,
```
a-veitch: Suggest having SCALAR first, as it's the only dimensionless value.

core/java/com/google/instrumentation/stats/MeasurementDescriptor.java
```diff
@@ -0,0 +1,146 @@
...
+  @Override
+  public String toString() {
+    return name;
+  }
+
+  /**
+   * Fundamental units of measurement.
+   */
+  public enum BasicUnit {
+    UNKNOWN,
```
a-veitch: Remove UNKNOWN - this is only in the proto format to distinguish it as uninitialized. In this context (people using this API) it should never be set/used.

core/java/com/google/instrumentation/stats/RpcConstants.java
```diff
@@ -24,34 +30,78 @@
...
+      new MeasurementDescriptor(
+          "/rpc/client/response_bytes",
+          "Response MB/s",
+          new MeasurementUnit(6, bytes, secs));
+  // TODO(dpo): verify this encoding
+  public static final MeasurementDescriptor RPC_CLIENT_ROUNDTRIP_LATENCY =
+      new MeasurementDescriptor(
+          "/rpc/client/roundtrip_latency",
+          "RPC roundtrip latency us",
+          new MeasurementUnit(1, scalar));
```
a-veitch: Shouldn't this be MeasurementUnit(-6, secs)

core/java/com/google/instrumentation/stats/RpcConstants.java
```diff
@@ -24,34 +30,78 @@
...
+  public static final MeasurementDescriptor RPC_SERVER_RESPONSE_BYTES =
+      new MeasurementDescriptor(
+          "/rpc/server/response_bytes",
+          "Response MB/s",
+          new MeasurementUnit(6, bytes, secs));
+  public static final MeasurementDescriptor RPC_SERVER_SERVER_LATENCY =
+      new MeasurementDescriptor(
+          "/rpc/server/server_latency",
+          "Latency in msecs",
+          new MeasurementUnit(3, scalar));
```
a-veitch: This should be MeasurementUnit(-3, secs)

But why is is us for client/roundtrip_latency and ms for server/server_latency? Should use consistent units.

core/java/com/google/instrumentation/stats/MeasurementDescriptor.java
```diff
@@ -0,0 +1,146 @@
...
+
+  /**
+   * Detailed description of the measurement, used in documentation.
+   */
+  public final String description;
+
+  /**
+   * The units in which {@link MeasurementDescriptor} values are measured.
+   */
+  public final MeasurementUnit unit;
```
sebright: These fields should be private with getters, in case we need to change the representation in the future.

core/java/com/google/instrumentation/stats/MeasurementDescriptor.java
```diff
@@ -0,0 +1,146 @@
...
+
+  /**
+   * Construtor.
+   */
+  public MeasurementDescriptor(String name, String description, MeasurementUnit unit) {
```
sebright: A public factory method would give us more flexibility for future changes.

core/java/com/google/instrumentation/stats/MeasurementDescriptor.java
```diff
@@ -0,0 +1,146 @@
...
+   * Fundamental units of measurement.
+   */
+  public enum BasicUnit {
+    UNKNOWN,
+    BITS,
+    BYTES,
+    SCALAR,
+    SECS,
+    CORES,
+    MAX_UNITS;
```
sebright: MAX_UNITS is only needed in the proto format.

core/java/com/google/instrumentation/stats/MeasurementDescriptor.java
```diff
@@ -0,0 +1,146 @@
...
+    this.name = StringUtil.sanitize(name);
+    this.description = description;
+    this.unit = unit;
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    return (obj instanceof MeasurementDescriptor)
+        && name.equals(((MeasurementDescriptor) obj).name);
+  }
```
sebright: Shouldn't `equals()` compare all fields?  We could have a separate method to only compare names.

core/java/com/google/instrumentation/stats/MeasurementDescriptor.java
```diff
@@ -0,0 +1,146 @@
...
+   *     numerator: BYTES
+   *     denominator: SECS
+   *
+   * <p>- nanoseconds
+   *     power10: -9
+   *     numerator: SECS
+   */
+  public static final class MeasurementUnit {
+    /**
+     * Unit multiplier.
```
sebright: I think this comment could be clearer, because `power10` is the base-10 log of the multiplier.

core/java/com/google/instrumentation/stats/RpcConstants.java
```diff
@@ -24,34 +30,78 @@
...
-      new MetricName("/rpc/server/uncompressed_request_bytes");
-  public static final MetricName RPC_SERVER_UNCOMPRESSED_RESPONSE_BYTES =
-      new MetricName("/rpc/server/uncompressed_response_bytes");
-
-  //Visible for testing
+  public static final MeasurementDescriptor RPC_CLIENT_ERROR_COUNT =
+      new MeasurementDescriptor(
+          "/rpc/client/error_count",
+          "RPC Errors",
+          new MeasurementUnit(1, scalar));
```
sebright: Shouldn't this be `new MeasurementUnit(0, scalar)`?

core/java/com/google/instrumentation/stats/RpcConstants.java
```diff
@@ -24,34 +30,78 @@
...
+      new MeasurementDescriptor(
+          "/rpc/client/uncompressed_response_bytes",
+          "Uncompressed Request MB/s",
+          new MeasurementUnit(6, bytes, secs));
+
+  public static final MeasurementDescriptor RPC_SERVER_ERROR_COUNT =
+      new MeasurementDescriptor(
+          "/rpc/server/error_count",
+          "RPC Errors",
+          new MeasurementUnit(1, scalar));
```
sebright: I think this should be 0, too.

core/java/com/google/instrumentation/stats/MeasurementDescriptor.java
```diff
@@ -0,0 +1,146 @@
...
+   * <p>- MB/s (i.e. megabytes / s):
+   *     power10: 6
+   *     numerator: BYTES
+   *     denominator: SECS
+   *
+   * <p>- nanoseconds
+   *     power10: -9
+   *     numerator: SECS
+   */
+  public static final class MeasurementUnit {
```
sebright: This class should also have getters and factory methods.

core/java/com/google/instrumentation/stats/RpcConstants.java
dinooliva: Agreed - I'm following the specifications in resource_definitions.txt but I'll verify the units with Kun.

core/java/com/google/instrumentation/stats/RpcConstants.java
dinooliva: Yes, thanks for catching this, I misread the comments on  units.

core/java/com/google/instrumentation/stats/MeasurementDescriptor.java
dinooliva: Done

core/java/com/google/instrumentation/stats/MeasurementDescriptor.java
dinooliva: I initially considered defining getters as you suggest but this class is specifically to define an immutable value - all the fields are immutable and contain immutable values - in this case, the utility of getters is questionable.

I'm open to changing to the getters approach but I'd like a compelling reason for doing so.

core/java/com/google/instrumentation/stats/MeasurementDescriptor.java
dinooliva: My understanding of the semantics of a MeasurementDescriptor is that the name should be the only discriminator (i.e. it is an programming error to define two descriptors with the same name).

core/java/com/google/instrumentation/stats/MeasurementDescriptor.java
dinooliva: Same point as previous.

I don't see the utility of factory methods.

core/java/com/google/instrumentation/stats/MeasurementDescriptor.java
dinooliva: It's actually described in the class-level comment but it's probably worth restating here too.

core/java/com/google/instrumentation/stats/RpcConstants.java
dinooliva: Agreed.

core/java/com/google/instrumentation/stats/RpcConstants.java
dinooliva: Agreed.

core/java/com/google/instrumentation/stats/CensusContext.java
```diff
@@ -41,12 +41,12 @@ public final CensusContext with(
   }
 
   /**
-   * Records the given metrics against this {@link CensusContext}.
+   * Records the given measurements against this {@link CensusContext}.
```
adriancole: I like this name much better

core/java/com/google/instrumentation/stats/MeasurementDescriptor.java
adriancole: personally, I prefer the style here (except not with an exposed public ctor). It is simpler and still allows new fields. Representations can change without changing the canonical struct (Ex we do the same thing in zipkin)

core/java/com/google/instrumentation/stats/MeasurementDescriptor.java
```diff
@@ -0,0 +1,148 @@
...
+     *
+     * <p>Note: This list is unmodifiable, attempts to update it will throw an
+     * UnsupportedOperationException.
+     */
+    public final List<BasicUnit> denominators;
+
+    /**
+     * Basic constructor that takes all fields.
+     */
+    public MeasurementUnit(int power10, List<BasicUnit> numerators, List<BasicUnit> denominators) {
```
adriancole: constructors like these are fine to start with, but they become laborious when you add fields later. factory and/or builder, or even auto-value are preferred imho

core/java/com/google/instrumentation/stats/MeasurementDescriptor.java
dinooliva: After some thought, I've decided that it's probably better to go with standard practices for the open source community (for both getters and factory methods).

core/java/com/google/instrumentation/stats/MeasurementDescriptor.java
dinooliva: Changed, as per my previous comment.

core/java/com/google/instrumentation/stats/MeasurementDescriptor.java
dinooliva: I've gone with the factory method approach.

core/java/com/google/instrumentation/stats/CensusContext.java
dinooliva: Yes, hopefully this will help make things clear.

PR #52, opened 2016-11-22
Adds Aggregation and View descriptor classes.

dinooliva:


core/java/com/google/instrumentation/common/MatchFunction.java
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.instrumentation.common;
+
+/**
+ * Encodes functions {@code func: A -> B} for use in encoding disjoint-unions.
```
adriancole: food for thought. this description will appeal to those who tend towards math and might repel others. An example in terms of this project (measurement aggergations) could bring the latter in. Otherwise, you could make this an inner type of AggregationDescriptor which makes the context a little more understood.

core/java/com/google/instrumentation/common/MatchFunction.java
```diff
@@ -0,0 +1,21 @@
...
+ */
+public interface MatchFunction<A, B> {
```
adriancole: sometimes I match the sig of j.u.Function with a comment saying we are defining a custom type due to language level

core/java/com/google/instrumentation/stats/AggregationDescriptor.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    /**
+     * The (possibly null) histogram bucket boundaries for a distribution.
+     *
+     * <p>Note: The returned list is unmodifiable, attempts to update it will throw an
+     * UnsupportedOperationException.
+     */
+    @Nullable
+    public List<Double> getBucketBoundaries() {
```
sebright: When would this be null?

core/java/com/google/instrumentation/stats/AggregationDescriptor.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    /**
+     * The (possibly null) time intervals to record for Intervals.
+     *
+     * <p>Note: The returned list is unmodifiable, attempts to update it will throw an
+     * UnsupportedOperationException.
+     */
+    @Nullable
+    public List<Duration> getIntervalSizes() {
```
sebright: Same as above

core/java/com/google/instrumentation/common/MatchFunction.java
sebright: Another option is to use visitor pattern terminology and have an `AggregationDescriptorVisitor` interface with two methods inside `AggregationDescriptor`.  The downside is that users wouldn't be able to take advantage of lambdas with Java 8.

core/javatests/com/google/instrumentation/stats/AggregationDescriptorTest.java
```diff
@@ -0,0 +1,121 @@
...
+    AggregationDescriptor descriptor =
+        DistributionAggregationDescriptor.create(Arrays.asList(buckets));
+    assertThat(descriptor.match(
+        new MatchFunction<DistributionAggregationDescriptor, Boolean> () {
+          @Override public Boolean func(DistributionAggregationDescriptor dDescriptor) {
+            if (dDescriptor.getBucketBoundaries().size() != buckets.length) {
+              return false;
+            }
+            for (int i = 0; i < buckets.length; i++) {
+              if (!dDescriptor.getBucketBoundaries().get(i).equals(buckets[i])) {
```
sebright: Should this test also use `isWithin()`?

core/javatests/com/google/instrumentation/stats/AggregationDescriptorTest.java
```diff
@@ -0,0 +1,121 @@
...
+         IntervalAggregationDescriptor.create(Arrays.asList(intervals));
+    assertThat(iDescriptor.getIntervalSizes()).isNotNull();
+    assertThat(iDescriptor.getIntervalSizes().size()).isEqualTo(intervals.length);
+    for (int i = 0; i < intervals.length; i++) {
+      assertThat(iDescriptor.getIntervalSizes().get(i)).isEqualTo(intervals[i]);
+    }
+  }
+
+  @Test
+  public void testDistributionAggregationDescriptorMatch() {
```
sebright: It isn't very clear what this is testing.  I think that adding a comment or using simpler `func()` methods would help.

core/java/com/google/instrumentation/common/MatchFunction.java
dinooliva: I've updated the description slightly and added a link to AggregationDescriptor and will consider how to show a simple example.

I've put this class in common because several of our datastructures will need to encode tagged unions (aka sum types, discriminated unions ...) in Java.

I've considered a number of approaches (matching functions, the visitor pattern, istanceof, and so on) and I find this to be the most clear and type safe - I am open to alternatives if there are compelling arguments.

core/java/com/google/instrumentation/common/MatchFunction.java
dinooliva: That's a really good point. I've update the class name and method to match j.u.Function and added a pointer to it.

core/java/com/google/instrumentation/stats/AggregationDescriptor.java
dinooliva: The histogram bucket boundaries are optional, which is described in the class description and  factory methods. I've updated the comment to make it more clear here.

core/javatests/com/google/instrumentation/stats/AggregationDescriptorTest.java
dinooliva: Done.

core/javatests/com/google/instrumentation/stats/AggregationDescriptorTest.java
dinooliva: No longer an issue.

core/javatests/com/google/instrumentation/stats/AggregationDescriptorTest.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.Arrays;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/**
+ * Tests for {@link AggregationDescriptor}
+ */
+@RunWith(JUnit4.class)
+public final class AggregationDescriptorTest {
```
adriancole: still think test classes are a great opportunity to show how the types are used (as opposed to just showing 3 things in = 3 things out). Sometimes I add a TestData class or similar which has some realistic data in it, even if only 5 measurements. This is sometimes punted to a separate LearningTest class, but yeah one way or another would be nice.

dinooliva: For these tests, I just want to verify that the data is wired up as expected.

Once the descriptors and associated data classes are checked in, I plan to add some examples that I hope will demonstrate how everything works together.

adriancole:
thanks!

PR #55, opened 2016-11-29
Remove context

dinooliva:
Removes CensusGrpcContext, which is not used and will be replaced by a more generic mechanism.


adriancole:
cool, but favor to ask :) please raise or describe the replacement by tomorrow if possible? I've a long flight (Yow! melbourne) and it would be cool to toy with whatever that is.

dinooliva:
Unfortunately, it is still a work in progress but I figured it was worth removing this code before anyone depended upon it.

adriancole:
fair enough

PR #67, opened 2016-12-07
Adds factory methods to TagKey and TagValue

dinooliva:


core/java/com/google/instrumentation/stats/RpcConstants.java
```diff
@@ -58,8 +60,13 @@
   public static final MeasurementDescriptor RPC_CLIENT_ROUNDTRIP_LATENCY =
       MeasurementDescriptor.create(
           "/rpc/client/roundtrip_latency",
-          "RPC roundtrip latency us",
-          MeasurementUnit.create(-6, scalar));
+          "RPC roundtrip latency msecs",
+          MeasurementUnit.create(-3, scalar));
```
sebright: Why is this in milliseconds?

dinooliva: That shouldn't be part of this change - I've reverted it for now.

PR #69, opened 2016-12-07
Changes serialize() to take an OutputStream as a parameter.

dinooliva:
If we decide to go with this change, there will be a dual change to deserialize to take an InputStream rather than a ByteBuffer.

### DIFF
--- a/core/java/com/google/instrumentation/stats/StatsContext.java
+++ b/core/java/com/google/instrumentation/stats/StatsContext.java
@@ -13,6 +13,7 @@
 
 package com.google.instrumentation.stats;
 
+import java.io.IOException;
 import java.io.OutputStream;
 
 /**
@@ -55,9 +56,8 @@ public abstract class StatsContext {
    * the {@link StatsContext} protobuf representation.
    *
    * @param output the {@link OutputStream} to add the serialized form of this {@link StatsContext}.
-   * @return the given {@link OutputStream}
    */
-  public abstract OutputStream serialize(OutputStream output);
+  public abstract void serialize(OutputStream output) throws IOException;
 
   /**
    * Builder for {@link StatsContext}.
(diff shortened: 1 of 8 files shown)


core/java/com/google/instrumentation/stats/StatsContext.java
```diff
@@ -51,12 +51,13 @@ public final StatsContext with(
...
-   * <p>The inverse of {@link StatsContextFactory#deserialize(ByteBuffer)} and should be based on
+   * <p>The inverse of {@link StatsContextFactory#deserialize(OutputStream)} and should be based on
    * the {@link StatsContext} protobuf representation.
    *
-   * @return serialized bytes.
+   * @param output the {@link OutputStream} to add the serialized form of this {@link StatsContext}.
+   * @return the given {@link OutputStream}
    */
-  public abstract ByteBuffer serialize();
+  public abstract OutputStream serialize(OutputStream output);
```
sebright: Shouldn't this be void?

core_impl/java/com/google/instrumentation/stats/StatsSerializer.java
```diff
@@ -40,17 +42,22 @@ static ByteBuffer serialize(StatsContextImpl context) {
           .append(TAG_DELIM)
           .append(tag.getValue());
     }
-    return ByteBuffer.wrap(StatsContextProto.StatsContext
-        .newBuilder().setTags(builder.toString()).build().toByteArray());
+    try {
+      output.write(StatsContextProto.StatsContext
+          .newBuilder().setTags(builder.toString()).build().toByteArray());
```
sebright: This can call `StatsContextProto.StatsContext.writeTo(OutputStream)` now.

core/java/com/google/instrumentation/stats/StatsContext.java
dinooliva: I have it return the output stream in case someone wants to chain calls together.

I am thinking of adding a 'throws IOException' in case writing to the stream has a problem.

core_impl/java/com/google/instrumentation/stats/StatsSerializer.java
dinooliva: Nice.

dinooliva:
I don't think there's any point to - we don't need them for backward compatibility so we're probably better off getting rid of them entirely.

PR #71, opened 2016-12-07
Checks for unused imports with Checkstyle.

sebright:
This commit runs Checkstyle with two separate configurations, Google's Java style guide and a configuration file within the repository.


sebright:
Thanks!

PR #76, opened 2016-12-21
Adds visitor classes to View and ViewDescriptor.

sebright:


core/java/com/google/instrumentation/stats/View.java
```diff
@@ -32,15 +32,48 @@
...
   /**
+   * Visitor for the View type.
+   */
+  public abstract static class Visitor<T> {
+    public abstract T apply(DistributionView view);
+
+    public abstract T apply(IntervalView view);
+
+    /**
+     * Creates a Visitor with two {@link Function} objects.
```
adriancole: heh does it now? :P

core/java/com/google/instrumentation/stats/View.java
```diff
@@ -32,15 +32,48 @@
...
+     *
+     * <p>Example:
```
adriancole: eventhough the local name says what this does, I'd still spell it out in the Example text.

ex. `<p>For example, the following visitor counts the aggregations in a view:`

core/java/com/google/instrumentation/stats/View.java
```diff
@@ -32,15 +32,48 @@
...
+     *
+     * <pre>{@code
+     *   int numAggregations = myView.match(View.Visitor.create(
+     *       d -> d.getDistributionAggregations().size(),
+     *       i -> i.getIntervalAggregations().size()));
+     *  }</pre>
+     */
+    public static <T> Visitor<T> create(
```
adriancole: move this one level higher and it is a little less awkward. ex `View.createVisitor`

core/java/com/google/instrumentation/stats/ViewDescriptor.java
```diff
@@ -81,6 +79,43 @@ private ViewDescriptor(
   }
 
   /**
+   * Visitor for the ViewDescriptor type.
```
adriancole: is javadoc complaining at you? :P

core/java/com/google/instrumentation/stats/ViewDescriptor.java
```diff
@@ -81,6 +79,43 @@ private ViewDescriptor(
...
+   */
+  public abstract static class Visitor<T> {
+    public abstract T apply(DistributionViewDescriptor descriptor);
+
+    public abstract T apply(IntervalViewDescriptor descriptor);
+
+    /**
+     * Creates a Visitor with two {@link Function} objects.
```
adriancole: same comments. explaining the example is higher value than saying the vistor's params are functions :)

same also for createVisitor

core/java/com/google/instrumentation/stats/View.java
sebright: Oh, you're right. It's all in the signature. Is the new comment more useful?

core/java/com/google/instrumentation/stats/View.java
sebright: Yes, that looks better.

core/java/com/google/instrumentation/stats/View.java
sebright: Good idea.

core/java/com/google/instrumentation/stats/ViewDescriptor.java
sebright: I'm not sure what else to say here, since it's already a design pattern.

core/java/com/google/instrumentation/stats/ViewDescriptor.java
sebright: Done

core/java/com/google/instrumentation/stats/ViewDescriptor.java
adriancole: I suppose I'm wondering if we can just not add boiler-plate javadoc when there's nothing notable to say

sebright: I wanted to make it clear that I applied the visitor pattern to ViewDescriptor.  Do you think that it's already clear enough from the code?

sebright:
Thanks!

core/java/com/google/instrumentation/stats/ViewDescriptor.java
adriancole: yeah I think because it is an inner type of ViewDescriptor it is implicit that it is a visitor for that. Anyway, no biggie just was wondering if there was a requirement to doc all symbols.

PR #83, opened 2017-01-27
Changes StatsManager.getView(...) to take only the name of the ViewDescriptor.

sebright:


sebright:
getView(...) should take a ViewDescriptor.Name instead of a String.

PR #85, opened 2017-01-29
Checks the commit history of pull requests in CI.

sebright:
The script runs on Travis and only checks the commit history when the commit is
from a pull request. It fails the build if the pull request contains merge
commits ~~or does not start from the branch that it is being merged into~~. If there
is an unexpected exception, the script does not fail the build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ install:
     fi
 
 script:
+  - python check-git-history.py
   - case "$TASK" in
       "BUILD")
         case "$TRAVIS_JDK_VERSION" in
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,9 @@ matrix:
 
   - env: TASK=CHECK_GIT_HISTORY
 
+  allow_failures:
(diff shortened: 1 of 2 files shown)


bogdandrutu:
I think this is a great thing to have.

sebright:
Thanks.  I just updated the script to use TRAVIS_PULL_REQUEST to detect pull requests, which is simpler.

bogdandrutu:
Please re-sync and fix the conflicts in .travis.yml.

sebright:
Rebased

bogdandrutu:
Can we print a message that everything was successful?

sebright:
I just added the check to "Allowed failures" so that we can try it out before making it fail the build.

> Can we print a message that everything was successful?

It currently prints "Checked pull request history." when it finishes.  Do you think it should print more details?

bogdandrutu:
I would print:
Checked pull request history: SUCCEED (or FAILED)

sebright:
Thanks.  I updated the message.

PR #86, opened 2017-01-30
Improve error handling in Provider.

zhangkun83:
1. Log impl loading results.
2. Fall back to default only if class not found, which is considered a
normal case.  If the class does exist, it is intended to be loaded
successfuly, thus other exceptions should be treated as error and thrown
to the caller.


zhangkun83:
@dinooliva

PR #87, opened 2017-01-31
Applies minor fixes to RpcConstants.

dinooliva:


core/java/com/google/instrumentation/stats/RpcConstants.java
```diff
@@ -116,12 +116,12 @@
           MeasurementUnit.create(6, bytes));
 
   // Common histogram bucket boundaries for bytes received/sets DistributionViewDescriptors.
-  private static final List<Double> RPC_BYTES_BUCKET_BOUNDARIES = Arrays.asList(
```
a-veitch: The List used here is not immutable, so the ALL_CAPS style should not apply. Either change the name, or use Collections.unmodifiableList. (also other similar uses).

core/java/com/google/instrumentation/stats/RpcConstants.java
```diff
@@ -159,7 +159,7 @@
           Arrays.asList(RPC_CLIENT_METHOD));
   public static final DistributionViewDescriptor RPC_CLIENT_RESPONSE_BYTES_VIEW =
       DistributionViewDescriptor.create(
-          "/rpc/client/response_bytes",
+          "rpc client response_bytes",
```
a-veitch: Have we decided this is the type of naming we want? (we need to have the general view and resource naming discussion at some point).

dinooliva: No, I think this is still open for discussion - I've made the name here consistent with the other names but they may be changed.

core/java/com/google/instrumentation/stats/RpcConstants.java
dinooliva: They are meant to be constants but that's a reasonable point. I'll change them to be UnmodifiableLists in the next pr.

Issue #88, opened 2017-02-02
Inconsistent time units in RpcConstants


zhangkun83:
`RPC_CLIENT_SERVER_ELAPSED_TIME`, `RPC_SERVER_SERVER_ELAPSED_TIME` and `RPC_SERVER_SERVER_LATENCY` are documented to be in `msec`, which is milliseconds.

`RPC_CLIENT_ROUNDTRIP_LATENCY` is documented to be `RPC roundtrip latency us`, which is microseconds. There is no reason for it to be different.

sebright:
There is also an inconsistency in units between the `MeasurementDescriptor` and `ViewDescriptors` for client roundtrip latency:

https://github.com/google/instrumentation-java/blob/7f0ce697c792f6c539c6cfcbd85e253040cb7454/core/java/com/google/instrumentation/stats/RpcConstants.java#L58-L62

https://github.com/google/instrumentation-java/blob/7f0ce697c792f6c539c6cfcbd85e253040cb7454/core/java/com/google/instrumentation/stats/RpcConstants.java#L139-L145

sebright:
This was fixed by #130.

PR #92, opened 2017-02-06
Adds views for interval stats to RpcConstants.

dinooliva:


core/java/com/google/instrumentation/stats/RpcConstants.java
```diff
@@ -218,6 +221,128 @@
...
+  static final Duration MINUTE = Duration.create(60, 0);
+  static final Duration HOUR = Duration.create(60 * 60, 0);
+
+  /**
+   * Census defined rpc client {@link IntervalViewDescriptor}s.
+   */
+  public static final IntervalViewDescriptor RPC_CLIENT_ROUNDTRIP_LATENCY_INTERVAL_VIEW =
+      IntervalViewDescriptor.create(
+          "rpc client roundtrip latency",
+          "Latency in msecs",
```
sebright: The units are inconsistent with RPC_CLIENT_ROUNDTRIP_LATENCY, which uses microseconds.  I don't know which is correct, though.  This is related to #88.

core/java/com/google/instrumentation/stats/RpcConstants.java
```diff
@@ -218,6 +221,128 @@
...
+          RPC_CLIENT_ROUNDTRIP_LATENCY,
+          IntervalAggregationDescriptor.create(Arrays.asList(MINUTE, HOUR)),
+          Arrays.asList(RPC_CLIENT_METHOD));
+
+  public static final IntervalViewDescriptor RPC_CLIENT_REQUEST_BYTES_INTERVAL_VIEW =
+      IntervalViewDescriptor.create(
+          "rpc client request_bytes",
```
sebright: Why do some of the views have underscores in the names?

core/java/com/google/instrumentation/stats/RpcConstants.java
```diff
@@ -218,6 +221,128 @@
...
+          "Uncompressed Request MB",
+          RPC_CLIENT_UNCOMPRESSED_REQUEST_BYTES,
+          IntervalAggregationDescriptor.create(Arrays.asList(MINUTE, HOUR)),
+          Arrays.asList(RPC_CLIENT_METHOD));
+
+  public static final IntervalViewDescriptor RPC_CLIENT_UNCOMPRESSED_RESPONSE_BYTES_INTERVAL_VIEW =
+      IntervalViewDescriptor.create(
+          "rpc client uncompressed_response_bytes",
+          "Uncompressed Response MB",
+          RPC_CLIENT_RESPONSE_BYTES,
```
sebright: Shouldn't this be `RPC_CLIENT_UNCOMPRESSED_RESPONSE_BYTES`?

core/java/com/google/instrumentation/stats/RpcConstants.java
```diff
@@ -218,6 +221,128 @@
...
+      IntervalViewDescriptor.create(
+          "rpc client server_elapsed_time",
+          "Server elapsed time in msecs",
+          RPC_CLIENT_SERVER_ELAPSED_TIME,
+          IntervalAggregationDescriptor.create(Arrays.asList(MINUTE, HOUR)),
+          Arrays.asList(RPC_CLIENT_METHOD));
+
+  /**
+   * Census defined rpc server {@link IntervalViewDescriptor}s.
+   */
```
sebright: My comments in the RPC client section apply here, too, except for the comment about units.

core/java/com/google/instrumentation/stats/RpcConstants.java
```diff
@@ -218,6 +221,128 @@
...
           Arrays.asList(RPC_SERVER_METHOD));
 
+  // Interval Stats
+  static final Duration MINUTE = Duration.create(60, 0);
+  static final Duration HOUR = Duration.create(60 * 60, 0);
+
+  /**
+   * Census defined rpc client {@link IntervalViewDescriptor}s.
+   */
+  public static final IntervalViewDescriptor RPC_CLIENT_ROUNDTRIP_LATENCY_INTERVAL_VIEW =
```
dinooliva: done

core/java/com/google/instrumentation/stats/RpcConstants.java
dinooliva: These definitions are consistent with what's defined internally. This will need to be resolved as part of the resolution of 88.

core/java/com/google/instrumentation/stats/RpcConstants.java
dinooliva: The format of view names is still under discussion but I've gone ahead and made everything consistent.

core/java/com/google/instrumentation/stats/RpcConstants.java
dinooliva: Good catch!

core/java/com/google/instrumentation/stats/RpcConstants.java
dinooliva: done

core/java/com/google/instrumentation/stats/RpcConstants.java
```diff
@@ -218,6 +221,128 @@
...
+          IntervalAggregationDescriptor.create(Arrays.asList(MINUTE, HOUR)),
+          Arrays.asList(RPC_CLIENT_METHOD));
+
+  public static final IntervalViewDescriptor RPC_CLIENT_ERROR_COUNT_INTERVAL_VIEW =
+      IntervalViewDescriptor.create(
+          "rpc client error count",
+          "RPC Errors",
+          RPC_CLIENT_ERROR_COUNT,
+          IntervalAggregationDescriptor.create(Arrays.asList(MINUTE, HOUR)),
+          Arrays.asList(RPC_CLIENT_METHOD));
```
dinooliva: Interval stats do not include it internally. If it's necessary, we can add it later.

core/java/com/google/instrumentation/stats/RpcConstants.java
```diff
@@ -13,20 +13,21 @@
...
 /**
  * Constants for collecting rpc stats.
  */
 public final class RpcConstants {
-  /**
-   * Rpc tag keys.
-   */
+  // Rpc tag keys.
   public static final TagKey RPC_STATUS = TagKey.create("OpStatus");
   public static final TagKey RPC_CLIENT_METHOD = TagKey.create("method");
```
a-veitch: Who about rpc_method instead of just method as a key name?

core/java/com/google/instrumentation/stats/RpcConstants.java
```diff
@@ -13,20 +13,21 @@
...
 import java.util.List;
 
 /**
  * Constants for collecting rpc stats.
  */
 public final class RpcConstants {
-  /**
-   * Rpc tag keys.
-   */
+  // Rpc tag keys.
```
a-veitch: Add documentation for each one of the constants?

core/java/com/google/instrumentation/stats/RpcConstants.java
```diff
@@ -36,9 +37,7 @@
   private static final List<BasicUnit> scalar = Arrays.asList(BasicUnit.SCALAR);
   private static final List<BasicUnit> seconds = Arrays.asList(BasicUnit.SECONDS);
 
-  /**
-   * Census defined rpc client {@link MeasurementDescriptor}s.
-   */
+  // Census defined rpc client {@link MeasurementDescriptor}s.
```
a-veitch: Why change all of these from javadoc?

core/java/com/google/instrumentation/stats/RpcConstants.java
```diff
@@ -116,29 +113,27 @@
...
+  static final List<Double> RPC_MILLIS_BUCKET_BOUNDARIES = Collections.unmodifiableList(
+      Arrays.asList(0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 8.0, 10.0, 13.0, 16.0, 20.0, 25.0, 30.0,
+          40.0, 50.0, 65.0, 80.0, 100.0, 130.0, 160.0, 200.0, 250.0, 300.0, 400.0, 500.0, 650.0,
+          800.0, 1000.0, 2000.0, 5000.0, 10000.0, 20000.0, 50000.0, 100000.0));
+
+  // Census defined rpc client {@link ViewDescriptor}s.
   public static final DistributionViewDescriptor RPC_CLIENT_ERROR_COUNT_VIEW =
       DistributionViewDescriptor.create(
-          "rpc client error count",
+          "rpc client error_count",
```
a-veitch: Why have underscores just for the last bit of the name? I'd expect all underscores, or none.

core/java/com/google/instrumentation/stats/RpcConstants.java
```diff
@@ -218,6 +211,124 @@
...
 
+  // Interval Stats
+  static final Duration MINUTE = Duration.create(60, 0);
+  static final Duration HOUR = Duration.create(60 * 60, 0);
+
+  // Census defined rpc client {@link IntervalViewDescriptor}s.
+  public static final IntervalViewDescriptor RPC_CLIENT_ROUNDTRIP_LATENCY_INTERVAL_VIEW =
+      IntervalViewDescriptor.create(
+          "rpc client roundtrip_latency",
+          "Latency in msecs",
```
a-veitch: I would add mention of minute-hour stats in the description, to avoid confusion with other similar names. Maybe also should be reflected in the name?

core/java/com/google/instrumentation/stats/RpcConstants.java
dinooliva: As Kristen notes, the javadoc only associates with one definition. Eventually we should probably document each of the views but that should probably happen in the proto-based version of this file once we have finalized all of the view.

core/java/com/google/instrumentation/stats/RpcConstants.java
dinooliva: I have updated the descriptions.

core/java/com/google/instrumentation/stats/RpcConstants.java
dinooliva: I think going with OP_STATUS and METHOD probably makes the most sense but is out of scope for this pr. I'll put together something separately.

core/java/com/google/instrumentation/stats/RpcConstants.java
dinooliva: The associated metric name is rpc/client/error_count - this convention changes the '/' to ' ' - we can decide if there is another convention that makes more sense as part of the proto representation work.

PR #93, opened 2017-02-08
Updates wire-encoding of tags to use varints

dinooliva:


adriancole:
is this only used for length prefixing? or is this also to be used somewhere else

is the motivation size?

dinooliva:
Hi Adrian - Yes, our encoding is <key_size><key_bytes><value_size><value_bytes> ...

Originally we considered using a byte to encode size but decided to use VarInt so that we're not overly constrained on the size of keys and values.

sebright:
I pushed a commit that adds a Maven build file.

core_impl/java/com/google/instrumentation/stats/StatsSerializer.java
```diff
@@ -14,77 +14,71 @@
...
+      int numTags = VarInt.getVarInt(buffer);
+      HashMap<String, String> tags = new HashMap<String, String>(numTags);
+      for (int i = 0; i < numTags; i++) {
+        String key = decode(buffer);
+        String val = decode(buffer);
+        tags.put(key, val);
+      }
+      return new StatsContextImpl(tags);
+    } catch (BufferUnderflowException exn) {
+      throw new IOException(exn.toString());
```
sebright: Would it be better to pass the whole exception to IOException?

shared/java/com/google/io/base/VarInt.java
```diff
@@ -0,0 +1,290 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.nio.ByteBuffer;
+
+/**
+ * Common methods to encode and decode varints and varlongs into ByteBuffers and
+ * arrays.
+ * <p>
+ * Note: this class is package-private because it is meant to be replaced by an
```
sebright: I think this comment is out of date.

core/javatests/com/google/instrumentation/stats/StatsContextFactoryTest.java
```diff
@@ -25,7 +26,8 @@
  */
 @RunWith(JUnit4.class)
 public class StatsContextFactoryTest {
-  @Test
+
+  @Test(expected = IOException.class)
   public void testDeserializeEmpty() throws Exception {
     assertThat(Stats.getStatsContextFactory().deserialize(new ByteArrayInputStream(new byte[0])))
         .isEqualTo(Stats.getStatsContextFactory().getDefault());
```
dinooliva: done.

shared/java/com/google/io/base/VarInt.java
dinooliva: Done

core_impl/java/com/google/instrumentation/stats/StatsSerializer.java
dinooliva: Done.

core_impl/java/com/google/instrumentation/stats/StatsSerializer.java
```diff
@@ -14,77 +14,71 @@
...
+    Set<Entry<String, String>> tags = context.tags.entrySet();
+    VarInt.putVarInt(tags.size(), buffer);
+    for (Entry<String, String> tag : tags) {
+      encode(tag.getKey(), buffer);
+      encode(tag.getValue(), buffer);
     }
-    StatsContextProto.StatsContext.newBuilder().setTags(builder.toString()).build().writeTo(output);
+    int length = buffer.capacity() - buffer.remaining();
+    buffer.rewind();
+    ByteString encodedTags = ByteString.copyFrom(buffer, length);
```
dinooliva: Not as far as I've been able to find. If the cost is prohibitive, we may ave to look into developing our own subclass implementation but we'll need some evidence first.

sebright: okay

PR #94, opened 2017-02-09
Move stats examples to the equivalent maven path.

bogdandrutu:


bogdandrutu:
Yes we would like to move all the code to follow the maven preferred path.

PR #99, opened 2017-02-17
Downgrades default guava version to 19.0, and tests with guava 19.0, 20.0, and 21.0.

sebright:
This change also requires downgrading truth to 0.30.

See https://github.com/grpc/grpc-java/issues/2688#issuecomment-280205113 .  I tested with multiple `guava` versions by adding a Maven property for the version, because I couldn't find a way to specify the version of a dependency on the command line.  Let me know if there is a better way to test multiple versions.

/cc @zhangkun83 @dinooliva

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,11 +42,14 @@ install:
         cd .. ;;
     esac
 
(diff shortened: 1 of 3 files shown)


sebright:
Thanks!

PR #105, opened 2017-02-28
Add few usage examples for the trace API.

bogdandrutu:


adriancole:
cc @reta

examples/src/main/java/com/google/instrumentation/trace/BasicTracing.java
```diff
@@ -0,0 +1,30 @@
...
+
+package com.google.instrumentation.trace;
+
+/** Example showing how to create a {@link Span} and add annotations. */
+public final class BasicTracing {
+  // Per class Tracer.
+  private static final Tracer tracer = Tracer.getTracer();
+
+  private static void doWork() {
+    try (Span span = tracer.startSpan(null, "MyRootSpan", StartSpanOptions.getDefault())) {
```
adriancole: so in this case, this is try-with closeable. In brave, I unimplemented closeable on span for some various reasons, though no-one else reviewed the change and @reta asked about this practice... we can consider reverting it.

https://github.com/openzipkin/brave/pull/352

what were your thoughts on having span be closeable? Assuming for simplicity or maybe because opentracing does it?

examples/src/main/java/com/google/instrumentation/trace/MultiSpansContextTracing.java
```diff
@@ -0,0 +1,53 @@
...
+
+import com.google.instrumentation.common.NonThrowingCloseable;
+
+/**
+ * Example showing how to create a child {@link Span}, install it to the current context and add
+ * annotations.
+ */
+public final class MultiSpansContextTracing {
+  // Per class Tracer.
+  private static final Tracer tracer = Tracer.getTracer();
```
adriancole: note: this is nice for an example, though not a test (like a learning test), as it doesn't show the intended result actually happened. no action needed, just saying

examples/src/main/java/com/google/instrumentation/trace/MultiSpansScopedTracing.java
```diff
@@ -0,0 +1,48 @@
...
+
+package com.google.instrumentation.trace;
+
+import com.google.instrumentation.common.NonThrowingCloseable;
+
+/**
+ * Example showing how to create a child {@link Span} using {@link ScopedSpan}, install it in the
+ * current context, and add annotations.
+ */
+public final class MultiSpansScopedTracing {
```
adriancole: maybe an example of async. For example, the scoping part is implicitly attaching the span as current. sometimes we need to set current from a span created on another thread.

adriancole:
nice work!

examples/src/main/java/com/google/instrumentation/trace/BasicTracing.java
bogdandrutu: It is easier to use and less error-prone for the case when users want to instrument their code without using the context propagation.

examples/src/main/java/com/google/instrumentation/trace/MultiSpansContextTracing.java
bogdandrutu: I think these are examples not tests. We will have tests separately.

examples/src/main/java/com/google/instrumentation/trace/MultiSpansScopedTracing.java
bogdandrutu: Coming next. You can add this to the issues as a request.

PR #106, opened 2017-02-28
Make all Maven build files extend com.google:google directly.

sebright:
`core/pom.xml` and `core_impl/pom.xml` no longer extend the top-level `pom.xml`, which
simplifies the build.  However, a few fields need to be duplicated between the
`pom.xml` files.

/cc @bogdandrutu

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -8,18 +8,46 @@
   <version>0.1.0-SNAPSHOT</version>
 
   <name>Census Java Core</name>
+
+  <description>
+    Census provides a framework to define and collect stats against metrics and
+    to break those stats down across user-defined dimensions.
+  </description>
   <url>https://github.com/google/instrumentation-java</url>
 
+  <packaging>pom</packaging>
+
   <parent>
-    <groupId>com.google.census</groupId>
-    <artifactId>census-parent</artifactId>
-    <version>0.1.0-SNAPSHOT</version>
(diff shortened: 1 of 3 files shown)


core/pom.xml
```diff
@@ -8,18 +8,46 @@
   <version>0.1.0-SNAPSHOT</version>
 
   <name>Census Java Core</name>
+
+  <description>
+    Census provides a framework to define and collect stats against metrics and
```
bogdandrutu: I think we should update this description:

Instrumentation provides a library that includes functionality for tracing, tagging and stats recording.

core_impl/pom.xml
```diff
@@ -8,16 +8,32 @@
...
+  <description>
+    Census provides a framework to define and collect stats against metrics and
+    to break those stats down across user-defined dimensions.
+  </description>
   <url>https://github.com/google/instrumentation-java</url>
 
+  <packaging>pom</packaging>
+
   <parent>
-    <groupId>com.google.census</groupId>
```
bogdandrutu: Should we create a shared parent that is a child of the google pom but sets the common things for core/core_impl/etc.?

sebright:
I fixed the parent.relativePath warning without removing the build file inheritance in #109.

core/pom.xml
sebright: Thanks.  I updated the description in #109.

PR #107, opened 2017-02-28
Move the loading with reflection logic from the Tracer to Provider.

bogdandrutu:
We want to share the same logic between stats and trace.

Add more tests for the loading code.

### DIFF
--- a/core/src/test/java/com/google/instrumentation/common/ProviderTest.java
+++ b/core/src/test/java/com/google/instrumentation/common/ProviderTest.java
@@ -15,47 +15,113 @@ package com.google.instrumentation.common;
 
 import static com.google.common.truth.Truth.assertThat;
 
+import java.util.ServiceConfigurationError;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
 
-/**
- * Tests for {@link Provider}
- */
+/** Tests for {@link Provider} */
 @RunWith(JUnit4.class)
 public class ProviderTest {
-  @Test
-  public void testGoodClass() throws Exception {
-    GetGen getGen0 =
-        Provider.newInstance("com.google.instrumentation.common.ProviderTest$GetGen", null);
-    assertThat(getGen0).isNotNull();
-    assertThat(getGen0.getGen()).isEqualTo(0);
-    for (int i = 1; i < 10; i++) {
-      GetGen getGenI =
-          Provider.newInstance("com.google.instrumentation.common.ProviderTest$GetGen", null);
-      assertThat(getGenI).isNotNull();
-      assertThat(getGenI.getGen()).isEqualTo(i);
-    }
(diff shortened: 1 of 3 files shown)


core/src/test/java/com/google/instrumentation/common/ProviderTest.java
```diff
@@ -15,47 +15,111 @@
...
+            true,
+            Provider.getCorrectClassLoader(ProviderTest.class)),
+        NoDefaultConstructorClass.class);
+  }
 
-    public GetGen() {
-      gen = genCount++;
-    }
+  @Test(expected = ServiceConfigurationError.class)
+  public void createInstance_ThrowsErrorWhenNoSubclass() throws Exception {
```
sebright: I think this test name could be clearer, for example, `createInstance_ThrowsErrorWhenClassIsNotASubclass` or `createInstance_ThrowsErrorWhenClassCannotBeCast`.

core/src/test/java/com/google/instrumentation/common/ProviderTest.java
```diff
@@ -15,47 +15,111 @@
...
-    assertThat(getGen0.getGen()).isEqualTo(0);
-    for (int i = 1; i < 10; i++) {
-      GetGen getGenI =
-          Provider.newInstance("com.google.instrumentation.common.ProviderTest$GetGen", null);
-      assertThat(getGenI).isNotNull();
-      assertThat(getGenI.getGen()).isEqualTo(i);
-    }
-    assertThat(getGen0.getGen()).isEqualTo(0);
+    assertThat(
+            Provider.<GoodClass>newInstance(
```
sebright: Is `createInstance` meant to replace `newInstance`?

bogdandrutu: Yes.
Reasons are:
* We need to call Class.forName with a literal otherwise shading tools will not be able to replace the path.
* Android class-loader http://stackoverflow.com/questions/13407006

core/src/test/java/com/google/instrumentation/common/ProviderTest.java
bogdandrutu: Done

core/src/test/java/com/google/instrumentation/common/ProviderTest.java
```diff
@@ -70,7 +70,8 @@ public void createInstance_ThrowsErrorWhenPrivateClass() throws Exception {
   }
 
   @Test(expected = ServiceConfigurationError.class)
-  public void createInstance_ThrowsErrorWhenPrivateConstructor() throws Exception {
+  public void createInstance_ThrowsErrorWhenClassDoesNotHavePrivateConstructor()
```
sebright: Shouldn't this be `createInstance_ThrowsErrorWhenClassHasPrivateConstructor`?

PR #110, opened 2017-03-01
Add implementation for ContextSpanHandler.

bogdandrutu:
Add also a getter for the io.grpc.Context.Key.


core_context_impl/src/test/java/com/google/instrumentation/trace/ContextSpanHandlerImplTest.java
```diff
@@ -0,0 +1,79 @@
...
+    assertThat(new ContextSpanHandlerImpl().getCurrentSpan()).isNull();
+    assertThat(tracer.getCurrentSpan()).isSameAs(BlankSpan.getInstance());
+  }
+
+  @Test
+  public void getCurrentSpan() {
+    Context origContext =
+        Context.current().withValue(GrpcTraceUtils.getContextSpanKey(), span).attach();
+    assertThat(tracer.getCurrentSpan()).isSameAs(span);
+    Context.current().detach(origContext);
```
sebright: Is this line cleaning up for the following tests?  I think it should be in a finally block.

core_context_impl/src/test/java/com/google/instrumentation/trace/ContextSpanHandlerImplTest.java
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  @Test
+  public void withSpan() {
```
sebright: Should this test test that the original span is restored?

core_context_impl/src/main/java/com/google/instrumentation/trace/GrpcTraceUtils.java
```diff
@@ -0,0 +1,37 @@
...
+
+package com.google.instrumentation.trace;
+
+import io.grpc.Context;
+
+/**
+ * Util methods/functionality for gRPC system.
+ */
+public final class GrpcTraceUtils {
+  // Reserved {@link io.grpc.Context.Key} for the {@link Span}.
```
bogdandrutu: Done. I think the agreement is to not use javadocs for private things. in the IDE you can click on the next line directly on the type.

core_context_impl/src/main/java/com/google/instrumentation/trace/GrpcTraceUtils.java
```diff
@@ -0,0 +1,37 @@
...
+  private static final Context.Key<Span> CONTEXT_SPAN_KEY =
+      Context.key("instrumentation-trace-key");
+
+  /**
+   * Returns The {@link io.grpc.Context.Key} used to interact with {@link io.grpc.Context}.
+   *
+   * @return The {@code io.grpc.Context.Key} used to interact with {@code io.grpc.Context}.
```
bogdandrutu: I looked at the generated javadoc. If this is public then we must add them. It is much easier to read. Happy to chat offline about this.

core_context_impl/src/main/java/com/google/instrumentation/trace/ContextSpanHandlerImpl.java
```diff
@@ -0,0 +1,72 @@
...
+
+  @Override
+  public NonThrowingCloseable withSpan(Span span) {
+    return new WithSpan(span, GrpcTraceUtils.getContextSpanKey());
+  }
+
+  // Defines an arbitrary scope of code as a traceable operation. Supports try-with-resources idiom.
+  private static class WithSpan implements NonThrowingCloseable {
+    // The original {@link io.grpc.Context};
+    private final Context origContext;
```
bogdandrutu: No actually no need for a comment name is suggestive.

core_context_impl/src/main/java/com/google/instrumentation/trace/ContextSpanHandlerImpl.java
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * Constructs a new {@link WithSpanImpl}.
```
bogdandrutu: Done.

core_context_impl/src/main/java/com/google/instrumentation/trace/ContextSpanHandlerImpl.java
```diff
@@ -0,0 +1,72 @@
...
+    return GrpcTraceUtils.getContextSpanKey().get(Context.current());
+  }
+
+  @Override
+  public NonThrowingCloseable withSpan(Span span) {
+    return new WithSpan(span, GrpcTraceUtils.getContextSpanKey());
+  }
+
+  // Defines an arbitrary scope of code as a traceable operation. Supports try-with-resources idiom.
+  private static class WithSpan implements NonThrowingCloseable {
```
bogdandrutu: Done.

core_context_impl/src/test/java/com/google/instrumentation/trace/ContextSpanHandlerImplTest.java
bogdandrutu: Done.

core_context_impl/src/test/java/com/google/instrumentation/trace/ContextSpanHandlerImplTest.java
bogdandrutu: Added test that the context restored. It is not for the following tests. But want to have examples that if users copies from tests they are the expected way to use this :)

sebright: I still think that the test should restore the context in a finally block, as an example for users, and to prevent a failure in this test from causing other tests to fail.

BUILD
```diff
@@ -43,6 +43,17 @@ java_library(
 )
 
 java_library(
+    name = "trace-core_context_impl",
```
bogdandrutu: Done.

core_context_impl/src/test/java/com/google/instrumentation/trace/ContextSpanHandlerImplTest.java
bogdandrutu: Done.

PR #113, opened 2017-03-01
Modify serialize and deserialize to match the new tag encodings.

songy23:
* Removed type byte sequence according to the discussion in Doc.

Have run bazel test StatsContextTest && bazel test StatsContextFactoryTest, all tests passed.

Refer to: https://docs.google.com/document/d/1z6ixVD1VDlDWyPE1o6KoquU_yNJ4E79PrvFiR6RD95s/edit#heading=h.za0m8gfrsbdh


core_impl/src/main/java/com/google/instrumentation/stats/StatsSerializer.java
```diff
@@ -30,14 +30,25 @@
...
 final class StatsSerializer {
+  //  * tag_type is one byte, and is used to describe the format of value_bytes.
+  //    In particular, the low 2 bits of this byte are used as follows:
+  //    00 (value 0): string (UTF-8) encoding
+  //    01 (value 1): integer (varint int64 encoding).
+  //    10 (value 2): boolean format.
+  //    TODO(songya): Currently we only support encoding on string type.
+  static final int TYPE_STRING = 0;
+  static final int TYPE_INTEGER = 1;
+  static final int TYPE_BOOLEAN = 2;
```
dinooliva: VALUE_TYPE

It may make sense to make this an enum - e.g. VALUE_TYPE.STRING.

songy23: Yep enum should make more sense. Will update VALUE_TYPE

core_impl/src/main/java/com/google/instrumentation/stats/StatsSerializer.java
```diff
@@ -48,17 +59,28 @@ static void serialize(StatsContextImpl context, OutputStream output) throws IOEx
...
+        int type = VarInt.getVarInt(buffer);
+        switch (type) {
+          case VALUE_TYPE_STRING:
+            String key = decode(buffer);
+            String val = decode(buffer);
+            tags.put(key, val);
+            break;
+          case VALUE_TYPE_INTEGER:
+          case VALUE_TYPE_BOOLEAN:
+            // TODO(songya): add support for tag types integer and boolean
```
dinooliva: If we run into these cases while they are unsupported, we should through an IOException (otherwise we're going to try to read the tag data as a type in the next iteration).

core/src/test/java/com/google/instrumentation/stats/StatsContextFactoryTest.java
```diff
@@ -24,8 +24,14 @@
  */
 @RunWith(JUnit4.class)
 public class StatsContextFactoryTest {
-  @Test(expected = IOException.class)
+  @Test
```
dinooliva: So this no longer throws an exception? That's unexpected - is there a reason it shouldn't?

core_impl/src/main/java/com/google/instrumentation/stats/StatsSerializer.java
dinooliva: You should also add unit tests to verify that an exception is thrown when the type is not string.

songy23: Got it, will update the logic here and add more tests on it.

core/src/test/java/com/google/instrumentation/stats/StatsContextFactoryTest.java
songy23: Yes, I talked with Alistair yesterday and we think since the encoding has changed to [tag_metadata key_len key_bytes value_len value_bytes]*, an empty byte array will indicate that user doesn't specify any tag and uses the default tags. Therefore this case is allowed now.

While in the past the encoding format is num_tags [key_len key_bytes value_len value_bytes]*, even if a user doesn't specify a tag, there will be one type indicating the number of tags (0 in this case), and the input byte array should never be empty.

dinooliva: Makes sense, though we should document that somewhere.

This test should then be updated to check that deserializing an empty byte array returns the default stats context.

songy23: Yep, that case is covered in StatsContextTest.testSerialize()

core/src/test/java/com/google/instrumentation/stats/StatsContextFactoryTest.java
```diff
@@ -24,8 +27,81 @@
...
+
+  private static final String KEY1 = "Key";
+  private static final String VALUE_STRING = "String";
+  private static final int VALUE_INT = 10;
+  private static final boolean VALUE_BOOL = true;
+
+  @Test
   public void testDeserializeEmpty() throws Exception {
-    Stats.getStatsContextFactory().deserialize(new ByteArrayInputStream(new byte[0]));
+    testDeserialize(new ByteArrayInputStream(new byte[0]));
```
dinooliva: Since this is a test and doesn't throw an exception, you should test that the value is the expected value. If you think this is redundant with other tests, you can consider removing it.

songy23: Sure, just add an assert here to verify deserialize returns a default stats context.

core_impl/src/main/java/com/google/instrumentation/stats/StatsSerializer.java
```diff
@@ -30,14 +30,28 @@
...
+  private static final int VALUE_TYPE_STRING = 0;
+  private static final int VALUE_TYPE_INTEGER = 1;
+  private static final int VALUE_TYPE_BOOLEAN = 2;
+
   // Serializes a StatsContext by transforming it into a StatsContextProto. The
   // encoded tags are of the form:
-  //   num_tags [key_len key_bytes value_len value_bytes]*
+  //   [tag_metadata key_len key_bytes value_len value_bytes]*
   static void serialize(StatsContextImpl context, OutputStream output) throws IOException {
     ByteBuffer buffer = ByteBuffer.allocate(1024);
```
dinooliva: This is not your issue but, since ByteBuffers don't resize automatically, this code could cause problems. Add a TODO() to fix this issue.

songy23: Yeah added a note here. Will also do a rebase to merge my previous commits.

PR #114, opened 2017-03-02
Add a default implementation for TimestampFactory.

bogdandrutu:
Default implementation uses System#currentTimeMillis.
Add a way to provide other implementation using java8.
Remove the Tracer#timeNow.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -15,7 +15,10 @@
 java_library(
     name = "common-core",
     srcs = glob(["core/src/main/java/com/google/instrumentation/common/*.java"]),
-    deps = ["@jsr305//jar"],
+    deps = [
+        "@guava//jar",
+        "@jsr305//jar",
+    ],
 )
 
 java_library(
--- a/BUILD
+++ b/BUILD
@@ -359,6 +359,19 @@ java_test(
 )
 
 java_test(
+    name = "TimestampFactoryTest",
+    srcs = ["core/src/test/java/com/google/instrumentation/common/TimestampFactoryTest.java"],
+    deps = [
+        ":common-core",
+        "@guava//jar",
+        "@guava_testlib//jar",
+        "@jsr305//jar",
+        "@junit//jar",
+        "@truth//jar",
+    ],
+)
+
+java_test(
     name = "ViewDescriptorTest",
     srcs = ["core/src/test/java/com/google/instrumentation/stats/ViewDescriptorTest.java"],
     deps = [
--- a/core/src/main/java/com/google/instrumentation/common/Timestamp.java
+++ b/core/src/main/java/com/google/instrumentation/common/Timestamp.java
@@ -24,7 +24,7 @@ public final class Timestamp {
   private static final long MAX_SECONDS = 315576000000L;
   private static final int MAX_NANOS = 999999999;
   private static final long NUM_MILLIS_PER_SECOND = 1000L;
-  private static final int NUM_NANOS_PER_MILLI = 1000000;
+  private static final int NUM_NANOS_PER_MILLI = 1000 * 1000;
   private final long seconds;
   private final int nanos;
 
--- a/core/src/main/java/com/google/instrumentation/common/Timestamp.java
+++ b/core/src/main/java/com/google/instrumentation/common/Timestamp.java
@@ -16,8 +16,25 @@ package com.google.instrumentation.common;
 /**
  * A representation of an instant in time. The instant is the number of nanoseconds after the number
  * of seconds since the Unix Epoch.
+ *
+ * <p>Use {@link TimestampFactory#now} to get the current timestamp since epoch
+ * (1970-01-01T00:00:00Z).
  */
-public class Timestamp {
+public final class Timestamp {
+  private static final long MAX_SECONDS = 315576000000L;
+  private static final int MAX_NANOS = 999999999;
+  private static final long NUM_MILLIS_PER_SECOND = 1000L;
+  private static final int NUM_NANOS_PER_MILLI = 1000000;
+  private final long seconds;
+  private final int nanos;
+
+  private Timestamp(long seconds, int nanos) {
+    this.seconds = seconds;
+    this.nanos = nanos;
+  }
+
+  // TODO(bdrutu): Make create and fromMillis package-protected.
+
   /**
    * Creates a new timestamp from given seconds and nanoseconds.
    *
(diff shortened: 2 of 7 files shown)


BUILD
```diff
@@ -15,7 +15,10 @@
 java_library(
     name = "common-core",
     srcs = glob(["core/src/main/java/com/google/instrumentation/common/*.java"]),
-    deps = ["@jsr305//jar"],
+    deps = [
+        "@guava//jar",
```
sebright: We have been trying to avoid a dependency on guava in the API. The discussion is in #13.  It looks like you only use it for `@VisibleForTesting` on a non-public method; could you add a comment instead?

core/src/main/java/com/google/instrumentation/common/Timestamp.java
```diff
@@ -16,8 +16,25 @@
...
+ *
+ * <p>Use {@link TimestampFactory#now} to get the current timestamp since epoch
+ * (1970-01-01T00:00:00Z).
  */
-public class Timestamp {
+public final class Timestamp {
+  private static final long MAX_SECONDS = 315576000000L;
+  private static final int MAX_NANOS = 999999999;
+  private static final long NUM_MILLIS_PER_SECOND = 1000L;
+  private static final int NUM_NANOS_PER_MILLI = 1000000;
```
sebright: `1000 * 1000` might be clearer.

core/src/main/java/com/google/instrumentation/common/TimestampFactory.java
```diff
@@ -13,16 +13,52 @@
...
+  static class Handler {
+    protected Handler() {};
+
+    protected Timestamp timeNow() {
+      return Timestamp.fromMillis(System.currentTimeMillis());
+    }
+  }
+
+  @VisibleForTesting
+  static Handler loadTimestamFactoryHandler(ClassLoader classLoader) {
```
sebright: s/stam/stamp

core/src/main/java/com/google/instrumentation/common/TimestampFactory.java
```diff
@@ -13,16 +13,52 @@
...
-   * @return the new timestamp.
+   * @return the current instant using the system clock, not null.
    */
-  Timestamp now();
+  public static Timestamp now() {
+    return HANDLER.timeNow();
+  }
+
+  // This class cannot be final because in case of java8 runtime we can have a better granularity
+  // timestamp.
```
sebright: Will `TimestampFactory` also be able to mock time in tests?

core/src/main/java/com/google/instrumentation/common/TimestampFactory.java
```diff
@@ -13,16 +13,52 @@
...
+  @Immutable
+  static class Handler {
```
sebright: I think that it would be slightly cleaner to split this class into an interface and an immutable final class.  Then the Java 8 class could implement the interface.

core/src/test/java/com/google/instrumentation/common/TimestampFactoryTest.java
```diff
@@ -0,0 +1,54 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/** Unit tests for {@link TimestampFactory}. */
+@RunWith(JUnit4.class)
+public class TimestampFactoryTest {
+  private static final int NUM_NANOS_PER_MILLI = 1000000;
```
sebright: Same here, I think `1000 * 1000` is clearer.

core/src/test/java/com/google/instrumentation/common/TimestampFactoryTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/** Unit tests for {@link TimestampFactory}. */
+@RunWith(JUnit4.class)
+public class TimestampFactoryTest {
```
sebright: Does this test suite have a build target?

BUILD
bogdandrutu: We discussed this with gRPC people. As long as we depend on the same guava version is good. We use guava in trace a lot for checkNotNull, checkArgument, etc.

core/src/main/java/com/google/instrumentation/common/TimestampFactory.java
bogdandrutu: done.

core/src/test/java/com/google/instrumentation/common/TimestampFactoryTest.java
bogdandrutu: now it does.

core/src/test/java/com/google/instrumentation/common/TimestampFactoryTest.java
bogdandrutu: done.

core/src/main/java/com/google/instrumentation/common/Timestamp.java
bogdandrutu: done.

core/src/main/java/com/google/instrumentation/common/TimestampFactory.java
bogdandrutu: Too much Java question in a 10 words sentence. It is possible by injecting a mock handler. When we will need that we add a way to inject a different implementation for the Handler.

core/src/main/java/com/google/instrumentation/common/TimestampFactory.java
bogdandrutu: done.

core/src/main/java/com/google/instrumentation/common/TimestampFactory.java
```diff
@@ -41,16 +41,18 @@ public static Timestamp now() {
   // This class cannot be final because in case of java8 runtime we can have a better granularity
   // timestamp.
   @Immutable
-  static class Handler {
-    protected Handler() {};
+  static interface Handler {
```
sebright: Static doesn't have an effect on interfaces.

core/src/main/java/com/google/instrumentation/common/TimestampFactory.java
sebright: Sounds good.

core/src/main/java/com/google/instrumentation/common/TimestampFactory.java
```diff
@@ -41,16 +41,18 @@ public static Timestamp now() {
...
+    Timestamp timeNow();
+  }
 
-    protected Timestamp timeNow() {
+  static final class DefaultHandler implements Handler {
+    public Timestamp timeNow() {
```
sebright: I think this should have `@Override`, though I'm surprised that CI didn't catch it.

core/src/main/java/com/google/instrumentation/common/TimestampFactory.java
bogdandrutu: Done.

core/src/main/java/com/google/instrumentation/common/TimestampFactory.java
bogdandrutu: Done.

PR #118, opened 2017-03-03
Enable all Error Prone warnings in Travis Bazel build.

sebright:
/cc @dinooliva @bogdandrutu


bogdandrutu:
Please fix the travis build.

sebright:
I need to figure out how to exclude the generated Protocol Buffer classes.

sebright:
I just disabled the three warnings. I think it will be easier to only suppress warnings from Protocol Buffers once we switch to the new `java_proto_library` build rule.

PR #120, opened 2017-03-04
Update gitignore to ignore IDEA and clasify all ignores by source.

bogdandrutu:


.gitignore
```diff
@@ -1,6 +1,26 @@
...
+
+# IntelliJ IDEA
+.idea
+*.iml
+
+# Eclipse
 .classpath
 .project
 .settings
+.gitignore
```
sebright: I don't think we should ignore the .gitignore file while we store it in the main repository.

bogdandrutu: You are correct. Bad source :).

PR #123, opened 2017-03-07
Add a SpanBuilder class.

bogdandrutu:
To avoid exponential increase in the number of startSpan[XXX] methods in the Tracer class.

### DIFF
--- a/core/src/main/java/com/google/instrumentation/trace/SpanBuilder.java
+++ b/core/src/main/java/com/google/instrumentation/trace/SpanBuilder.java
@@ -1,0 +1,277 @@
+/*
+ * Copyright 2017, Google Inc.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.instrumentation.trace;
+
+import com.google.instrumentation.common.NonThrowingCloseable;
+import com.google.instrumentation.common.Timestamp;
+
+import java.util.List;
+import javax.annotation.Nullable;
+
+/**
+ * {@link SpanBuilder} is used to construct {@link Span} instances which define arbitrary scopes of
+ * code that are sampled for distributed tracing as a single atomic unit.
+ *
+ * <p>This is a simple example where all the work is being done within a single scope and the
+ * Context is manually propagated:
+ *
+ * <pre>{@code
+ * class MyClass {
+ *   private static final Tracer tracer = Tracer.getTracer();
+ *   void doWork(Span parent) {
+ *     // Create a Span as a child of the given parent.
+ *     try (Span span = tracer.spanBuilder(parent, "MyChildSpan").startSpan()) {
+ *       span.addAnnotation("my annotation");
+ *       doSomeWork(span);
+ *     }
+ *     // "span" will be ended here.
+ *   }
+ * }
+ * }</pre>
+ *
+ * <p>This is a simple example where all the work is being done within a single scope and a single
+ * thread and the Context is automatically propagated:
+ *
+ * <pre>{@code
+ * class MyClass {
+ *   private static final Tracer tracer = Tracer.getTracer();
+ *   void doWork {
+ *     // Create a Span as a child of the current Span.
+ *     try (NonThrowingCloseable ss = tracer.spanBuilder("MyChildSpan").startScopedSpan()) {
+ *       tracer.getCurrentSpan().addAnnotation("my annotation");
+ *       doSomeWork();  // Here the new span is in the current Context.
+ *     }
+ *   }
+ * }
+ * }</pre>
+ *
+ * <p>There might be cases where you do not perform all the work inside one static scope and the
+ * Context is automatically propagated:
+ *
+ * <pre>{@code
+ * class MyWorkClass {
+ *   private static final Tracer tracer = Tracer.getTracer();
+ *   Span mySpan;
+ *
+ *   MyWorkClass() {
+ *     // Create a Span as a child of the current Span.
+ *     mySpan = tracer.spanBuilder("MyChildSpan").startSpan();
+ *   }
+ *
+ *   void doInitialWork {
+ *     try (NonThrowingCloseable ws = tracer.withSpan(mySpan)) {
+ *       tracer.getCurrentSpan().addAnnotation("my annotation in initial work");
+ *       doSomeInitialWork(); // Here the new span is in the current Context.
+ *     }
+ *   }
+ *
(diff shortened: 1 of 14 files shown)


core/src/main/java/com/google/instrumentation/trace/SpanBuilder.java
```diff
@@ -0,0 +1,276 @@
...
+  public SpanBuilder setRecordEvents(@Nullable Boolean recordEvents) {
+    startSpanOption.setRecordEvents(recordEvents);
+    return this;
+  }
+
+  /**
+   * If called this will force the newly created {@code Span} to be a root span. As a consequence,
+   * any parent specified (or inherited from the Context) will be ignored (N.B. does not apply to
+   * linked parents set through {@link #setParentLinks}).
+   *
```
adriancole: add a `<p>` and explain why you'd ever do this? :)

core/src/main/java/com/google/instrumentation/trace/SpanBuilder.java
```diff
@@ -0,0 +1,277 @@
...
+ * Context is manually propagated:
+ *
+ * <pre>{@code
+ * class MyClass {
+ *   private static final Tracer tracer = Tracer.getTracer();
+ *   void doWork(Span parent) {
+ *     // Create a Span as a child of the given parent.
+ *     try (Span span = tracer.spanBuilder(parent, "MyChildSpan").startSpan()) {
+ *       span.addAnnotation("my annotation");
+ *       doSomeWork(span);
```
adriancole: maybe something similar to this to make it more explicit that we are manually propagating.

```java
doSomeWork(span /* manually propagate a span to onto the stack */);
```

core/src/main/java/com/google/instrumentation/trace/SpanBuilder.java
```diff
@@ -0,0 +1,277 @@
...
+ * thread and the Context is automatically propagated:
+ *
+ * <pre>{@code
+ * class MyClass {
+ *   private static final Tracer tracer = Tracer.getTracer();
+ *   void doWork {
+ *     // Create a Span as a child of the current Span.
+ *     try (NonThrowingCloseable ss = tracer.spanBuilder("MyChildSpan").startScopedSpan()) {
+ *       tracer.getCurrentSpan().addAnnotation("my annotation");
+ *       doSomeWork();  // Here the new span is in the current Context.
```
adriancole: something similar to reinforce the impact of being in the current context..

// Here the new span is in the current Context, so it can be used implicitly anywhere down the stack

core/src/main/java/com/google/instrumentation/trace/SpanBuilder.java
```diff
@@ -0,0 +1,277 @@
...
+ *     }
+ *   }
+ * }
+ * }</pre>
+ *
+ * <p>There might be cases where you do not perform all the work inside one static scope and the
+ * Context is automatically propagated:
+ *
+ * <pre>{@code
+ * class MyWorkClass {
```
adriancole: make this hit harder by pretending you are an interceptor!  Ex choose a realistic rpc interceptor and pare it down.

core/src/main/java/com/google/instrumentation/trace/SpanBuilder.java
```diff
@@ -0,0 +1,277 @@
...
+  }
+
+  /**
+   * Sets the start time for the {@link Span}.
+   *
+   * @param startTime The start time for the {@code Span}. If {@code null} is used, then the current
+   *     system time at the point at which {@link #startSpan} is called will be used.
+   * @return this.
+   */
+  public SpanBuilder setStartTime(@Nullable Timestamp startTime) {
```
adriancole: personal taste, but I usually leave off "set" prefix, and don't add fillers like "return this"

core/src/main/java/com/google/instrumentation/trace/SpanBuilder.java
```diff
@@ -0,0 +1,277 @@
...
+  public SpanBuilder becomeRoot() {
+    parentSpanContext = null;
+    hasRemoteParent = false;
+    return this;
+  }
+
+  /**
+   * Creates and starts a new {@link Span}.
+   *
+   * <p>If used without try-with-resources <b>must</b> manually call {@link Span#end}.
```
adriancole: I think this will be a less used option, as it implies manual propagation. If we remove closeable from span, it will further encourage startScoped, which more often what someone wants. For those with manual propagation, they rarely get to close in the same block as it was started.

core/src/main/java/com/google/instrumentation/trace/SpanBuilder.java
```diff
@@ -0,0 +1,277 @@
...
+   *   private static final Tracer tracer = Tracer.getTracer();
+   *   void DoWork() {
+   *     try (Span span = tracer.spanBuilder(null, "MyRootSpan").startSpan()) {
+   *       span.addAnnotation("We did X");
+   *     }
+   *   }
+   * }
+   * }</pre>
+   *
+   * <p>Prior to Java SE 7, you can use a finally block to ensure that a resource is closed
```
adriancole: not sure you need to remind how to use try/finally before java 7. If you think about how to make each word count.. probably reasonable to consider axing this and just saying don't leak spans or something

core/src/main/java/com/google/instrumentation/trace/SpanBuilder.java
```diff
@@ -0,0 +1,277 @@
...
+   *
+   * <p>Example of usage:
+   *
+   * <pre>{@code
+   * class MyClass {
+   *   private static final Tracer tracer = Tracer.getTracer();
+   *   void doWork {
+   *     // Create a Span as a child of the current Span.
+   *     try (NonThrowingCloseable ss = tracer.spanBuilder("MyChildSpan").startScopedSpan()) {
+   *       tracer.getCurrentSpan().addAnnotation("my annotation");
```
adriancole: good point to show how to get the span, as adding annotation is the most likely thing as soon as it has started.

maybe put a comment above saying anything contained by this closure can call tracer.getCurrentSpan

core/src/main/java/com/google/instrumentation/trace/SpanBuilder.java
```diff
@@ -0,0 +1,277 @@
...
+   *
+   * @return the newly created {@code Span}.
+   */
+  public Span startSpan() {
+    return start();
+  }
+
+  // TODO(bdrutu): Add error_prone annotation @MustBeClosed when the 2.0.16 jar is fixed.
+  /**
+   * Creates and starts a new {@link Span} and returns an object that defines a scope where the
```
adriancole: I'd simplify this comment and move the more literal one below the example. It would make it more easily accessible.

```
Starts a new new span and sets it as the {@link Tracer#getCurrentSpan current span}.
```

core/src/main/java/com/google/instrumentation/trace/Tracer.java
```diff
@@ -29,25 +29,34 @@
  *
  * <p>This can be used similarly to {@link java.util.logging.Logger}.
  *
- * <p>Example usage:
+ * <p>Users may choose to use manual or automatic Context propagation. Because of that this class
+ * offers APIs to facilitate both usages.
+ *
+ * <p>Example usage with manual context propagation:
```
adriancole: flip the examples, as manual is for more advanced users

PR #126, opened 2017-03-08
Change TraceId and SpanId to use byte[].

bogdandrutu:
Refactor these classes to use a factory method.

### DIFF
--- a/core/src/main/java/com/google/instrumentation/trace/SpanId.java
+++ b/core/src/main/java/com/google/instrumentation/trace/SpanId.java
@@ -42,10 +42,10 @@ public final class SpanId implements Comparable<SpanId> {
   public static final SpanId INVALID = new SpanId(new byte[SPAN_ID_SIZE]);
 
   /**
-   * Creates a new {@code SpanId} whose value is taken from the given param.
+   * Returns a {@code SpanId} whose representation is given param.
    *
-   * @param bytes the 8-bytes representation.
-   * @return a new {@code SpanId}.
+   * @param bytes the representation of the {@code SpanId}.
+   * @return a {@code SpanId} whose representation is given param.
    * @throws NullPointerException if bytes is null.
    * @throws IllegalArgumentException if bytes length is not 8.
    */
--- a/core/src/main/java/com/google/instrumentation/trace/SpanId.java
+++ b/core/src/main/java/com/google/instrumentation/trace/SpanId.java
@@ -45,7 +45,7 @@ public final class SpanId implements Comparable<SpanId> {
    * Creates a new {@code SpanId} whose value is taken from the given param.
    *
    * @param bytes the 8-bytes representation.
-   * @return a new {@code SpanId} or {@link SpanId#INVALID} if all bytes are '\0'.
+   * @return a new {@code SpanId}.
    * @throws NullPointerException if bytes is null.
    * @throws IllegalArgumentException if bytes length is not 8.
    */
--- a/core/src/main/java/com/google/instrumentation/trace/SpanId.java
+++ b/core/src/main/java/com/google/instrumentation/trace/SpanId.java
@@ -13,33 +13,80 @@
 
 package com.google.instrumentation.trace;
 
-import com.google.common.base.MoreObjects;
-import com.google.common.base.Objects;
+import static com.google.common.base.Preconditions.checkArgument;
 
-import java.util.Formatter;
-import java.util.Locale;
+import com.google.common.base.MoreObjects;
+import com.google.common.io.BaseEncoding;
+
+import java.util.Arrays;
+import java.util.Random;
 import javax.annotation.concurrent.Immutable;
 
 /**
- * A class that represents a span identifier. A span identifier is a 64-bit, unsigned integer.
+ * A class that represents a span identifier. A valid span identifier is a 8-bytes array with
+ * at least one non-zero byte.
  */
 @Immutable
-public final class SpanId {
-  private static final long INVALID_SPAN_ID = 0;
-  private final long spanId;
+public final class SpanId implements Comparable<SpanId> {
+  // The size in bytes of the span id.
+  private static final int SPAN_ID_SIZE = 8;
+  private final byte[] bytes;
+
+  private SpanId(byte[] bytes) {
+    this.bytes = bytes;
+  }
 
   /**
-   * The invalid {@code SpanId}.
+   * The invalid {@code SpanId}. All bytes are 0.
    */
-  public static final SpanId INVALID = new SpanId(INVALID_SPAN_ID);
+  public static final SpanId INVALID = new SpanId(new byte[SPAN_ID_SIZE]);
 
   /**
-   * Creates a span identifier whose value is taken from the given {@code long}.
+   * Creates a new {@code SpanId} whose value is taken from the given param.
    *
-   * @param spanId a long used as the span identifier.
+   * @param bytes the 8-bytes representation.
+   * @return a new {@code SpanId} or {@link SpanId#INVALID} if all bytes are '\0'.
+   * @throws NullPointerException if bytes is null.
+   * @throws IllegalArgumentException if bytes length is not 8.
    */
-  SpanId(long spanId) {
-    this.spanId = spanId;
+  public static SpanId fromBytes(byte[] bytes) {
+    checkArgument(bytes.length == SPAN_ID_SIZE, "bytes");
+    byte[] bytesCopy = Arrays.copyOf(bytes, SPAN_ID_SIZE);
+    return Arrays.equals(bytesCopy, INVALID.bytes) ? INVALID : new SpanId(bytes);
+  }
+
+  /**
+   * Generates a new random {@code SpanId}.
+   *
+   * @param random The random number generator.
+   * @return a valid new {@code SpanId}.
+   */
+  public static SpanId generateRandomId(Random random) {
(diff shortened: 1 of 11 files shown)


core/src/test/java/com/google/instrumentation/trace/SamplersTest.java
```diff
@@ -25,9 +26,9 @@
 public class SamplersTest {
   @Test
   public void alwaysSampleSampler_AlwaysReturnTrue() {
-    TraceId traceId = new TraceId(2, 3);
-    SpanId parentSpanId = new SpanId(17);
-    SpanId spanId = new SpanId(42);
+    TraceId traceId = TraceId.generateRandomId(new SecureRandom());
+    SpanId parentSpanId = SpanId.generateRandomId(new SecureRandom());
+    SpanId spanId = SpanId.generateRandomId(new SecureRandom());
```
sebright: Would it be better to make this test deterministic?  You could create a `java.util.Random` with a fixed seed and then reuse it throughout the test.

core/src/test/java/com/google/instrumentation/trace/SpanTest.java
```diff
@@ -27,7 +28,10 @@
 @RunWith(JUnit4.class)
 public class SpanTest {
   private static final SpanContext spanContext =
-      new SpanContext(new TraceId(10, 20), new SpanId(30), TraceOptions.getDefault());
+      new SpanContext(
+          TraceId.generateRandomId(new SecureRandom()),
+          SpanId.generateRandomId(new SecureRandom()),
+          TraceOptions.getDefault());
```
sebright: same comment about random numbers

core/src/test/java/com/google/instrumentation/trace/SpanIdTest.java
```diff
@@ -23,37 +23,49 @@
...
+    assertThat(second.compareTo(first)).isLessThan(0);
+    assertThat(first.compareTo(SpanId.fromBytes(firstBytes))).isEqualTo(0);
   }
 
   @Test
   public void traceId_EqualsAndHashCode() {
     EqualsTester tester = new EqualsTester();
     tester.addEqualityGroup(SpanId.INVALID, SpanId.INVALID);
-    tester.addEqualityGroup(first, new SpanId(123));
+    tester.addEqualityGroup(first, SpanId.fromBytes(firstBytes));
```
sebright: I think it would be better to use a new array instead of `firstBytes`, to ensure that the arrays are compared correctly.

core/src/test/java/com/google/instrumentation/trace/TracerTest.java
```diff
@@ -224,8 +225,11 @@ public void startChildSpan() {
   @Test
   public void startSpanWitRemoteParent() {
     Tracer mockTracer = newTracerWithMocks();
-    SpanContext spanContext = new SpanContext(new TraceId(10, 20), new SpanId(30),
-        TraceOptions.getDefault());
+    SpanContext spanContext =
+        new SpanContext(
+            TraceId.generateRandomId(new SecureRandom()),
+            SpanId.generateRandomId(new SecureRandom()),
+            TraceOptions.getDefault());
```
sebright: same comment about random numbers

core/src/test/java/com/google/instrumentation/trace/TraceIdTest.java
```diff
@@ -23,57 +23,51 @@
...
   }
 
   @Test
   public void traceId_EqualsAndHashCode() {
     EqualsTester tester = new EqualsTester();
     tester.addEqualityGroup(TraceId.INVALID, TraceId.INVALID);
-    tester.addEqualityGroup(first, new TraceId(123, 456));
-    tester.addEqualityGroup(second, new TraceId(0, 789));
-    tester.addEqualityGroup(third, new TraceId(987, 0));
+    tester.addEqualityGroup(first, TraceId.fromBytes(firstBytes));
```
sebright: I think this should also create a new array.

core/src/main/java/com/google/instrumentation/trace/SpanId.java
```diff
@@ -13,33 +13,80 @@
...
+  /**
+   * The invalid {@code SpanId}. All bytes are 0.
+   */
+  public static final SpanId INVALID = new SpanId(new byte[SPAN_ID_SIZE]);
+
+  /**
+   * Creates a new {@code SpanId} whose value is taken from the given param.
+   *
+   * @param bytes the 8-bytes representation.
+   * @return a new {@code SpanId} or {@link SpanId#INVALID} if all bytes are '\0'.
```
bogdandrutu: Done.

core/src/test/java/com/google/instrumentation/trace/SpanIdTest.java
bogdandrutu: Done.

core/src/test/java/com/google/instrumentation/trace/TraceIdTest.java
bogdandrutu: Done.

core/src/test/java/com/google/instrumentation/trace/SamplersTest.java
bogdandrutu: Done.

core/src/test/java/com/google/instrumentation/trace/TracerTest.java
bogdandrutu: Done

core/src/test/java/com/google/instrumentation/trace/SpanTest.java
bogdandrutu: Done.

bogdandrutu:
PTAL

core/src/test/java/com/google/instrumentation/trace/SamplersTest.java
sebright: If I understand the documentation correctly, `SecureRandom` can produce non-deterministic results even if the seed is set, so I would use `java.util.Random`.  I would also create a new `Random` at the start of every test case, either as a local variable or an instance variable that can be initialized in a `@Before` method, so that test order doesn't affect the results.

core/src/main/java/com/google/instrumentation/trace/SpanId.java
sebright: The Javadoc still says that it creates new `SpanId`s, in the summary and in the `@return` tag.

bogdandrutu: I think the github review tool is a bit broken. look at the latest change.

core/src/test/java/com/google/instrumentation/trace/SamplersTest.java
bogdandrutu: Changed to manually create one Random in the functions we need.

adriancole:
like that this is explicit. it also gets you better accuracy on random (as nextLong isn't really 64bit). in most cases json encoders would need to be custom anyway, so yeah lg

core/src/main/java/com/google/instrumentation/trace/SpanId.java
```diff
@@ -13,33 +13,80 @@
...
 
-import java.util.Formatter;
-import java.util.Locale;
+import java.util.Arrays;
+import java.util.Random;
 import javax.annotation.concurrent.Immutable;
 
 /**
- * A class that represents a span identifier. A span identifier is a 64-bit, unsigned integer.
+ * A class that represents a span identifier. A valid span identifier is a 8-bytes array with
```
a-veitch: Grammar nit: s/a 8-bytes/an 8-byte/

core/src/main/java/com/google/instrumentation/trace/SpanId.java
```diff
@@ -13,33 +13,80 @@
...
+    return Arrays.equals(bytesCopy, INVALID.bytes) ? INVALID : new SpanId(bytes);
+  }
+
+  /**
+   * Generates a new random {@code SpanId}.
+   *
+   * @param random The random number generator.
+   * @return a valid new {@code SpanId}.
+   */
+  public static SpanId generateRandomId(Random random) {
```
a-veitch: Was Random introduced in Java SE 7? If so, what do we do to support earlier versions?

core/src/main/java/com/google/instrumentation/trace/SpanId.java
```diff
@@ -53,35 +100,31 @@ public boolean equals(Object obj) {
...
-   * Returns the long that represents the span identifier.
-   *
-   * @return the span identifier.
-   */
-  public long getSpanId() {
-    return spanId;
+  @Override
+  public int compareTo(SpanId that) {
+    for (int i = 0; i < SPAN_ID_SIZE; i++) {
+      if (bytes[i] == that.bytes[i]) {
```
a-veitch: You could save a line here:

if (bytes[i] != that.bytes[i]) {
  return bytes[i] < that.bytes[i] ? -1 : 1;
}

core/src/main/java/com/google/instrumentation/trace/SpanId.java
bogdandrutu: Random is from java 1.0

core/src/main/java/com/google/instrumentation/trace/SpanId.java
bogdandrutu: Thanks.

PR #127, opened 2017-03-08
Remove the try-with-resources support for the Span.

bogdandrutu:
This is important because we want to discourage the usage of the Span and encourage the usage of ScopedSpan.

Now SpanBuild#startSpan is only for manual propagation and there is no confusion that the try-with-resources objects will interact with the Context and Span does not.

### DIFF
--- a/core/src/main/java/com/google/instrumentation/trace/ScopedSpanHandle.java
+++ b/core/src/main/java/com/google/instrumentation/trace/ScopedSpanHandle.java
@@ -18,7 +18,7 @@ import com.google.instrumentation.common.NonThrowingCloseable;
 /**
  * Defines a scope of code where the given {@link Span} is in the current context. The scope is
  * exited when the object is closed and the previous context is restored. When the scope exits the
- * given {@link Span} will be ended using {@link Span#end}.
+ * given {@code Span} will be ended using {@link Span#end}.
  *
  * <p>Supports try-with-resource idiom.
  */
@@ -27,7 +27,7 @@ final class ScopedSpanHandle implements NonThrowingCloseable {
   private final NonThrowingCloseable withSpan;
 
   /**
-   * Creates a {@link ScopedSpanHandle}
+   * Creates a {@code ScopedSpanHandle}
    *
    * @param span The span that will be installed in the current context.
    * @param contextSpanHandler The handler that is used to interact with the current context.
@@ -38,12 +38,11 @@ final class ScopedSpanHandle implements NonThrowingCloseable {
   }
 
   /**
-   * Uninstalls the {@code Span} from the current context and ends it with default options. If the
-   * span has been already ended then {@link Span#end()} is a no-op.
+   * Uninstalls the {@code Span} from the current context and ends it by calling {@link Span#end()}.
    */
   @Override
   public void close() {
     withSpan.close();
-    span.end(EndSpanOptions.DEFAULT);
+    span.end();
   }
 }
--- a/core/src/main/java/com/google/instrumentation/trace/SpanBuilder.java
+++ b/core/src/main/java/com/google/instrumentation/trace/SpanBuilder.java
@@ -31,11 +31,10 @@ import javax.annotation.Nullable;
  *   private static final Tracer tracer = Tracer.getTracer();
  *   void doWork(Span parent) {
  *     // Create a Span as a child of the given parent.
- *     try (Span span = tracer.spanBuilder(parent, "MyChildSpan").startSpan()) {
- *       span.addAnnotation("my annotation");
- *       doSomeWork(span); // Manually propagate the new span down the stack.
- *     }
- *     // "span" will be ended here.
+ *     Span span = tracer.spanBuilder(parent, "MyChildSpan").startSpan();
+ *     span.addAnnotation("my annotation");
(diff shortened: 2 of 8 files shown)


core/src/main/java/com/google/instrumentation/trace/SpanBuilder.java
```diff
@@ -31,11 +31,10 @@
...
  *     // Create a Span as a child of the given parent.
- *     try (Span span = tracer.spanBuilder(parent, "MyChildSpan").startSpan()) {
- *       span.addAnnotation("my annotation");
- *       doSomeWork(span); // Manually propagate the new span down the stack.
- *     }
- *     // "span" will be ended here.
+ *     Span span = tracer.spanBuilder(parent, "MyChildSpan").startSpan();
+ *     span.addAnnotation("my annotation");
+ *     doSomeWork(span); // Manually propagate the new span down the stack.
+ *     span.end();  // Manually end the span.
```
sebright: Does this mean that it isn't necessary to end the span if an exception is thrown?

bogdandrutu: Good point @adriancole this is another reason why try-with-resource may work better because users do not have to think about exceptions.

adriancole:
>
> Good point @adriancole <https://github.com/adriancole> this is another reason why try-with-resource may work better because users do not have to think about exceptions.
>
> well, they would still want to note the exception regardless of
try/finally or try-with-resources (if closing a span) I think

core/src/main/java/com/google/instrumentation/trace/ScopedSpanHandle.java
```diff
@@ -38,12 +38,12 @@
   }
 
   /**
-   * Uninstalls the {@code Span} from the current context and ends it with default options. If the
-   * span has been already ended then {@link Span#end()} is a no-op.
+   * Uninstalls the {@code Span} from the current context and ends by calling {@link Span#end()}
+   * it with default options.
```
a-veitch: Move the "it" in the comment before "by".

bogdandrutu: Thanks. also removed the "with default options" because that is what Span#end does.

bogdandrutu: Typo. Thanks for catching this.

core/src/main/java/com/google/instrumentation/trace/SpanBuilder.java
bogdandrutu: Made the example exception safe.

sebright: Do you mean the example above?  I don't see the change.

core/src/test/java/com/google/instrumentation/trace/SpanTest.java
```diff
@@ -68,13 +68,6 @@ public void endCallsEndWithDefaultOptions() {
     verify(span).end(same(EndSpanOptions.DEFAULT));
   }
 
-  @Test
-  public void closeCallsEndWithDefaultOptions() {
-    Span span = Mockito.spy(new NoopSpan(spanContext, spanOptions));
-    span.close();
-    verify(span).end(same(EndSpanOptions.DEFAULT));
-  }
```
sebright: Isn't this test still relevant, even if the user has to call `close()`?

examples/src/main/java/com/google/instrumentation/trace/BasicTracing.java
```diff
@@ -19,9 +19,9 @@
   private static final Tracer tracer = Tracer.getTracer();
 
   private static void doWork() {
-    try (Span span = tracer.spanBuilder(null, "MyRootSpan").startSpan()) {
-      span.addAnnotation("This annotation is added directly to the span.");
-    }
+    Span span = tracer.spanBuilder(null, "MyRootSpan").startSpan();
+    span.addAnnotation("This annotation is added directly to the span.");
+    span.end();
```
sebright: Should this example use a try-finally?

examples/src/main/java/com/google/instrumentation/trace/MultiSpansTracing.java
```diff
@@ -19,13 +19,13 @@
...
-      }
-      rootSpan.addAnnotation("Annotation to the root Span after child is ended.");
-    }
+    Span rootSpan = tracer.spanBuilder(null, "MyRootSpan").startSpan();
+    rootSpan.addAnnotation("Annotation to the root Span before child is created.");
+    Span childSpan = tracer.spanBuilder(rootSpan, "MyChildSpan").startSpan();
+    childSpan.addAnnotation("Annotation to the child Span");
+    childSpan.end();
+    rootSpan.addAnnotation("Annotation to the root Span after child is ended.");
+    rootSpan.end();
```
sebright: Same here

core/src/main/java/com/google/instrumentation/trace/SpanBuilder.java
bogdandrutu: Forgot that there are 2 examples (one for the class comments one for the method).

PR #130, opened 2017-03-09
Fixes time units for rpc client latency

dinooliva:


core/src/main/java/com/google/instrumentation/stats/RpcConstants.java
```diff
@@ -53,11 +53,10 @@
           "/rpc/client/response_bytes",
           "Response MB",
           MeasurementUnit.create(6, bytes));
-  // TODO(dpo): verify this encoding
   public static final MeasurementDescriptor RPC_CLIENT_ROUNDTRIP_LATENCY =
       MeasurementDescriptor.create(
           "/rpc/client/roundtrip_latency",
-          "RPC roundtrip latency us",
+          "RPC roundtrip latency msec",
           MeasurementUnit.create(-6, seconds));
```
a-veitch: For millisec, shouldn't this be -3?

dinooliva:
Good catch - thanks!

adriancole:
Aww was hoping that was intentionally us!

dinooliva:
Lol - gRpc may add support for finer-grain at some point but these views reflect what they have now.

PR #131, opened 2017-03-09
Fixes units for rpc data sizes to correctly be bytes rather than MB

dinooliva:


core/src/main/java/com/google/instrumentation/stats/RpcConstants.java
```diff
@@ -251,15 +251,15 @@
   public static final IntervalViewDescriptor RPC_CLIENT_REQUEST_BYTES_INTERVAL_VIEW =
       IntervalViewDescriptor.create(
           "rpc client request_bytes",
-          "Minute and Hour stats for request in MBs",
+          "Minute and Hour stats for request in bytes",
```
sebright: I think this should be "request size in bytes".

core/src/main/java/com/google/instrumentation/stats/RpcConstants.java
```diff
@@ -251,15 +251,15 @@
...
           RPC_CLIENT_REQUEST_BYTES,
           IntervalAggregationDescriptor.create(Arrays.asList(MINUTE, HOUR)),
           Arrays.asList(RPC_CLIENT_METHOD));
 
   public static final IntervalViewDescriptor RPC_CLIENT_RESPONSE_BYTES_INTERVAL_VIEW =
       IntervalViewDescriptor.create(
           "rpc client response_bytes",
-          "Minute and Hour stats for response size MBs",
+          "Minute and Hour stats for response size bytes",
```
sebright: Should this be "in bytes"?

core/src/main/java/com/google/instrumentation/stats/RpcConstants.java
dinooliva: Done.

core/src/main/java/com/google/instrumentation/stats/RpcConstants.java
dinooliva: Done.

PR #132, opened 2017-03-11
Starts using AutoValue for classes in stats-core.

sebright:
This PR applies AutoValue to `MeasurementDescriptor` and `MeasurementUnit` and adds some tests.

### DIFF
--- a/core/src/main/java/com/google/instrumentation/stats/MeasurementDescriptor.java
+++ b/core/src/main/java/com/google/instrumentation/stats/MeasurementDescriptor.java
@@ -13,6 +13,8 @@
 
 package com.google.instrumentation.stats;
 
+import com.google.auto.value.AutoValue;
+
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
@@ -22,7 +24,8 @@ import java.util.List;
  *
  * <p>Note: MeasurementDescriptor names are {@link String}s with enforced restrictions.
  */
-public final class MeasurementDescriptor {
+@AutoValue
+public abstract class MeasurementDescriptor {
   public static final int MAX_LENGTH = StringUtil.MAX_LENGTH;
 
   /**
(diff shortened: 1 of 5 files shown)


dinooliva:
Before approving this change, I need to understand the implications of going to autovalue. A few questions that come to mind:
- How well supported is it by non-bazel build systems?
- Is there any performance overhead compared to the hand-written class?
- If we are going this route for MeasurementDescriptor, shouldn't we consistently do this for the other View-related data structures?

sebright:
I still need to debug the warning in the Bazel build.

sebright:
I added a commit to fix the build:

#### Suppress annotation processor warning in Bazel CI build.

This commit suppresses a javac "processing" warning that was enabled with
`--javacopt=-Xlint:all`.  The warning was `"warning: No processor claimed any of
these annotations: com.google.auto.value.AutoValue,javax.annotation.Nullable"`
According to the response in
https://groups.google.com/forum/#!topic/bazel-discuss/_R3A9TJSoPM, the best
solution is to ignore the warning.

pom.xml
```diff
@@ -170,6 +170,12 @@
       <version>3.0.0</version>
     </dependency>
     <dependency>
+      <groupId>com.google.auto.value</groupId>
+      <artifactId>auto-value</artifactId>
+      <version>1.3</version>
```
adriancole: 1.4 is latest (now :) )

sebright:
@adriancole Thanks! I upgraded to 1.4.

sebright:
@bogdandrutu @dinooliva Do you mind if I merge this?

core/src/main/java/com/google/instrumentation/stats/MeasurementDescriptor.java
```diff
@@ -22,47 +24,34 @@
  *
  * <p>Note: MeasurementDescriptor names are {@link String}s with enforced restrictions.
  */
-public final class MeasurementDescriptor {
+@AutoValue
+public abstract class MeasurementDescriptor {
   public static final int MAX_LENGTH = StringUtil.MAX_LENGTH;
 
+  MeasurementDescriptor() {}
```
bogdandrutu: Please add a comment:
// Disallow extending this class outside this package ?

sebright: I thought that that type of comment would become redundant if we did the same thing for all value classes.  Do you think it is still worth documenting within each class?

sebright: We discussed this in person and decided against the documentation.

sebright:
Thanks for the reviews!

PR #138, opened 2017-03-13
Starts building tests with Maven.

sebright:
I skipped building the `core` tests, because they currently create a circular dependency between the `core` and `core_impl` directories.


.travis.yml
```diff
@@ -63,6 +65,8 @@ script:
         case "$TRAVIS_JDK_VERSION" in
           "oraclejdk8")
             mvn verify -P dev -Dguava.version=$GUAVA_VERSION ;;
+          "openjdk6")
+            mvn verify -Dmaven.test.skip -Dguava.version=$GUAVA_VERSION ;;
           *)
```
bogdandrutu: This should be "oraclejdk7") now correct?

pom.xml
```diff
@@ -167,5 +169,29 @@
       <artifactId>jsr305</artifactId>
       <version>3.0.0</version>
     </dependency>
+    <dependency>
```
bogdandrutu: There is a bad interaction between mockito and junit because of org.hamcrest. See threads like: http://stackoverflow.com/questions/5569394/how-to-use-junit-and-hamcrest-together/21097967

    <dependency>
      <groupId>com.google.truth</groupId>
      <artifactId>truth</artifactId>
      <version>0.30</version>
      <scope>test</scope>
    </dependency>
    <dependency>
      <groupId>org.hamcrest</groupId>
      <artifactId>hamcrest-core</artifactId>
      <version>1.3</version>
      <scope>test</scope>
    </dependency>
    <dependency>
      <groupId>org.hamcrest</groupId>
      <artifactId>hamcrest-library</artifactId>
      <version>1.3</version>
      <scope>test</scope>
    </dependency>
    <dependency>
      <groupId>org.mockito</groupId>
      <artifactId>mockito-core</artifactId>
      <version>1.9.5</version>
      <scope>test</scope>
    </dependency>
    <dependency>
      <groupId>junit</groupId>
      <artifactId>junit</artifactId>
      <version>4.11</version>
      <scope>test</scope>
    </dependency>

.travis.yml
sebright: Yes. I thought it also was a good default, though.  Do you think it would be better to handle "oraclejdk7" as a separate case and fail in the default case?

pom.xml
sebright: Thanks!  I updated the build file.

.travis.yml
bogdandrutu: Yes. That way we are sure that once we test java9 for example we do not forget and use the default rule.

PR #141, opened 2017-03-15
Add benchmarks for HttpPropagationUtil using jmh.

bogdandrutu:


benchmark/src/main/java/com/google/instrumentation/trace/HttpPropagationUtilBenchmark.java
```diff
@@ -0,0 +1,109 @@
...
+  public SpanContext toFromHttpFormatSpanContext() {
+    return HttpPropagationUtil.fromHttpHeaderValue(
+        HttpPropagationUtil.toHttpHeaderValue(spanContext));
+  }
+
+  @Benchmark
+  @BenchmarkMode(Mode.SampleTime)
+  @OutputTimeUnit(TimeUnit.NANOSECONDS)
+  public String base16EncodeSpanContext() {
+    return BaseEncoding.base16().lowerCase().encode(spanContextBytes);
```
sebright: Why do the last three benchmarks test base-16 encoding and decoding?  Is that because it is part of the implementation of one of the other methods that operates on `SpanContext`?  I think it's worth a comment.

pom.xml
```diff
@@ -206,4 +206,24 @@
...
+    <profile>
+      <id>default</id>
+      <activation>
+        <activeByDefault>true</activeByDefault>
+      </activation>
+    </profile>
+    <profile>
+      <id>dev</id>
+      <modules>
+        <module>benchmark</module>
```
sebright: I think it would be clearer to put the default modules under the default profile, to avoid overriding them.

benchmark/src/main/java/com/google/instrumentation/trace/HttpPropagationUtilBenchmark.java
bogdandrutu: Done. I feel they are a bit redundant with the class comments (HttpPropagationUtil) but I added them.

pom.xml
bogdandrutu: Done.

PR #144, opened 2017-03-17
Update bazel to 0.4.5 and protobuf to 3.2.0 (same as gRPC).

bogdandrutu:


.bazelrc
```diff
@@ -112,21 +104,29 @@ build:error_prone_experimental_warnings \
...
+    # --javacopt='-Xep:TypeParameterShadowing:ERROR' \
+    # --javacopt='-Xep:LiteralClassName:ERROR' \
+    # --javacopt='-Xep:OperatorPrecedence:ERROR' \
+    # --javacopt='-Xep:MultiVariableDeclaration:ERROR' \
+    # --javacopt='-Xep:RemoveUnusedImports:ERROR' \
+    # --javacopt='-Xep:BoxedPrimitiveConstructor:ERROR' \
+    # --javacopt='-Xep:ImmutableEnumChecker:ERROR' \
+    # --javacopt='-Xep:WildcardImport:ERROR' \
+    # --javacopt='-Xep:ThrowsUncheckedException:ERROR' \
+    # --javacopt='-Xep:ArgumentParameterMismatch:ERROR' \
```
sebright: Could you please split the warning suppression changes into a separate commit, so that it will be easier to revert once we remove the warnings?

It looks like the these warnings come from the proto dependency, `com_google_protobuf_java`, not the generated classes, so I think they should go under a different comment.

.travis.yml
```diff
@@ -60,7 +60,7 @@ install:
 script:
   - case "$BUILD" in
       "BAZEL")
-        bazel test ... --javacopt=-Werror --javacopt=-Xlint:all --javacopt=-Xlint:-cast --javacopt=-Xlint:-deprecation --javacopt=-Xlint:-try --config=error_prone_warnings --config=error_prone_experimental_warnings ;;
+        bazel test ... --javacopt=-Werror --javacopt=-Xlint:all --javacopt=-Xlint:-cast --javacopt=-Xlint:-deprecation --javacopt=-Xlint:-try --javacopt=-Xlint:-rawtypes --javacopt=-Xlint:-serial --javacopt=-Xlint:-unchecked --config=error_prone_warnings --config=error_prone_experimental_warnings ;;
```
sebright: Could you please add a comment explaining why we need to suppress the three new warnings?

.bazelrc
bogdandrutu: Split them into a separate block with a different comment to allow us to keep track of all the commented warnings.

.travis.yml
bogdandrutu: Added a separate comment for these 3 warnings.

PR #154, opened 2017-03-19
Change the trace API to directly depend on the Context implementation.

bogdandrutu:
Other changes in this PR:
Enable maven tests for core (common and trace).
Make trace tests in core to compile with java6.

Reasons to change to directly use the io.grpc.Context API:
* We would need to have a different API that allows users to pass the generic context between threads and via Executors (similar with io.grpc.Context).
* io.grpc.Context is an API the Storage is decoupled and can be backed by any mechanism not necessary thread-local.
* Support for Deadline and Cancellation is a nice to have.
* Other products may already support propagation of the io.grpc.Context.
* It would have been very hard to make binaries work correctly when they have libraries that are using io.grpc.Context and libraries that are using something different.


core/pom.xml
```diff
@@ -76,4 +79,13 @@
       </build>
     </profile>
   </profiles>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.google.io</groupId>
+      <artifactId>shared</artifactId>
+      <version>0.1.0-SNAPSHOT</version>
```
adriancole: maybe leave an XML comment where this is and when it will be non-snapshot?

core/src/main/java/com/google/instrumentation/trace/Tracer.java
```diff
@@ -31,14 +31,21 @@
  * <p>Users may choose to use manual or automatic Context propagation. Because of that this class
  * offers APIs to facilitate both usages.
  *
+ * <p>The automatic context propagation is done using {@link io.grpc.Context} which is a gRPC
```
adriancole: too bad the package name of Context can't be changed from grpc, as might raise questions over time.

core/src/main/java/com/google/instrumentation/trace/Tracer.java
```diff
@@ -60,25 +73,39 @@
...
    */
   public static Tracer getTracer() {
     return INSTANCE;
   }
 
+  // Any provider that may be used for SpanFactory can be added here.
+  @VisibleForTesting
+  static SpanFactory loadSpanFactory(ClassLoader classLoader) {
+    try {
+      // Because of shading tools we must call Class.forName with the literal string name of the
```
adriancole: supernit. tighten sentence so it doesn't wrap :D

core/src/main/java/com/google/instrumentation/trace/Tracer.java
```diff
@@ -60,25 +73,39 @@
...
+      // class.
+      return Provider.createInstance(
+          Class.forName("com.google.instrumentation.trace.SpanFactoryImpl", true, classLoader),
+          SpanFactory.class);
+    } catch (ClassNotFoundException e) {
+      logger.log(Level.FINE, "Using default implementation for SpanFactory.", e);
```
adriancole: should fallback to noop be a fine level?

core/pom.xml
bogdandrutu: I made a change to show that is the current version for the library ${version}. Does that explain?

core/src/main/java/com/google/instrumentation/trace/Tracer.java
bogdandrutu: We are working on this. Please file an issue if you are concerned about this because it may help use to prioritize the work.

core/src/main/java/com/google/instrumentation/trace/Tracer.java
bogdandrutu: Yes. The expected behavior is that all the libraries like gRPC, google client APIs for cloud, etc will depend only on the API and the service owner will link the implementation.

If the service owner does not want tracing I think this should be fine, also all the tests for the libraries will print a warning message because they do not depend on the implementation.

core/src/main/java/com/google/instrumentation/trace/Tracer.java
bogdandrutu: Done.

core/src/test/java/com/google/instrumentation/trace/ContextUtilsTest.java
```diff
@@ -38,47 +33,55 @@ public void setUp() {
...
+    ;
+    NonThrowingCloseable ws = ContextUtils.withSpan(span);
+    try {
+      assertThat(ContextUtils.getCurrentSpan()).isSameAs(span);
+    } finally {
+      ws.close();
     }
-    assertThat(tracer.getCurrentSpan()).isNotSameAs(span);
+    assertThat(ContextUtils.getCurrentSpan()).isNull();
+    ;
```
adriancole: nit extra semi

core/pom.xml
```diff
@@ -76,4 +79,13 @@
       </build>
     </profile>
   </profiles>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.google.io</groupId>
+      <artifactId>shared</artifactId>
+      <version>${version}</version>
```
adriancole: `${project.version}` is more commonly used I think.. I didn't know there was a plain "version" property!

core/src/test/java/com/google/instrumentation/trace/ContextUtilsTest.java
bogdandrutu: Done. Thanks.

core/pom.xml
```diff
@@ -33,7 +42,10 @@
               Don't build the tests, because they currently create a circular
               dependency between the 'core' and 'core_impl' directories.
               -->
-              <skip>true</skip>
+              <skip>false</skip>
+              <testExcludes>
+                <exclude>**/stats/*.java</exclude>
+              </testExcludes>
```
bogdandrutu: Done.

core/src/main/java/com/google/instrumentation/trace/ContextUtils.java
```diff
@@ -14,40 +14,42 @@
...
-    return new WithSpan(span, GrpcTraceUtils.getContextSpanKey());
+  /**
+   * Enters the scope of code where the given {@link Span} is in the current context, and returns an
+   * object that represents that scope. The scope is exited when the returned object is closed.
+   *
+   * <p>Supports try-with-resource idiom.
+   *
+   * @param span The {@code Span} to be set to the current context.
+   * @return An object that defines a scope where the given {@code Span} will be set to the current
+   *     context.
```
bogdandrutu: Done. changed to present :)

core/src/main/java/com/google/instrumentation/trace/Tracer.java
```diff
@@ -60,25 +73,38 @@
...
+  }
+
   /**
-   * Returns the {@link Tracer} with the provided implementations for {@link ContextSpanHandler} and
-   * {@link SpanFactory}.
-   *
-   * <p>If no implementation is provided for any of the {@link Tracer} modules then no-op
-   * implementations will be used.
+   * Returns the {@link Tracer}. If no implementation is provided for any of the {@code Tracer}
+   * modules then no-op implementations will be used.
```
bogdandrutu: Updated.

core/src/test/java/com/google/instrumentation/trace/TracerTest.java
```diff
@@ -49,14 +52,42 @@ public void defaultGetCurrentSpan() {
 
   @Test(expected = NullPointerException.class)
   public void withSpan_NullSpan() {
-    try (NonThrowingCloseable ws = tracer.withSpan(null)) {}
+    NonThrowingCloseable ws = tracer.withSpan(null);
```
bogdandrutu: Done. Thanks.

core/src/test/java/com/google/instrumentation/trace/TracerTest.java
```diff
@@ -49,14 +52,42 @@ public void defaultGetCurrentSpan() {
...
+                    public void run() {
+                      assertThat(tracer.getCurrentSpan()).isSameAs(span);
+                    }
+                  });
+    } finally {
+      ws.close();
+    }
+    assertThat(tracer.getCurrentSpan()).isNotSameAs(span);
+    // When we run the runnable we will have the span in the current Context.
+    runnable.run();
```
bogdandrutu: Done.

core/src/main/java/com/google/instrumentation/trace/Tracer.java
adriancole: done https://github.com/google/instrumentation-java/issues/162

core/src/main/java/com/google/instrumentation/trace/Tracer.java
adriancole: thx for the explanation

PR #155, opened 2017-03-20
Cleanup pom.xml in our implementation packages.

bogdandrutu:
Use the current version for dependencies between different internal packages.
Avoid generating javadocs for implementation packages.

### DIFF
--- a/benchmarks/pom.xml
+++ b/benchmarks/pom.xml
@@ -16,11 +16,15 @@
     <version>0.1.0-SNAPSHOT</version>
   </parent>
 
+  <properties>
+    <maven.javadoc.skip>true</maven.javadoc.skip>
+  </properties>
+
   <dependencies>
     <dependency>
       <groupId>com.google.census</groupId>
       <artifactId>census-core</artifactId>
-      <version>0.1.0-SNAPSHOT</version>
+      <version>${project.version}</version>
     </dependency>
     <dependency>
       <groupId>org.openjdk.jmh</groupId>
--- a/benchmarks/pom.xml
+++ b/benchmarks/pom.xml
@@ -16,10 +16,6 @@
     <version>0.1.0-SNAPSHOT</version>
   </parent>
 
-  <properties>
-    <maven.javadoc.skip>true</maven.javadoc.skip>
-  </properties>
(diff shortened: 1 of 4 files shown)


benchmarks/pom.xml
```diff
@@ -16,11 +16,15 @@
     <version>0.1.0-SNAPSHOT</version>
   </parent>
 
+  <properties>
+    <maven.javadoc.skip>true</maven.javadoc.skip>
+  </properties>
+
```
sebright: Why are you controlling Javadoc generation with a property?  Do you still want to generate Javadoc sometimes, such as during CI?

benchmarks/pom.xml
```diff
@@ -16,11 +16,15 @@
...
   <dependencies>
     <dependency>
       <groupId>com.google.census</groupId>
       <artifactId>census-core</artifactId>
-      <version>0.1.0-SNAPSHOT</version>
+      <version>${project.version}</version>
```
sebright: Nice!

benchmarks/pom.xml
bogdandrutu: In my opinion we do not want to generate or export the implementation java documentation. We should generate and export the API documentation but the internal documentation is not necessary to be exported (I would say that can actually make our life harder because users may start to depend on the implementation rather than on the API.)

benchmarks/pom.xml
bogdandrutu: Thanks.

benchmarks/pom.xml
sebright: I agree that we shouldn't export Javadoc for anything but the public APIs.  I had enabled Javadoc generation for everything because we have Javadoc-style comments in many classes, and I wanted to check that they were correctly formatted in CI.  If we continue to check the formatting, I think that it would be better to just put the `maven-javadoc-plugin` section under the `dev` profile, to be more consistent with the other CI-only checks.

adriancole:
ps source and javadoc jars are required for maven central (even if the jars
are empty iirc!). we usually have a profile that opts out of building these
unless publishing snapshot or release

PR #156, opened 2017-03-20
Add sample code for reading proto binaries on descriptor constants during run time.

songy23:
This change demonstrates how to read and parse binary text-proto files and retrieve RPC constants during run time. This should not be merged until we have finally decided where to place the proto binary file (rpc_constants.pb).

### DIFF
--- a/core/src/main/java/com/google/instrumentation/stats/RpcConstantsFromProtoBinary.java
+++ b/core/src/main/java/com/google/instrumentation/stats/RpcConstantsFromProtoBinary.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2016, Google Inc.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.instrumentation.stats;
+
+import com.google.common.io.Files;
+import com.google.instrumentation.stats.proto.CensusProto;
+import com.google.instrumentation.stats.proto.ViewDescriptorConstantsProto;
+import com.google.instrumentation.stats.proto.ViewDescriptorConstantsProto.ViewDescriptorConstants;
+import java.io.File;
+import java.io.IOException;
+import java.util.List;
+
+/**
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/google/instrumentation/stats/RpcConstantsFromProtoBinary.java
```diff
@@ -0,0 +1,56 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.instrumentation.stats;
+
+import com.google.common.io.Files;
```
adriancole: I was trying to follow the dependency changes.. does this imply guava is a strict dependency of instrumentation-java? (sorry if noticing this late)

core/src/test/java/com/google/instrumentation/stats/RpcConstantsFromProtoBinaryTest.java
```diff
@@ -0,0 +1,26 @@
...
+@RunWith(JUnit4.class)
+public class RpcConstantsFromProtoBinaryTest {
+  private List<CensusProto.MeasurementDescriptor> measurementDescriptorList =
+      RpcConstantsFromProtoBinary.getMeasurementDescriptorList();
+  private List<CensusProto.ViewDescriptor> viewDescriptorList =
+      RpcConstantsFromProtoBinary.getViewDescriptorList();
+
+  @Test
+  public void testConstants() {
+    assertThat(measurementDescriptorList).isNotEmpty();
```
adriancole: maybe test for at least old constants? That would not imply maintenance (as you are only spot-checking some vs all) allowing a more reasonable sanity check.

alternatively, you could do a qualitative check like that the names are all non-empty

core/src/main/java/com/google/instrumentation/stats/RpcConstantsFromProtoBinary.java
songy23: Sorry for the confusion, this PR is intended to demonstrate one way to read pre-defined constants from protobuf binaries. With this approach we need to add guava and proto as dependency to stats_core. However after talking with Dino, we find this is not a desirable approach, because for some reason stats_core cannot directly depend on proto. I'll try not to add additional dependencies to stats_core and update this PR later.

core/src/test/java/com/google/instrumentation/stats/RpcConstantsFromProtoBinaryTest.java
songy23: That's a good advice, I'll definitely add some tests to verify the constants in the list.

PR #158, opened 2017-03-20
Add support for gradle to build the instrumentation-java.

bogdandrutu:


build.gradle
```diff
@@ -0,0 +1,137 @@
+// TODO(bdrutu): Default build directory is build but on MacOs with a case-insensitive
+// File System would conflict with Bazel BUILD file.  Remove this and update the
+// .gitignore entry once Bazel is removed.
+buildDir = 'build_gradle'
```
sebright: Should build_gradle be added to .gitignore?

build.gradle
```diff
@@ -0,0 +1,137 @@
...
+    sourceCompatibility = 1.6
+    targetCompatibility = 1.6
+
+    repositories {
+        mavenCentral()
+        mavenLocal()
+    }
+
+    [compileJava, compileTestJava].each() {
+        it.options.compilerArgs += ["-Xlint:all", "-Xlint:-try", "-Xlint:-path"]
```
sebright: Why do we need to suppress path warnings?

build.gradle
```diff
@@ -0,0 +1,137 @@
...
+                libraries.mockito,
+                libraries.truth
+
+        // The ErrorProne plugin defaults to the latest, which would break our
+        // build if error prone releases a new version with a new check
+        errorprone 'com.google.errorprone:error_prone_core:2.0.15'
+    }
+
+    checkstyle {
+        configFile = file("$rootDir/checkstyle.xml")
```
sebright: Could we use the default Checkstyle Google style checks?

.travis.yml
```diff
@@ -64,6 +72,8 @@ script:
   - case "$BUILD" in
       "BAZEL")
         bazel test ... --javacopt=-Werror --javacopt=-Xlint:all --javacopt=-Xlint:-cast --javacopt=-Xlint:-deprecation --javacopt=-Xlint:-try --javacopt=-Xlint:-rawtypes --javacopt=-Xlint:-serial --javacopt=-Xlint:-unchecked --config=error_prone_warnings --config=error_prone_experimental_warnings ;;
+      "GRADLE")
+        ./gradlew check :instrumentation-all:jacocoTestReport --stacktrace ;;
```
bogdandrutu: Because here we want to have jacoco for code coverage. https://github.com/jacoco/jacoco

https://docs.gradle.org/current/userguide/java_plugin.html#sec:java_tasks
assemble & check are the minimum to make sure code compiles & passes tests (and they are part of the java plugin).

The :instrumentation-all:jacocoTestReport makes sure that checkstyle, and all the plugins that we have run and produce the reports (if any).

.travis.yml
```diff
@@ -77,3 +87,22 @@ script:
...
     esac
+
+after_success:
+  - case "$BUILD" in
+      "GRADLE")
+        ./gradlew :instrumentation-all:coveralls --stacktrace ;
+   	bash <(curl -s https://codecov.io/bash) ;;
+       *)
+        echo "No GRADLE" ;
+        exit 0 ;;
```
bogdandrutu: I don't think so. It was a bit more clear to me. Removed.

build.gradle
bogdandrutu: Removed. Not a problem here because we do not build anything in the root directory that can conflict with the BAZEL file.

build.gradle
bogdandrutu: No. It is a bit hard to configure severity level and ignoring things like SingleJavadocLine. gRPC did the same thing copying the checkstyle from the repository. So this should be fine. @zhangkun83 is this the same thing you did correct?

build.gradle
bogdandrutu: Was an error in one of the jar in the package com.google.errorprone:error_prone_core:2.0.15. Updated to com.google.errorprone:error_prone_core:2.0.19 works fine now without removing the path warning.

build.gradle
```diff
@@ -0,0 +1,132 @@
+buildscript {
+    repositories {
+        mavenCentral()
+        mavenLocal()
+        maven {
+            url "https://plugins.gradle.org/m2/"
+        }
+    }
+    dependencies {
+        classpath 'com.google.gradle:osdetector-gradle-plugin:1.4.0'
```
zhangkun83: It doesn't seem you need this plugin.

build.gradle
```diff
@@ -0,0 +1,132 @@
...
+    apply plugin: 'maven'
+    apply plugin: 'idea'
+    apply plugin: 'java'
+    apply plugin: "signing"
+    apply plugin: "jacoco"
+    // TODO(bdrutu): enable all checks.
+    apply plugin: "net.ltgt.errorprone"
+
+    group = "com.google.instrumentation"
+    version = "0.1.0-SNAPSHOT" // CURRENT_INSTRUMENTATION_VERSION
```
zhangkun83: The latest release is 0.3.0. It should be either 0.3.1-SNAPSHOT or 0.4.0-SNAPSHOT.

.travis.yml
sebright: Thanks

build.gradle
bogdandrutu: Done.

build.gradle
bogdandrutu: Done.

bogdandrutu:
PTAL.

PR #159, opened 2017-03-21
Fix a bug in the Span where the returned options was not unmodifiable.

bogdandrutu:
Remove forgotten method from a previous PR.
Add check for consistency between SpanContext sampled and RECORD_EVENTS.

### DIFF
--- a/core/src/main/java/com/google/instrumentation/trace/Span.java
+++ b/core/src/main/java/com/google/instrumentation/trace/Span.java
@@ -48,7 +48,8 @@ public abstract class Span {
     RECORD_EVENTS;
   }
 
-  private static final EnumSet<Options> DEFAULT_OPTIONS = EnumSet.noneOf(Options.class);
+  private static final Set<Options> DEFAULT_OPTIONS =
+      Collections.unmodifiableSet(EnumSet.noneOf(Options.class));
 
   /**
    * Creates a new {@code Span}.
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/google/instrumentation/trace/Span.java
```diff
@@ -46,17 +48,25 @@
...
    *     options will be set.
    * @throws NullPointerException if context is {@code null}.
+   * @throws IllegalArgumentException is the {@code SpanContext} is sampled but no RECORD_EVENTS
+   *     options.
    */
   protected Span(SpanContext context, @Nullable EnumSet<Options> options) {
     this.context = checkNotNull(context, "context");
-    this.options = options == null ? EnumSet.noneOf(Options.class) : EnumSet.copyOf(options);
+    this.options =
+        Collections.unmodifiableSet(options == null ? DEFAULT_OPTIONS : EnumSet.copyOf(options));
```
sebright: Suggestion: I think it's simpler to use Collections.unmodifiableSet to create DEFAULT_OPTIONS and then have:
```java
 this.options =
         options == null ? DEFAULT_OPTIONS : Collections.unmodifiableSet(EnumSet.copyOf(options));
```

core/src/main/java/com/google/instrumentation/trace/Span.java
```diff
@@ -46,17 +48,25 @@
...
+  private static final EnumSet<Options> DEFAULT_OPTIONS = EnumSet.noneOf(Options.class);
+
   /**
    * Creates a new {@code Span}.
    *
    * @param context The context associated with this {@code Span}.
    * @param options The options associated with this {@code Span}. If {@code null} then default
    *     options will be set.
    * @throws NullPointerException if context is {@code null}.
+   * @throws IllegalArgumentException is the {@code SpanContext} is sampled but no RECORD_EVENTS
```
a-veitch: s/is/if

core/src/main/java/com/google/instrumentation/trace/Span.java
```diff
@@ -46,17 +48,25 @@
...
+   *     options.
    */
   protected Span(SpanContext context, @Nullable EnumSet<Options> options) {
     this.context = checkNotNull(context, "context");
-    this.options = options == null ? EnumSet.noneOf(Options.class) : EnumSet.copyOf(options);
+    this.options =
+        Collections.unmodifiableSet(options == null ? DEFAULT_OPTIONS : EnumSet.copyOf(options));
+    checkArgument(
+        !context.getTraceOptions().isSampled() || (this.options.contains(Options.RECORD_EVENTS)),
+        "Span without record_events but sampled.");
```
a-veitch: Think message would be slightly clearer as "Span is sampled, but does not have RECORD_EVENTS set."

core/src/main/java/com/google/instrumentation/trace/Span.java
bogdandrutu: Thanks.

core/src/main/java/com/google/instrumentation/trace/Span.java
bogdandrutu: Done.

core/src/main/java/com/google/instrumentation/trace/Span.java
bogdandrutu: Done.

bogdandrutu:
Added tests as well.

PR #161, opened 2017-03-22
Add EventQueue classes

a-veitch:
Functionality to enable doing background updates of instrumentation data (e.g. trace annotations or usage measurements) so as to minimally impact foreground task latency.


core_impl/src/main/java/com/google/instrumentation/common/EventQueue.java
```diff
@@ -0,0 +1,165 @@
...
+    @Override
+    public void onEvent(InstrumentationEvent event, long sequence, boolean endOfBatch) {
+      event.getEntry().process();
+    }
+  }
+
+  // The single instance of the class.
+  private static final EventQueue eventQueue = new EventQueue();
+  // The event queue is built on this {@link Disruptor}.
+  private final Disruptor<InstrumentationEvent> disruptor;
```
adriancole: might want to somehow make disruptor optional. For example having the event queue type abstract and/or use shade to repackage disruptor. in my experience, libraries that are to be used at the lowest level of the platform easily conflict with app deps.

bogdandrutu: Our model is that every library (gRPC for example) will link only the API (which should not depend on any fancy library except jsr305 and guava(19).

The implementation will be linked only by the service owner. Which agree can still conflict with other libraries but it is easier to shade the disruptor (or any other dependency) when it is in only one jar.

I think having a interface makes less sense (or maybe I do not understand your point) when the implementation jar will be linked only once.

adriancole: sorry I wasn't meaning about an extra interface, just a base implementation (even if abstract for how to implement a part of queuing) which could allow more code reuse without a dep like this. basically batteries included but swappable (which might not be the approach you are looking for based on what you said).

adriancole: interestingly this type isn't used, yet (and the core-impl package is otherwise little code), so maybe just keep in mind that as the impl part grows, there will be parts others can harvest. guava is usually used by others pretty well understood etc. disruptor has a relatively low change frequency at the moment, and likely stable enough to have a core dep, but it is less understood.

Probably only async folks (who also use disruptor) who want to use this impl will need to make a custom build which shades it and publishes a variant which doesn't conflict. We can always wait until they complain (which may never happen)

build.gradle
```diff
@@ -69,6 +69,7 @@ subprojects {
         grpcContextVersion = '1.1.2'
 
         libraries = [
+	        disruptor    : 'com.lmax:disruptor:3.3.6',
```
bogdandrutu: Indentation?

core_impl/build.gradle
```diff
@@ -3,6 +3,7 @@ description = 'Instrumentation Core Impl'
 dependencies {
     compile project(':instrumentation-core'),
             project(':proto'),
+	    libraries.disruptor,
```
bogdandrutu: indentation?

core_impl/src/main/java/com/google/instrumentation/common/EventQueue.java
```diff
@@ -0,0 +1,165 @@
...
+    disruptor = new Disruptor<InstrumentationEvent>(new InstrumentationEventFactory(), bufferSize,
+        Executors.newSingleThreadExecutor(), ProducerType.MULTI, new SleepingWaitStrategy());
+    disruptor.handleEventsWith(new InstrumentationEventHandler());
+    disruptor.start();
+    ringBuffer = disruptor.getRingBuffer();
+  }
+
+  /**
+   * Get the {@link EventQueue} instance.
+   * @return the singleton {@code EventQueue} instance.
```
bogdandrutu: Please use the formatter I suggested it will add an extra line here.

core_impl/src/main/java/com/google/instrumentation/common/EventQueue.java
```diff
@@ -0,0 +1,165 @@
...
+    // ensures that the event handler can take unsynchronized actions whenever possible.
+    disruptor = new Disruptor<InstrumentationEvent>(new InstrumentationEventFactory(), bufferSize,
+        Executors.newSingleThreadExecutor(), ProducerType.MULTI, new SleepingWaitStrategy());
+    disruptor.handleEventsWith(new InstrumentationEventHandler());
+    disruptor.start();
+    ringBuffer = disruptor.getRingBuffer();
+  }
+
+  /**
+   * Get the {@link EventQueue} instance.
```
bogdandrutu: s/Get/Returns

core_impl/src/test/java/com/google/instrumentation/common/EventQueueTest.java
```diff
@@ -0,0 +1,108 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/**
+ * Unit tests for {@link EventQueue}.
+ */
+@RunWith(JUnit4.class)
+public class EventQueueTest {
+  /**
+   * Simple class to use that keeps an incrementing counter. Will fail with an assertion if
```
bogdandrutu: Don't need javadoc format for private class.

core_impl/src/test/java/com/google/instrumentation/common/EventQueueTest.java
```diff
@@ -0,0 +1,108 @@
...
+    private int count;
+    private volatile long id;  // stores thread ID used in first increment operation.
+
+    public Counter() {
+      count = 0;
+      id = -1;
+    }
+
+    /**
+     * Increment counter by 1. Will fail in assertion if multiple different threads are used
```
bogdandrutu: s/Increment/Increments

core_impl/src/main/java/com/google/instrumentation/common/EventQueue.java
```diff
@@ -0,0 +1,165 @@
...
+ *</pre>
+ */
+
+@ThreadSafe
+public final class EventQueue {
+  // An event in the {@link EventQueue}. Just holds a reference to an EventQueueEntry.
+  private static final class InstrumentationEvent {
+    private EventQueueEntry entry = null;
+
+    // Set the EventQueueEntry associated with this InstrumentationEvent.
```
bogdandrutu: s/Set/Sets

core_impl/src/main/java/com/google/instrumentation/common/EventQueue.java
```diff
@@ -0,0 +1,165 @@
...
+ *     int measurement = doSomeWork();
+ *     // Instead of calling update() directly, create an event to do the processing, and insert
+ *     // it into the EventQueue. It will be processed in a background thread, and doSomething()
+ *     // can return immediately.
+ *     EventQueue.getInstance.enqueue(new SomeClassUpdateEvent(this, measurement));
+ *   }
+ * }
+ *</pre>
+ */
+
```
bogdandrutu: Extra empty line. Please format the code using https://github.com/google/google-java-format (if you don't use an IDE that you can configure with this formatter). For the moment use it only for the files you added.

core_impl/src/main/java/com/google/instrumentation/common/EventQueue.java
```diff
@@ -0,0 +1,165 @@
...
+  // An event in the {@link EventQueue}. Just holds a reference to an EventQueueEntry.
+  private static final class InstrumentationEvent {
+    private EventQueueEntry entry = null;
+
+    // Set the EventQueueEntry associated with this InstrumentationEvent.
+    void setEntry(EventQueueEntry entry) {
+      this.entry = entry;
+    }
+
+    // Get the EventQueueEntry associated with this InstrumentationEvent.
```
bogdandrutu: s/Get/Returns

core_impl/src/main/java/com/google/instrumentation/common/EventQueue.java
```diff
@@ -0,0 +1,165 @@
...
+  // Factory for InstrumentationEvent.
+  private static final class InstrumentationEventFactory
+      implements EventFactory<InstrumentationEvent> {
+    @Override
+    public InstrumentationEvent newInstance() {
+      return new InstrumentationEvent();
+    }
+  }
+
+  /**
```
bogdandrutu: Don't need javadoc for private class. We decided to do it only for at least package-protected classes/methods.

core_impl/src/main/java/com/google/instrumentation/common/EventQueue.java
a-veitch: More classes that use it coming soon. Re shading, lets wait and see; I suspect we may want to do this, but if so, will do a separate PR.

core_impl/src/test/java/com/google/instrumentation/common/EventQueueTest.java
a-veitch: Done.

core_impl/src/test/java/com/google/instrumentation/common/EventQueueTest.java
a-veitch: Done

core_impl/src/main/java/com/google/instrumentation/common/EventQueue.java
a-veitch: Done.

core_impl/src/main/java/com/google/instrumentation/common/EventQueue.java
a-veitch: Done.

core_impl/src/main/java/com/google/instrumentation/common/EventQueue.java
a-veitch: Done.

core_impl/src/main/java/com/google/instrumentation/common/EventQueue.java
a-veitch: Done.

core_impl/src/main/java/com/google/instrumentation/common/EventQueue.java
a-veitch: Done.

core_impl/src/main/java/com/google/instrumentation/common/EventQueue.java
a-veitch: Done.

core_impl/build.gradle
a-veitch: Another TAB :-(

build.gradle
a-veitch: Looks like a TAB snuck in there.

a-veitch:
PTAL

adriancole:
SGTM

PR #166, opened 2017-03-23
Remove maven support for instrumentation-java.

bogdandrutu:
Add to animalsniffer to gradle and force signature for 1.6 in core and core-impl.
Build with java7/8 gradle.
Build with java8 bazel.


build.gradle
```diff
@@ -36,7 +42,9 @@ subprojects {
     [compileJava, compileTestJava].each() {
         it.options.compilerArgs += ["-Xlint:all", "-Xlint:-try"]
         it.options.encoding = "UTF-8"
-        it.options.compilerArgs += ["-Werror"]
+        if (JavaVersion.current().isJava8Compatible()) {
+            it.options.compilerArgs += ["-Werror"]
```
sebright: Why doesn't -Werror work with Java 7?

bogdandrutu: Added a todo to fix and re-enable.

bogdandrutu:
PTAL.

adriancole:
sniff sniff.. bye maven, you will be missed.

(seriously, this is fine.. there's a lot of good reasons to use gradle for
this ecosystem)

PR #171, opened 2017-03-24
Rename Labels to Attributes.

bogdandrutu:
I did git mv on all the files, but don't know why git decided to not see all of them as moved files.


core/src/main/java/com/google/instrumentation/trace/AttributeValue.java
```diff
@@ -19,38 +19,39 @@
 import javax.annotation.concurrent.Immutable;
 
 /**
- * A class that represents all the possible values for a label in {@link Labels}.. A label can have
- * 3 types of values: {@link String}, {@link Boolean} or {@link Long}.
+ * A class that represents all the possible values for a attribute in {@link Attributes}. A
```
a-veitch: s/a attribute/an attribute/

sebright:
> I did git mv on all the files, but don't know why git decided to not see all of them as moved files.

git only infers renames from the similarity of files, so `git mv` is just a shortcut for removing and then adding a file: https://git.wiki.kernel.org/index.php/GitFaq#Why_does_Git_not_.22track.22_renames.3F .  I think that git would be more likely to show files as moved if you moved them and edited them in separate commits.

PR #172, opened 2017-03-24
Add Parser for parsing view constants proto binary.

songy23:
This PR is dependent on PR https://github.com/google/instrumentation-proto/pull/24.

Since stats-core cannot depend on proto, I used an independent library util to do the translation (from proto binary to POJOs). Currently I'm assuming that we could separate the RPC constants from stats-core, however if those constants absolutely need to reside in stats-core, I'll try other alternative approaches (converting to JSON, etc.).

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -3,6 +3,7 @@ include ":instrumentation-java-all"
 include ":instrumentation-java-benchmarks"
 include ":instrumentation-java-core"
 include ":instrumentation-java-core-impl"
+include ":instrumentation-java-core-util"
 include ":proto"
 include ":shared"
 
@@ -10,6 +11,7 @@ project(':instrumentation-java-all').projectDir = "$rootDir/all" as File
 project(':instrumentation-java-benchmarks').projectDir = "$rootDir/benchmarks" as File
 project(':instrumentation-java-core').projectDir = "$rootDir/core" as File
 project(':instrumentation-java-core-impl').projectDir = "$rootDir/core_impl" as File
+project(':instrumentation-java-core-util').projectDir = "$rootDir/util" as File
 project(':proto').projectDir = "$rootDir/proto" as File
 project(':shared').projectDir = "$rootDir/shared" as File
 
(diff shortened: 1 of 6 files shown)


songy23:
> What has changed is just how these constants are initialized, which is just implementation detail and doesn't change the nature of the API.
The proto constants will server as a single source of truth, although for Java this doesn't make much changes, they could be reused for other language platforms.

> You still have hard-coded constants. 
I believe this PR is meant to provide a util for reading and initializing proto constants, and RPC constants is a special use case on how we define constants in proto and read it into Java objects. Later other users will be able to define and use constants in the same way as RPC constants. 

> The constants also have to call out the types of the view descriptors, which goes against the idea of defining everything in the proto.
In addition, in fact we do have proto definitions on measure and view descriptors, but to use those proto definitions we need to add instrumentation-proto as a dependency. Based on our previous conversation, I'm assuming gRPC doesn't want to depend on instrumentation-proto, but can depend on stats-core. Thus I translate the proto-generated descriptor objects into those data structures defined in stats-core.

zhangkun83:
> Based on our previous conversation, I'm assuming gRPC doesn't want to depend on instrumentation-proto, but can depend on stats-core. Thus I translate the proto-generated descriptor objects into those data structures defined in stats-core.

Right, and gRPC nowadays can depend on RpcConstants safely because it's proto-free. Since RpcConstantsFromProto depends on ViewConstantsProtoBinaryParser which depends on proto, it's no longer proto-free. It is backward without any gain.

I discussed with @ejona86 about the implications of allowing Census to have proto dependency. Even though we can split the Census dependency out of gRPC core, it will still be a hassle for both gRPC and Census team, because both of us would have to provide separate artifacts for Android and non-Android uses. So it's preferable to avoid proto dependency in our libraries.

Since all you want is the proto to be the source of truth, how about creating a code-generator that generates proto-free RpcConstants from the textpb as part of your build process? The constant names can be derived from the names from the proto, which will be different from the current constant names, but I can update gRPC code to accommodate that.

songy23:
Thanks for the clarification, I see the problem here. I'll check to see if and how to generate those constants during build-time.

PR #173, opened 2017-03-27
Add Annotation and Link as classes in the trace API.

bogdandrutu:
Change Span to accept Links and Annotations.

This change allows to expose later an interface to export SpanData without depending on protos which is mandatory when using Stackdriver tracer.

### DIFF
--- a/core/src/main/java/com/google/instrumentation/trace/Attributes.java
+++ b/core/src/main/java/com/google/instrumentation/trace/Attributes.java
@@ -28,6 +28,10 @@ import javax.annotation.concurrent.Immutable;
  */
 @Immutable
 public final class Attributes {
+  /** Empty {@code Attributes} value. */
+  public static final Attributes EMPTY =
+      new Attributes(Collections.<String, AttributeValue>emptyMap());
+
   // This object is an immutable Map.
   private final Map<String, AttributeValue> attributes;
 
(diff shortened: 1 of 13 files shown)


core/src/main/java/com/google/instrumentation/trace/Attributes.java
```diff
@@ -28,6 +28,10 @@
  */
 @Immutable
 public final class Attributes {
```
adriancole: no auto-value?

core/src/main/java/com/google/instrumentation/trace/Span.java
```diff
@@ -73,53 +73,60 @@ protected Span(SpanContext context, @Nullable EnumSet<Options> options) {
...
    */
   public abstract void addAttributes(Attributes attributes);
 
   /**
    * Adds an annotation to the {@code Span}.
    *
-   * @param description The description of the annotation time event.
+   * @param description the description of the annotation time event.
    */
   public abstract void addAnnotation(String description);
```
adriancole: is this the same as attribute with a description and empty attributes? If so, you could make this a concrete, but overridable method.

core/src/main/java/com/google/instrumentation/trace/Attributes.java
bogdandrutu: I think public API should not be auto-generated.

core/src/main/java/com/google/instrumentation/trace/Span.java
bogdandrutu: I can do that separately.

PR #174, opened 2017-03-28
Add SpanData and TimedEvent classes.

bogdandrutu:
Add toString methods for Annotation and Link.
The SpanData class si the main class that will expose Span's data outside the internal library.

### DIFF
--- a/core_impl/src/main/java/com/google/instrumentation/trace/SpanData.java
+++ b/core_impl/src/main/java/com/google/instrumentation/trace/SpanData.java
@@ -1,0 +1,169 @@
+/*
+ * Copyright 2017, Google Inc.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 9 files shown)


core_impl/src/main/java/com/google/instrumentation/trace/SpanData.java
```diff
@@ -0,0 +1,169 @@
...
+import com.google.instrumentation.common.Timestamp;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.Immutable;
+
+/** Immutable representation of all data collected by the {@link Span} class. */
+@Immutable
+public final class SpanData {
```
adriancole: suppose this would be an input object zipkin folks could convert from similar to how we do in htrace.. 
https://github.com/apache/incubator-htrace/tree/master/htrace-zipkin

bogdandrutu: Yes that is a possible case.

PR #175, opened 2017-03-28
Add build rule for examples to generate executable binaries.

bogdandrutu:
Fix javadoc errors in examples.
Move all the examples in a different package than the implementation to avoid using (by mistake) package-protected methods that others cannot use.

### DIFF
--- a/examples/src/main/java/com/google/instrumentation/examples/trace/BasicScopedTracing.java
+++ b/examples/src/main/java/com/google/instrumentation/examples/trace/BasicScopedTracing.java
@@ -11,21 +11,30 @@
  * limitations under the License.
  */
 
-package com.google.instrumentation.trace;
+package com.google.instrumentation.examples.trace;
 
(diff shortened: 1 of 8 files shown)


examples/src/main/java/com/google/instrumentation/examples/trace/BasicScopedTracing.java
```diff
@@ -11,12 +11,14 @@
...
-package com.google.instrumentation.trace;
+package com.google.instrumentation.examples.trace;
 
 import com.google.instrumentation.common.NonThrowingCloseable;
+import com.google.instrumentation.trace.Span;
+import com.google.instrumentation.trace.Tracer;
 
 /**
- * Example showing how to create a {@link Span} using {@link ScopedSpan}, install it in the current
+ * Example showing how to create a {@link Span} using scoped Span}, install it in the current
```
sebright: There is an extra right bracket.

PR #182, opened 2017-03-29
Mention Windows build command and git history check in CONTRIBUTING.md.

sebright:


CONTRIBUTING.md
```diff
@@ -54,9 +54,15 @@ We also follow these project-specific guidelines:
   constructors to all AutoValue classes to prevent classes in other packages
   from extending them.
 
+## Building instrumentation-java
+
+Run `./gradlew clean assemble check` on OS X or Linux, or run
```
a-veitch: Note, you can use 'build' instead of 'assemble check'. I also mostly skip the 'clean' step, except for a last check before getting review.

sebright: I just copied the command from .travis.yml.  What is the difference between `build` and `assemble check`?

a-veitch: AFAICT they do exactly the same actions (I ran both, compared output). But there may be some difference in the future (?) so probably better to keep the `assemble check` explicit.

sebright:
Thanks!

PR #186, opened 2017-03-31
Remove the attributes class and use directly a map.

bogdandrutu:
The reason is that before we were forced to do an extra copy of the map when we implement AddAttributes (once in the Attributes class and once in the implementation).


adriancole:
Sgtm

core/src/main/java/com/google/instrumentation/trace/AttributeValue.java
```diff
@@ -13,30 +13,51 @@
...
+
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Objects;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 
 /**
- * A class that represents all the possible values for an attribute in {@link Attributes}. A
- * attribute can have 3 types of values: {@code String}, {@code Boolean} or {@code Long}.
+ * A class that represents all the possible values for an attribute . A attribute can have 3 types
```
a-veitch: grammar nit: s/A attribute/An attribute/

core/src/test/java/com/google/instrumentation/trace/BlankSpanTest.java
```diff
@@ -30,13 +32,13 @@ public void hasInvalidContextAndDefaultSpanOptions() {
 
   @Test
   public void doNotCrash() {
+    Map<String, AttributeValue> attributes = new HashMap<String, AttributeValue>();
+    attributes.put(
```
a-veitch: Maybe add one test with multiple attributes, different types.

bogdandrutu:
Updated. Waiting for tests to pass.

PR #187, opened 2017-03-31
Change TraceOptions to be little-endian ordered.

bogdandrutu:


adriancole:
Maybe explain why little vs big?

core/src/main/java/com/google/instrumentation/trace/HttpPropagationUtil.java
```diff
@@ -25,24 +25,24 @@
...
- *       </ul>
+ * <li>Header name: Trace-Context
+ * <li>Header value: base16(&lt;version_id&gt;)&lt;version_format&gt;
+ * <li>version_id: 1-byte representing the version id.
+ * <li>For version_id = 0:
+ *     <ul>
+ *     <li>version_format: base16(&lt;trace-id&gt;&lt;span-id&gt;&lt;trace-options&gt;)
+ *     <li>trace-id: 16-byte array representing the trace_id.
+ *     <li>span-id: 8-byte array representing the span_id.
+ *     <li>trace-options: 4-byte array representing the trace_options.
```
a-veitch: Should document that this is in little-endian order, if represented as an int.

core/src/main/java/com/google/instrumentation/trace/HttpPropagationUtil.java
```diff
@@ -25,24 +25,24 @@
...
+ *     <li>Valid value example:
+ *         <ul>
+ *         <li>"00404142434445464748494A4B4C4D4E4F616263646566676800000001"
+ *         <li>version_id = 0;
+ *         <li>trace_id = {64, 65, 66, 67, 68, 69, 70, 71, 72, 73, 74, 75, 76, 77, 78, 79}
+ *         <li>span_id = {97, 98, 99, 100, 101, 102, 103, 104};
+ *         <li>trace_options = {0, 0, 0, 1} == 1 &lt;&lt; 24;
```
a-veitch: maybe change this to just {1, 0, 0, 0} and drop the shift; it's not needed if you have the little-endian doc above.

a-veitch:
> Maybe explain why little vs big?

Because most of the worlds systems are little-endian.

core/src/main/java/com/google/instrumentation/trace/HttpPropagationUtil.java
bogdandrutu: Done.

core/src/main/java/com/google/instrumentation/trace/HttpPropagationUtil.java
bogdandrutu: Done.

adriancole:
>
> Because most of the worlds systems are little-endian.
>
:) ok I will use that answer too when someone miscodes this

PR #189, opened 2017-04-04
Only load one stats class with reflection.

sebright:
Previously, the `Stats` class used reflection to instantiate subclasses for both
`StatsManager` and `StatsContextFactory`.  That design forced the `StatsManager`
and `StatsContextFactory` subclasses to only interact through their public APIs.

This commit changes `Stats` so that it only instantiates `StatsManager`.  The
`StatsManager` instance then creates the one `StatsContextFactory` instance by
calling a constructor.  The main difference is that `StatsManager` now knows
what subclass of `StatsContextFactory` is in use, and all of the stats
implementation classes can interact through package-private methods.

### DIFF
--- a/core/src/main/java/com/google/instrumentation/stats/Stats.java
+++ b/core/src/main/java/com/google/instrumentation/stats/Stats.java
@@ -19,17 +19,17 @@ import com.google.instrumentation.common.Provider;
  * {@link Stats}.
  */
 public final class Stats {
-  private static final StatsContextFactory CONTEXT_FACTORY = Provider.newInstance(
-      "com.google.instrumentation.stats.StatsContextFactoryImpl", null);
-
   private static final StatsManager STATS_MANAGER = Provider.newInstance(
       "com.google.instrumentation.stats.StatsManagerImpl", null);
 
   /**
    * Returns the default {@link StatsContextFactory}.
+   *
+   * @deprecated Use {@link StatsManager#getStatsContextFactory()} instead.
    */
+  @Deprecated
   public static StatsContextFactory getStatsContextFactory() {
-    return CONTEXT_FACTORY;
+    return STATS_MANAGER.getStatsContextFactory();
   }
 
   /**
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/google/instrumentation/stats/Stats.java
```diff
@@ -19,17 +19,17 @@
  * {@link Stats}.
  */
 public final class Stats {
-  private static final StatsContextFactory CONTEXT_FACTORY = Provider.newInstance(
-      "com.google.instrumentation.stats.StatsContextFactoryImpl", null);
-
   private static final StatsManager STATS_MANAGER = Provider.newInstance(
```
bogdandrutu: This is not a constant:
 https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names

I think the name should be statsManager

core/src/main/java/com/google/instrumentation/stats/StatsManager.java
```diff
@@ -28,4 +28,9 @@
    * Returns the current stats data, {@link View}, associated with the given {@link ViewDescriptor}.
    */
   public abstract View getView(ViewDescriptor viewDescriptor);
+
+  /**
+   * Returns the default {@link StatsContextFactory}.
+   */
+  public abstract StatsContextFactory getStatsContextFactory();
```
bogdandrutu: Please do not make this public until we decide how the contextFactory interface should look like and if it should be in the core as we have it in the trace. This means do not deprecate yet the old function and make this method protected.

core/src/main/java/com/google/instrumentation/stats/Stats.java
sebright: You're right, it will soon start storing the stats.

core/src/main/java/com/google/instrumentation/stats/StatsManager.java
sebright: `StatsContextFactory` is already exposed and is used by gRPC.  I agree that it probably needs to be changed, though, so I'll save this change for another PR.

sebright:
Thanks!

PR #190, opened 2017-04-04
Skeleton for mapping from MeasurementDescriptors to Views.

songy23:
These are some very rough skeletons for implementing the class that maps MeasurementDescriptors to a set of Views. The changes on StatsManagerImpl should be built upon PR https://github.com/google/instrumentation-java/pull/189.

I'm not sure whether we shall extend current [View](https://github.com/google/instrumentation-java/blob/master/core/src/main/java/com/google/instrumentation/stats/View.java) and have additional fields on Distributions (this will make View a mutable class), or just create a new MutableView class.


core_impl/src/main/java/com/google/instrumentation/stats/StatsManagerImpl.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package com.google.instrumentation.stats;
+
+import java.util.Set;
+
+/**
+ * Native Implementation of {@link StatsManager}.
+ */
+public class StatsManagerImpl extends StatsManager {
```
sebright: This class can be final.

core_impl/src/main/java/com/google/instrumentation/stats/MeasurementDescriptorToViewMap.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+public final class MeasurementDescriptorToViewMap {
+
+  // We are using a preset measurement {@link RpcConstants.RPC_CLIENT_ROUNDTRIP_LATENCY}
+  // for this prototype.
+  // The prototype does not support setting measurements or views dynamically.
+  // TODO(songya): remove the hard-coded preset measurement later.
+  // TODO(songya): replace View with a new MutableView class.
+  private static final Map<MeasurementDescriptor, Set<View>> mutableMap =
+      new ConcurrentHashMap<MeasurementDescriptor, Set<View>>();
```
sebright: Since `MeasurementDescriptor` names are expected to be unique, you could just use the names as keys.

core_impl/src/main/java/com/google/instrumentation/stats/MeasurementDescriptorToViewMap.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * A singleton class that stores a mapping from MeasurementDescriptors to a set of Views.
+ */
+public final class MeasurementDescriptorToViewMap {
```
sebright: This class doesn't need to be public.

core_impl/src/main/java/com/google/instrumentation/stats/StatsManagerImpl.java
```diff
@@ -0,0 +1,39 @@
...
+
+  @Override
+  public void registerView(ViewDescriptor viewDescriptor) {
+    Set<View> viewSet = MeasurementDescriptorToViewMap.getMap()
+        .get(viewDescriptor.getMeasurementDescriptor());
+    if (viewSet == null) {
```
sebright: Shouldn't this line also check whether the `ViewDescriptor` matches a `View` in the set?

core_impl/src/main/java/com/google/instrumentation/stats/MeasurementDescriptorToViewMap.java
songy23: Makes sense, as long as we only access it through StatsManagerImpl

core_impl/src/main/java/com/google/instrumentation/stats/MeasurementDescriptorToViewMap.java
songy23: Yeah that's a better implementation, I'll inject MeasurementDescriptorToViewMap in StatsManagerImpl.

songy23: Sure, will take a look at Multimap

core_impl/src/main/java/com/google/instrumentation/stats/MeasurementDescriptorToViewMap.java
```diff
@@ -0,0 +1,55 @@
...
+      new ConcurrentHashMap<MeasurementDescriptor, Set<View>>();
+
+  static {
+    mutableMap.put(RpcConstants.RPC_CLIENT_ROUNDTRIP_LATENCY,
+        Collections.synchronizedSet(new HashSet<View>()));
+  }
+
+  // TODO(songya): support setting measurements and views dynamically.
+  private static final Map<MeasurementDescriptor, Set<View>> map =
+      Collections.unmodifiableMap(mutableMap);
```
songy23: This unmodifiable map is just for this prototype, and it is just a snapshot of the real mutable map. We don't want the map to have other entries except for RPC_CLIENT_ROUNDTRIP_LATENCY, and using a unmodifiable map helps us enforce that (no one can access the real mutable map). Later when we want to support dynamically setting measurement descriptors, this unmodifiable map should be removed.

core_impl/src/main/java/com/google/instrumentation/stats/MeasurementDescriptorToViewMap.java
songy23: Not sure if using the names has significant benefit over using MeasurementDescriptor itself. I think using MeasurementDescriptor will be more straightforward.

core_impl/src/main/java/com/google/instrumentation/stats/StatsManagerImpl.java
songy23: Fixed.

core_impl/src/main/java/com/google/instrumentation/stats/StatsManagerImpl.java
songy23: Checking that will need to iterate all the views in the set in a loop, and I think this is for registering new views so the ViewDescriptor should not match any View in the set. I'll add the checking in else{} block.

core_impl/src/main/java/com/google/instrumentation/stats/MeasurementDescriptorToViewMap.java
sebright: I just merged #192, so you can use the new key type if you want.

songy23: Thx, will update

core_impl/src/main/java/com/google/instrumentation/stats/StatsManagerImpl.java
```diff
@@ -13,21 +13,88 @@
...
+      for (View view : views) {
+        if (view.getViewDescriptor().equals(viewDescriptor)) {
+          throw new IllegalArgumentException("View has already been registered.");
+        }
+      }
+    }
+
+    // TODO(songya): Extend View class and construct View with internal Distributions.
+    View newView;
+    if (viewDescriptor instanceof DistributionViewDescriptor) {
```
sebright: This set of if/else blocks can use [ViewDescriptor.match(...)](https://github.com/google/instrumentation-java/blob/2e16e06b33d259591117383eb59552ffa15950d2/core/src/main/java/com/google/instrumentation/stats/ViewDescriptor.java#L62) instead of `instanceof` and casting.

core_impl/src/test/java/com/google/instrumentation/stats/StatsManagerImplTest.java
```diff
@@ -0,0 +1,70 @@
...
+public class StatsManagerImplTest {
+
+  @Rule
+  public final ExpectedException thrown = ExpectedException.none();
+
+  private StatsManager statsManager = new StatsManagerImpl();
+
+  @Ignore
+  @Test
+  public void testRegisterAndGetView() throws Exception {
```
sebright: Why is this test ignored?  It looks like the methods are implemented now.

core_impl/src/main/java/com/google/instrumentation/stats/MeasurementDescriptorToViewMap.java
```diff
@@ -0,0 +1,43 @@
...
+   * A synchronized singleton map that storing the one-to-many mapping from MeasurementDescriptors
+   * to Views.
+   */
+  private static final Multimap<MeasurementDescriptor.Name, View> mutableMap =
+      Multimaps.synchronizedMultimap(HashMultimap.<MeasurementDescriptor.Name, View>create());
+
+  /**
+   * Returns the singleton MeasurementDescriptors to Views map.
+   */
+  static final Multimap<MeasurementDescriptor.Name, View> getMap() {
```
sebright: I think it would be better to not expose the map, if possible.  One reason is that this class should ensure that no code modifies the mutable views without synchronizing on the map.

core_impl/src/main/java/com/google/instrumentation/stats/StatsManagerImpl.java
```diff
@@ -13,21 +13,88 @@
...
 public final class StatsManagerImpl extends StatsManager {
 
+  private final Multimap<MeasurementDescriptor.Name, View> measurementDescriptorToViewMap =
+      MeasurementDescriptorToViewMap.getMap();
   private final StatsContextFactoryImpl statsContextFactory = new StatsContextFactoryImpl();
 
   @Override
   public void registerView(ViewDescriptor viewDescriptor) {
-    throw new UnsupportedOperationException("Not implemented yet.");
+    // We are using a preset measurement {@link RpcConstants.RPC_CLIENT_ROUNDTRIP_LATENCY}
```
sebright: `@link` has no effect in a non-Javadoc comment.

core_impl/src/main/java/com/google/instrumentation/stats/StatsManagerImpl.java
```diff
@@ -13,21 +13,88 @@
...
+    }
+
+    MeasurementDescriptor measurementDescriptor = viewDescriptor.getMeasurementDescriptor();
+    Collection<View> views = measurementDescriptorToViewMap.get(
+        measurementDescriptor.getMeasurementDescriptorName());
+
+    if (views != null) {
+      for (View view : views) {
+        if (view.getViewDescriptor().equals(viewDescriptor)) {
+          throw new IllegalArgumentException("View has already been registered.");
```
sebright: Is there any problem with registering the same view twice? Can we just ignore it the second time?

core_impl/src/main/java/com/google/instrumentation/stats/StatsManagerImpl.java
```diff
@@ -13,21 +13,88 @@
...
+    if (!viewDescriptor.equals(RpcConstants.RPC_CLIENT_ROUNDTRIP_LATENCY_VIEW)) {
+      throw new UnsupportedOperationException(
+          "The prototype will only support Distribution View RPC_CLIENT_ROUNDTRIP_LATENCY_VIEW.");
+    }
+
+    MeasurementDescriptor measurementDescriptor = viewDescriptor.getMeasurementDescriptor();
+    Collection<View> views = measurementDescriptorToViewMap.get(
+        measurementDescriptor.getMeasurementDescriptorName());
+
+    if (views != null) {
```
sebright: `Multimap.get` returns an empty collection when the key is not present.

core_impl/src/test/java/com/google/instrumentation/stats/StatsManagerImplTest.java
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * Tests for {@link StatsManagerImpl}.
+ */
+@RunWith(JUnit4.class)
+public class StatsManagerImplTest {
+
+  @Rule
+  public final ExpectedException thrown = ExpectedException.none();
+
+  private StatsManager statsManager = new StatsManagerImpl();
```
sebright: This field can be final.

core_impl/src/main/java/com/google/instrumentation/stats/MeasurementDescriptorToViewMap.java
```diff
@@ -0,0 +1,43 @@
...
+ * A class that stores a singleton map from MeasurementDescriptors to Views.
+ */
+final class MeasurementDescriptorToViewMap {
+
+  /**
+   * A synchronized singleton map that storing the one-to-many mapping from MeasurementDescriptors
+   * to Views.
+   */
+  private static final Multimap<MeasurementDescriptor.Name, View> mutableMap =
+      Multimaps.synchronizedMultimap(HashMultimap.<MeasurementDescriptor.Name, View>create());
```
songy23: So this map should be singleton in terms of each StatsManagerImpl? I was thinking it is a singleton across all the stats managers.

core_impl/src/main/java/com/google/instrumentation/stats/StatsManagerImpl.java
songy23: Fixed

core_impl/src/main/java/com/google/instrumentation/stats/StatsManagerImpl.java
songy23: Fixed

core_impl/src/main/java/com/google/instrumentation/stats/StatsManagerImpl.java
songy23: Either ignoring it or throwing an exception makes sense to me, I can update and document the logic here.

core_impl/src/main/java/com/google/instrumentation/stats/StatsManagerImpl.java
songy23: Sure, that's a better solution. Will update

core_impl/src/test/java/com/google/instrumentation/stats/StatsManagerImplTest.java
songy23: Fixed

core_impl/src/test/java/com/google/instrumentation/stats/StatsManagerImplTest.java
songy23: Removed the Ignore tag. The implementation is not complete yet but this test should pass anyway.

core_impl/src/main/java/com/google/instrumentation/stats/MeasurementDescriptorToViewMap.java
sebright: Yes.  Normally there will only be one StatsManagerImpl, so there will only be one instance of MeasurementDescriptorToViewMap.  However, in test code, we will be able to create a new StatsManagerImpl for each test and ensure that the tests are independent.

core_impl/src/main/java/com/google/instrumentation/stats/MeasurementDescriptorToViewMap.java
songy23: I'm not very clear on your suggestion here: are you suggesting to just expose APIs to put & get values from the map (like getView() and putView()), instead of exposing the map itself?

core_impl/src/main/java/com/google/instrumentation/stats/MeasurementDescriptorToViewMap.java
songy23: That makes sense, will update here

core_impl/src/main/java/com/google/instrumentation/stats/MeasurementDescriptorToViewMap.java
sebright: Yes, I thought that we could add methods to MeasurementDescriptorToViewMap to handle all operations that we need to perform on the map of views.

core_impl/src/main/java/com/google/instrumentation/stats/StatsManagerImpl.java
sebright: I just thought that the API would be easier to use if client code didn't have to ensure that views were only added once.

core_impl/src/main/java/com/google/instrumentation/stats/MeasurementDescriptorToViewMap.java
songy23: Makes sense, thx

core_impl/src/test/java/com/google/instrumentation/stats/StatsManagerImplTest.java
```diff
@@ -45,7 +43,7 @@ public void testRegisterAndGetView() throws Exception {
         Timestamp.fromMillis(System.currentTimeMillis()));
     statsManager.registerView(RpcConstants.RPC_CLIENT_ROUNDTRIP_LATENCY_VIEW);
     View actual = statsManager.getView(RpcConstants.RPC_CLIENT_ROUNDTRIP_LATENCY_VIEW);
-    assertThat(actual).isEqualTo(expected);
+    assertThat(actual.getViewDescriptor()).isEqualTo(expected.getViewDescriptor());
```
sebright: Could this assertion compare to RpcConstants.RPC_CLIENT_ROUNDTRIP_LATENCY_VIEW?

core_impl/src/test/java/com/google/instrumentation/stats/StatsManagerImplTest.java
```diff
@@ -55,14 +53,6 @@ public void testRegisterUnsupportedViewDesciptor() throws Exception {
   }
 
   @Test
-  public void testRegisterDuplicatedViewDesciptor() throws Exception {
-    statsManager.registerView(RpcConstants.RPC_CLIENT_ROUNDTRIP_LATENCY_VIEW);
-
-    thrown.expect(IllegalArgumentException.class);
-    statsManager.registerView(RpcConstants.RPC_CLIENT_ROUNDTRIP_LATENCY_VIEW);
-  }
```
sebright: I think this test is still useful for checking that the second call doesn't throw an exception.

core_impl/src/main/java/com/google/instrumentation/stats/MeasurementDescriptorToViewMap.java
```diff
@@ -16,28 +16,38 @@
...
-  static final Multimap<MeasurementDescriptor.Name, View> getMap() {
-    return mutableMap;
+  final View getView(ViewDescriptor viewDescriptor) {
+    Collection<View> views = mutableMap.get(
+        viewDescriptor.getMeasurementDescriptor().getMeasurementDescriptorName());
+    for (View view : views) {
+      if (view.getViewDescriptor().equals(viewDescriptor)) {
+        return view;
+      }
+    }
```
songy23: I think `Multimaps.synchronizedMultimap()` has already returned us a thread-safe Multi Map. Should we synchronized(views) here?

songy23: I think `Multimaps.synchronizedMultimap()` has already returned us a thread-safe Multi Map. Should we synchronized(views) here?

core_impl/src/test/java/com/google/instrumentation/stats/StatsManagerImplTest.java
songy23: Sure, updated.

core_impl/src/test/java/com/google/instrumentation/stats/StatsManagerImplTest.java
songy23: Makes sense, put this test back.

core_impl/src/main/java/com/google/instrumentation/stats/MeasurementDescriptorToViewMap.java
sebright: The Javadoc says that the user needs to synchronize on the map when accessing its collection views: https://google.github.io/guava/releases/19.0/api/docs/com/google/common/collect/Multimaps.html#synchronizedMultimap(com.google.common.collect.Multimap)

songy23: Fixed, thanks for pointing out

songy23:
Sure, I'll squash the commits into one. Thanks!

PR #191, opened 2017-04-04
Remove start/end timestamps from our public interface.

bogdandrutu:
Until we have a better design on how are we going to support them we should not have this functionality in our public API.


adriancole:
Completely agree. At least in brave if it werent for opentracing needing
this I wouldnt have added support for it either (at the api level.. obv
reporting model would)

PR #192, opened 2017-04-05
Adds a class to represent a MeasurementDescriptor name.

sebright:
We need to associate data with MeasurementDescriptors in several places in the
stats implementation.  For example, MeasurementMap maps MeasurementDescriptors
to recorded values.  Using MeasurementDescriptor as the key is not ideal,
because it is not obvious how the keys should be compared.  The keys could be
compared by name, since names are unique, or they could be compared by all
fields.

I think that we should always compare MeasurementDescriptors by name.  It is
more efficient, and it means that two MeasurementDescriptors with slightly
different descriptions won't be treated as distinct.  Having a separate key type
will make it clearer how the MeasurementDescriptors will be compared.

This commit just adds the class and doesn't change any existing APIs.

### DIFF
--- a/core/src/main/java/com/google/instrumentation/stats/MeasurementDescriptor.java
+++ b/core/src/main/java/com/google/instrumentation/stats/MeasurementDescriptor.java
@@ -34,6 +34,14 @@ public abstract class MeasurementDescriptor {
    * Constructs a new {@link MeasurementDescriptor}.
    */
   public static MeasurementDescriptor create(
+      MeasurementDescriptor.Name name, String description, MeasurementUnit unit) {
+    return new AutoValue_MeasurementDescriptor(name.asString(), description, unit);
+  }
+
+  /**
+   * Constructs a new {@link MeasurementDescriptor}.
+   */
+  public static MeasurementDescriptor create(
       String name, String description, MeasurementUnit unit) {
     return new AutoValue_MeasurementDescriptor(StringUtil.sanitize(name), description, unit);
   }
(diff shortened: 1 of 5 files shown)


songy23:
I agree with @sebright, the name of MeasurementDescriptor needs to be unique (https://github.com/google/instrumentation-java/blob/master/core/src/main/java/com/google/instrumentation/stats/MeasurementDescriptor.java#L42) and comparing by name should be enough.

core/src/main/java/com/google/instrumentation/stats/MeasurementDescriptor.java
```diff
@@ -34,13 +34,22 @@
...
+   */
+  public static MeasurementDescriptor create(
       String name, String description, MeasurementUnit unit) {
     return new AutoValue_MeasurementDescriptor(StringUtil.sanitize(name), description, unit);
   }
 
   /**
    * Name of measurement, e.g. rpc_latency, cpu. Must be unique.
    */
+  // TODO(sebright): Change the type of this field to MeasurementDescriptor.Name.
```
bogdandrutu: I think it will be hard to just change this return type. Maybe better if we add a new method that returns the MeasurementDescriptor.Name and this will just return the name.asString.

sebright: > I think it will be hard to just change this return type.

We could change the return type in two steps, if we wanted to.

> Maybe better if we add a new method that returns the MeasurementDescriptor.Name and this will just return the name.asString.

I added a new getter that returns the Name type and left the return type of getName() as String.

sebright:
Thanks!

PR #193, opened 2017-04-05
Apply AutoValue to Tag, TagKey, and TagValue.

sebright:


core/src/test/java/com/google/instrumentation/stats/TagTest.java
```diff
@@ -52,9 +52,9 @@ public void testHashCode() {
 
   @Test
   public void testToString() {
-    assertThat(Tag.create(K1, V1).toString()).isEqualTo("Tag<k1,v1>");
-    assertThat(Tag.create(K2, V1).toString()).isEqualTo("Tag<k2,v1>");
-    assertThat(Tag.create(K1, V2).toString()).isEqualTo("Tag<k1,v2>");
+    assertThat(Tag.create(K1, V1).toString()).isEqualTo("Tag{key=k1, value=v1}");
+    assertThat(Tag.create(K2, V1).toString()).isEqualTo("Tag{key=k2, value=v1}");
+    assertThat(Tag.create(K1, V2).toString()).isEqualTo("Tag{key=k1, value=v2}");
```
sebright: Should we remove the toString() tests, now that we are using AutoValue?  Another option is to just test that the output of toString() is not the default, as suggested here: http://stackoverflow.com/a/12528424

core/src/main/java/com/google/instrumentation/stats/TagKey.java
```diff
@@ -30,27 +35,14 @@
...
+    return new AutoValue_TagKey(StringUtil.sanitize(key));
   }
 
-  @Override
-  public int hashCode() {
-    return key.hashCode();
-  }
+  public abstract String asString();
 
+  // TODO(sebright) Use the default AutoValue toString(), which is more useful for debugging.
```
bogdandrutu: Why do we not use that?

core/src/test/java/com/google/instrumentation/stats/TagTest.java
bogdandrutu: I think if the class uses AutoValue we should do that.

core/src/main/java/com/google/instrumentation/stats/TagKey.java
sebright: gRPC is currently calling toString() to access the key and value fields.  I added that to the comment.

core/src/test/java/com/google/instrumentation/stats/TagTest.java
sebright: I removed it.

PR #194, opened 2017-04-06
Roll back the change to have different factory methods for spans and spans with remote parent.

bogdandrutu:
Reason is that this way we can ensure an order between events recorded in different spans from the same trace.

We will calculate the timestamp using delay since the start time of the initial Span in the trace in the process (root span or a span with remote parent).

### DIFF
--- a/core/src/main/java/com/google/instrumentation/trace/SpanBuilder.java
+++ b/core/src/main/java/com/google/instrumentation/trace/SpanBuilder.java
@@ -102,18 +102,17 @@ public final class SpanBuilder {
   private final SpanFactory spanFactory;
   private final String name;
   private final StartSpanOptions startSpanOption = new StartSpanOptions();
+  private Span parentSpan;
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/google/instrumentation/trace/SpanBuilder.java
```diff
@@ -102,18 +102,17 @@
   private final SpanFactory spanFactory;
   private final String name;
   private final StartSpanOptions startSpanOption = new StartSpanOptions();
+  private Span parentSpan;
   private SpanContext parentSpanContext;
```
adriancole: personally I like remote parent being a part of the context, as that allows you to separate the concern cleanly (context being demarshalled from a remote source is pretty easy to identify)

bogdandrutu: I think that makes sense. We will probably do that change later.

PR #195, opened 2017-04-06
Add support for binary encoding.

bogdandrutu:
This change allows us to change the handler that we use for serialization and deserialization at runtime even for older libraries that they use an old version of the library.


core/src/main/java/com/google/instrumentation/trace/PropagationUtil.java
```diff
@@ -0,0 +1,230 @@
...
+     */
+    public abstract byte[] toBinaryFormat(SpanContext spanContext);
+
+    /**
+     * Parses the {@link SpanContext} from the binary format.
+     *
+     * @param bytes a binary encoded buffer from which the {@code SpanContext} will be parsed.
+     * @return the parsed {@code SpanContext}.
+     */
+    public abstract SpanContext fromBinaryFormat(byte[] bytes);
```
sebright: Would it be better to use `InputStream` and `OutputStream` for these two methods, so that they are more general?  That is what `StatsSerializer` does: https://github.com/google/instrumentation-java/blob/2e16e06b33d259591117383eb59552ffa15950d2/core_impl/src/main/java/com/google/instrumentation/stats/StatsSerializer.java#L50-89

core/src/main/java/com/google/instrumentation/trace/PropagationUtil.java
```diff
@@ -0,0 +1,230 @@
...
+    checkNotNull(bytes, "bytes");
+    return handler.fromBinaryFormat(bytes);
+  }
+
+  /**
+   * Sets a new {@link Handler}.
+   *
+   * @param newHandler the new {@code Handler} to be set.
+   */
+  public static void setHandler(Handler newHandler) {
```
sebright: When will this method be called?  Would it be better to make the Handler an argument to the serialization and deserialization methods?

bogdandrutu: ACK. Made it package-protected for the moment. We use it in tests only for the moment.

core/src/main/java/com/google/instrumentation/trace/PropagationUtil.java
bogdandrutu: We can add these variants later. I will try to get this PR with bytes for the moment and I will ask gRPC how do they need this (byte[] or InputStream)

core/src/main/java/com/google/instrumentation/trace/PropagationUtil.java
```diff
@@ -0,0 +1,230 @@
...
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
+
+import com.google.common.io.BaseEncoding;
+import java.util.concurrent.atomic.AtomicReferenceArray;
+
+/**
+ * This is a helper class for {@link SpanContext} propagation on the wire. Allows changing the
+ * defautl handler at runtime using {@link #setHandler(Handler)}.
```
a-veitch: s/defautl/default/

bogdandrutu: Removed the entire sentence because for the moment setHandler is package-protected.

PR #196, opened 2017-04-07
Add Distribution class

a-veitch:
This PR adds a functional Distribution class, to be used for collecting stats.

### DIFF
--- a/core/src/main/java/com/google/instrumentation/stats/Distribution.java
+++ b/core/src/main/java/com/google/instrumentation/stats/Distribution.java
@@ -1,0 +1,235 @@
+/*
+ * Copyright 2017, Google Inc.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.instrumentation.stats;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import javax.annotation.Nullable;
+
+/**
+ * A distribution contains summary statistics for a population of values and, optionally, a
+ * histogram representing the distribution of those values across a specified set of histogram
+ * buckets.
+ *
+ * <p>The bucket boundaries for that histogram are described by {@link getBucketBoundaries}, which
+ * defines {@code getBucketBoundaries.size() + 1 (= N)} buckets. The boundaries for bucket index i
+ * are:
+ *
+ * <ul>
+ *   <li>[-infinity, bounds[i]) for i == 0
+ *   <li>[bounds[i-1], bounds[i]) for 0 &lt; i &lt; N-2
+ *   <li>[bounds[i-1], +infinity) for i == N-1
+ * </ul>
+ *
+ * <p>i.e. an underflow bucket (number 0), zero or more finite buckets (1 through N - 2, and an
+ * overflow bucket (N - 1), with inclusive lower bounds and exclusive upper bounds.
+ *
+ * <p>Note: If N = 1, there are no finite buckets and the single bucket is both the overflow and
+ * underflow bucket.
+ *
+ * <p>Although not forbidden, it is generally a bad idea to include non-finite values (infinities or
+ * NaNs) in the population of values, as this will render the {@code mean} meaningless.
+ */
+public final class Distribution {
+  private long count = 0; // The number of values in the population.
+  private double sum = 0.0; // The sum of the values in the population.
+  private Range range = new Range(); // Range of values in the population.
+  private final ArrayList<Double> bucketBoundaries; // Histogram boundaries; null means no histogram
+  private Long[] bucketCounts; // Counts for each histogram bucket
+
+  /**
+   * Constructs a new, empty {@link Distribution}.
+   *
+   * @return a new, empty {@code Distribution}.
+   */
+  public static final Distribution create() {
+    return new Distribution(null);
+  }
+
+  /**
+   * Constructs a new {@link Distribution} with specified {@code bucketBoundaries}.
+   *
+   * @param bucketBoundaries the boundaries for the buckets in the underlying {@code Distribution}.
+   * @return a new, empty {@code Distribution} with the specified boundaries.
+   * @throws NullPointerException if {@code bucketBoundaries} is null.
+   * @throws IllegalArgumentException if {@code bucketBoundaries} are not sorted.
+   */
+  public static final Distribution create(List<Double> bucketBoundaries) {
+    checkNotNull(bucketBoundaries, "bucketBoundaries");
+    double lower = bucketBoundaries.get(0);
+    for (int i = 1; i < bucketBoundaries.size(); i++) {
+      double next = bucketBoundaries.get(i);
+      checkArgument(lower < next, "Bucket boundaries not sorted.");
+      lower = next;
+    }
+    return new Distribution(bucketBoundaries);
+  }
+
+  // Returns true if the distribution has histogram buckets.
+  private boolean hasBuckets() {
+    return (bucketBoundaries != null);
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/google/instrumentation/stats/Distribution.java
```diff
@@ -0,0 +1,236 @@
...
+   */
+  public static final Distribution create(List<Double> bucketBoundaries) {
+    checkNotNull(bucketBoundaries, "bucketBoundaries");
+    checkArgument(bucketBoundaries.size() > 0, "Zero length bucket boundaries");
+    double lower = bucketBoundaries.get(0);
+    for (int i = 1; i < bucketBoundaries.size(); i++) {
+      double next = bucketBoundaries.get(i);
+      checkArgument(lower < next, "Bucket boundaries not sorted.");
+      lower = next;
+    }
```
sebright: These invariants should be checked after the list is copied, in case the list changes.

core/src/main/java/com/google/instrumentation/stats/Distribution.java
```diff
@@ -0,0 +1,236 @@
...
+
+  // Returns true if the distribution has histogram buckets.
+  private boolean hasBuckets() {
+    return (bucketBoundaries != null);
+  }
+
+  // Construct a new Distribution with given boundaries.
+  private Distribution(List<Double> bucketBoundaries) {
+    this.bucketBoundaries =
+        (bucketBoundaries == null) ? null : new ArrayList<Double>(bucketBoundaries.size());
```
sebright: bucketBoundaries could be made immutable with `Collections.unmodifiableList(new ArrayList<Double>(bucketBoundaries))`.

core/src/test/java/com/google/instrumentation/stats/DistributionTest.java
```diff
@@ -0,0 +1,124 @@
...
+      assertThat(empty.getBucketCounts().get(i)).isEqualTo(0);
+    }
+  }
+
+  @Test(expected = NullPointerException.class)
+  public void testNullBoundaries() {
+    Distribution distribution = Distribution.create(null);
+  }
+
+  @Test(expected = IllegalArgumentException.class)
```
songy23: Travis-CI is complaining about this, might use ExpectedException instead.

core/src/main/java/com/google/instrumentation/stats/Distribution.java
```diff
@@ -0,0 +1,236 @@
...
+   *
+   * @return The sum of values in the population.
+   */
+  public double getSum() {
+    return sum;
+  }
+
+  /**
+   * The range of the population values. If {@link #getCount()} is zero then this returned range is
+   * implementation-dependent. @ return The {code Range} representing the range of population
```
sebright: Should `@return` be on the next line?

core/src/main/java/com/google/instrumentation/stats/Distribution.java
```diff
@@ -0,0 +1,236 @@
...
+   * boundaries.
```
sebright: Should this say "range of population values"?

core/src/main/java/com/google/instrumentation/stats/Distribution.java
```diff
@@ -0,0 +1,236 @@
...
+   *
+   * <p>The size of {@link #getBucketCounts()} must be no greater than N as defined in {@link
+   * DistributionAggregationDescriptor#getBucketBoundaries()}.
+   *
+   * <p>Any suffix of trailing buckets containing only zero may be omitted.
+   *
+   * <p>{@link #getBucketCounts()} will return null iff the associated {@link
+   * DistributionAggregationDescriptor#getBucketBoundaries()} returns null.
+   *
+   * @return The count of population boundaries in each histogram bucket.
```
sebright: Should this say "values" instead of "boundaries"?

core/src/main/java/com/google/instrumentation/stats/Distribution.java
```diff
@@ -0,0 +1,236 @@
...
+   */
+  public Range getRange() {
+    return range;
+  }
+
+  /**
+   * A Distribution may optionally contain a histogram of the values in the population. The
+   * histogram is given in {@link #getBucketCounts()} as counts of values that fall into one of a
+   * sequence of non-overlapping buckets, described by {@link
+   * DistributionAggregationDescriptor#getBucketBoundaries()}. The sum of the values in {@link
```
sebright: Should the link point to Distribution instead of DistributionAggregationDescriptor, since the Javadoc for Distribution also explains the buckets?

core/src/main/java/com/google/instrumentation/stats/Distribution.java
```diff
@@ -0,0 +1,236 @@
...
+    }
+    bucketCounts[bucketCounts.length - 1]++;
+  }
+
+  /**
+   * Put a new value into the Distribution.
+   *
+   * @param value new value to be added to population
+   */
+  public void put(double value) {
```
sebright: I have only seen "put" used for Java maps, so I think "add" would be a little clearer.

core/src/main/java/com/google/instrumentation/stats/Distribution.java
```diff
@@ -0,0 +1,236 @@
...
+    }
+  }
+
+  /** Describes a range of population values. */
+  public static final class Range {
+    private double min = Double.POSITIVE_INFINITY;
+    private double max = Double.NEGATIVE_INFINITY;
+
+    /** Construct a new, empty {@code Range}. */
+    public Range() {}
```
sebright: We have been exposing factory methods instead of constructors for the other stats classes.

core/src/main/java/com/google/instrumentation/stats/Distribution.java
```diff
@@ -0,0 +1,236 @@
...
+    range.put(value);
+    if (hasBuckets()) {
+      putIntoBucket(value);
+    }
+  }
+
+  /** Describes a range of population values. */
+  public static final class Range {
+    private double min = Double.POSITIVE_INFINITY;
+    private double max = Double.NEGATIVE_INFINITY;
```
sebright: Shouldn't the max always be greater than the min?

core/src/main/java/com/google/instrumentation/stats/Distribution.java
```diff
@@ -0,0 +1,236 @@
...
+    public double getMax() {
+      return max;
+    }
+
+    /**
+     * Put a new value into the Range. Sets min and max values if appropriate.
+     *
+     * @param value the new value
+     */
+    public void put(double value) {
```
sebright: Same here, I think "add" would be clearer.

core/src/main/java/com/google/instrumentation/stats/DistributionAggregationDescriptor.java
```diff
@@ -18,6 +18,7 @@
 import java.util.List;
 import javax.annotation.Nullable;
 
+// TODO(aveitch) This class should be changed to use Distributions.
```
sebright: I think this class only needs the bucket boundaries from Distribution.

core/src/test/java/com/google/instrumentation/stats/DistributionTest.java
```diff
@@ -0,0 +1,124 @@
...
+    distribution.put(1.0);
+    assertThat(distribution.getCount()).isEqualTo(1);
+    assertThat(distribution.getSum()).isEqualTo(1.0);
+    assertThat(distribution.getMean()).isEqualTo(1.0);
+    assertThat(distribution.getRange().getMin()).isEqualTo(1.0);
+    assertThat(distribution.getRange().getMax()).isEqualTo(1.0);
+    distribution.put(1.0);
+    assertThat(distribution.getCount()).isEqualTo(2);
+    assertThat(distribution.getSum()).isEqualTo(2.0);
+    assertThat(distribution.getMean()).isEqualTo(1.0);
```
sebright: Should these assertions do fuzzy comparisions on the doubles (with `isWithin`)?

core/src/test/java/com/google/instrumentation/stats/DistributionTest.java
a-veitch: Thanks, I've changed to ExpectedException

adriancole:
Looks good. Is there an associated proto or otherwise wire format yet?

sebright:
@adriancole This class represents part of the DistributionAggregation proto: https://github.com/google/instrumentation-proto/blob/cc428c7bc049175ce677d910bbc3a23461c306a1/stats/census.proto#L149.  I think we should also refactor that proto so that it has two fields, `distribution` and `tags`.

sebright:
Subsumed by #198.

PR #197, opened 2017-04-08
Update to use the new proposed encoding.

bogdandrutu:
Change the PropagationUtil to use the new binary encoding.
Change the TraceOptions to be 1B instead of 4B.
Remove the HTTP encoding for the moment until we define the common encoding between projects.

### DIFF
--- a/core/src/main/java/com/google/instrumentation/trace/PropagationUtil.java
+++ b/core/src/main/java/com/google/instrumentation/trace/PropagationUtil.java
@@ -34,17 +34,17 @@ import com.google.common.io.BaseEncoding;
  *         representing the trace_id.
  *     <li>SpanId: (filed_id = 1, len = 8, default = “00000000”) - 8-byte array representing the
  *         span_id.
- *     <li>TraceOptions: (filed_id = 2, len = 4, default = “0000”) - 4-byte array representing the
+ *     <li>TraceOptions: (filed_id = 2, len = 4, default = “0”) - 1-byte array representing the
  *         trace_options. It is in little-endian order, if represented as an int.
  *     <li>Fields MUST be encoded using the filed id order (smaller to higher).
  *     <li>Valid value example:
  *         <ul>
  *         <li>{0, 0, 64, 65, 66, 67, 68, 69, 70, 71, 72, 73, 74, 75, 76, 77, 78, 79, 1, 97, 98, 99,
- *             100, 101, 102, 103, 104, 2, 1, 0, 0, 0}
+ *             100, 101, 102, 103, 104, 2, 1}
  *         <li>version_id = 0;
  *         <li>trace_id = {64, 65, 66, 67, 68, 69, 70, 71, 72, 73, 74, 75, 76, 77, 78, 79}
  *         <li>span_id = {97, 98, 99, 100, 101, 102, 103, 104};
- *         <li>trace_options = {1, 0, 0, 0} == 1;
+ *         <li>trace_options = {1};
  *         </ul>
  *     </ul>
  *
@@ -59,11 +59,11 @@ import com.google.common.io.BaseEncoding;
  * <li>It is strongly encouraged to use this format when using HTTP as a RPC transport.
  * <li>Valid value example:
  *     <ul>
- *     <li>"0000404142434445464748494A4B4C4D4E4F0161626364656667680201000000"
+ *     <li>"0000404142434445464748494A4B4C4D4E4F0161626364656667680201"
  *     <li>version_id = 0;
  *     <li>trace_id = {64, 65, 66, 67, 68, 69, 70, 71, 72, 73, 74, 75, 76, 77, 78, 79}
  *     <li>span_id = {97, 98, 99, 100, 101, 102, 103, 104};
- *     <li>trace_options = {1, 0, 0, 0} == 1;
+ *     <li>trace_options = {1};
  *     </ul>
  *
  * </ul>
--- a/core/src/main/java/com/google/instrumentation/trace/PropagationUtil.java
+++ b/core/src/main/java/com/google/instrumentation/trace/PropagationUtil.java
@@ -15,10 +15,8 @@ package com.google.instrumentation.trace;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
-import static com.google.common.base.Preconditions.checkState;
 
 import com.google.common.io.BaseEncoding;
-import java.util.concurrent.atomic.AtomicReferenceArray;
 
 /**
  * This is a helper class for {@link SpanContext} propagation on the wire.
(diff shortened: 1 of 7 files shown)


adriancole:
While binary format is fine, the text format is not usable compared to
others. Operators routinely copy paste trace id from headers and the
numeric field encoding will make this much harder to support for routine
questions.

One suggestion could be to blacklist a field number associated with a
delimiter character like '/' and use that as a substitute for explicit
field numbers. Ex when '/' is a prefix use the declaration order as index
which should always be correct anyway especially in version 1

adriancole:
adding in some (but not all bc lazy) "B3 folks" and/or folks who've been shared on the google doc but might not have been nagged yet :)

@nicmunroe @basvanbeek @jcarres-mdsol @felixbarny @eirslett @michaelsembwever @marcingrzejszczak @fedj @kevinoliver

This is an application of a trace context/header format discussed at a recent tracing workshop: 
https://docs.google.com/document/d/1Mrw7hxVAkj7h98hvgRixDt1RrEJ5fsqqddPofIHRVvI/edit

Notably, the above format assumes only the "parent span" is propagated. So, this means the B3/zipkin convention of propagating the "current span" isn't supported. I explained this model here:

https://github.com/openzipkin/b3-propagation#why-is-parentspanid-propagated

The largest impact of this format is that users would need to fork a child on receipt, which at least in zipkin isn't fully supported at the moment.

https://github.com/openzipkin/zipkin/issues/1480

Key other things to notice about this format is that there's a coordinated schema (ex what fields 0-3 mean), and it allows for future updates to the spec (ex field 4), although doesn't support variants or extensions as identifiers are reserved for the spec owner. This format is mostly fixed length as something else will be out for "baggage" (arbitrary key/values to go across the wire).

This format has an specific trace ID shape that's compatible with zipkin.  Notably, "debug flag" isn't encodable here, although sampling is. Some tracers, such as brave, sleuth, and opentracing+zipkin variants, have "plug-ins" to affect propagation format while holding the in-process representation of Trace Context constant. Experimentation here will be a bit easier, limited to zipkin's support of single-host spans discussed earlier.

Realistically speaking, I'd expect B3 to remain in use while some tracers experiment supporting this. To be more clear, I'd expect B3 to co-exist with this because not everyone will be able to or desire to change their propagation format across their entire architecture. B3 will remain in place indefinitely at the same time as some experiment with this.

I pinged you in case you are interested in this, or have opinions, now that there's some code (albeit in java) to see. So, please take a look and give feedback if interested!

bogdandrutu:
@adriancole I think there are a lot of questions about the HTTP format in the https://github.com/TraceContext/tracecontext-spec/pull/1. I will move forward with this PR and I will also remove the support for the HTTP format in a followup PR until we decide on all the decisions.

Our binary format will be like in this PR but the HTTP format will follow the decisions from the TraceContext repo specs.

Is that fine for you?

adriancole:
Sgtm

core/src/test/java/com/google/instrumentation/trace/PropagationUtilTest.java
```diff
@@ -30,46 +32,47 @@
...
+    assertWithMessage("Binary propagated context is not equal with the initial context.")
+        .that(propagatedBinarySpanContext)
+        .isEqualTo(spanContext);
   }
 
-  @Test
-  public void getHttpHeaderName() {
-    assertThat(PropagationUtil.HTTP_HEADER_NAME).isEqualTo("Trace-Context");
+  // For valid requests to avoid using try-catch everywhere.
+  private static SpanContext fromBinaryValue(byte[] bytes) {
```
sebright: Why can't the tests just throw the exception?

core/src/test/java/com/google/instrumentation/trace/PropagationUtilTest.java
```diff
@@ -84,180 +87,108 @@ public void propagate_SpanContextNoTracing() {
...
-        .isEqualTo("0000000000000000000000000000000061FF0000000000000001340012");
-  }
-
-  @Test(expected = NullPointerException.class)
-  public void toHttpHeaderValue_NullSpanContext() {
-    PropagationUtil.toHttpHeaderValue(null);
-  }
-
-  @Test
   public void toHttpHeaderValue_InvalidSpanContext() {
```
sebright: Should this test name be updated?

core/src/test/java/com/google/instrumentation/trace/PropagationUtilTest.java
```diff
@@ -30,46 +32,47 @@
...
+  private static final byte[] traceOptionsBytes = new byte[] {1};
   private static final TraceOptions traceOptions = TraceOptions.fromBytes(traceOptionsBytes);
   private static final byte[] EXAMPLE_BYTES =
       new byte[] {
-        0, 64, 65, 66, 67, 68, 69, 70, 71, 72, 73, 74, 75, 76, 77, 78, 79, 97, 98, 99, 100, 101,
-        102, 103, 104, 1, 0, 0, 0
+        0, 0, 64, 65, 66, 67, 68, 69, 70, 71, 72, 73, 74, 75, 76, 77, 78, 79, 1, 97, 98, 99, 100,
+        101, 102, 103, 104, 2, 1
       };
   private static final SpanContext EXAMPLE_SPAN_CONTEXT =
```
sebright: The static final fields have inconsistent capitalization.

core/src/test/java/com/google/instrumentation/trace/PropagationUtilTest.java
```diff
@@ -84,180 +87,108 @@ public void propagate_SpanContextNoTracing() {
...
+  public void parseBinaryValue_EmptyInput() throws IOException {
+    PropagationUtil.fromBinaryValue(new byte[0]);
+  }
 
-    assertThat(
-            PropagationUtil.fromHttpHeaderValue(
-                "0000000000000000000000000000000061FF0000000000000000000000"))
-        .isEqualTo(new SpanContext(traceId, spanId, TraceOptions.DEFAULT));
+  @Test(expected = IOException.class)
+  public void parseBinaryValue_UnsupportedVersionId() throws IOException {
```
sebright: Should this test check the exception's error message, to make sure that it was thrown for the right reason?

core/src/main/java/com/google/instrumentation/trace/PropagationUtil.java
```diff
@@ -30,139 +28,76 @@
...
- *     <li>span-id: 8-byte array representing the span_id.
- *     <li>trace-options: 4-byte array representing the trace_options. It is in little-endian order,
- *         if represented as an int.
+ *     <li>version_format: &lt;field&gt;&lt;field&gt;...
+ *     <li>field_format: &lt;field_id&gt;&lt;field_format&gt;
+ *     <li>TraceId: (filed_id = 0, len = 16, default = “0000000000000000”) - 16-byte array
+ *         representing the trace_id.
+ *     <li>SpanId: (filed_id = 1, len = 8, default = “00000000”) - 8-byte array representing the
+ *         span_id.
+ *     <li>TraceOptions: (filed_id = 2, len = 4, default = “0”) - 1-byte array representing the
```
bogdandrutu: Done

core/src/main/java/com/google/instrumentation/trace/PropagationUtil.java
```diff
@@ -30,139 +28,76 @@
...
+ *         trace_options. It is in little-endian order, if represented as an int.
+ *     <li>Fields MUST be encoded using the filed id order (smaller to higher).
```
bogdandrutu: Done. Thanks

core/src/main/java/com/google/instrumentation/trace/PropagationUtil.java
```diff
@@ -30,139 +28,76 @@
...
    * @param bytes a binary encoded buffer from which the {@code SpanContext} will be parsed.
    * @return the parsed {@code SpanContext}.
-   * @throws NullPointerException if the {@code input} is null.
-   * @throws IllegalArgumentException if the {@code input} is invalid.
-   * @throws IllegalStateException if the version is not supported.
+   * @throws NullPointerException if the {@code input} is {@code null}.
+   * @throws IOException if the version is not supported or the input is invalid
    */
-  public static SpanContext fromBinaryValue(byte[] bytes) {
+  public static SpanContext fromBinaryValue(byte[] bytes) throws IOException {
```
bogdandrutu: Done.

core/src/test/java/com/google/instrumentation/trace/PropagationUtilTest.java
bogdandrutu: Done.

core/src/test/java/com/google/instrumentation/trace/PropagationUtilTest.java
bogdandrutu: Done.

core/src/test/java/com/google/instrumentation/trace/PropagationUtilTest.java
bogdandrutu: Done.

core/src/main/java/com/google/instrumentation/trace/PropagationUtil.java
```diff
@@ -30,139 +28,76 @@
  * <li>version_id: 1-byte representing the version id.
  * <li>For version_id = 0:
  *     <ul>
- *     <li>version_format: &lt;trace-id&gt;&lt;span-id&gt;&lt;trace-options&gt;
- *     <li>trace-id: 16-byte array representing the trace_id.
- *     <li>span-id: 8-byte array representing the span_id.
- *     <li>trace-options: 4-byte array representing the trace_options. It is in little-endian order,
- *         if represented as an int.
+ *     <li>version_format: &lt;field&gt;&lt;field&gt;...
```
bogdandrutu: What should I say? Any proposal?

sebright: I'm not sure.  I thought it was something like BNF at first, but then I didn't understand "`TraceId: (field_id = 0, len = 16, default = "0000000000000000")`".  Maybe it would help if you emphasized that TraceId, SpanId, and TraceOptions are the fields described in the first line, for example:

```
...
* field_format: <field_id><field_format>
* Fields:
  0. TraceId: 16-byte array representing the trace_id.  Default = "0000000000000000"
  1. SpanId: 8-byte array representing the span_id.  Default = "00000000"
  2. TraceOptions: 1-byte array representing the trace_options. Default = "0"
* Fields MUST be encoded using the field id order (smaller to higher).
...
```

bogdandrutu: Done.

sebright: The numbering in the Javadoc starts at 1, so you might need to use `<ol start="0">` or remove the numbering.

bogdandrutu:
Everything is done. Thanks for the review.

zhangkun83:
Since the size of the byte array returned by `getBytes()` is documented on the javadoc, it should be considered as part of the API guarantee and any change to the size is API-breaking. Then why not just return a byte value?

zhangkun83:
Also, `ServerStats.create()` requires a byte as the "traceOption", which further solidifies the one-byte option format. There isn't any good reason to keep using a `byte[]` here.

PR #198, opened 2017-04-10
Address comments for Distribution class

songy23:
Merged code changes from @a-veitch 's PR  #196, made some modifications and addressed most of the review comments from @sebright. Also added a new BucketBoundaries class for storing the histogram bucket boundaries.

### DIFF
--- a/core/src/main/java/com/google/instrumentation/stats/BucketBoundaries.java
+++ b/core/src/main/java/com/google/instrumentation/stats/BucketBoundaries.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2017, Google Inc.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.instrumentation.stats;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * The optional histogram bucket boundaries for a {@link Distribution}.
+ */
+public final class BucketBoundaries {
+  private final List<Double> bucketBoundaries;
+
+  /**
+   * @param bucketBoundaries the boundaries for the buckets in the underlying {@link Distribution}.
+   * @return a new {@code BucketBoundaries} with the specified boundaries.
+   * @throws NullPointerException if {@code bucketBoundaries} is null.
+   * @throws IllegalArgumentException if {@code bucketBoundaries} is not sorted, or has zero length.
+   */
+  public static final BucketBoundaries create(List<Double> bucketBoundaries) {
+    checkNotNull(bucketBoundaries, "bucketBoundaries list should not be null.");
+    List<Double> bucketBoundariesCopy = new ArrayList<Double>(bucketBoundaries);  // Deep copy.
+    checkArgument(bucketBoundariesCopy.size() > 0, "Zero length bucket boundaries");
+    double lower = bucketBoundariesCopy.get(0);
+    for (int i = 1; i < bucketBoundariesCopy.size(); i++) {
+      double next = bucketBoundariesCopy.get(i);
+      checkArgument(lower < next, "Bucket boundaries not sorted.");
+      lower = next;
+    }
+
+    return new BucketBoundaries(Collections.unmodifiableList(bucketBoundariesCopy));
+  }
+
+  /**
+   * @return a list of histogram bucket boundaries.
+   */
+  public final List<Double> getBoundaries() {
+    return bucketBoundaries;
+  }
+
+  private BucketBoundaries(List<Double> bucketBoundaries) {
+    this.bucketBoundaries = bucketBoundaries;
+  }
+}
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/google/instrumentation/stats/Distribution.java
```diff
@@ -0,0 +1,230 @@
...
+  }
+
+  /**
+   * Constructs a new {@link Distribution} with specified {@code bucketBoundaryList}.
+   *
+   * @param bucketBoundaryList the boundaries for the buckets in the underlying
+   * {@code Distribution}.
+   * @return a new, empty {@code Distribution} with the specified boundaries.
+   */
+  public static final Distribution create(List<Double> bucketBoundaryList) {
```
sebright: This method should still check the argument for null.

core/src/main/java/com/google/instrumentation/stats/Distribution.java
```diff
@@ -0,0 +1,230 @@
...
+    return (bucketBoundaries != null);
+  }
+
+  // Construct a new Distribution with given boundaries.
+  private Distribution(BucketBoundaries bucketBoundaries) {
+    this.bucketBoundaries = bucketBoundaries;
+    if (hasBuckets()) {
+      bucketCounts = new long[bucketBoundaries.getList().size() + 1];
+      for (int i = 0; i < bucketCounts.length; i++) {
+        bucketCounts[i] = 0L;
```
sebright: Initializing the elements isn't necessary now that they are primitive longs.

core/src/main/java/com/google/instrumentation/stats/Distribution.java
```diff
@@ -0,0 +1,230 @@
...
+ *   <li>[-infinity, bounds[i]) for i == 0
+ *   <li>[bounds[i-1], bounds[i]) for 0 &lt; i &lt; N-1
+ *   <li>[bounds[i-1], +infinity) for i == N-1
+ * </ul>
+ *
+ * <p>i.e. an underflow bucket (number 0), zero or more finite buckets (1 through N - 2, and an
+ * overflow bucket (N - 1), with inclusive lower bounds and exclusive upper bounds.
+ *
+ * <p>Note: If N = 1, there are no finite buckets and the single bucket is both the overflow and
+ * underflow bucket.
```
sebright: Could you please add a TODO to determine whether N = 1 is valid?  Then we should make this comment consistent with the validation of the bucket boundaries.

core/src/main/java/com/google/instrumentation/stats/Distribution.java
```diff
@@ -0,0 +1,230 @@
...
+   * bucket has number 0; the finite buckets, if any, have numbers 1 through N-2; the overflow
+   * bucket has number N-1).
+   *
+   * <p>The size of {@link #getBucketCounts()} must be no greater than N as defined in {@link
+   * BucketBoundaries}.
+   *
+   * <p>Any suffix of trailing buckets containing only zero may be omitted.
+   *
+   * <p>{@link #getBucketCounts()} will return null iff the associated {@link BucketBoundaries}
+   * returns null.
```
sebright: I think this should be "is null" instead of "returns null".

core/src/main/java/com/google/instrumentation/stats/BucketBoundaries.java
```diff
@@ -0,0 +1,62 @@
...
+   * @param bucketBoundaries the boundaries for the buckets in the underlying {@link Distribution}.
+   * @return a new {@code BucketBoundaries} with the specified boundaries.
+   * @throws NullPointerException if {@code bucketBoundaries} is null.
+   * @throws IllegalArgumentException if {@code bucketBoundaries} is not sorted, or has zero length.
+   */
+  static final BucketBoundaries create(List<Double> bucketBoundaries) {
+    checkNotNull(bucketBoundaries, "bucketBoundaries");
+    checkArgument(bucketBoundaries.size() > 0, "Zero length bucket boundaries");
+
+    List<Double> bucketBoundariesCopy = new ArrayList<Double>(bucketBoundaries.size());
```
sebright: The list could be copied more efficiently with `List<Double> bucketBoundariesCopy = new ArrayList<Double>(bucketBoundaries);`.

core/src/main/java/com/google/instrumentation/stats/BucketBoundaries.java
```diff
@@ -0,0 +1,61 @@
...
+
+  /**
+   * @param bucketBoundaries the boundaries for the buckets in the underlying {@link Distribution}.
+   * @return a new {@code BucketBoundaries} with the specified boundaries.
+   * @throws NullPointerException if {@code bucketBoundaries} is null.
+   * @throws IllegalArgumentException if {@code bucketBoundaries} is not sorted, or has zero length.
+   */
+  static final BucketBoundaries create(List<Double> bucketBoundaries) {
+    checkNotNull(bucketBoundaries, "bucketBoundaries list should not be null.");
+    checkArgument(bucketBoundaries.size() > 0, "Zero length bucket boundaries");
```
sebright: Now this check isn't necessary.

core/src/main/java/com/google/instrumentation/stats/BucketBoundaries.java
```diff
@@ -0,0 +1,62 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * The optional histogram bucket boundaries for a {@link Distribution}.
+ */
+final class BucketBoundaries {
```
sebright: When a public method of a public class uses a type in its signature, that type should also be public, to make the API easier to use.  We can make this class public later, when we need to expose it in the API.

songy23: That makes sense, now that we're using `BucketBoundaries` to construct `Distribution` and `DistributionAggregationDescriptor`, I'll make this class public.

core/src/main/java/com/google/instrumentation/stats/BucketBoundaries.java
songy23: Removed.

songy23:
Hi @sebright, I've rebase and squash the commits. If you got time please feel free to take a look again. Thx

PR #199, opened 2017-04-11
Use the new proposed encoding and remove protobuf dependency for stats-impl.

songy23:


core_impl/src/main/java/com/google/instrumentation/stats/StatsSerializer.java
```diff
@@ -31,52 +29,81 @@
...
+  // Deserializes input to StatsContext based on the binary format standard.
+  // The encoded tags are of the form: <version_id><encoded_tags>
   static StatsContextImpl deserialize(InputStream input) throws IOException {
     try {
-      StatsContextProto.StatsContext context = StatsContextProto.StatsContext.parseFrom(input);
-      ByteBuffer buffer = context.getTags().asReadOnlyByteBuffer();
+      byte[] bytes = ByteStreams.toByteArray(input);
       HashMap<String, String> tags = new HashMap<String, String>();
+      if (bytes.length == 0) {
+        // Return a default StatsContext on empty input.
```
sebright: This comment seems inconsistent with the documentation at the top, since an empty byte array doesn't contain the version.

core_impl/src/test/java/com/google/instrumentation/stats/StatsContextFactoryTest.java
```diff
@@ -29,14 +31,19 @@
  */
 @RunWith(JUnit4.class)
 public class StatsContextFactoryTest {
```
sebright: It looks like there aren't any tests that directly test the output from the serialize method.  I think there should also be some tests involving multiple tags.

core_impl/src/main/java/com/google/instrumentation/stats/StatsSerializer.java
```diff
@@ -31,52 +29,81 @@
...
-  //    The high 6 bits are reserved for future use. The value_bytes type is
-  //    encoded as:
-  //    00 (value 0): string (UTF-8) encoding
-  //    01 (value 1): integer (varint int64 encoding).
-  //    10 (value 2): boolean format.
-  //    11 (value 3): reserved for future use.
+
+  // This describes the encoding of stats context information (tags)
+  // for passing across RPC's.
+  // Tags are encoded in single byte sequence. The format is:
```
sebright: Sorry, I meant "version 0".

core_impl/src/test/java/com/google/instrumentation/stats/StatsContextFactoryTest.java
sebright: > Since StatsSerializer.serialize() will only be called by StatsContextImpl.serialize(), tests for serialize() will be under StatsContextTest (https://github.com/google/instrumentation-java/blob/master/core_impl/src/test/java/com/google/instrumentation/stats/StatsContextTest.java#L143).

The tests that are currently in StatsContextTest.java are all round-trip tests, so they aren't testing the encoding.  I meant that it would be nice to have some tests that create a StatsContext, serialize it, and compare it against a byte[] with the expected encoding.  They could also go in StatsContextTest.java.

> I'll add tests for mixed tags.

I meant that we should have a test for multiple String tags, since we can't test the other types yet.

core_impl/src/test/java/com/google/instrumentation/stats/StatsContextFactoryTest.java
```diff
@@ -93,18 +101,34 @@ private static StatsContext testDeserialize(InputStream inputStream) throws IOEx
...
+    for (int i = 0; i <= VALUE_TYPE_FALSE; i++) {
+      encodeSingleTypeTagToOutputStream(i, byteArrayOutputStream);
+    }
+    return new ByteArrayInputStream(byteArrayOutputStream.toByteArray());
+  }
+  
+  private static void encodeSingleTypeTagToOutputStream(
+      int valueType, ByteArrayOutputStream byteArrayOutputStream) throws IOException {
+    byteArrayOutputStream.write(valueType);
+    encodeString(KEY + valueType, byteArrayOutputStream);  // encode <tag_key_len><tag_key>
```
sebright: Does this line append `valueType` to create a unique tag key?  I think it would help to mention that in a comment.

core_impl/src/main/java/com/google/instrumentation/stats/StatsSerializer.java
```diff
@@ -31,52 +29,81 @@
...
+  //         <tag_val> == tag_val_len bytes comprising UTF-8 string
+  //     <tag_encoding> (tag_field_id == 1) ==
+  //       <tag_key_len><tag_key><int_tag_val>
+  //         <tag_key_len> == varint encoded integer
+  //         <tag_key> == tag_key_len bytes comprising tag key name
+  //         <int_tag_value> == 8 bytes, little-endian integer
+  //     <tag_encoding> (tag_field_id == 2 || tag_field_id == 3) ==
+  //       <tag_key_len><tag_key>
+  //         <tag_key_len> == varint encoded integer
+  //         <tag_key> == tag_key_len bytes comprising tag key name
```
bogdandrutu: The only requirement is that newly added ids must be added at the end (appended) because the old implementations don't know about the format of that field. For the initial known tags no restriction for that.

Also you may want to look at PropagationUtil (from trace) on how to structure/format the comment to be consistent between these.

core_impl/src/main/java/com/google/instrumentation/stats/StatsSerializer.java
bogdandrutu: for empty byte array I would return a ParseException. Also I think it is better to use ParseException instead of IOException to be consistent with the trace.

core_impl/src/main/java/com/google/instrumentation/stats/StatsSerializer.java
songy23: Style fixed.

core_impl/src/main/java/com/google/instrumentation/stats/StatsSerializer.java
songy23: Sure, updated to throw ParseException instead.

core_impl/src/test/java/com/google/instrumentation/stats/StatsContextFactoryTest.java
songy23: Done.

core_impl/src/test/java/com/google/instrumentation/stats/StatsContextFactoryTest.java
songy23: Done.

core_impl/src/main/java/com/google/instrumentation/stats/StatsSerializer.java
sebright: StatsSerializer is reading and writing from OutputStream and InputStream, which can do IO.  I don't think we should turn any of the IOExceptions into ParseExceptions.  We should probably throw both ParseException and IOException from `deserialize` and IOException from `serialize`.  However, gRPC is already calling `deserialize`, so we would have to coordinate the change with them: https://github.com/grpc/grpc-java/blob/c8dff0d974631822978a7d637c6941ed9545ddd2/core/src/main/java/io/grpc/internal/CensusStreamTracerModule.java#L115.  We could provide methods that work with byte[] that don't throw IOException, if necessary.

core/src/main/java/com/google/instrumentation/stats/StatsContext.java
```diff
@@ -57,7 +57,7 @@ public final StatsContext with(
    *
    * @param output the {@link OutputStream} to add the serialized form of this {@link StatsContext}.
    */
-  public abstract void serialize(OutputStream output) throws IOException;
+  public abstract void serialize(OutputStream output) throws ParseException;
```
sebright: I don't think `serialize` should throw a ParseException.  (See my comment in StatsSerializer.)

core_impl/src/main/java/com/google/instrumentation/stats/StatsContextImpl.java
```diff
@@ -40,10 +40,10 @@ public StatsContextImpl record(MeasurementMap stats) {
   /**
    * Serializes a {@link StatsContextImpl} into {@code CensusContextProto} serialized format.
    *
-   * <p>The encoded tags are of the form: {@code <tag prefix> + 'key' + <tag delim> + 'value'}*
+   * <p>The encoded tags are of the form: {@code <version_id>;<encoded_tags>;}
```
sebright: Same here, I don't think the semicolons are necessary.

core_impl/src/main/java/com/google/instrumentation/stats/StatsSerializer.java
```diff
@@ -21,38 +21,50 @@
...
+ * <p>Encoding of stats context information (tags) for passing across RPC's:</p>
+ * 
+ * <ul>
+ *   <li>Tags are encoded in single byte sequence. The version 0 format is:
+ *   <li>{@code <version_id>;<encoded_tags>}
+ *   <li>{@code <version_id>; == a single byte, value 0}
+ *   <li>{@code <encoded_tags>; == (<tag_field_id>;<tag_encoding>;)*}
+ *       <ul>
+ *       <li>{@code <tag_field_id>} == a single byte, value 0 (string tag), 1 (int tag),
+ *           2 (true bool tag), 3 (false bool flag)
```
sebright: s/flag/tag

core_impl/src/main/java/com/google/instrumentation/stats/StatsSerializer.java
```diff
@@ -21,38 +21,50 @@
...
 /**
  * Native implementation {@link StatsContext} serialization.
+ * 
+ * <p>Encoding of stats context information (tags) for passing across RPC's:</p>
+ * 
+ * <ul>
+ *   <li>Tags are encoded in single byte sequence. The version 0 format is:
+ *   <li>{@code <version_id>;<encoded_tags>}
+ *   <li>{@code <version_id>; == a single byte, value 0}
+ *   <li>{@code <encoded_tags>; == (<tag_field_id>;<tag_encoding>;)*}
```
sebright: Same here

core_impl/src/main/java/com/google/instrumentation/stats/StatsContextFactoryImpl.java
```diff
@@ -26,10 +26,10 @@
   /**
    * Deserializes a {@link StatsContextImpl} from a serialized {@code CensusContextProto}.
    *
-   * <p>The encoded tags are of the form: {@code <tag prefix> + 'key' + <tag delim> + 'value'}*
+   * <p>The encoded tags are of the form: {@code <version_id>;<encoded_tags>;}
```
songy23: Removed. Seems to be a copy and paste problem.

core_impl/src/main/java/com/google/instrumentation/stats/StatsSerializer.java
songy23: Done.

core/src/main/java/com/google/instrumentation/stats/StatsContext.java
songy23: Yeah that also makes sense. I'll update `serialize` to throw `IOException` as before.

songy23:
Talk with @sebright offline, we would still use IOException for serialize and deserialize for now (for backwards consistency), and might discuss whether to use ParseException later.

PR #200, opened 2017-04-11
Start implementing StatsManagerImpl.record(...).

sebright:
Changes in this PR:
- Store `StatsContextImpl`'s map of tags in an unmodifiable map so that it is safe to access from other classes.
- Add a field for an event queue to `StatsManagerImpl`.
- Create an `EventQueue` interface to allow testing with a simple `EventQueue`.  I renamed the original `EventQueue` class to `DisruptorEventQueue`.
- Add a `StatsEvent` class that can be added to the `EventQueue`.
- Add a record method to `StatsManagerImpl`.  It doesn't do anything yet.

It is much easier to review the individual commits, because git didn't follow the rename of `EventQueue`.

/cc @a-veitch @dinooliva @bogdandrutu

### DIFF
--- a/core_impl/src/main/java/com/google/instrumentation/stats/MeasurementDescriptorToViewMap.java
+++ b/core_impl/src/main/java/com/google/instrumentation/stats/MeasurementDescriptorToViewMap.java
@@ -17,6 +17,7 @@ import com.google.common.collect.HashMultimap;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Multimaps;
 import java.util.Collection;
+import java.util.Map;
 
 /**
  * A class that stores a singleton map from {@link MeasurementDescriptor.Name} to {@link View}.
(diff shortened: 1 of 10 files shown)


core_impl/src/main/java/com/google/instrumentation/stats/MeasurementDescriptorToViewMap.java
```diff
@@ -52,4 +53,23 @@ final View getView(ViewDescriptor viewDescriptor) {
   void putView(MeasurementDescriptor.Name name, View view) {
     mutableMap.put(name, view);
   }
+
+  // Records stats with a set of tags.
+  void record(Map<String, String> tags, MeasurementMap stats) {
+    for (MeasurementValue mv : stats) {
+      if (mv.getMeasurement()
```
songy23: Could we check for supported measure descriptor before we enqueue a new `StatsEvent` (instead of checking here)?

core_impl/src/main/java/com/google/instrumentation/stats/StatsManagerImpl.java
```diff
@@ -81,6 +82,35 @@ StatsContextFactoryImpl getStatsContextFactory() {
...
+   */
+  void record(StatsContextImpl tags, MeasurementMap measurementValues) {
+    queue.enqueue(new StatsEvent(measurementDescriptorToViewMap, tags.tags, measurementValues));
+  }
+
+  // An EventQueue entry that records the stats from one call to StatsManager.record(...).
+  private static final class StatsEvent implements EventQueueEntry {
+    private final Map<String, String> tags;
+    private final MeasurementMap stats;
+    private final MeasurementDescriptorToViewMap statsMap;
```
songy23: I'm not sure whether we would like to use `StatsEvent` for `registerView()` and `getView()` as well. If that's the case, it would be better to store a reference to `StatsManagerImpl` instead of just `MeasurementDescriptorToViewMap`.

sebright: Done

sebright:
Thanks!

PR #201, opened 2017-04-12
Add a new class Trace which holds the global state for the trace implementation.

bogdandrutu:
* Tracer changed to not contain any global state.
* PropagationUtil changed to not contain any global state.
* TraceService contains the global instances for Tracer and BinaryPropagationHandler.
* PropagationUtil split between BinaryPropagationHandler (for the interface definition) and BinaryPropagationHandlerImpl (for the implementation).

Major changes:
* TraceService/TraceServiceImpl
* Tracer no longer load the implementation using reflection.
* PropagationUtil split.
* Everything else is code moved.

### DIFF
--- a/core/src/main/java/com/google/instrumentation/trace/Trace.java
+++ b/core/src/main/java/com/google/instrumentation/trace/Trace.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2016, Google Inc.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.instrumentation.trace;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.instrumentation.common.Provider;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/** Class that manages a global instance of the {@link TraceService}. */
+public final class Trace {
+  private static final Logger logger = Logger.getLogger(Tracer.class.getName());
+  private static final TraceService traceService =
+      loadTraceService(Provider.getCorrectClassLoader(TraceService.class));
+
+  /**
+   * Returns the global {@link Tracer}.
+   *
+   * @return the global {@code Tracer}.
+   */
+  public static Tracer getTracer() {
+    return traceService.getTracer();
+  }
+
+  /**
+   * Returns the global {@link BinaryPropagationHandler}.
+   *
+   * @return the global {@code BinaryPropagationHandler}.
+   */
+  public static BinaryPropagationHandler getBinaryPropagationHandler() {
+    return traceService.getBinaryPropagationHandler();
+  }
+
+  // Any provider that may be used for TraceService can be added here.
+  @VisibleForTesting
+  static TraceService loadTraceService(ClassLoader classLoader) {
+    try {
+      // Call Class.forName with literal string name of the class to help shading tools.
+      return Provider.createInstance(
+          Class.forName("com.google.instrumentation.trace.TraceServiceImpl", true, classLoader),
+          TraceService.class);
+    } catch (ClassNotFoundException e) {
+      logger.log(Level.FINE, "Using default implementation for TraceService.", e);
+    }
+    return TraceService.getNoopTraceService();
+  }
+}
--- a/core/src/main/java/com/google/instrumentation/trace/TraceService.java
+++ b/core/src/main/java/com/google/instrumentation/trace/TraceService.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2016, Google Inc.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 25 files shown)


core/src/main/java/com/google/instrumentation/trace/TraceService.java
```diff
@@ -0,0 +1,67 @@
...
+  private final Tracer tracer;
+  private final BinaryPropagationHandler binaryPropagationHandler;
+
+  /**
+   * Returns the {@link Tracer} with the provided implementations. If no implementation is provided
+   * then no-op implementations will be used.
+   *
+   * @return the {@code Tracer} implementation.
+   */
+  public Tracer getTracer() {
```
adriancole: how do they know if it is noop?

core/src/main/java/com/google/instrumentation/trace/Trace.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+package com.google.instrumentation.trace;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.instrumentation.common.Provider;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/** Class that manages a global instance of the {@link TraceService}. */
```
adriancole: are objects returned here safe to cache? If so, maybe note that.

Also, static singletons is to some an anti-pattern reserved for edge cases where there's no other option. Is that the thinking here? a "last resort" or are you expecting tracing users to instantiate a JVM singleton normally? Either way, might be worthwhile documenting the intended use.

core/src/main/java/com/google/instrumentation/trace/TraceService.java
```diff
@@ -0,0 +1,67 @@
...
+ * limitations under the License.
+ */
+
+package com.google.instrumentation.trace;
+
+/**
+ * Class that holds the implementation instances for {@link Tracer} and {@link
+ * BinaryPropagationHandler}.
+ *
+ * <p>This implementation of this class is loaded using reflection by the {@link Trace}.
```
adriancole: is this intentionally the only pattern? or is it likely that in dependency injection, like Guice or Dagger, one would return a TraceService as an object graph that includes the tracer and propagation handler? Or would we expect in DI, that they just configure and inject Tracer and propagation handler directly (which implies TraceService is only used for JVM/classloader singletons)

core/src/main/java/com/google/instrumentation/trace/TraceService.java
```diff
@@ -0,0 +1,67 @@
...
+ */
+public abstract class TraceService {
```
adriancole: if this is only used for testing, I'd be inclined to inline its code back into Trace.

you can use Powermock to mock static methods. That eliminates the need to explain the otherwise public supporting type here. Powermock can do fancy things as it often actually wraps the callsite as opposed to changing the target. We use it in zipkin sometimes.. I like it a lot better than putting test code in main, but it is a judgement call.

ex.

```java
@RunWith(PowerMockRunner.class)
@PrepareForTest({MyCode.class}) // implicitly intercepts the call site
public class PlatformTest {
  MyCode code = new CodeThatUsesTraceDotGetTracer();
  Tracer tracer = newFooTracer();

  @Before public void setup() {
    mockStatic(Trace.class);
    when(Trace.getTracer()).thenReturn(mockTracer);
   }
```

```xml
    <dependency>
      <groupId>org.powermock</groupId>
      <artifactId>powermock-api-mockito</artifactId>
      <version>${powermock.version}</version>
      <scope>test</scope>
    </dependency>
```

core/src/main/java/com/google/instrumentation/trace/TraceService.java
bogdandrutu: Why is that important? We can always add a isNoopTracer method but I would like to hear what is the use-case.

core/src/main/java/com/google/instrumentation/trace/Trace.java
bogdandrutu: Yes objects are safe to cache and encouraged to be used as Logger:
private static final Tracer tracer = Trace.getTracer();

A lot of blog-posts about why singletons are "evil". An alternative of this would be to manually propagate an instance of a TraceService within your code, which seems much worse from an integration point of view.

The reason why we went the path of reflection loading is that we expect users to do 0 initialization (just link the implementation to their binary), or minimum initialization work.

core/src/main/java/com/google/instrumentation/trace/TraceService.java
bogdandrutu: I am not that familiar with Dagger (or Guice) (too much Java involved). I think it should be fine to allow users to inject the implementation using this frameworks. Do I have to change anything to allow users to do that?

core/src/main/java/com/google/instrumentation/trace/TraceService.java
bogdandrutu: So one reason to have a separate class that holds the instances instead of having Tracer keeping it's own instance and propagation handler doing the same is because we try to have only one class that keeps the "global state".

Another reason is that we want to add more functionality to this class like:
* let users register "exporters"
* etc.
that in my opinion does not belong to tracer (or it may belong but then Tracer become a TraceService which includes multiple different functionalities).

In the current desing Tracer deals with Span creation and propagation in the process which are the most used functionalities.

core/src/test/java/com/google/instrumentation/trace/ScopedSpanHandleTest.java
```diff
@@ -28,7 +28,7 @@
 /** Unit tests for {@link ScopedSpanHandle}. */
 @RunWith(JUnit4.class)
 public class ScopedSpanHandleTest {
-  private static final Tracer tracer = Tracer.getTracer();
+  private static final Tracer tracer = Trace.getTracer();
```
sebright: Suggestion: You could call `Tracer.getNoopTracer()` (or a constructor if it's not the NoopTracer) to ensure that you use the right class.

PR #206, opened 2017-04-13
Split Distribution into mutable and immutable classes.

sebright:
Distribution is an immutable AutoValue class, and MutableDistribution is a
mutable version of Distribution.  We can use MutableDistribution to store stats
as they are recorded and then copy them into a Distribution before returning
them in a View.  We can also use Distribution to store the distribution data in
DistributionAggregation.

I copied/renamed files in separate commits to try to make the diffs smaller.

/cc @a-veitch @dinooliva

### DIFF
--- a/core/src/main/java/com/google/instrumentation/stats/Distribution.java
+++ b/core/src/main/java/com/google/instrumentation/stats/Distribution.java
@@ -28,8 +28,8 @@ import javax.annotation.concurrent.Immutable;
  * buckets.
  *
  * <p>The bucket boundaries for that histogram are described by {@link BucketBoundaries}, which
- * defines {@code BucketBoundaries.getBoundaries().size() + 1 (= N)} buckets.
- * The boundaries for bucket index i are:
+ * defines {@code BucketBoundaries.getBoundaries().size() + 1 (= N)} buckets. The boundaries for
+ * bucket index i are:
  *
  * <ul>
  *   <li>[-infinity, bounds[i]) for i == 0
@@ -42,11 +42,11 @@ import javax.annotation.concurrent.Immutable;
  *
  * <p>Note: If N = 1, there are no finite buckets and the single bucket is both the overflow and
  * underflow bucket.
- * TODO(songya): needs to determine whether N = 1 is valid. (Currently we don't allow N = 1.)
  *
  * <p>Although not forbidden, it is generally a bad idea to include non-finite values (infinities or
  * NaNs) in the population of values, as this will render the {@code mean} meaningless.
  */
+//  TODO(songya): needs to determine whether N = 1 is valid. (Currently we don't allow N = 1.)
 @Immutable
 @AutoValue
 public abstract class Distribution {
@@ -122,9 +122,8 @@ public abstract class Distribution {
   /**
    * A Distribution may optionally contain a histogram of the values in the population. The
    * histogram is given in {@link #getBucketCounts()} as counts of values that fall into one of a
-   * sequence of non-overlapping buckets, described by {@link BucketBoundaries}.
-   * The sum of the values in {@link #getBucketCounts()} must equal the value in
-   * {@link #getCount()}.
+   * sequence of non-overlapping buckets, described by {@link BucketBoundaries}. The sum of the
+   * values in {@link #getBucketCounts()} must equal the value in {@link #getCount()}.
    *
    * <p>Bucket counts are given in order under the numbering scheme described above (the underflow
    * bucket has number 0; the finite buckets, if any, have numbers 1 through N-2; the overflow
@@ -135,8 +134,8 @@ public abstract class Distribution {
    *
    * <p>Any suffix of trailing buckets containing only zero may be omitted.
    *
-   * <p>{@link #getBucketCounts()} will return null iff the associated {@link BucketBoundaries}
-   * is null.
+   * <p>{@link #getBucketCounts()} will return null iff the associated {@link BucketBoundaries} is
+   * null.
    *
    * @return The count of population values in each histogram bucket.
    */
--- a/core/src/main/java/com/google/instrumentation/stats/Distribution.java
+++ b/core/src/main/java/com/google/instrumentation/stats/Distribution.java
@@ -15,7 +15,9 @@ package com.google.instrumentation.stats;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
+import com.google.auto.value.AutoValue;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/google/instrumentation/stats/Distribution.java
```diff
@@ -44,63 +46,51 @@
...
  * NaNs) in the population of values, as this will render the {@code mean} meaningless.
  */
-public final class Distribution {
-  private long count = 0; // The number of values in the population.
-  private double sum = 0.0; // The sum of the values in the population.
-  private Range range = Range.create(); // Range of values in the population.
-  private final BucketBoundaries bucketBoundaries; // Histogram boundaries; null means no histogram
-  private long[] bucketCounts; // Counts for each histogram bucket
+@AutoValue
+public abstract class Distribution {
```
bogdandrutu: Add "@Immutable" annotation.

core/src/main/java/com/google/instrumentation/stats/Distribution.java
```diff
@@ -44,63 +46,51 @@
...
-    return (bucketBoundaries != null);
-  }
-
-  // Construct a new Distribution with optional bucket boundaries.
-  private Distribution(@Nullable BucketBoundaries bucketBoundaries) {
-    this.bucketBoundaries = bucketBoundaries;
-    if (hasBuckets()) {
-      bucketCounts = new long[bucketBoundaries.getBoundaries().size() + 1];
+  private static List<Long> longArrayToImmutableList(long[] array) {
+    List<Long> boxedBucketCounts = new ArrayList<Long>(array.length);
```
bogdandrutu: return Collections.unmodifiableList(Arrays.asList(array))?

core/src/main/java/com/google/instrumentation/stats/MutableDistribution.java
```diff
@@ -0,0 +1,215 @@
...
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import javax.annotation.Nullable;
+
+/**
+ * Mutable version of {@link Distribution}.
```
bogdandrutu: Please run the google-java-formatter (will make this comment one line)

core/src/main/java/com/google/instrumentation/stats/MutableDistribution.java
```diff
@@ -0,0 +1,215 @@
...
+public final class MutableDistribution {
+  private long count = 0; // The number of values in the population.
+  private double sum = 0.0; // The sum of the values in the population.
+  private Range range = Range.create(); // Range of values in the population.
+
+  @Nullable
+  private final BucketBoundaries bucketBoundaries; // Histogram boundaries; null means no histogram
+
+  @Nullable
+  private long[] bucketCounts; // Counts for each histogram bucket
```
bogdandrutu: Why not final? You don't modify the reference do you?

core/src/main/java/com/google/instrumentation/stats/MutableDistribution.java
```diff
@@ -0,0 +1,215 @@
...
+
+  // Returns true if the distribution has histogram buckets.
+  private boolean hasBuckets() {
+    return (bucketBoundaries != null);
+  }
+
+  // Construct a new MutableDistribution with optional bucket boundaries.
+  private MutableDistribution(@Nullable BucketBoundaries bucketBoundaries) {
+    this.bucketBoundaries = bucketBoundaries;
+    if (hasBuckets()) {
```
bogdandrutu: I feel that is better in the constructor to not call hasBuckets() because it may change to be bucketCounts != null and the code will break.

bucketCounts = bucketBoundaries != null ? new long[bucketBoundaries.getBoundaries().size() + 1] : null;

core/src/main/java/com/google/instrumentation/stats/MutableDistribution.java
```diff
@@ -0,0 +1,215 @@
...
+   * The bucket counts of the optional histogram.
+   *
+   * @see Distribution#getBucketCounts()
+   * @return The count of population values in each histogram bucket, or {@code null} if there is no
+   *     histogram.
+   */
+  @Nullable
+  public List<Long> getBucketCounts() {
+    if (hasBuckets()) {
+      List<Long> boxedBucketCounts = new ArrayList<Long>(bucketCounts.length);
```
bogdandrutu: Same comment about Arrays.asList.

core/src/main/java/com/google/instrumentation/stats/MutableDistribution.java
```diff
@@ -0,0 +1,215 @@
...
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import javax.annotation.Nullable;
+
+/**
+ * Mutable version of {@link Distribution}.
+ */
+public final class MutableDistribution {
```
bogdandrutu: Do you need this class to be public?

core/src/main/java/com/google/instrumentation/stats/MutableDistribution.java
```diff
@@ -0,0 +1,215 @@
...
+  // Construct a new MutableDistribution with optional bucket boundaries.
+  private MutableDistribution(@Nullable BucketBoundaries bucketBoundaries) {
+    this.bucketBoundaries = bucketBoundaries;
+    if (hasBuckets()) {
+      bucketCounts = new long[bucketBoundaries.getBoundaries().size() + 1];
+    }
+  }
+
+  /**
+   * The number of values in the population. Must be non-negative.
```
bogdandrutu: I would start this sentence with:

Returns the number ....
https://google.github.io/styleguide/javaguide.html#s7.2-summary-fragment

core/src/test/java/com/google/instrumentation/stats/MutableDistributionTest.java
```diff
@@ -0,0 +1,147 @@
...
+public class MutableDistributionTest {
+
+  @Rule
+  public final ExpectedException thrown = ExpectedException.none();
+
+  @Test
+  public void testEmptyDistribution() {
+    MutableDistribution empty = MutableDistribution.create();
+    assertThat(empty.getCount()).isEqualTo(0);
+    assertThat(empty.getSum()).isWithin(1e-5).of(0.0);
```
songy23: Minor: could you make `1e-5` a constant? Sorry I forgot to do that.

core/src/main/java/com/google/instrumentation/stats/MutableDistribution.java
```diff
@@ -0,0 +1,215 @@
...
+   *     MutableDistribution}.
+   * @return a new, empty {@code MutableDistribution} with the specified boundaries.
+   */
+  public static final MutableDistribution create(BucketBoundaries bucketBoundaries) {
+    checkNotNull(bucketBoundaries, "bucketBoundaries Object should not be null.");
+    return new MutableDistribution(bucketBoundaries);
+  }
+
+  // Returns true if the distribution has histogram buckets.
+  private boolean hasBuckets() {
```
songy23: This method might be useful for users, could we make it public?

sebright: I'd like to ask @a-veitch before we expose it.  I added a TODO for now.

core/src/test/java/com/google/instrumentation/stats/MutableDistributionTest.java
sebright: Done

adriancole:
I like this. I'd use it. Nice work!

core/src/main/java/com/google/instrumentation/stats/MutableDistribution.java
bogdandrutu: I think we should start with limited visibility (as much as possible) initially and open them when users request that (they need it).

core/src/main/java/com/google/instrumentation/stats/MutableDistribution.java
bogdandrutu: Sure but I think we should follow the style guide for the comments. Happy to review a different PR for that if you feel that is better.

core/src/main/java/com/google/instrumentation/stats/MutableDistribution.java
sebright: Thanks.  I made both classes package-private.

core/src/main/java/com/google/instrumentation/stats/MutableDistribution.java
sebright: I updated the Javadoc in https://github.com/google/instrumentation-java/pull/206/commits/8cce0516533263ae469ab2b557b86209059e10e8.

sebright:
I rebased and fixed the Error Prone warnings.  Thanks for the review!

PR #208, opened 2017-04-13
Throw ParseException for deserialize.

songy23:
Consider using ParseException instead of IOException for deserialize().

### DIFF
--- a/core_impl/src/main/java/com/google/instrumentation/stats/StatsSerializer.java
+++ b/core_impl/src/main/java/com/google/instrumentation/stats/StatsSerializer.java
@@ -21,6 +21,7 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.nio.BufferUnderflowException;
 import java.nio.ByteBuffer;
+import java.text.ParseException;
 import java.util.HashMap;
 import java.util.Map.Entry;
 import java.util.Set;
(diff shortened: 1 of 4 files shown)


core_impl/src/main/java/com/google/instrumentation/stats/StatsSerializer.java
```diff
@@ -88,18 +89,18 @@ static void serialize(StatsContextImpl context, OutputStream output) throws IOEx
 
   // Deserializes input to StatsContext based on the binary format standard.
   // The encoded tags are of the form: <version_id><encoded_tags>
-  static StatsContextImpl deserialize(InputStream input) throws IOException {
+  static StatsContextImpl deserialize(InputStream input) throws IOException, ParseException {
```
bogdandrutu: We discussed with the gRPC team and they suggested to use byte[] instead of Input[Output]Stream because that is what they need.

songy23: A problem with this: If they want to serialize() and deserialize() with byte[], for serialization they need to pass in an array that's large enough, but then there will probably be trailing 0s, and trying to deserialize that array will throw an exception. How are we supposed to resolve that?

bogdandrutu: I was thinking to have something like this:
byte[] serialize(StatsContext ctxt) {}
StatsContext deserialize(byte[]) {}

songy23: Create a separate issue for this: https://github.com/google/instrumentation-java/issues/213
Will make the code changes in another PR.

adriancole:
Personally i dont like using exception types that are close but no cigar.
For example, parse exception was made for a different but similar purpose.
Style choice of course but i would throw something different if you cannot
parse like illegalargument

adriancole:
Specifically this is a rather obscure text exception. It is valient to try
to reuse types but this one is a miss imho.

PR #210, opened 2017-04-13
Fix a few style issues.

songy23:
Fix style issues in order to import changes.


core_impl/src/test/java/com/google/instrumentation/stats/StatsManagerImplTest.java
```diff
@@ -36,6 +36,7 @@
 
   @Test
   public void testRegisterAndGetView() throws Exception {
+    @SuppressWarnings("unused")
```
bogdandrutu: What about removing this for the moment instead of using suppress warning

songy23: Sure that's also fine, will update this

Issue #41, opened 2016-10-04
Consider providing a no-op CensusContextFactory


zhangkun83:
There will be cases where a Census-instrumented product would want to disable Census, or need a place-holder when Census.getCenusContextFactory() returns null. Yes they could implement their own no-op CensusContextFactory, but since this may be a common use case, it makes sense that Census provides a no-op impl as part of the census-api artfiact.

dinooliva:
I have considered providing a default no-op implementation and have two concerns about do so:

1. We have a set of unit tests that are associated with the APIs and encode the behavior that any implementation of the APIs should provide (e.g. record some stats, grab the data, verify that the correct data is returned). A no-op implementation will not be able to pass those tests (since it will do nothing), which I believe will be confusing (to say the least).

2. Having a no-op implementation means that a client will not notice when their application has been built w/no stats support. I agree that having the application raise an exception during initialization of the stats library is not ideal but it is a clear indication that the application was built/linked incorrectly.

Thoughts?

zhangkun83:
> We have a set of unit tests that are associated with the APIs and encode the behavior that any implementation of the APIs should provide (e.g. record some stats, grab the data, verify that the correct data is returned).

IMHO, the CensusContext API is only a reporting API, it doesn't mandate how the reported data is going to be propagated or used. An no-op implementation doesn't seem to violate the API contract.

There may also be cases where customer wants to disable stats reporting in some cases. It would be nice that they could just switch to a no-op Census factory instead of having an if-branch on every call site to CensusContext.

> Having a no-op implementation means that a client will not notice when their application has been built w/no stats support. 

Not necessarily. Providing an no-op implementation doesn't mean it's used silently. The applications still have choices.  They would try loading a real implementation, and if that fails, have the option to break or fall back to a no-op implementation.

zhangkun83:
We have made our code to skip reporting logic if getStatsContextFactory() returns null, so we don't need the no-op impl any more.

adriancole:
I like return null vs noop as well. It is easier to reason with (in context
of a staticly bound default) and hopefully people dont rely on classloader
singletons routinely anyway.

bogdandrutu:
The problem with returning null vs noop is that every user library like gRPC has to write 3 lines to interact with everything:

StatsContextFactory statsContextFactory = Stats.getStatsContextFactory();
if (statsContextFactory != null) {
  statsContextFactory.DoSomething();
}

[...]

PR #211, opened 2017-04-13
Fix NullPointerException for Stats.getStatsContextFactory()

songy23:
Please note Provider.newInstance() is Nullable.

### DIFF
--- a/core/src/main/java/com/google/instrumentation/stats/Stats.java
+++ b/core/src/main/java/com/google/instrumentation/stats/Stats.java
@@ -14,6 +14,7 @@
 package com.google.instrumentation.stats;
 
 import com.google.instrumentation.common.Provider;
+import javax.annotation.Nullable;
 
 /**
  * {@link Stats}.
@@ -25,6 +26,7 @@ public final class Stats {
   /**
    * Returns the default {@link StatsContextFactory}.
    */
+  @Nullable
   public static StatsContextFactory getStatsContextFactory() {
     return statsManager == null ? null : statsManager.getStatsContextFactory();
   }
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/google/instrumentation/stats/Stats.java
```diff
@@ -26,7 +26,7 @@
    * Returns the default {@link StatsContextFactory}.
    */
   public static StatsContextFactory getStatsContextFactory() {
-    return statsManager.getStatsContextFactory();
+    return statsManager == null ? null : statsManager.getStatsContextFactory();
```
bogdandrutu: Shouldn't we return a no-op factory in this case? That is what we do in trace.

songy23: Create a separate issue for this: https://github.com/google/instrumentation-java/issues/215

bogdandrutu: Please use at least the "@Nullable" annotation.

sebright: There is already an issue requesting a no-op implementation: #41

/cc @zhangkun83

songy23: Fixed.

zhangkun83: @sebright We have made our code to skip reporting logic if `getStatsContextFactory()` returns null, so we don't need the no-op impl any more. I am going to close that issue.

PR #217, opened 2017-04-14
Update serialize and deserialize to use byte[] instead of byte stream.

songy23:
Related issue: https://github.com/google/instrumentation-java/issues/213
This PR is built upon PR https://github.com/google/instrumentation-java/pull/208.

Also need to update any external code that calls serialize() and deserialize(), since the signature of these methods now changes.

### DIFF
--- a/core/src/main/java/com/google/instrumentation/stats/StatsContext.java
+++ b/core/src/main/java/com/google/instrumentation/stats/StatsContext.java
@@ -14,7 +14,6 @@
 package com.google.instrumentation.stats;
 
 import java.io.IOException;
-import java.io.OutputStream;
 
 /**
  * An immutable context for stats operations.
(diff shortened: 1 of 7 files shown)


adriancole:
Do we still need IOException in this case? Seems out of date if the buffer
is a byte array

songy23:
@bogdandrutu Internally gRPC is using Java stats, we have a different implementation for them but it is based on the same interfaces in stats-core.

@adriancole Yeah IOException should be removed if we decided to go with byte array.

songy23:
Thanks @zhangkun83 for the suggestion. I've put back the original serialization interfaces and mark them as Deprecated. I didn't implement them in stats-core though, because I think it would be trivial to copy and paste the implementation from StatsSeriliazer to the core package.

zhangkun83:
gRPC implements `StatsContext` in tests. Unless you implement the new method, it's still an API-breaking change to us. Unless you implement the old method, our test code can't switch to implementing the new method. Therefore, use the mutual-reference implementation seems to be the only viable approach.

songy23:
I talked with @dinooliva on Monday about this. It seems initially we were using byte[] as the input and output for serialization, but later we were asked to use byte stream instead (since stream provides more flexibility over fix-sized arrays). Plus making this change will also need migration from gRPC side. I'm wondering if it's really necessary to switch back to the initial implementation. If no one objects, I'll close this PR and leave issue https://github.com/google/instrumentation-java/issues/213 open to see if this is really needed.

PR #220, opened 2017-04-14
Remove unused import.

songy23:


songy23:
I agree - that would be very helpful

PR #223, opened 2017-04-14
Update initial release instructions and add instructions for deployment.

bogdandrutu:


RELEASING.md
```diff
@@ -120,3 +123,45 @@ github repository.
...
+Deployment to Maven Central (or the snapshot repo) is for all of the artifacts
+from the project.
+
+### Branch
+
+Before building/deploying, be sure to switch to the appropriate branch or tag.
+For the current release use:
+
+```bash
+$ git checkout -b v$MAJOR.$MINOR.$PATCH
```
sebright: Is the `-b` correct?  I thought that you would need to check out the commit at the tag `v$MAJOR.$MINOR.$PATCH` to build the release, but `-b` creates a branch named `v$MAJOR.$MINOR.$PATCH` at the current commit.

bogdandrutu: Actually it worked correctly for me (maybe by luck) https://git-scm.com/docs/git-checkout. So I had one remote branch with that name (which was the tag) so it created a local branch equivalent with the remote branch.

So it depends a bit how people setup their git repo. I have mine set with origin (my fork of the project) and upstream (the main remote main instrumentation-java). If my way to use the repo is the normal one the correct command is:
git checkout -b v$MAJOR.$MINOR.$PATCH --track remote/v$MAJOR.$MINOR.$PATCH

Do you want me to modify to use this command?

bogdandrutu:
PTAL

RELEASING.md
sebright: So these two commands require you to have a remote branch `upstream/v$MAJOR.$MINOR.$PATCH`?  Could you use `git checkout -b v$MAJOR.$MINOR.$PATCH tags/v$MAJOR.$MINOR.$PATCH` so that only the tag is required?  The command only succeeds if the tag exists, so it is less likely to check out the wrong commit.

bogdandrutu: Done. Thanks.

PR #227, opened 2017-04-15
Move Gradle build from "install" to "script" section of .travis.yml.

sebright:
The Gradle build command is the command that should determine whether the build
succeeds or fails, so it should go in the "script" section.  Failed commands in
the "install" section mark the build as "Errored".


sebright:
Thanks!

PR #228, opened 2017-04-17
Fix a few style issues for tests.

songy23:


core_impl/src/test/java/com/google/instrumentation/stats/StatsContextFactoryTest.java
```diff
@@ -31,26 +31,21 @@
 @RunWith(JUnit4.class)
 public class StatsContextFactoryTest {
 
-  private static final int VERSION_ID = 0;
-  private static final int VALUE_TYPE_STRING = 0;
-  private static final int VALUE_TYPE_INTEGER = 1;
-  private static final int VALUE_TYPE_TRUE = 2;
-  private static final int VALUE_TYPE_FALSE = 3;
-
```
sebright: I think it's still important to test the constants.  You could add a test that compares each constant with the expected value.

songy23: Sure, added.

PR #229, opened 2017-04-17
Add a missing @Test annotation.

isturdy:


sebright:
I was wondering why Error Prone didn't catch this.  It looks like it only warns when a public no-arg method starts with "test" but doesn't have a Test annotation (http://errorprone.info/bugpattern/JUnit4TestNotRun), and this method doesn't start with "test".

bogdandrutu:
@sebright I think you are correct and probably we did a poor job naming our test methods. Looking now at the style guide:
https://google.github.io/styleguide/javaguide.html#s5.2.3-method-names

junit4 does not require testing methods to be named testXXX but actually the style guide encourages that.

We may want to fix this later.

sebright:
I thought the style guide was just showing an example of a test naming style.  I don't think we need to start all test methods with "test", unless we decide to add it to the project-specific style guide.  My guess is that Error Prone is just trying to detect JUnit 3 tests (which must start with "test") that weren't properly migrated to JUnit 4.

PR #234, opened 2017-04-19
Add MutableView, modify mapping and implement process.

songy23:
Use MutableView to record stats in distribution, and convert MutableView to View when getView() is called.

This is a skeleton of MutableView, it has not fully been implemented yet, since some of the implementation needs to depend on changes on DistributionAggregation: https://github.com/google/instrumentation-java/pull/233

### DIFF
--- a/core/src/main/java/com/google/instrumentation/stats/MutableView.java
+++ b/core/src/main/java/com/google/instrumentation/stats/MutableView.java
@@ -1,0 +1,155 @@
+/*
+ * Copyright 2017, Google Inc.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.instrumentation.stats;
+
+import com.google.instrumentation.common.Function;
+import com.google.instrumentation.common.Timestamp;
+import com.google.instrumentation.stats.View.DistributionView;
+import com.google.instrumentation.stats.ViewDescriptor.DistributionViewDescriptor;
+import com.google.instrumentation.stats.ViewDescriptor.IntervalViewDescriptor;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * A mutable version of {@link View}, used for recording stats and start/end time.
+ */
+abstract class MutableView {
+
+  /**
+   * The {@link ViewDescriptor} associated with this {@link View}.
+   */
+  abstract ViewDescriptor getViewDescriptor();
+
+  /**
+   * Applies the given match function to the underlying data type.
+   */
+  abstract <T> T match(
+      Function<MutableDistributionView, T> p0,
+      Function<MutableIntervalView, T> p1);
+
+  /**
+   * Record stats with the given tags.
+   */
+  abstract void record(Map<String, String> tags, double value);
+
+  /**
+   * Convert this {@link MutableView} to {@link View}.
+   */
+  abstract View toView();
+
+  private MutableView() {
+  }
+
+  /**
+   * A {@link MutableView} for recording stats on distribution-based aggregations.
+   */
+  static final class MutableDistributionView extends MutableView {
+
+    /**
+     * Constructs a new {@link MutableDistributionView}.
+     */
+    static MutableDistributionView create(DistributionViewDescriptor distributionViewDescriptor,
+        List<MutableDistribution> distributions, Timestamp start) {
+      return new MutableDistributionView(
+          distributionViewDescriptor, distributions, start);
+    }
+
+    @Override
+    ViewDescriptor getViewDescriptor() {
+      return distributionViewDescriptor;
+    }
+
+    @Override
+    <T> T match(
+        Function<MutableDistributionView, T> p0, Function<MutableIntervalView, T> p1) {
+      return p0.apply(this);
+    }
+
+    @Override
+    void record(Map<String, String> tags, double value) {
+      // TODO(songya): compare tags with tagKeys in ViewDescriptor
+      for (MutableDistribution distribution : mutableDistributions) {
+        distribution.add(value);
+      }
+    }
+
+    @Override
+    final View toView() {
+      final List<DistributionAggregation> distributionAggregations =
+          new ArrayList<DistributionAggregation>();
+      for (MutableDistribution distribution : mutableDistributions) {
+        distributionAggregations.add(
+            DistributionAggregation.create(distribution.getCount(), distribution.getMean(),
+                distribution.getSum(),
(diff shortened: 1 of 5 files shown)


core_impl/src/main/java/com/google/instrumentation/stats/MeasurementDescriptorToViewMap.java
```diff
@@ -20,25 +20,27 @@
...
+      Multimaps.synchronizedMultimap(
+          HashMultimap.<MeasurementDescriptor.Name, MutableView>create());
 
   /**
-   * Returns a {@link View} corresponding to the given {@link ViewDescriptor}.
+   * Returns a {@link MutableView} corresponding to the given {@link ViewDescriptor}.
    */
-  final View getView(ViewDescriptor viewDescriptor) {
-    Collection<View> views = mutableMap.get(
+  final MutableView getMutableView(ViewDescriptor viewDescriptor) {
```
sebright: I think this method should convert the MutableView to a View, so that it can do it in the synchronized block.

core/src/main/java/com/google/instrumentation/stats/MutableView.java
```diff
@@ -0,0 +1,155 @@
...
+
+    /**
+     * Returns start timestamp for this aggregation.
+     */
+    Timestamp getStart() {
+      return start;
+    }
+
+    private final DistributionViewDescriptor distributionViewDescriptor;
+    private final List<MutableDistribution> mutableDistributions;
```
sebright: Shouldn't mutableDistributions be a map from tag values to MutableDistributions, i.e., `Map<List<TagValue>, Distribution>`?  We will need the tag values when constructing the DistributionAggregations.  Since this PR doesn't use the tag values yet, you could just add a TODO here.

core_impl/src/main/java/com/google/instrumentation/stats/StatsManagerImpl.java
```diff
@@ -113,26 +114,27 @@ public void process() {
...
-          viewDescriptor,
-          null,
-          Timestamp.fromMillis(System.currentTimeMillis()),
+    public MutableView apply(DistributionViewDescriptor viewDescriptor) {
+      List<MutableDistribution> distributions = new ArrayList<MutableDistribution>();
+      List<Double> bucketBoundaries =
+          viewDescriptor.getDistributionAggregationDescriptor().getBucketBoundaries();
+      distributions.add(
+          MutableDistribution.create(
+              bucketBoundaries == null ? null : BucketBoundaries.create(bucketBoundaries)));
```
sebright: I think there should also be a TODO here saying that we should only add the MutableDistributions as we receive new combinations of tags.

core/src/main/java/com/google/instrumentation/stats/MutableView.java
```diff
@@ -0,0 +1,155 @@
...
+import com.google.instrumentation.stats.ViewDescriptor.DistributionViewDescriptor;
+import com.google.instrumentation.stats.ViewDescriptor.IntervalViewDescriptor;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * A mutable version of {@link View}, used for recording stats and start/end time.
+ */
+abstract class MutableView {
```
sebright: I think we can simplify this class later, because it isn't part of the API.  For example, we probably don't need the factory methods.  We also don't need to pass in the distributions at the start, because the class is mutable.  I'm not suggesting any changes for this PR, though.

songy23: Sure, I added a TODO here.

core/src/main/java/com/google/instrumentation/stats/MutableView.java
songy23: That's correct, mark here as TODO.

core_impl/src/main/java/com/google/instrumentation/stats/MeasurementDescriptorToViewMap.java
songy23: Added a `getView()` method. However we need to keep `getMutableView()` as well, since `record()` needs to call it and retrieve the corresponding `MutableView`.

core_impl/src/main/java/com/google/instrumentation/stats/StatsManagerImpl.java
songy23: Done.

songy23:
Thanks!

core_impl/src/test/java/com/google/instrumentation/stats/MeasurementDescriptorToViewMapTest.java
```diff
@@ -31,12 +32,11 @@
...
   public void testPutAndGetView() {
-    View expected = DistributionView.create(
+    MutableView expected = MutableDistributionView.create(
         RpcConstants.RPC_CLIENT_ROUNDTRIP_LATENCY_VIEW,
-        null,
-        Timestamp.fromMillis(System.currentTimeMillis()),
+        new ArrayList<MutableDistribution>(),
         Timestamp.fromMillis(System.currentTimeMillis()));
-    measurementDescriptorToViewMap.putView(
+    measurementDescriptorToViewMap.putMutableView(
```
adriancole: putMutableView is a mouthful. is elaborating this in the name very important, or can we rely on the type system to explain?

core_impl/src/test/java/com/google/instrumentation/stats/StatsManagerImplTest.java
```diff
@@ -63,10 +66,24 @@ public void testGetNonexistentView() throws Exception {
     statsManager.getView(RpcConstants.RPC_CLIENT_REQUEST_COUNT_VIEW);
   }
 
-  // TODO(sebright): Turn this into a real test once StatsManagerImpl.record(...) is implemented.
-  // It currently only shows that record(...) can be called with a SimpleEventQueue.
   @Test
   public void testRecord() {
+    statsManager.registerView(RpcConstants.RPC_CLIENT_ROUNDTRIP_LATENCY_VIEW);
+    statsManager.record(
+        StatsContextFactoryImpl.DEFAULT,
```
adriancole: nit this method and the DEFAULT parameter kindof distracting. it seems it is only giving tags, which makes a lot more sense than guessing what `StatsContextFactoryImpl.DEFAULT` is all about.

can this be refactored? It would be nice to have this whole method intuitive, and this part is not

sebright: I think we will refactor StatsContext so that it is a final class in the core directory.  Then it won't need to be referenced indirectly from StatsContextFactoryImpl.  I made an initial PR: #219

adriancole:
>
> I think we will refactor StatsContext so that it is a final class in the
> core directory. Then it won't need to be referenced indirectly from
> StatsContextFactoryImpl. I made an initial PR: #219
> <https://github.com/google/instrumentation-java/pull/219>
>
cool, thing I noticed was that StatsContextImpl is being passed as a
parameter name tags, then only tags is used. so whichever makes this
tighter sounds good.

core_impl/src/test/java/com/google/instrumentation/stats/MeasurementDescriptorToViewMapTest.java
songy23: I'm fine with naming it `putView()`, the argument type should be self-explanatory.

Updated in PR https://github.com/google/instrumentation-java/pull/235

PR #235, opened 2017-04-19
Finish implementation on MutableView, add several tests for record().

songy23:
/cc @bogdandrutu @dinooliva

### DIFF
--- a/core_impl/src/test/java/com/google/instrumentation/stats/StatsManagerImplTest.java
+++ b/core_impl/src/test/java/com/google/instrumentation/stats/StatsManagerImplTest.java
@@ -17,6 +17,9 @@ import static com.google.common.truth.Truth.assertThat;
 
 import com.google.instrumentation.common.SimpleEventQueue;
 import com.google.instrumentation.stats.View.DistributionView;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
@@ -33,6 +36,17 @@ public class StatsManagerImplTest {
   public final ExpectedException thrown = ExpectedException.none();
 
   private static final double TOLERANCE = 1e-5;
+  private static final TagKey tagKey = RpcConstants.RPC_CLIENT_METHOD;
+  private static final TagValue tagValue1 = TagValue.create("some client method");
+  private static final TagValue tagValue2 = TagValue.create("some other client method");
+  private static final Map<String, String> oneTag = new HashMap<String, String>();
+  private static final Map<String, String> anotherTag = new HashMap<String, String>();
+  private static final Map<String, String> wrongTag = new HashMap<String, String>();
+  static {
+    oneTag.put(tagKey.asString(), tagValue1.asString());
+    anotherTag.put(tagKey.asString(), tagValue2.asString());
+    wrongTag.put("Wrong Tag Key", tagValue1.asString());
+  }
 
   private final StatsManagerImpl statsManager = new StatsManagerImpl(new SimpleEventQueue());
 
@@ -42,7 +56,6 @@ public class StatsManagerImplTest {
     View actual = statsManager.getView(RpcConstants.RPC_CLIENT_ROUNDTRIP_LATENCY_VIEW);
     assertThat(actual.getViewDescriptor()).isEqualTo(
         RpcConstants.RPC_CLIENT_ROUNDTRIP_LATENCY_VIEW);
-    // TODO(songya): verify if the distributions of actual and expected view are equal.
   }
 
   @Test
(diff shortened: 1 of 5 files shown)


core_impl/src/test/java/com/google/instrumentation/stats/StatsManagerImplTest.java
```diff
@@ -69,23 +82,109 @@ public void testGetNonexistentView() throws Exception {
...
+        (DistributionView) statsManager.getView(RpcConstants.RPC_CLIENT_ROUNDTRIP_LATENCY_VIEW);
+    List<DistributionAggregation> distributionAggregations = view.getDistributionAggregations();
+    assertThat(distributionAggregations).hasSize(2);
+    DistributionAggregation distributionAggregation1 = distributionAggregations.get(0);
+    DistributionAggregation distributionAggregation2 = distributionAggregations.get(1);
+
+    // The order of distributionAggregation1 and distributionAggregation2 could be random
+    if (distributionAggregation1.getCount() == 2) {
+      DistributionAggregation distributionAggregation = distributionAggregation1;
+      distributionAggregation1 = distributionAggregation2;
```
sebright: Would it be simpler to copy distributionAggregations into an ArrayList and then sort it on count (with a Comparator) to give it a fixed order?  Then you wouldn't have to swap the variables.

core/src/main/java/com/google/instrumentation/stats/MutableView.java
```diff
@@ -19,8 +19,10 @@
 import com.google.instrumentation.stats.ViewDescriptor.DistributionViewDescriptor;
 import com.google.instrumentation.stats.ViewDescriptor.IntervalViewDescriptor;
 import java.util.ArrayList;
+import java.util.HashMap;
```
adriancole: nit: LinkedHashMap if nothing else, but for tests getting consistent ordering regardless of JRE. (not necessarily your tests, anyone who uses this in tests)

core/src/main/java/com/google/instrumentation/stats/MutableView.java
```diff
@@ -80,22 +81,43 @@ ViewDescriptor getViewDescriptor() {
 
     @Override
     void record(Map<String, String> tags, double value) {
-      // TODO(songya): compare tags with tagKeys in ViewDescriptor
-      for (MutableDistribution distribution : mutableDistributions) {
-        distribution.add(value);
+      final List<TagKey> tagKeys = this.distributionViewDescriptor.getTagKeys();
```
adriancole: maybe note that either uniqueness in tag key isn't important, or it is enforced elsewhere.

core_impl/src/test/java/com/google/instrumentation/stats/StatsManagerImplTest.java
```diff
@@ -69,23 +82,109 @@ public void testGetNonexistentView() throws Exception {
...
+    assertThat(distributionAggregation1.getTags().get(0).getKey()).isEqualTo(tagKey);
+    assertThat(distributionAggregation1.getTags().get(0).getValue()).isEqualTo(tagValue1);
+    assertThat(distributionAggregation2.getTags().get(0).getKey()).isEqualTo(tagKey);
+    assertThat(distributionAggregation2.getTags().get(0).getValue()).isEqualTo(tagValue2);
+  }
+
+  private static void verifyDistributionAggregation(
+      DistributionAggregation distributionAggregation, int bucketCountsSize, int count, double sum,
+      double mean, double min, double max, int tagsSize) {
+    assertThat(distributionAggregation.getBucketCounts().size()).isEqualTo(bucketCountsSize);
```
songy23: It looks like `DistributionAggregation.getBucketCounts()` won't omit trailing zeros. I got a list like [0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 1, 0, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0] by calling `getBucketCounts()` after I recorded value 10, 20, 30 and 40. I have hard-coded a test for verifying the buckets, not sure if there's cleaner way to do that.

core/src/main/java/com/google/instrumentation/stats/MutableView.java
sebright: I was wondering whether we should sort the Aggregations in a View.  @a-veitch thought that we shouldn't guarantee an order, so that we have the flexibility to change the implementation later.  If we don't guarantee the order, then I don't think we should test the order of the Aggregations.

I think we could deal with the unknown order by adding a class with some testing utilities, e.g., a method that takes a View and returns the Aggregations in sorted order.

songy23: Makes sense. I'll update to keep the order in tests instead of in the implementation.

core_impl/src/test/java/com/google/instrumentation/stats/StatsManagerImplTest.java
sebright: Sorry, I didn't finish reviewing the changes.

We don't omit the trailing buckets now, but we could change it.  Here is the comment: https://github.com/google/instrumentation-java/blob/55f477ece335dad5dc5ca949870f0f216d369931/core/src/main/java/com/google/instrumentation/stats/DistributionAggregation.java#L106 .  Since you added the check for the non-zero buckets, I think you could just remove this check.

songy23:
Thanks, I updated the check.

PR #237, opened 2017-04-20
Use StatsContext, TagKey, and TagValue types in more places.

sebright:
This just adds a little more type-safety.  ~~The first commit is from #236.~~  I don't want this PR to conflict with #233, so I will wait to merge it.


adriancole:
weird.. seems like deja vu.. wasnt' there something recently to convert to plain maps?

sebright:
@adriancole Do you mean switching from StatsContext to Map?  I don't remember a change like that.  We will be adding support for integer and boolean tag values, so I think we will need the specialized classes.

adriancole:
I was thinking about this one #186

sebright:
Thanks.  I think there is still a big benefit to having a class in this case, because it will let us change the representation if we need to, for performance.

sebright:
Thanks!  I rebased my branch.

PR #240, opened 2017-04-20
Change visibility for SpanFactory and StartSpanOptions

bogdandrutu:
We have to allow testing and implementation.
Also add a public SpanContext.create.

### DIFF
--- a/core/src/main/java/com/google/instrumentation/trace/SpanFactory.java
+++ b/core/src/main/java/com/google/instrumentation/trace/SpanFactory.java
@@ -16,7 +16,7 @@ package com.google.instrumentation.trace;
 import javax.annotation.Nullable;
 
 /** Factory class to create and start a {@link Span}. */
-abstract class SpanFactory {
+public abstract class SpanFactory {
   /**
    * Creates and starts a new child {@link Span} (or root if parent is {@code null}), with parent
    * being the designated {@code Span} and the given options.
--- a/core/src/main/java/com/google/instrumentation/trace/SpanFactory.java
+++ b/core/src/main/java/com/google/instrumentation/trace/SpanFactory.java
@@ -26,7 +26,7 @@ public abstract class SpanFactory {
    * @param options The options for the start of the {@code Span}.
    * @return A child {@code Span} that will have the name provided.
    */
-  abstract Span startSpan(@Nullable Span parent, String name, StartSpanOptions options);
+  protected abstract Span startSpan(@Nullable Span parent, String name, StartSpanOptions options);
 
   /**
    * Creates and starts a new child {@link Span} (or root if parent is {@code null}), with parent
@@ -39,6 +39,6 @@ public abstract class SpanFactory {
    * @param options The options for the start of the {@code Span}.
    * @return A child {@code Span} that will have the name provided.
    */
-  abstract Span startSpanWithRemoteParent(
+  protected abstract Span startSpanWithRemoteParent(
       @Nullable SpanContext remoteParent, String name, StartSpanOptions options);
 }
--- a/core/src/main/java/com/google/instrumentation/trace/StartSpanOptions.java
+++ b/core/src/main/java/com/google/instrumentation/trace/StartSpanOptions.java
@@ -23,7 +23,7 @@ import javax.annotation.Nullable;
  * overriding the {@link Sampler sampler}, the parent links, and option to record all the events
  * even if the {@code Span} is not sampled.
  */
-final class StartSpanOptions {
+public final class StartSpanOptions {
   private Sampler sampler;
   private List<Span> parentLinks;
   private Boolean recordEvents;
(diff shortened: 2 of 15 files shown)


core/src/test/java/com/google/instrumentation/trace/LinkTest.java
```diff
@@ -18,26 +18,17 @@
...
-    spanContext =
-        new SpanContext(
-            TraceId.generateRandomId(random),
-            SpanId.generateRandomId(random),
-            TraceOptions.DEFAULT);
-  }
+  private static final Random random = new Random(1234);
+  private static final SpanContext spanContext =
+      SpanContext.create(
+          TraceId.generateRandomId(random), SpanId.generateRandomId(random), TraceOptions.DEFAULT);
```
sebright: What is the reason for this change?  I thought it was better for the two fields to be non-static, because it made it clear that the tests are deterministic.  Even though `random` is only used to initialize `spanContext`, it looks like it could be used by individual tests, which could make the test results depend on the order in which they are run.  You could just make the fields non-static to ensure that they are initialized for each test.

core/src/main/java/com/google/instrumentation/trace/SpanFactory.java
```diff
@@ -16,7 +16,7 @@
 import javax.annotation.Nullable;
 
 /** Factory class to create and start a {@link Span}. */
-abstract class SpanFactory {
+public abstract class SpanFactory {
   /**
```
sebright: Should the methods be made public, too?

core/src/main/java/com/google/instrumentation/trace/StartSpanOptions.java
```diff
@@ -23,7 +23,7 @@
  * overriding the {@link Sampler sampler}, the parent links, and option to record all the events
  * even if the {@code Span} is not sampled.
  */
-final class StartSpanOptions {
+public final class StartSpanOptions {
```
sebright: Should any of these methods be public?

core/src/test/java/com/google/instrumentation/trace/SamplersTest.java
```diff
@@ -24,29 +24,33 @@
 /** Unit tests for {@link Samplers}. */
 @RunWith(JUnit4.class)
 public class SamplersTest {
+  private static final Random random = new Random(1234);
+  private static final TraceId traceId = TraceId.generateRandomId(random);
+  private static final SpanId parentSpanId = SpanId.generateRandomId(random);
+  private static final SpanId spanId = SpanId.generateRandomId(random);
```
sebright: I think it would be better if these fields were non-static, for the same reason as above.

core/src/test/java/com/google/instrumentation/trace/SpanBuilderTest.java
```diff
@@ -36,21 +36,17 @@
 public class SpanBuilderTest {
   private static final String SPAN_NAME = "MySpanName";
   private static final Tracer tracer = Tracing.getTracer();
+  private static final Random random = new Random(1234);
+  private static final SpanContext spanContext =
+      SpanContext.create(
+          TraceId.generateRandomId(random), SpanId.generateRandomId(random), TraceOptions.DEFAULT);
```
sebright: Same here

core_impl/src/test/java/com/google/instrumentation/trace/SpanDataTest.java
```diff
@@ -38,9 +38,11 @@
   private static final String DISPLAY_NAME = "MySpanDisplayName";
   private static final String ANNOTATION_TEXT = "MyAnnotationText";
   private static final Status status = Status.DEADLINE_EXCEEDED.withDescription("TooSlow");
-  private Random random;
-  private SpanContext spanContext;
-  private SpanId parentSpanId;
+  private static final Random random = new Random(1234);
```
sebright: Same here

core/src/main/java/com/google/instrumentation/trace/SpanFactory.java
bogdandrutu: Made them protected.

core/src/main/java/com/google/instrumentation/trace/StartSpanOptions.java
bogdandrutu: Made the getters public.

core/src/test/java/com/google/instrumentation/trace/SamplersTest.java
bogdandrutu: Done.

core/src/test/java/com/google/instrumentation/trace/SpanBuilderTest.java
bogdandrutu: Done.

core_impl/src/test/java/com/google/instrumentation/trace/SpanDataTest.java
bogdandrutu: Done.

core/src/test/java/com/google/instrumentation/trace/LinkTest.java
bogdandrutu: I think it is fine to have them non-static for the moment.

PR #247, opened 2017-04-21
Make the TraceOptions Builder methods public for testing.

bogdandrutu:


zhangkun83:
Not related to this change, but you may also want to make one of the two abstract `addAnnotation()` on `Span` final, because it seems one can delegate to the other.

core/src/main/java/com/google/instrumentation/trace/TraceOptions.java
```diff
@@ -183,7 +183,7 @@ private Builder(byte options) {
      *
      * @return this.
      */
-    Builder setIsSampled() {
+    public Builder setIsSampled() {
```
adriancole: one thing I like is to make an Internal class which makes it obvious that this is not exposed for others to use.

Ex. Internal.setIsSampled(options);

you can see examples of this in Brave and okhttp

zhangkun83:
@bogdandrutu sorry for those last-minute requirements. Apparently I didn't think through the testing requirements when I reviewed the API. If you'd prefer the APIs that we made public today to be non-public, you could consider providing fake implementations for classes such as `SpanFactory` and `SpanContext` so that you can keep their constructors and construction parameters private. We probably won't have time to explore that option now, but we can annotate them `@VisibleForTesting` today and add javadoc saying that they might be turned package-private or protected in the future.

PR #250, opened 2017-04-21
Use a 'Greedy' approach for aggregations on measurement.

songy23:


core_impl/src/test/java/com/google/instrumentation/stats/StatsManagerImplTest.java
```diff
@@ -192,16 +199,26 @@ public void testRecordNonExistentMeasurementDescriptor() {
...
         (DistributionView) statsManager.getView(RpcConstants.RPC_CLIENT_ROUNDTRIP_LATENCY_VIEW);
-    // Won't record stats if there are non existent tags.
-    assertThat(view.getDistributionAggregations()).hasSize(0);
+
+    DistributionAggregation distributionAggregation = view.getDistributionAggregations().get(0);
+    assertThat(view.getDistributionAggregations()).hasSize(1);
+    List<Tag> tags = distributionAggregation.getTags();
+    // Won't record the unregistered tag key, will use default tag instead:
+    // "method" : "unknown/not set".
+    assertThat(tags.get(0).getKey()).isNotEqualTo(wrongTagKey);
```
sebright: > assertThat(tags.get(0).getKey()).isNotEqualTo(wrongTagKey);

I think this assertion is redundant when we also have `assertThat(tags.get(0).getKey()).isEqualTo(tagKey);`.

core_impl/src/test/java/com/google/instrumentation/stats/StatsManagerImplTest.java
```diff
@@ -84,16 +85,24 @@ public void testGetNonexistentView() throws Exception {
   @Test
   public void testRecord() {
     statsManager.registerView(RpcConstants.RPC_CLIENT_ROUNDTRIP_LATENCY_VIEW);
+    Timestamp start = Timestamp.fromMillis(System.currentTimeMillis());
     for (double val : Arrays.<Double>asList(10.0, 20.0, 30.0, 40.0)) {
       statsManager.record(
           oneTag, MeasurementMap.of(RpcConstants.RPC_CLIENT_ROUNDTRIP_LATENCY, val));
+      // Simulate transmission delay.
+      try {
+        Thread.sleep(500);
```
sebright: We should use a fake clock instead, to keep the tests deterministic and fast.  Also, I think that any refactoring to use a fake clock should be done in a separate PR, because it would be a large change.

core_impl/src/test/java/com/google/instrumentation/stats/StatsManagerImplTest.java
```diff
@@ -192,16 +199,26 @@ public void testRecordNonExistentMeasurementDescriptor() {
...
+
   @Test
   public void testRecordNonExistentTag() {
     statsManager.registerView(RpcConstants.RPC_CLIENT_ROUNDTRIP_LATENCY_VIEW);
-    statsManager.record(
-        wrongTag, MeasurementMap.of(RpcConstants.RPC_CLIENT_ROUNDTRIP_LATENCY, 10.0));
+    statsManager.record(wrongTag, MeasurementMap.of(
+        RpcConstants.RPC_CLIENT_ROUNDTRIP_LATENCY, 10.0));
+    statsManager.record(wrongTag, MeasurementMap.of(
+        RpcConstants.RPC_CLIENT_ROUNDTRIP_LATENCY, 50.0));
```
sebright: I think it would be useful to check that stats for multiple StatsContexts with non-matching keys are combined.  You could use a second set of tags here or create a new test.

core_impl/src/test/java/com/google/instrumentation/stats/StatsManagerImplTest.java
```diff
@@ -192,16 +199,26 @@ public void testRecordNonExistentMeasurementDescriptor() {
     assertThat(view.getDistributionAggregations()).hasSize(0);
   }
 
-  // TODO(songya): update this test once we determine how to handle tags that aren't an exact match
-  // for the view.
+
   @Test
   public void testRecordNonExistentTag() {
```
sebright: This test's name seems less accurate now.  How about something like testRecord_TagNotMatchingView?

core_impl/src/main/java/com/google/instrumentation/stats/MutableView.java
```diff
@@ -29,6 +30,7 @@
  */
 // TODO(songya): remove or modify the methods of this class, since it's not part of the API.
 abstract class MutableView {
+  @VisibleForTesting static final TagValue UNKNOWN_TAG_VALUE = TagValue.create("unknown/not set");
```
songy23: Marked. The current value is from previous discussion (not sure if we've come to a conclusion on this).

core_impl/src/test/java/com/google/instrumentation/stats/StatsManagerImplTest.java
songy23: Makes sense, I'll leave a TODO here.

core_impl/src/test/java/com/google/instrumentation/stats/StatsManagerImplTest.java
```diff
@@ -180,7 +179,15 @@ public void testRecordWithEmptyStatsContext() {
...
     DistributionView view =
         (DistributionView) statsManager.getView(RpcConstants.RPC_CLIENT_ROUNDTRIP_LATENCY_VIEW);
-    assertThat(view.getDistributionAggregations()).hasSize(0);
+    assertThat(view.getDistributionAggregations()).hasSize(1);
+    DistributionAggregation distributionAggregation = view.getDistributionAggregations().get(0);
+    List<Tag> tags = distributionAggregation.getTags();
+    assertThat(tags.get(0).getKey()).isEqualTo(tagKey);
+    // Tag is missing for associated measurementValues, should use default tag value
+    // "unknown/not set".
+    assertThat(tags.get(0).getValue()).isEqualTo(MutableView.UNKNOWN_TAG_VALUE);
```
songy23: Add this test in MutableViewTest.

core_impl/src/test/java/com/google/instrumentation/stats/StatsManagerImplTest.java
songy23: Done

core_impl/src/test/java/com/google/instrumentation/stats/StatsManagerImplTest.java
songy23: Removed the first one.

core_impl/src/test/java/com/google/instrumentation/stats/StatsManagerImplTest.java
songy23: Sure, updated this test.

songy23:
Thanks!

PR #251, opened 2017-04-24
Start adding a "tags" package.

sebright:
This commit adds TagKey, TagChange, TagOp, and TagContext to the new package.
It does not modify any classes in "stats", for backwards compatibility.  TagKey
can be paired with values of type String, long, or boolean.  TagContext is
similar to StatsContext, except that it uses AutoValue and doesn't contain any
methods for recording stats or serializing/deserializing. It uses the TagChange
and TagOp classes to modify the tag values.

/cc @dinooliva @songy23 @bogdandrutu


sebright:
I updated this PR to use a builder for TagSet, and I removed TagChange and TagOp.

sebright:
I added a TagSet factory class and error handling for TagSet.insert(...).  We might need to split TagSet into an abstract class in `core` and an implementation class in `core_impl`, so that anyone can write a TagSet implementation.

sebright:
I split TagSet into an abstract class and an implementation.

sebright:
This PR is becoming too large, so I split the TagSet and TagKey changes into #276.

sebright:
Closing in favor of #276.

PR #253, opened 2017-04-24
Fix a flaky test in StatsContextTest.java.

songy23:
Fix issue https://github.com/google/instrumentation-java/issues/224.

### DIFF
--- a/core_impl/src/test/java/com/google/instrumentation/stats/StatsContextTest.java
+++ b/core_impl/src/test/java/com/google/instrumentation/stats/StatsContextTest.java
@@ -15,15 +15,17 @@ package com.google.instrumentation.stats;
 
 import static com.google.common.truth.Truth.assertThat;
 
+import com.google.common.collect.Collections2;
 import com.google.common.testing.EqualsTester;
-
 import com.google.io.base.VarInt;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Map.Entry;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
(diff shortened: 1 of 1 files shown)


core_impl/src/test/java/com/google/instrumentation/stats/StatsContextTest.java
```diff
@@ -167,17 +169,21 @@ public void testToString() {
...
+    for (Entry<TagKey, TagValue> entry : tagMap.entrySet()) {
       expected.write(VALUE_TYPE_STRING);
-      encodeString(tag.getKey().toString(), expected);
-      encodeString(tag.getValue().toString(), expected);
+      encodeString(entry.getKey().asString(), expected);
+      encodeString(entry.getValue().asString(), expected);
     }
-    builder.build().serialize(actual);
 
     assertThat(actual.toByteArray()).isEqualTo(expected.toByteArray());
```
sebright: It looks like the order of the tags in `actual` depends on the implementation of the StatsContextImpl constructor.  The test could fail if the constructor copied the map.  A more reliable way to test the serialized format is to create one ByteArrayOutputStream for each possible order of the tags and then assert that the result matches one of them.

songy23: That makes sense but I think it would be trivial to generate all the permutations of tags. The key point here is to eliminate the possible inconsistency on tag order between expected and actual, so how about modifying to traverse `statsContextImpl.tags.entrySet()` instead of `tagMap.entrySet()`? In that way even if the constructor copied (or even modified) the map, we could still guarantee the tag orders are the same between expected and actual.

sebright: I think it would be better to test through the StatsContext API, to make the test less fragile.  You could use Guava's permutations method to generate all permutations of `tags`: https://google.github.io/guava/releases/19.0/api/docs/com/google/common/collect/Collections2.html#permutations(java.util.Collection)

songy23: Updated. That method is super useful, thanks!

PR #255, opened 2017-04-25
Make auto-value compileOnly dependency. (#254)

bogdandrutu:
Because it's not needed at runtime.  Having it in runtime classpath
triggers warnings like `warning: No processor claimed any of these
annotations: ` with error-prone compiler (grpc/grpc-java/pull/2938)
(cherry picked from commit a19c013692a0c2a61652da67add51d10ba81ebf1)

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -1,10 +1,9 @@
 description = 'Instrumentation: API'
 
 dependencies {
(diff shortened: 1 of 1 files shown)


sebright:
Thanks!

PR #257, opened 2017-04-27
Add an interface for getting the time throughout the Instrumentation library.

sebright:
Clock is an abstract class. It can have different implementations for Java 7,
Java 8, and tests.


adriancole:
:) curious why not put the impls in the same PR?

bogdandrutu:
We will add it soon.

PR #258, opened 2017-04-27
Add Clock implementations for Java 6/7 and for testing.

sebright:
The two implementations are in the `internal` package for now.

/cc @adriancole

### DIFF
--- a/core/src/main/java/com/google/instrumentation/internal/TestClock.java
+++ b/core/src/main/java/com/google/instrumentation/internal/TestClock.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2017, Google Inc.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


core/src/test/java/com/google/instrumentation/internal/TestClockTest.java
```diff
@@ -0,0 +1,56 @@
...
+import com.google.instrumentation.common.Timestamp;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/**
+ * Tests for {@link TestClock}.
+ */
+@RunWith(JUnit4.class)
+public final class TestClockTest {
```
adriancole: run forest run!

core/src/main/java/com/google/instrumentation/internal/TestClock.java
```diff
@@ -0,0 +1,85 @@
...
+    clock.setTime(time);
+    return clock;
+  }
+
+  /**
+   * Sets the time.
+   *
+   * @param time the new time.
+   */
+  public void setTime(Timestamp time) {
```
bogdandrutu: Maybe instead of setTime we should use an AdvanceTime(Duration) or AdvanceTime(nanos)?

sebright: I initially added `setTime` because I found it to be more useful when I refactored the stats tests to use TestClock.  I added `advanceTime(Duration)` in the last commit.

PR #260, opened 2017-04-27
Split core_impl into directories for Android, Java 7, and Java 8.

sebright:
This directory structure will allow us to support the three Java versions but
also take advantage of the features available in each one.  Each directory has a
readme.

I refactored the stats code so that there is one StatsManagerImpl for each of
Android, Java 7, and Java 8.  StatsManagerImpl is loaded with reflection, and it
specifies the stats classes that should be used for the implementation.  I also
moved DisruptorEventQueue to core_impl_java, since it isn't Android-compatible.

This commit doesn't affect tracing.

/cc @dinooliva @songy23 @a-veitch 

@bogdandrutu I'm not sure if I updated the build files correctly.


sebright:
I noticed that the first coverage report was incomplete, so I listed the new directories in [all/build.gradle](https://github.com/google/instrumentation-java/pull/260/commits/77f3bedc976242a2681debf838ea7bebe71c7c7c#diff-a4f4a71bfa7154f576719a810f439c6eR20).  The duplicate StatsManagerImpls seemed to cause a NullPointerException in the Gradle javadoc command, though: https://travis-ci.org/google/instrumentation-java/jobs/226581574#L914

core_impl_android/README.md
```diff
@@ -0,0 +1,6 @@
+Instrumentation Android implementation
+======================================================
+
+* Java 6 and Android compatible.
```
bogdandrutu: This will be only android compatible. Once android move forward with supporting more java7 we will move this to java7. So I would say just "Android compatible".

all/build.gradle
```diff
@@ -14,6 +14,10 @@ buildscript {
 def subprojects = [
   project(':instrumentation-java-core'),
   project(':instrumentation-java-core-impl'),
+  project(':instrumentation-java-core-impl-android'),
```
bogdandrutu: For the moment I would not put android/java7/java8 here. We should probably have 3 code coverage instead of all (java7, java8, android) to avoid the duplicate class names issue.

core_impl_android/src/main/java/com/google/instrumentation/stats/StatsManagerImpl.java
```diff
@@ -0,0 +1,27 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.instrumentation.stats;
+
+import com.google.instrumentation.common.SimpleEventQueue;
```
bogdandrutu: I think SimpleEventQueue should also be moved into the Android implementation.

all/build.gradle
sebright: Thanks

core_impl/README.md
```diff
@@ -0,0 +1,5 @@
+Instrumentation implementation
+======================================================
+
+* The main implementation.
```
sebright: Done

core_impl_android/README.md
sebright: Done

core_impl_android/src/main/java/com/google/instrumentation/stats/StatsManagerImpl.java
sebright: SimpleEventQueue is used for testing, so it needs to be in core_impl.  I just used it in the Android implementation as a placeholder.  Do you think I should copy it and rename the original class to something like TestEventQueue?

bogdandrutu: I am fine with keeping it there.

PR #261, opened 2017-04-27
Add initial span implementation.

bogdandrutu:
* The implementation is not completed but allows us to implement the StartEndHandler in parallel with the rest of the implementation.

### DIFF
--- a/core_impl/src/main/java/com/google/instrumentation/trace/SpanImpl.java
+++ b/core_impl/src/main/java/com/google/instrumentation/trace/SpanImpl.java
@@ -22,6 +22,7 @@ import java.util.Map;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.annotation.Nullable;
+import javax.annotation.concurrent.GuardedBy;
 import javax.annotation.concurrent.ThreadSafe;
 
 /** Implementation for the {@link Span} class. */
@@ -29,17 +30,34 @@ import javax.annotation.concurrent.ThreadSafe;
 final class SpanImpl extends Span {
   private static final Logger logger = Logger.getLogger(Tracer.class.getName());
 
+  // The parent SpanId of this span. Null if this is a root.
   private final SpanId parentSpanId;
+  // Handler called when the span starts and ends.
   private final StartEndHandler startEndHandler;
+  // The displayed name of the span.
   private final String name;
-  // All the following variables are initialized iff the Span has RECORD_EVENTS option.
+  // The clocked used to get the time.
   private final Clock clock;
-  private TimestampConverter timestampConverter;
-  private long startNanoTime;
+  // The time converter used to convert nano time to Timestamp. This is needed because java has
+  // milliseconds granularity for Timestamp and tracing events are recorded more often.
+  private final TimestampConverter timestampConverter;
+  // The start time of the span. Set when the span is created iff the RECORD_EVENTS options is
+  // set, otherwise 0.
+  private final long startNanoTime;
+  // The status of the span. Set when the span is ended iff the RECORD_EVENTS options is set.
+  @GuardedBy("this")
   private Status status;
+  // The end time of the span. Set when the span is ended iff the RECORD_EVENTS options is set,
+  // otherwise 0.
+  @GuardedBy("this")
   private long endNanoTime;
+  // True if the span is ended.
+  @GuardedBy("this")
   private boolean hasBeenEnded;
 
+  // Creates and starts a span with the given configuration. TimestampConverter is null if the
+  // Span is a root span or the parent is not sampled. If the parent is sampled we should use the
+  // same converter to ensure ordering between tracing events.
   static SpanImpl startSpan(SpanContext context,
       @Nullable EnumSet<Options> options,
       String name,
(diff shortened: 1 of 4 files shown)


core_impl/src/main/java/com/google/instrumentation/trace/SpanImpl.java
```diff
@@ -0,0 +1,170 @@
...
+@ThreadSafe
+final class SpanImpl extends Span {
+  private static final Logger logger = Logger.getLogger(Tracer.class.getName());
+
+  private final SpanId parentSpanId;
+  private final StartEndHandler startEndHandler;
+  private final String name;
+  // All the following variables are initialized iff the Span has RECORD_EVENTS option.
+  private final Clock clock;
+  private TimestampConverter timestampConverter;
```
a-veitch: Related: please provide a comment line for each private var documenting its purpose.

core_impl/src/main/java/com/google/instrumentation/trace/SpanImpl.java
```diff
@@ -0,0 +1,170 @@
...
+  /**
+   * Returns an immutable representation of all the data from this {@code Span}.
+   *
+   * @return an immutable representation of all the data from this {@code Span}.
+   * @throws IllegalStateException if the Span doesn't have RECORD_EVENTS option.
+   */
+  SpanData toSpanData() {
+    checkState(
+        getOptions().contains(Options.RECORD_EVENTS),
+        "Getting SpanData for a Span without RECORD_EVENTS options.");
```
a-veitch: Nit: remove s from "options"

core_impl/src/main/java/com/google/instrumentation/trace/SpanImpl.java
```diff
@@ -0,0 +1,170 @@
...
+import javax.annotation.concurrent.ThreadSafe;
+
+/** Implementation for the {@link Span} class. */
+@ThreadSafe
+final class SpanImpl extends Span {
+  private static final Logger logger = Logger.getLogger(Tracer.class.getName());
+
+  private final SpanId parentSpanId;
+  private final StartEndHandler startEndHandler;
+  private final String name;
```

core_impl/src/main/java/com/google/instrumentation/trace/SpanImpl.java
bogdandrutu: Added comments. Also timestamp converter is used for multiple reasons:
* Better granularity than ms for tracing events.
* Ensure tracing events ordering between multiple spans from the same trace. We use the same converter for all the Spans that are part of the same trace to ensure for example that parent start time < child start time.

core_impl/src/main/java/com/google/instrumentation/trace/SpanImpl.java
```diff
@@ -0,0 +1,170 @@
...
+
+  /**
+   * Abstract class to handle the start and end operations for a {@link Span} only when the {@code
+   * Span} has {@link Span.Options#RECORD_EVENTS} option.
+   *
+   * <p>Implementation must avoid high overhead work in any of the methods because the code is
+   * executed on the critical path.
+   */
+  @ThreadSafe
+  abstract static class StartEndHandler {
```
bogdandrutu: Done. Actually one instance of the handler can be (it will be) shared by multiple Spans at the same time so multiple threads can call into OpStart/OpEnd at the same time.

core_impl/src/test/java/com/google/instrumentation/trace/TimestampConverterTest.java
```diff
@@ -16,18 +16,23 @@
...
   public void convertNanoTime() {
-    Timestamp timestamp = Timestamp.create(1234, 5678);
-    TimestampConverter timeConverter = new TimestampConverter(timestamp, 2345);
-    assertThat(timeConverter.convertNanoTime(1234)).isEqualTo(timestamp.addNanos(-1111));
-    assertThat(timeConverter.convertNanoTime(3456)).isEqualTo(timestamp.addNanos(1111));
+    TimestampConverter timeConverter = TimestampConverter.now(testClock);
+    assertThat(timeConverter.convertNanoTime(testClock.nowNanos() - 1234))
+        .isEqualTo(timestamp.addNanos(-1234));
+    assertThat(timeConverter.convertNanoTime(testClock.nowNanos() + 1234))
+        .isEqualTo(timestamp.addNanos(1234));
```
bogdandrutu: Not sure I understand exactly your comment. Can you clarify a bit?

bogdandrutu:
PTAL.

core_impl/src/main/java/com/google/instrumentation/trace/SpanImpl.java
```diff
@@ -0,0 +1,195 @@
...
+final class SpanImpl extends Span {
+  private static final Logger logger = Logger.getLogger(Tracer.class.getName());
+
+  // The parent SpanId of this span. Null if this is a root.
+  private final SpanId parentSpanId;
+  // Handler called when the span starts and ends.
+  private final StartEndHandler startEndHandler;
+  // The displayed name of the span.
+  private final String name;
+  // The clocked used to get the time.
```
a-veitch: s/ed//

core_impl/src/test/java/com/google/instrumentation/trace/TimestampConverterTest.java
sebright: I just had an idea for making the test more similar to the expected use case, by mocking Clock and giving it low precision.  I think the existing test is fine, though.  Here are the steps:

Set the mock's result for its two methods like this:
now(): TimeStamp.create(1234, 0)
nowNanos(): 100

Create the TimestampConverter, and then advance the clock to:
now(): TimeStamp.create(1234, 0)
nowNanos(): 300

Call convertNanoTime with 300, and check that the time is TimeStamp.create(1234, 200).

bogdandrutu:
Done. Merging soon.

PR #267, opened 2017-04-28
Change AttributeValue, EndSpanOptions and NetworkEvent to use autovalue.

bogdandrutu:


core/src/main/java/com/google/instrumentation/trace/NetworkEvent.java
```diff
@@ -69,91 +62,63 @@ public static Builder builder(Type type, long messageId) {
...
      * Sets the kernel timestamp.
      *
      * @param kernelTimestamp The kernel timestamp of the event.
      * @return this.
      */
-    public Builder setKernelTimestamp(@Nullable Timestamp kernelTimestamp) {
-      this.kernelTimestamp = kernelTimestamp;
-      return this;
-    }
+    public abstract Builder setKernelTimestamp(Timestamp kernelTimestamp);
```
sebright: Should this still be Nullable?

PR #266, opened 2017-04-28
Use Clock class throughout stats implementation.

sebright:

PR #268, opened 2017-04-29
Copy TraceComponentImpl into implementation directories and add Clock field. 

sebright:
~~The first six commits are from #266.~~

Three main commits:

#### Copy TraceComponentImpl into Android, Java 7, and Java 8 directories.

This change will allow TraceComponentImpl to have different implementations for
the different Java versions.

#### Add a Clock field to TraceComponent.

This is the clock that will be used throughout the tracing implementation.  It
is exposed so that users can record times, too.

#### Deprecate TimestampFactory in favor of Clock.
_________________________________________________

@bogdandrutu This is the rest of my branch that splits up the core_impl directory.  Is the tracing part correct?  Is this the right way to expose the time functionality?  I didn't update TimestampConverter because I saw that you already changed it in #261.

/cc @dinooliva @songy23 @a-veitch


core/src/main/java/com/google/instrumentation/trace/TraceComponent.java
```diff
@@ -61,6 +71,12 @@ public BinaryPropagationHandler getBinaryPropagationHandler() {
       return BinaryPropagationHandler.getNoopBinaryPropagationHandler();
     }
 
+    @Override
+    public Clock getClock() {
+      // TODO(sebright): Is this the right implementation, or should it return a constant?
+      return MillisClock.getInstance();
```
bogdandrutu: Probably we should return a ZeroTimeClock for no-op.

core/src/test/java/com/google/instrumentation/common/TimestampFactoryTest.java
```diff
@@ -8,7 +8,7 @@
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
 
-/** Unit tests for {@link TimestampFactory}. */
+/** Unit tests for {@code TimestampFactory}. */
```
bogdandrutu: Keep the link because this is how the template generates the comment.

core_impl_java_7/src/test/java/com/google/instrumentation/trace/TraceComponentImplTest.java
```diff
@@ -0,0 +1,36 @@
...
+@RunWith(JUnit4.class)
+public class TraceComponentImplTest {
+  @Test
+  public void implementationOfTracer() {
+    // TODO(bdrutu): Change this when TracerImpl is available.
+    assertThat(Tracing.getTracer()).isSameAs(Tracer.getNoopTracer());
+  }
+
+  @Test
+  public void implementationOfBinaryPropagationHandler() {
```
bogdandrutu: Maybe add a test that the clock is the MillisClock?

core_impl_java_8/src/test/java/com/google/instrumentation/trace/TraceComponentImplTest.java
```diff
@@ -0,0 +1,36 @@
...
+  public void implementationOfTracer() {
+    // TODO(bdrutu): Change this when TracerImpl is available.
+    assertThat(Tracing.getTracer()).isSameAs(Tracer.getNoopTracer());
+  }
+
+  @Test
+  public void implementationOfBinaryPropagationHandler() {
+    assertThat(Tracing.getBinaryPropagationHandler())
+        .isSameAs(BinaryPropagationHandlerImpl.INSTANCE);
+  }
```
bogdandrutu: Same here a test that the clock is InstantClock.

core/src/test/java/com/google/instrumentation/common/TimestampFactoryTest.java
sebright: What template uses this?

core_impl/src/main/java/com/google/instrumentation/trace/TraceComponentImplBase.java
```diff
@@ -14,11 +14,14 @@
 package com.google.instrumentation.trace;
 
 /** Implementation of the {@link TraceComponent}. */
-public final class TraceComponentImpl extends TraceComponent {
+public class TraceComponentImplBase extends TraceComponent {
```
sebright: StatsManagerImplBase isn't abstract, because it is used for testing in core_impl.  For example, [StatsManagerImplTest](https://github.com/google/instrumentation-java/blob/0c9fcffe501ecf6e583d917159f41c72c81d8841/core_impl/src/test/java/com/google/instrumentation/stats/StatsManagerImplTest.java#L58) creates a StatsManagerImplBase with a test clock and event queue and then tests the functionality that doesn't depend on Java version.  I assumed that the core_impl tests would also need to create TraceComponentImpls.

core/src/main/java/com/google/instrumentation/trace/TraceComponent.java
sebright: Done

core_impl_java_7/src/test/java/com/google/instrumentation/trace/TraceComponentImplTest.java
sebright: Good idea

core_impl_java_8/src/test/java/com/google/instrumentation/trace/TraceComponentImplTest.java
sebright: Done

core/src/test/java/com/google/instrumentation/common/TimestampFactoryTest.java
sebright: I changed it back to `@link`.

core_impl/src/main/java/com/google/instrumentation/trace/TraceComponentImplBase.java
```diff
@@ -0,0 +1,43 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.instrumentation.trace;
+
+import com.google.instrumentation.common.Clock;
+
+/** Implementation of the {@link TraceComponent}. */
```
bogdandrutu: Probably better to say "Base implementation of ..."

sebright: Done

PR #269, opened 2017-05-01
Implement StatsContext.record(...) using StatsManagerImplBase.record(...).

sebright:
This change requires StatsContextImpl to store a reference to the
StatsManagerImplBase that is used to record the stats. It is a temporary change
that is only needed until we add the new tagging API.

This commit also updates the StatsContext tests so that they only record stats
for the one supported view.

____________________________________

This change is not backwards compatible, because StatsContext.record(...) is no longer a no-op.

/cc @zhangkun83 @dinooliva @a-veitch @bogdandrutu


core_impl/src/main/java/com/google/instrumentation/stats/StatsManagerImplBase.java
```diff
@@ -76,7 +79,11 @@ public View getView(ViewDescriptor viewDescriptor) {
 
   @Override
   StatsContextFactoryImpl getStatsContextFactory() {
-    return statsContextFactory;
+    StatsContextFactoryImpl factory = statsContextFactory;
```
songy23: The local variable `factory` seems redundant, could just use `statsContextFactory`.

core_impl/src/test/java/com/google/instrumentation/stats/StatsContextTest.java
```diff
@@ -74,53 +84,94 @@
...
       MeasurementMap measurements = MeasurementMap.of(descriptor, value);
       context.record(measurements);
-      //verify(context.context).record(measurements);
+      // TODO(sebright): Check the values in the view.
+      View view = statsManager.getView(RpcViewConstants.RPC_CLIENT_ROUNDTRIP_LATENCY_VIEW);
+      view.match(
+          new Function<DistributionView, Void>() {
+            @Override
+            public Void apply(DistributionView view) {
+              assertThat(view.getDistributionAggregations()).hasSize(1);
```
songy23: Nit: Verify the value of `DistributionAggregation`?

core_impl/src/test/java/com/google/instrumentation/stats/StatsContextTest.java
```diff
@@ -74,53 +84,94 @@
...
       MeasurementMap measurements = builder.put(descriptor, value).build();
       context.record(measurements);
-      //verify(context.context).record(measurements);
+      // TODO(sebright): Check the values in the view.
+      View view = statsManager.getView(RpcViewConstants.RPC_CLIENT_ROUNDTRIP_LATENCY_VIEW);
+      view.match(
+          new Function<DistributionView, Void>() {
+            @Override
+            public Void apply(DistributionView view) {
+              assertThat(view.getDistributionAggregations()).hasSize(1);
```
songy23: Same as above

core_impl/src/main/java/com/google/instrumentation/stats/StatsContextFactoryImpl.java
```diff
@@ -15,13 +15,20 @@
...
 /**
  * Native Implementation of {@link StatsContextFactory}.
  */
 final class StatsContextFactoryImpl extends StatsContextFactory {
-  static final StatsContextImpl DEFAULT = new StatsContextImpl(new HashMap<TagKey, TagValue>(0));
+  private final StatsManagerImplBase statsManager;
+  private final StatsContextImpl defaultStatsContext;
+
+  StatsContextFactoryImpl(StatsManagerImplBase statsManager) {
+    this.statsManager = statsManager;
```
songy23: `checkNotNull` on `statsManager` here?

core_impl/src/main/java/com/google/instrumentation/stats/StatsContextImpl.java
```diff
@@ -23,19 +23,22 @@
  * Native Implementation of {@link StatsContext}.
  */
 final class StatsContextImpl extends StatsContext {
+  private final StatsManagerImplBase statsManager;
   final Map<TagKey, TagValue> tags;
 
-  StatsContextImpl(Map<TagKey, TagValue> tags) {
+  StatsContextImpl(StatsManagerImplBase statsManager, Map<TagKey, TagValue> tags) {
+    this.statsManager = statsManager;
```
songy23: Same, `checkNotNull`?

core_impl/src/main/java/com/google/instrumentation/stats/StatsContextImpl.java
```diff
@@ -23,19 +23,22 @@
...
 
   @Override
   public Builder builder() {
-    return new Builder(tags);
+    return new Builder(statsManager, tags);
   }
 
   @Override
   public StatsContextImpl record(MeasurementMap stats) {
+    statsManager.record(this, stats);
```
songy23: Or if we allow `statsManager` to be Nullable, we need to add a check here.

PR #270, opened 2017-05-01
Add all the informations that we have in the Span to the SpanData.

bogdandrutu:


adriancole:
Is the dropped events something for noop/unsampled? Or are you anticipating
tracking errors parsing events? Or is it something else?

core_impl/src/main/java/com/google/instrumentation/trace/SpanData.java
```diff
@@ -144,4 +146,133 @@ public static SpanData create(
...
+     * Returns a new immutable {@code TimedEvent<T>}.
+     *
+     * @param timestamp the {@code Timestamp} of this event.
+     * @param event the event.
+     * @param <T> can be one of {@code Annotation} or {@code NetworkEvent}.
+     * @return a new immutable {@code TimedEvent<T>}
+     */
+    public static <T> TimedEvent<T> create(Timestamp timestamp, T event) {
+      return new AutoValue_SpanData_TimedEvent<T>(
+          checkNotNull(timestamp, "timestamp"), checkNotNull(event, "event"));
```
sebright: AutoValue already calls checkNotNull if the argument doesn't have `@Nullable`.

bogdandrutu:
The library will allow to configure the maximum number of events (of each type) to be kept in memory for each Span. If the number of recorded events is larger than the maximum set in the config then events will be dropped (probably we will keep the latest N events initially).

core_impl/src/test/java/com/google/instrumentation/trace/TimedEventTest.java
```diff
@@ -1,62 +0,0 @@
...
-  public void timedEvent_WithNetworkEvent() {
-    TimedEvent<NetworkEvent> timedEvent = new TimedEvent<NetworkEvent>(timestamp, networkEvent);
-    assertThat(timedEvent.getTimestamp()).isEqualTo(timestamp);
-    assertThat(timedEvent.getEvent()).isEqualTo(networkEvent);
-    assertThat(timedEvent.toString()).contains(timestamp.toString());
-    assertThat(timedEvent.toString()).contains(networkEvent.toString());
-  }
-
-  @Test
-  public void link_EqualsAndHashCode() {
```
sebright: It might be worth adding a simple test for SpanData.equals, to make sure that it and all of its fields override equals.

core_impl/src/main/java/com/google/instrumentation/trace/SpanData.java
bogdandrutu: Done.

core_impl/src/main/java/com/google/instrumentation/trace/SpanData.java
```diff
@@ -144,4 +146,133 @@ public static SpanData create(
...
+
+  @Immutable
+  @AutoValue
+  public abstract static class TimedEvents<T> {
+    /**
+     * Returns a new immutable {@code TimedEvents<T>}.
+     *
+     * @param events the list of events.
+     * @param droppedEventsCount the number of dropped events.
+     * @param <T> can be one of {@code Annotation} or {@code NetworkEvent}.
```
bogdandrutu: Done.

core_impl/src/main/java/com/google/instrumentation/trace/SpanData.java
```diff
@@ -144,4 +146,133 @@ public static SpanData create(
...
+  /** A timed event representation. It can be a timed {@link Annotation} or {@link NetworkEvent}. */
+  @Immutable
+  @AutoValue
+  public abstract static class TimedEvent<T> {
+    /**
+     * Returns a new immutable {@code TimedEvent<T>}.
+     *
+     * @param timestamp the {@code Timestamp} of this event.
+     * @param event the event.
+     * @param <T> can be one of {@code Annotation} or {@code NetworkEvent}.
```
bogdandrutu: Done.

PR #276, opened 2017-05-03
Start adding a "tags" package.

sebright:
This commit adds TagKey and TagSet.  The classes are not used yet.

_______________________________
This change is part of #251.

/cc @dinooliva @songy23 @bogdandrutu


core/src/main/java/com/google/instrumentation/tags/TagKey.java
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * @param <TagValueT> The type of value that can be paired with this {@code TagKey}.
+ */
+@Immutable
+@AutoValue
+public abstract class TagKey<TagValueT> {
+  /** The maximum length for a tag key name. */
+  public static final int MAX_LENGTH = StringUtil.MAX_LENGTH;
+
+  enum TagType {
```
bogdandrutu: Isn't the type encoded into the generic type?

core/src/main/java/com/google/instrumentation/tags/TagKey.java
```diff
@@ -0,0 +1,88 @@
...
+import static com.google.instrumentation.tags.TagKey.TagType.TAG_BOOL;
+import static com.google.instrumentation.tags.TagKey.TagType.TAG_INT;
+import static com.google.instrumentation.tags.TagKey.TagType.TAG_STRING;
+
+import com.google.auto.value.AutoValue;
+import com.google.instrumentation.internal.StringUtil;
+import javax.annotation.concurrent.Immutable;
+
+/**
+ * Tag key.
```
bogdandrutu: Please explain more about where is used and how is used. The current comment does not add any value.

core/src/main/java/com/google/instrumentation/tags/TagSet.java
```diff
@@ -0,0 +1,177 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.instrumentation.tags;
+
+import com.google.instrumentation.internal.StringUtil;
+import javax.annotation.Nullable;
+
+/** A set of tags. */
```
bogdandrutu: What is a tag?

core/src/main/java/com/google/instrumentation/tags/TagSet.java
```diff
@@ -0,0 +1,177 @@
...
+   * @param key the key to look up.
+   * @return {@code true} if the key is present.
+   */
+  public abstract boolean tagKeyExists(TagKey<?> key);
+
+  // TODO(sebright): Which is better, the generic "get" method or three specialized "get" methods?
+  // Should these methods take defaults or throw exceptions to handle missing keys?
+
+  /**
+   * Gets a value of type {@code String}.
```
bogdandrutu: s/Gets/Returns

Maybe change everywhere in this commit?

core/src/main/java/com/google/instrumentation/tags/TagSet.java
```diff
@@ -0,0 +1,177 @@
...
+    public abstract Builder insert(TagKey<String> key, String value);
+
+    /**
+     * Adds the key/value pair if the key is not present. If the key is present, it logs an error.
+     *
+     * @param key the key to look up.
+     * @param value the value to insert for the given key.
+     * @return this
+     */
+    public abstract Builder insert(TagKey<Long> key, long value);
```
bogdandrutu: We should follow java map or set (probably map) standards and use put or add instead.

core/src/main/java/com/google/instrumentation/tags/TagKey.java
sebright: I wasn't sure what to write, but I tried to add an explanation in the TagSet Javadoc.  Are the new comments accurate?

sebright:
I have a few questions about the API, which are in TODOs in TagSet.

core/src/main/java/com/google/instrumentation/tags/TagSet.java
adriancole: some of these imply the caller remembering if something was there before or a potentially racy check. Seems laborious vs put semantics. If this labor is worthwhile, we should probably explain in the type level docs why it is.

sebright: > some of these imply the caller remembering if something was there before or a potentially racy check.

What is wrong with a racy check?  It is done in the builder, which is already not thread-safe.

> Seems laborious vs put semantics.  If this labor is worthwhile, we should probably explain in the type level docs why it is.

"set" has the same behavior as "put".  I think that the others are just convenience methods to save the user from having to call tagKeyExists() when they want to do something that depends on the existing value of a key.

/cc @a-veitch

core/src/main/java/com/google/instrumentation/tags/TagKey.java
```diff
@@ -0,0 +1,89 @@
...
+ * A key to a value stored in a {@link TagSet}.
+ *
+ * @param <TagValueT> The type of value that can be paired with this {@code TagKey}. {@code TagKey}s
+ *     can only be instantiated with types {@code String}, {@code Long}, and {@code Boolean}.
+ */
+@Immutable
+@AutoValue
+public abstract class TagKey<TagValueT> {
+  /** The maximum length for a tag key name. */
+  public static final int MAX_LENGTH = StringUtil.MAX_LENGTH;
```
a-veitch: I suggest making the max a much smaller number for initial purposes - maybe 255?

core/src/main/java/com/google/instrumentation/tags/TagKey.java
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  TagKey() {}
+
+  /**
+   * Constructs a {@code TagKey<String>} from the given string. The string will be sanitized such
+   * that:
+   *
+   * <ol>
+   *   <li>length is restricted to {@link #MAX_LENGTH}, strings longer than that will be truncated.
```
a-veitch: We should discuss what policies we want here. I think it would be better to raise an exception, rather than silently truncate or replace characters.

core/src/main/java/com/google/instrumentation/tags/TagKey.java
```diff
@@ -0,0 +1,89 @@
...
+   *   <li>characters are restricted to printable ascii characters, non-printable characters will be
+   *       replaced by an underscore '_'.
+   * </ol>
+   */
+  public static TagKey<Boolean> createBoolean(String key) {
+    return new AutoValue_TagKey<Boolean>(StringUtil.sanitize(key), TAG_BOOL);
+  }
+
+  abstract String getName();
```
a-veitch: Don't need either getName or getTagType in the v1 API.

core/src/main/java/com/google/instrumentation/tags/TagSet.java
```diff
@@ -0,0 +1,188 @@
...
+/**
+ * A set of key-value pairs that can be used to label anything that is associated with a specific
+ * operation.
+ *
+ * <p>For example, {@code TagSet}s can be used to label stats, log messages, or debugging
+ * information.
+ */
+public abstract class TagSet {
+  /** The maximum length for a string tag value. */
+  public static final int MAX_STRING_LENGTH = StringUtil.MAX_LENGTH;
```
a-veitch: Why is this here? Raises potential for conflict with the same constant in TagKey. Suggest having this in one place only (probably TagKey).

core/src/main/java/com/google/instrumentation/tags/TagSet.java
```diff
@@ -0,0 +1,188 @@
...
+
+  /**
+   * Determines whether a key is present.
+   *
+   * @param key the key to look up.
+   * @return {@code true} if the key is present.
+   */
+  public abstract boolean tagKeyExists(TagKey<?> key);
```
a-veitch: This implies that only one TagKey of a given name can exist at a time, which is not the case (you can have two tags with same key name, but different types). Need to match both name and type.

core/src/main/java/com/google/instrumentation/tags/TagSet.java
```diff
@@ -0,0 +1,188 @@
...
+
+  // TODO(sebright): Which is better, the generic "get" method or three specialized "get" methods?
+  // The specialized "get" methods can return primitives, which could be less expensive.
```
a-veitch: I think specialization is better, but don't feel strongly about it.

core/src/main/java/com/google/instrumentation/tags/TagSet.java
```diff
@@ -0,0 +1,188 @@
...
+
+  // TODO(sebright): Should these methods take defaults or throw exceptions to handle missing keys?
```
a-veitch: The trouble with default values is that there may not be one - especially for bool. I think that leaves your only other option as being throwing; if users are want to avoid that, they can use tagKeyExists first.

core/src/main/java/com/google/instrumentation/tags/TagSet.java
sebright: Thanks.  I removed the generic "get" method.  The set/update/insert methods were already specialized.

core/src/main/java/com/google/instrumentation/tags/TagSet.java
sebright: I removed the default parameters.  I also made getStringTagValue throw an exception if the key is missing, for consistency.

core_impl/src/test/java/com/google/instrumentation/tags/TagSetTest.java
```diff
@@ -0,0 +1,180 @@
...
+import org.junit.runners.JUnit4;
+
+/** Tests for {@link TagSet}. */
+// TODO(sebright): Add more tests once the API is finalized.
+@RunWith(JUnit4.class)
+public class TagSetTest {
+
+  private static final TagKey<String> KS1 = TagKey.createString("k1");
+  private static final TagKey<String> KS2 = TagKey.createString("k2");
+  private static final TagKey<Long> KI = TagKey.createInt("k2");
```
a-veitch: Should add a test doing something similar to (same key name, different types):

newBuilder().set(KS2, "v1").set(KI, 5).build()

core_impl/src/main/java/com/google/instrumentation/internal/logging/Logger.java
```diff
@@ -0,0 +1,28 @@
+/*
```
a-veitch: I understand after reading further - log on errors. I think we can do away with this for now, since we have defined error-less operations for v1, and even when we do have them, we will (probably) want to use exceptions instead.

core_impl/src/test/java/com/google/instrumentation/tags/TagSetTest.java
sebright: Good catch!

sebright:
I rebased onto master and updated error handling for invalid tag key names.  Now the public TagKey factory methods throw IllegalArgumentException when the name is invalid.

sebright:
I temporarily removed all TagMap methods, except for `set` and `clear`, since it looks like we will definitely need those operations.  I can add other operations back in a separate PR, once we decide what we need.  This PR is ready for more review.

core/src/main/java/com/google/instrumentation/tags/TagMap.java
```diff
@@ -0,0 +1,129 @@
...
+    /**
+     * Adds the key/value pair regardless of whether the key is present.
+     *
+     * @param key the key to look up.
+     * @param value the value to set for the given key.
+     * @return this
+     */
+    public Builder set(TagKey<String> key, String value) {
+      Preconditions.checkArgument(key.getTagType() == TagType.TAG_STRING);
+      Preconditions.checkArgument(StringUtil.isValid(value));
```
sebright: Should we add a class to represent a string tag value, so that we can validate the string once and use it many times?

core/src/main/java/com/google/instrumentation/tags/TagKey.java
```diff
@@ -0,0 +1,126 @@
...
+
+  TagKey() {}
+
+  /**
+   * Constructs a {@code TagKey<String>} from the given string. The string must meet the following
+   * requirements:
+   *
+   * <ol>
+   *   <li>It cannot be longer than {@link #MAX_LENGTH}.
+   *   <li>It can only contain printable ascii characters.
```
a-veitch: Nit: upper-case ASCII

core/src/main/java/com/google/instrumentation/tags/TagKey.java
```diff
@@ -0,0 +1,126 @@
...
+   *
+   * @param name the name of the key.
+   * @throws IllegalArgumentException if the name is not valid.
+   */
+  public static TagKey<String> createString(String name) {
+    Preconditions.checkArgument(StringUtil.isValid(name));
+    return createStringInternal(name);
+  }
+
+  // Creates a TagKey<String> by sanitizing the given String.
```
a-veitch: Not sure at all we want/need this functionality.

core/src/main/java/com/google/instrumentation/tags/TagKey.java
```diff
@@ -0,0 +1,126 @@
...
+   *
+   * <ol>
+   *   <li>It cannot be longer than {@link #MAX_LENGTH}.
+   *   <li>It can only contain printable ascii characters.
+   * </ol>
+   *
+   * @param name the name of the key.
+   * @throws IllegalArgumentException if the name is not valid.
+   */
+  public static TagKey<Long> createInt(String name) {
```
a-veitch: I think the name should be createLong as well - otherwise it's confusing.

core/src/main/java/com/google/instrumentation/tags/TagMap.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    Builder() {
+      this.tags = new HashMap<TagKey<?>, Object>();
+    }
+
+    /**
+     * Adds the key/value pair regardless of whether the key is present.
+     *
+     * @param key the key to look up.
```
a-veitch: Suggest: @param key the {@code TagKey} which will be set

core/src/main/java/com/google/instrumentation/tags/TagMap.java
```diff
@@ -0,0 +1,129 @@
...
+     * @param value the value to set for the given key.
```
a-veitch: Need to add @throws documentation.

core/src/main/java/com/google/instrumentation/tags/TagKey.java
a-veitch: Also other names, e.g. s/TAG_INT/TAG_LONG/

core/src/main/java/com/google/instrumentation/tags/TagMap.java
a-veitch: That would be fine as an optimization. Up to you whether you want to do this now, or add a todo.

core/src/main/java/com/google/instrumentation/tags/TagKey.java
sebright: Done

core/src/main/java/com/google/instrumentation/tags/TagKey.java
sebright: I removed it.

core/src/main/java/com/google/instrumentation/tags/TagKey.java
sebright: Done

core/src/main/java/com/google/instrumentation/tags/TagMap.java
sebright: Thanks.  I added a TODO.

core/src/main/java/com/google/instrumentation/tags/TagMap.java
sebright: Done

core/src/main/java/com/google/instrumentation/tags/TagMap.java
sebright: Done

sebright:
I renamed the `tags` package to `io.opencensus.tags`.  The name isn't finalized, but it makes it clear that the package is part of the new API.

core/src/main/java/io/opencensus/tags/TagKey.java
```diff
@@ -0,0 +1,111 @@
...
+ * @param <TagValueT> The type of value that can be paired with this {@code TagKey}. {@code TagKey}s
+ *     can only be instantiated with types {@code String}, {@code Long}, and {@code Boolean}.
+ */
+@Immutable
+@AutoValue
+public abstract class TagKey<TagValueT> {
+  /** The maximum length for a tag key name. */
+  public static final int MAX_LENGTH = StringUtil.MAX_LENGTH;
+
+  enum TagType {
```
bogdandrutu: I thought we are going to support only String type initially.

core/src/main/java/io/opencensus/tags/TagKey.java
```diff
@@ -0,0 +1,111 @@
...
+   *
+   * @param name the name of the key.
+   * @throws IllegalArgumentException if the name is not valid.
+   */
+  public static TagKey<String> createString(String name) {
+    Preconditions.checkArgument(StringUtil.isValid(name));
+    return createStringInternal(name);
+  }
+
+  private static TagKey<String> createStringInternal(String name) {
```
bogdandrutu: Why do you need this extra private method?

core/src/main/java/io/opencensus/tags/TagKey.java
```diff
@@ -0,0 +1,111 @@
...
+  enum TagType {
+    TAG_STRING,
+    TAG_LONG,
+    TAG_BOOLEAN
+  }
+
+  TagKey() {}
+
+  /**
+   * Constructs a {@code TagKey<String>} from the given string. The string must meet the following
```
bogdandrutu: The comment is confusing "from the given string". Probably better to say something like "with the given {@code name}" or something similar.

Also I would move the second sentence on a different line:
<p>The string must meet the following requirements:

core/src/main/java/io/opencensus/tags/TagKey.java
```diff
@@ -0,0 +1,111 @@
...
+   *
+   * <ol>
+   *   <li>It cannot be longer than {@link #MAX_LENGTH}.
+   *   <li>It can only contain printable ASCII characters.
+   * </ol>
+   *
+   * @param name the name of the key.
+   * @throws IllegalArgumentException if the name is not valid.
+   */
+  public static TagKey<String> createString(String name) {
```
bogdandrutu: s/createString/createStringKey?

core/src/main/java/io/opencensus/tags/TagKey.java
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * A key to a value stored in a {@link TagMap}.
+ *
+ * @param <TagValueT> The type of value that can be paired with this {@code TagKey}. {@code TagKey}s
+ *     can only be instantiated with types {@code String}, {@code Long}, and {@code Boolean}.
+ */
+@Immutable
+@AutoValue
+public abstract class TagKey<TagValueT> {
```
bogdandrutu: The current implementation defines equals/hashCode <name, type>. I think we haven't decided on this but you should be aware of this.

core/src/test/java/io/opencensus/tags/TagKeyTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+/** Tests for {@link TagKey} */
+@RunWith(JUnit4.class)
+public final class TagKeyTest {
+
+  @Rule public final ExpectedException thrown = ExpectedException.none();
+
+  @Test
+  public void testGetName() {
+    assertThat(TagKey.createString("foo").getName()).isEqualTo("foo");
```
bogdandrutu: Maybe a test for getType?

core/src/main/java/io/opencensus/tags/TagKey.java
sebright: Yes, it shouldn't matter while we are only allowing String tags.

core/src/main/java/io/opencensus/tags/TagKey.java
sebright: I made the factory methods for `TagKey<Long>` and `TagKey<Boolean>` non-public.

core/src/main/java/io/opencensus/tags/TagKey.java
sebright: Done

core/src/main/java/io/opencensus/tags/TagKey.java
sebright: Done.  I was trying to keep the name short, but createStringKey isn't much longer.

core/src/main/java/io/opencensus/tags/TagKey.java
sebright: Previously, there were two factory methods for each type of key, but the private methods aren't needed now.

core/src/test/java/io/opencensus/tags/TagKeyTest.java
sebright: Done

core/src/main/java/io/opencensus/tags/TagKey.java
```diff
@@ -0,0 +1,104 @@
...
+   * <ol>
+   *   <li>It cannot be longer than {@link #MAX_LENGTH}.
+   *   <li>It can only contain printable ASCII characters.
+   * </ol>
+   *
+   * @param name the name of the key.
+   * @throws IllegalArgumentException if the name is not valid.
+   */
+  public static TagKey<String> createStringKey(String name) {
+    Preconditions.checkArgument(StringUtil.isValid(name));
```
bogdandrutu: nit: Why not static import checkArguments? This is how we do everywhere else.

core/src/main/java/io/opencensus/tags/TagKey.java
```diff
@@ -0,0 +1,104 @@
...
+    TAG_LONG,
+    TAG_BOOLEAN
+  }
+
+  TagKey() {}
+
+  /**
+   * Constructs a {@code TagKey<String>} with the given name.
+   *
+   * <p>The string must meet the following requirements:
```
bogdandrutu: s/string/{@code name}

Probably everywhere else.

core/src/main/java/io/opencensus/tags/TagMap.java
```diff
@@ -0,0 +1,137 @@
...
+      return this;
+    }
+
+    /**
+     * Creates a {@code TagMap} from this builder.
+     *
+     * @return a {@code TagMap} with the same tags as this builder.
+     */
+    public TagMap build() {
+      return new TagMap(new HashMap<TagKey<?>, Object>(tags));
```
bogdandrutu: collections.unmodifiablemap to avoid any mistake?

core/src/main/java/io/opencensus/tags/TagMap.java
```diff
@@ -0,0 +1,137 @@
...
+     * @param value the value to set for the given key.
+     * @return this
+     * @throws IllegalArgumentException if either argument is null, the key is the wrong type, or
+     *     the value contains unprintable characters or is longer than {@link
+     *     TagMap#MAX_STRING_LENGTH}.
+     */
+    public Builder set(TagKey<String> key, String value) {
+      Preconditions.checkArgument(key.getTagType() == TagType.TAG_STRING);
+
+      // TODO(sebright): Consider adding a TagValue class to avoid validating the String every time
```
bogdandrutu: I think this is a good requirement to the design.

core/src/main/java/io/opencensus/tags/TagMap.java
```diff
@@ -0,0 +1,137 @@
...
+
+    /**
+     * Adds the key/value pair regardless of whether the key is present.
+     *
+     * @param key the {@code TagKey} which will be set.
+     * @param value the value to set for the given key.
+     * @return this
+     * @throws IllegalArgumentException if the key is null or the key is the wrong type.
+     */
+    public Builder set(TagKey<Boolean> key, boolean value) {
```
bogdandrutu: not public please.

core/src/main/java/io/opencensus/tags/TagMap.java
```diff
@@ -0,0 +1,137 @@
...
+
+    /**
+     * Adds the key/value pair regardless of whether the key is present.
+     *
+     * @param key the {@code TagKey} which will be set.
+     * @param value the value to set for the given key.
+     * @return this
+     * @throws IllegalArgumentException if the key is null or the key is the wrong type.
+     */
+    public Builder set(TagKey<Long> key, long value) {
```
bogdandrutu: not public please.

core/src/main/java/io/opencensus/tags/TagMap.java
```diff
@@ -0,0 +1,137 @@
...
+   * Returns a builder based on this {@code TagMap}.
+   *
+   * @return a builder based on this {@code TagMap}.
+   */
+  public Builder toBuilder() {
+    return new Builder(getTags());
+  }
+
+  /** Builder for the {@link TagMap} class. */
+  public static final class Builder {
```
bogdandrutu: At one point I thought that we will support something like:

public static TagMap build(TagMap parent, List<TagMutation>);

I think this way we can avoid 1 of the 2 copies of the tags.

Maybe not in this PR because we want to move forward but might worth playing with that idea.

sebright:
Thanks!

core/src/main/java/io/opencensus/tags/TagMap.java
sebright: @dinooliva said that users probably won't add the same tag values every time, so it's probably not worth adding the `TagMap build(TagMap parent, List)` method.

core/src/main/java/io/opencensus/tags/TagKey.java
```diff
@@ -0,0 +1,104 @@
...
+   * <ol>
+   *   <li>It cannot be longer than {@link #MAX_LENGTH}.
+   *   <li>It can only contain printable ASCII characters.
+   * </ol>
+   *
+   * @param name the name of the key.
+   * @throws IllegalArgumentException if the name is not valid.
+   */
+  // TODO(sebright): Make this public once we support types other than String.
+  static TagKey<Long> createLongKey(String name) {
```
adriancole: to save future pull request/issue comments, I'd put a comment (not necessarily javadoc) rationale on choosing long vs float or something else to represent numbers. There are good reasons to choose Long, but in the future when people ask, this will help others who may not have made the decision help explain why.

sebright: /cc @dinooliva

sebright: I created an issue: #640

PR #277, opened 2017-05-04
Simplify StatsManager by delegating stats operations to new a class, ViewManager.

sebright:
Classes in the StatsManager class hierarchy should just specify the classes
that are used for different stats implementations.  This commit moves all
stats-handling logic out of StatsManagerImplBase into a new class, ViewManager.

### DIFF
--- a/core_impl/src/main/java/com/google/instrumentation/stats/StatsManagerImplBase.java
+++ b/core_impl/src/main/java/com/google/instrumentation/stats/StatsManagerImplBase.java
@@ -15,66 +15,32 @@ package com.google.instrumentation.stats;
 
 import com.google.instrumentation.common.Clock;
 import com.google.instrumentation.common.EventQueue;
-import com.google.instrumentation.common.Function;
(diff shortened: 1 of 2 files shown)


core_impl/src/main/java/com/google/instrumentation/stats/StatsManagerImplBase.java
```diff
@@ -15,66 +15,32 @@
...
-  private final EventQueue queue;
 
-  // clock used throughout the stats implementation
-  private final Clock clock;
-
-  private final MeasurementDescriptorToViewMap measurementDescriptorToViewMap =
-      new MeasurementDescriptorToViewMap();
+  // StatsManagerImplBase delegates all operations related to stats to ViewManager in order to keep
+  // StatsManagerImplBase simple.
+  private final ViewManager statsCollector;
```
songy23: Minor: rename this variable to `viewManager` to be consistent?

sebright: Thanks, I missed that when I renamed it.

PR #278, opened 2017-05-05
MeasurementDescriptorToViewMap synchronization

sebright:
Two main commits:

#### Refactor MeasurementDescriptorToViewMap.putView(...).

Previously, ViewManager.registerView() called
MeasurementDescriptorToViewMap.getView(...) followed by
MeasurementDescriptorToViewMap.putView(...) to insert a view only if it was not
present.  There was no locking, so the view could be created multiple times.
This commit refactors putView so that it covers the full transaction on the
underlying multimap.  It creates an empty MutableView if the view is not present
and inserts it into the multimap.

#### Use intrinsic locking in MeasurementDescriptorToViewMap.

MeasurementDescriptorToViewMap previously used a SynchronizedMultimap for
synchronization of its Multimap.  However, every method ended up needing to
access the Multimap within a synchronized block anyway, because the Multimap's
values were not synchronized.  It is simpler to just use a HashMultimap and
synchronize on the whole MeasurementDescriptorToViewMap object.

### DIFF
--- a/core_impl/src/main/java/com/google/instrumentation/stats/MeasurementDescriptorToViewMap.java
+++ b/core_impl/src/main/java/com/google/instrumentation/stats/MeasurementDescriptorToViewMap.java
@@ -17,11 +17,16 @@ import com.google.common.collect.HashMultimap;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Multimaps;
 import com.google.instrumentation.common.Clock;
+import com.google.instrumentation.common.Function;
+import com.google.instrumentation.stats.MutableView.MutableDistributionView;
+import com.google.instrumentation.stats.MutableView.MutableIntervalView;
+import com.google.instrumentation.stats.ViewDescriptor.DistributionViewDescriptor;
+import com.google.instrumentation.stats.ViewDescriptor.IntervalViewDescriptor;
 import java.util.Collection;
 
 /**
- * A class that stores a singleton map from {@link MeasurementDescriptor.Name}s to
- * {@link MutableView}s.
+ * A class that stores a singleton map from {@link MeasurementDescriptor.Name}s to {@link
+ * MutableView}s.
  */
 final class MeasurementDescriptorToViewMap {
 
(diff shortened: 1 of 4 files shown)


core_impl/src/main/java/com/google/instrumentation/stats/MeasurementDescriptorToViewMap.java
```diff
@@ -34,55 +34,49 @@
...
-        return;
-      }
-      MutableView mutableView =
-          viewDescriptor.match(
-              new CreateMutableDistributionViewFunction(clock),
-              new CreateMutableIntervalViewFunction());
-      mutableMap.put(
-          viewDescriptor.getMeasurementDescriptor().getMeasurementDescriptorName(), mutableView);
+  synchronized void registerView(ViewDescriptor viewDescriptor, Clock clock) {
+    if (getView(viewDescriptor, clock) != null) {
```
songy23: Minor: this could be changed to `getMutableView()`

core_impl/src/main/java/com/google/instrumentation/stats/ViewManager.java
```diff
@@ -82,42 +65,15 @@ void record(StatsContextImpl tags, MeasurementMap measurementValues) {
     private final MeasurementMap stats;
     private final ViewManager statsCollector;
 
-    StatsEvent(
-        ViewManager statsCollector, StatsContextImpl tags, MeasurementMap stats) {
+    StatsEvent(ViewManager statsCollector, StatsContextImpl tags, MeasurementMap stats) {
```
songy23: Same

core_impl/src/main/java/com/google/instrumentation/stats/ViewManager.java
```diff
@@ -82,42 +65,15 @@ void record(StatsContextImpl tags, MeasurementMap measurementValues) {
     private final MeasurementMap stats;
     private final ViewManager statsCollector;
```
songy23: minor: statsCollector -> viewManager

sebright: Sorry, I missed this occurrence in the last PR.

core_impl/src/main/java/com/google/instrumentation/stats/MeasurementDescriptorToViewMap.java
sebright: Thanks, I also refactored getView to use getMutableView.

PR #281, opened 2017-05-05
Add a class to represent a ViewDescriptor name.

sebright:
This class can be used to query Views, since ViewDescriptor names must be
unique. This commit also prevents a view from being registered if it has the
same name as an existing view but is not the same view.

/cc @dinooliva @bogdandrutu @a-veitch

### DIFF
--- a/core/src/main/java/com/google/instrumentation/stats/ViewDescriptor.java
+++ b/core/src/main/java/com/google/instrumentation/stats/ViewDescriptor.java
@@ -28,30 +28,24 @@ public abstract class ViewDescriptor {
   /**
    * Name of view. Must be unique.
    */
-  public final Name getViewDescriptorName() {
-    return name;
-  }
+  public abstract Name getViewDescriptorName();
 
   /**
    * Name of view, as a {@code String}.
    */
   public final String getName() {
-    return name.asString();
+    return getViewDescriptorName().asString();
   }
 
   /**
    * More detailed description, for documentation purposes.
    */
-  public final String getDescription() {
-    return description;
-  }
+  public abstract String getDescription();
 
   /**
    * Measurement type of this view.
    */
-  public final MeasurementDescriptor getMeasurementDescriptor() {
-    return measurementDescriptor;
-  }
+  public abstract MeasurementDescriptor getMeasurementDescriptor();
 
   /**
    * Tag keys to match with the associated {@link MeasurementDescriptor}. If no keys are specified,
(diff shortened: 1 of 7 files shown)


sebright:
I also changed ViewDescriptor to use AutoValue, so that it overrides equals(...).

core_impl/src/main/java/com/google/instrumentation/stats/MeasurementDescriptorToViewMap.java
```diff
@@ -38,29 +41,46 @@
...
     Collection<MutableView> views =
         mutableMap.get(viewDescriptor.getMeasurementDescriptor().getMeasurementDescriptorName());
     for (MutableView view : views) {
-      if (view.getViewDescriptor().equals(viewDescriptor)) {
+      if (view.getViewDescriptor().getViewDescriptorName().equals(viewName)) {
         return view;
       }
     }
-    return null;
+    throw new AssertionError("registeredViews and mutableMap contain different views: "
```
songy23: Will this ever be reached? (since we already performed check in `registerView()`) 
Also since users don't know implementation details on `registeredViews` and `mutableMap`, this error message could be changed to something like `"View mismatch: Not recording stats for view: " + viewName`

sebright: > Will this ever be reached? (since we already performed check in registerView())

The line should be unreachable, unless there is a bug in this class.  I couldn't think of a way to avoid it.

> Also since users don't know implementation details on registeredViews and mutableMap, this error message could be changed to something like "View mismatch: Not recording stats for view: " + viewName

I tried to make the message more user friendly but also indicate that it is an internal error.  I wanted to keep the variables at the end of the message for debugging, in case this error message ever appears in a bug report.

core/src/main/java/com/google/instrumentation/stats/ViewDescriptor.java
```diff
@@ -93,80 +104,95 @@ public static DistributionViewDescriptor create(
...
+     * Constructs a new {@link DistributionViewDescriptor}.
+     */
+    public static DistributionViewDescriptor create(
+        Name name,
+        String description,
+        MeasurementDescriptor measurementDescriptor,
+        DistributionAggregationDescriptor distributionAggregationDescriptor,
+        List<TagKey> tagKeys) {
+      return new AutoValue_ViewDescriptor_DistributionViewDescriptor(
+          name,
```
sebright: AutoValue inserts null checks for any parameters that don't have `@Nullable`.

sebright:
Thanks!

PR #282, opened 2017-05-05
Update BinaryPropagationHandler to not be a singleton to be consistent with other implementations.

bogdandrutu:


core_impl/src/main/java/com/google/instrumentation/trace/TraceComponentImplBase.java
```diff
@@ -19,7 +19,7 @@
 public class TraceComponentImplBase extends TraceComponent {
   private static final Tracer tracer = Tracer.getNoopTracer();
   private static final BinaryPropagationHandler binaryPropagationHandler =
-      BinaryPropagationHandlerImpl.INSTANCE;
+      new BinaryPropagationHandlerImpl();
```
sebright: Should this field be non-static, too?

bogdandrutu: Done.

PR #283, opened 2017-05-07
Add hasRemoteParent property to SpanImpl and SpanData.

bogdandrutu:


adriancole:
What is the use case for this? For dependency link filtering?

PR #284, opened 2017-05-08
Add span factory implementation.

bogdandrutu:


core_impl/src/main/java/com/google/instrumentation/trace/SpanFactoryImpl.java
```diff
@@ -0,0 +1,131 @@
...
+import com.google.instrumentation.trace.Span.Options;
+import com.google.instrumentation.trace.TraceConfig.TraceParams;
+import java.util.EnumSet;
+import java.util.List;
+import java.util.Random;
+import javax.annotation.Nullable;
+
+/** Implementation of the {@link SpanFactory}. */
+final class SpanFactoryImpl extends SpanFactory {
+  private static final EnumSet<Span.Options> noRecordEventOptions =
```
a-veitch: I would remove both of the RecordEventOptions values. At the moment, we only have the one option, but in the future we may have more, in which case this code will need to be changed anyway - may as well assume that now. See comments later whn these values are used. to create spanOptions.

core_impl/src/main/java/com/google/instrumentation/trace/SpanFactoryImpl.java
```diff
@@ -0,0 +1,131 @@
...
+    Sampler sampler =
+        startSpanOptions.getSampler() != null
+            ? startSpanOptions.getSampler()
+            : activeTraceParams.getSampler();
+    if (sampler.shouldSample(
+        parent, hasRemoteParent, traceId, spanId, name, startSpanOptions.getParentLinks())) {
+      traceOptionsBuilder.setIsSampled();
+    }
+    TraceOptions traceOptions = traceOptionsBuilder.build();
+    EnumSet<Span.Options> spanOptions = noRecordEventOptions;
```
a-veitch: Change this to:

EnumSet<Span.Options> spanOptions = EnumSet.noneOf(Span.Options.class);

To make it clear that there are no options set (doesn't assume that record events is the only option)

core_impl/src/main/java/com/google/instrumentation/trace/SpanFactoryImpl.java
```diff
@@ -0,0 +1,131 @@
...
+    if (traceOptions.isSampled()
+        || (startSpanOptions.getRecordEvents() != null && startSpanOptions.getRecordEvents())) {
+      spanOptions = recordEventOptions;
```
a-veitch: And change this to:

spanOptions.add(Options.RECORD_EVENTS);

core_impl/src/main/java/com/google/instrumentation/trace/SpanFactoryImpl.java
bogdandrutu: Done.

core_impl/src/main/java/com/google/instrumentation/trace/SpanFactoryImpl.java
bogdandrutu: Done.

core_impl/src/main/java/com/google/instrumentation/trace/SpanFactoryImpl.java
bogdandrutu: Done.

core_impl/src/main/java/com/google/instrumentation/trace/SpanImpl.java
```diff
@@ -91,6 +91,15 @@ String getName() {
   }
 
   /**
+   * Returns the {@code TimestampConverter} used by this {@code Span}.
+   *
+   * @return the {@code TimestampConverter} used by this {@code Span}.
+   */
+  TimestampConverter getTimestampConverter() {
```

core_impl/src/main/java/com/google/instrumentation/trace/SpanFactoryImpl.java
```diff
@@ -0,0 +1,131 @@
...
+      String name,
+      StartSpanOptions startSpanOptions,
+      TimestampConverter timestampConverter) {
+    TraceParams activeTraceParams = traceConfig.getActiveTraceParams();
+    Random random = randomHandler.current();
+    TraceId traceId;
+    SpanId spanId = SpanId.generateRandomId(random);
+    SpanId parentSpanId = null;
+    TraceOptions.Builder traceOptionsBuilder;
+    if (parent == null || !parent.isValid()) {
```
bogdandrutu: So currently if the user explicitly set the option for this to become a root we pass null, but if no current Span in the context and the user does not specify the parent (or force to be a root) the context is invalid and we will pass the invalid context, or the last option is if the user cannot parse a receiving context from the wire they may pass invalid context instead of null. 

I know it is horrible and maybe I should change the API to always pass null.

sebright: Thanks, that makes sense.

PR #285, opened 2017-05-08
Use AutoValue to implement BucketBoundaries and DistributionAggregationDescriptor.

sebright:


core/src/main/java/com/google/instrumentation/stats/BucketBoundaries.java
```diff
@@ -16,15 +16,16 @@
...
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
 /**
  * The optional histogram bucket boundaries for a {@link Distribution}.
  */
-public final class BucketBoundaries {
-  private final List<Double> bucketBoundaries;
+@AutoValue
```
bogdandrutu: Use "@Immutable" for these classes.

sebright: Thanks.  I added the annotation to all of the AutoValue stats classes.

PR #286, opened 2017-05-08
Add an export service handler which logs all the SpanData.

bogdandrutu:


core/src/main/java/com/google/instrumentation/trace/TraceExporter.java
```diff
@@ -75,6 +77,31 @@
...
+   * <pre>{@code
+   * public static void main(String[] args) {
+   *   Tracing.getTraceExporter().registerServiceHandler(
+   *       "com.google.instrumentation.LoggingServiceHandler", new LoggingServiceHandler());
+   *   // ...
+   * }
+   * }</pre>
+   *
+   */
+  public static final class LoggingServiceHandler extends ServiceHandler {
```
sebright: If this class is part of the API, I think it should have a factory method and a private constructor.

bogdandrutu: Agreed. Changed.

PR #287, opened 2017-05-09
Add StartEndHandler implementation. Supports only export for the moment.

bogdandrutu:


core/src/main/java/com/google/instrumentation/trace/testing/FakeServiceHandler.java
```diff
@@ -0,0 +1,63 @@
...
+
+  /**
+   * Waits until we received numberOfSpans spans to export. Returns the list of exported {@link
+   * SpanData} objects, otherwise {@code null} if the current thread is interrupted.
+   *
+   * @param numberOfSpans the number of minimum spans to be collected.
+   * @return the list of exported {@link SpanData} objects, otherwise {@code null} if the current
+   *     thread is interrupted.
+   */
+  public List<SpanData> waitForExport(int numberOfSpans) {
```
sebright: This method could use `@Nullable`.

core_impl/src/test/java/com/google/instrumentation/trace/StartEndHandlerImplTest.java
```diff
@@ -0,0 +1,75 @@
...
+  @Test
+  public void exportSpans() {
+    SpanImpl span1 = createEndedSpan(SPAN_NAME_1);
+    SpanImpl span2 = createEndedSpan(SPAN_NAME_2);
+    List<SpanData> exported = serviceHandler.waitForExport(2);
+    assertThat(exported.size()).isEqualTo(2);
+    assertThat(exported.get(0)).isEqualTo(span1.toSpanData());
+    assertThat(exported.get(1)).isEqualTo(span2.toSpanData());
+  }
+}
```
sebright: Should there be a test that checks that non-sampled spans are not exported?

PR #288, opened 2017-05-09
Change TraceExporterImpl to implement also the StartEndHandler.

bogdandrutu:
Supports only service export for the moment.

Different design for https://github.com/google/instrumentation-java/pull/287

### DIFF
--- a/core_impl/src/main/java/com/google/instrumentation/trace/TraceExporterImpl.java
+++ b/core_impl/src/main/java/com/google/instrumentation/trace/TraceExporterImpl.java
@@ -14,6 +14,8 @@
 package com.google.instrumentation.trace;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.instrumentation.common.EventQueue;
+import com.google.instrumentation.trace.SpanImpl.StartEndHandler;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.LinkedList;
@@ -24,11 +26,17 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.annotation.concurrent.GuardedBy;
 
-/** Implementation of the {@link TraceExporter}. */
-final class TraceExporterImpl extends TraceExporter {
+/**
+ * Implementation of the {@link TraceExporter}, implements also {@link StartEndHandler}.
+ *
+ * <p>Uses the provided {@link EventQueue} to defer processing/exporting of the {@link SpanData}
+ * to avoid impact on the critical path.
+ */
+final class TraceExporterImpl extends TraceExporter implements StartEndHandler {
   private static final Logger logger = Logger.getLogger(TraceExporter.class.getName());
 
-  private final WorkerThread workerThread;
+  private final ServiceExporterThread serviceExporterThread;
+  private final EventQueue eventQueue;
 
   /**
    * Constructs a {@code TraceExporterImpl} that exports the {@link SpanData} asynchronously.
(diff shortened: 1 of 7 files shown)


core_impl/src/main/java/com/google/instrumentation/trace/TraceExporterImpl.java
```diff
@@ -132,10 +144,13 @@ private void onBatchExport(List<SpanData> spanDataList) {
       }
     }
 
-    private WorkerThread(int bufferSize, long scheduleDelayMillis) {
+    private ServiceExporterThread(int bufferSize, long scheduleDelayMillis) {
+      super();
```
sebright: `super()` isn't necessary.

core_impl/src/main/java/com/google/instrumentation/trace/TraceExporterImpl.java
```diff
@@ -24,11 +26,17 @@
 import java.util.logging.Logger;
 import javax.annotation.concurrent.GuardedBy;
 
-/** Implementation of the {@link TraceExporter}. */
-final class TraceExporterImpl extends TraceExporter {
+/**
+ * Implementation of the {@link TraceExporter}, implements also {@link StartEndHandler}.
+ *
+ * <p>Uses the provided {@link EventQueue} to defer processing/exporting of the {@link SpanData}
+ * to avoid impact on the critical path.
```
a-veitch: s/ on/ing/

core_impl/src/main/java/com/google/instrumentation/trace/TraceExporterImpl.java
```diff
@@ -40,41 +48,45 @@
...
-   * Adds a {@link SpanImpl} to be exported.
-   *
-   * @param span the {@code SpanImpl} that will be exported.
-   */
-  void addSpan(SpanImpl span) {
-    workerThread.addSpan(span);
+
+  @Override
+  public void onStart(SpanImpl span) {
+    // Do nothing.
```
a-veitch: Is there nothing we can do int he background for starting?

core_impl/src/test/java/com/google/instrumentation/trace/TraceExporterImplTest.java
```diff
@@ -149,30 +114,72 @@ public void serviceHandlerThrowsException() {
...
     assertThat(exported2.size()).isEqualTo(2);
-    assertThat(exported1.get(0).getDisplayName()).isEqualTo(SPAN_NAME_1);
-    assertThat(exported2.get(0).getDisplayName()).isEqualTo(SPAN_NAME_1);
-    assertThat(exported1.get(1).getDisplayName()).isEqualTo(SPAN_NAME_2);
-    assertThat(exported2.get(1).getDisplayName()).isEqualTo(SPAN_NAME_2);
+    assertThat(exported1.get(0)).isEqualTo(span1.toSpanData());
+    assertThat(exported2.get(0)).isEqualTo(span1.toSpanData());
+    assertThat(exported1.get(1)).isEqualTo(span2.toSpanData());
+    assertThat(exported2.get(1)).isEqualTo(span2.toSpanData());
+  }
```
sebright: I think there should be a test that checks that non-sampled spans are not exported.

bogdandrutu: Done.

core_impl/src/main/java/com/google/instrumentation/trace/TraceExporterImpl.java
bogdandrutu: Done.

core_impl/src/test/java/com/google/instrumentation/trace/TraceExporterImplTest.java
```diff
@@ -141,6 +159,20 @@ public void exportSpansToMultipleServices() {
...
+  public void exportNotSampledSpans() {
+    SpanImpl span1 = createNotSampledEndedSpan(SPAN_NAME_1);
+    SpanImpl span2 = createSampledEndedSpan(SPAN_NAME_2);
+    // Spans are recorded and exported in the same order as they are ended, we test that a non
+    // sampled span is not exported by creating and ending a sampled span after a non sampled span
+    // and checking that the first exported span is the sampled span (the non sampled did not get
+    // exported).
+    List<SpanData> exported = serviceHandler.waitForExport(1);
+    assertThat(exported.size()).isEqualTo(1);
+    assertThat(exported.get(0)).isNotEqualTo(span1.toSpanData());
```
sebright: >  assertThat(exported.get(0)).isNotEqualTo(span1.toSpanData());

I think this assertion is redundant with the one below, since the two spans have different names.

bogdandrutu: I need to use the variable :). I can simply remove the variable span1, but people may get confused and think that the span is not in the exported list because we do not keep a reference to it.

sebright: What if you created the spans separately from ending them, to make the export more explicit?

```
    SpanImpl span1 = createNotSampledSpan(SPAN_NAME_1);
    SpanImpl span2 = createSampledSpan(SPAN_NAME_2);
    span1.end();
    span2.end();
```

createSampledEndedSpan could call createSampledSpan.  It's also fine as it is.

PR #293, opened 2017-05-10
Use AutoValue to implement DistributionAggregation.

sebright:


core/src/main/java/com/google/instrumentation/stats/DistributionAggregation.java
```diff
@@ -29,64 +31,61 @@
...
+  public abstract double getSum();
 
   /**
    * The range of the population values. If {@link #getCount()} is zero then this returned range is
    * implementation-dependent.
    */
-  public Range getRange() {
-    return range;
-  }
+  public abstract Range getRange();
```
bogdandrutu: Should we comment something like:
"Returns the range ..."

Also probably better to copy the same in "@return ...", even if it repeats the same thing the javadoc looks better.

sebright:
I made a PR to clean up the Javadocs: #294

PR #297, opened 2017-05-11
Add benchmarks for start/end span and record trace events.

bogdandrutu:
Benchmarks ran with https://github.com/google/instrumentation-java/pull/296 (mean values)

BinaryPropagationHandlerImplBenchmark.fromBinaryValueSpanContext      66.611 ±  1.568  ns/op
BinaryPropagationHandlerImplBenchmark.toBinaryValueSpanContext          67.977 ±  2.524  ns/op
BinaryPropagationHandlerImplBenchmark.toFromBinarySpanContext           146.850 ± 39.152  ns/op
RecordTraceEventsNonSampledSpanBenchmark.addAnnotation                  55.014 ±  1.872  ns/op
RecordTraceEventsNonSampledSpanBenchmark.addLink                             46.806 ±  1.752  ns/op
RecordTraceEventsNonSampledSpanBenchmark.addNetworkEvent             49.202 ±  1.996  ns/op
RecordTraceEventsSampledSpanBenchmark.addAnnotation                         79.445 ± 40.390  ns/op
RecordTraceEventsSampledSpanBenchmark.addLink                                    59.594 ± 17.770  ns/op
RecordTraceEventsSampledSpanBenchmark.addNetworkEvent                    48.520 ±  1.840  ns/op
StartEndSpanBenchmark.startEndNonSampledChildSpan                             68.181 ±  8.417  ns/op
StartEndSpanBenchmark.startEndNonSampledRootSpan                              69.895 ±  3.243  ns/op
StartEndSpanBenchmark.startEndRecordEventsChildSpan                            61.785 ± 24.352  ns/op
StartEndSpanBenchmark.startEndRecordEventsRootSpan                             58.410 ±  3.025  ns/op
StartEndSpanBenchmark.startEndSampledChildSpan                                     61.010 ± 12.419  ns/op
StartEndSpanBenchmark.startEndSampledRootSpan                                      61.113 ±  3.164  ns/op

### DIFF
--- a/benchmarks/src/jmh/java/com/google/instrumentation/trace/RecordTraceEventsNonSampledSpanBenchmark.java
+++ b/benchmarks/src/jmh/java/com/google/instrumentation/trace/RecordTraceEventsNonSampledSpanBenchmark.java
@@ -14,7 +14,6 @@
 package com.google.instrumentation.trace;
 
 import java.util.HashMap;
-import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import org.openjdk.jmh.annotations.Benchmark;
 import org.openjdk.jmh.annotations.BenchmarkMode;
@@ -44,7 +43,7 @@ public class RecordTraceEventsNonSampledSpanBenchmark {
     linkedSpan.end();
   }
 
-  /** This benchmark attempts to measure performance of {@link Span#addAttributes(Map)}. */
+  /** This benchmark attempts to measure performance of adding an attribute to the span. */
   @Benchmark
   @BenchmarkMode(Mode.SampleTime)
   @OutputTimeUnit(TimeUnit.NANOSECONDS)
@@ -55,7 +54,7 @@ public class RecordTraceEventsNonSampledSpanBenchmark {
     return span;
   }
 
-  /** This benchmark attempts to measure performance of {@link Span#addAnnotation(String)}. */
+  /** This benchmark attempts to measure performance of adding an annotation to the span. */
   @Benchmark
   @BenchmarkMode(Mode.SampleTime)
   @OutputTimeUnit(TimeUnit.NANOSECONDS)
@@ -64,9 +63,7 @@ public class RecordTraceEventsNonSampledSpanBenchmark {
     return span;
   }
 
-  /**
-   * This benchmark attempts to measure performance of {@link Span#addNetworkEvent(NetworkEvent)}.
-   */
+  /** This benchmark attempts to measure performance of adding a network event to the span. */
   @Benchmark
   @BenchmarkMode(Mode.SampleTime)
   @OutputTimeUnit(TimeUnit.NANOSECONDS)
@@ -75,9 +72,7 @@ public class RecordTraceEventsNonSampledSpanBenchmark {
     return span;
   }
 
-  /**
-   * This benchmark attempts to measure performance of {@link Span#addLink(Link)}.
-   */
+  /** This benchmark attempts to measure performance of adding a link to the span. */
   @Benchmark
   @BenchmarkMode(Mode.SampleTime)
   @OutputTimeUnit(TimeUnit.NANOSECONDS)
--- a/benchmarks/src/jmh/java/com/google/instrumentation/trace/RecordTraceEventsNonSampledSpanBenchmark.java
+++ b/benchmarks/src/jmh/java/com/google/instrumentation/trace/RecordTraceEventsNonSampledSpanBenchmark.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2017, Google Inc.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.instrumentation.trace;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.TearDown;
+
+/** Benchmarks for {@link SpanImpl} to record trace events. */
+@State(Scope.Benchmark)
+public class RecordTraceEventsNonSampledSpanBenchmark {
+  private static final Tracer tracer = Tracing.getTracer();
+  private static final String SPAN_NAME = "MySpanName";
+  private static final String ANNOTATION_DESCRIPTION = "MyAnnotation";
+  private static final String ATTRIBUTE_KEY = "MyAttributeKey";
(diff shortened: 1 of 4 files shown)


adriancole:
I like that you measure in ns :P

benchmarks/src/jmh/java/com/google/instrumentation/trace/RecordTraceEventsNonSampledSpanBenchmark.java
```diff
@@ -0,0 +1,88 @@
...
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.TearDown;
+
+/** Benchmarks for {@link SpanImpl} to record trace events. */
+@State(Scope.Benchmark)
+public class RecordTraceEventsNonSampledSpanBenchmark {
+  private static final Tracer tracer = Tracing.getTracer();
```
sebright: I think it would be better to make the Tracer non-static, especially if it has any state.

bogdandrutu: Does it matter if it has static or non-static if it has any state? Tracing ensures that we have only one instance of the Tracer and we actually encourage user to use it this way because we don't want to pass it around.

If you feel strongly I can change it.

benchmarks/src/jmh/java/com/google/instrumentation/trace/RecordTraceEventsNonSampledSpanBenchmark.java
```diff
@@ -0,0 +1,88 @@
...
+  private static final String SPAN_NAME = "MySpanName";
+  private static final String ANNOTATION_DESCRIPTION = "MyAnnotation";
+  private static final String ATTRIBUTE_KEY = "MyAttributeKey";
+  private static final String ATTRIBUTE_VALUE = "MyAttributeValue";
+  private Span linkedSpan =
+      tracer.spanBuilder(SPAN_NAME).becomeRoot().setSampler(Samplers.alwaysSample()).startSpan();
```
bogdandrutu: Done. Thanks.

benchmarks/src/jmh/java/com/google/instrumentation/trace/RecordTraceEventsNonSampledSpanBenchmark.java
```diff
@@ -0,0 +1,88 @@
...
+    linkedSpan.end();
+  }
+
+  /** This benchmark attempts to measure performance of {@link Span#addAttributes(Map)}. */
+  @Benchmark
+  @BenchmarkMode(Mode.SampleTime)
+  @OutputTimeUnit(TimeUnit.NANOSECONDS)
+  public Span addAttributes() {
+    HashMap<String, AttributeValue> attributes = new HashMap<String, AttributeValue>();
+    attributes.put(ATTRIBUTE_KEY, AttributeValue.stringAttributeValue(ATTRIBUTE_VALUE));
```
bogdandrutu: I thought about this, but I think is better to measure the entire "code" that users will usually have to write in order to add an event.

benchmarks/src/jmh/java/com/google/instrumentation/trace/RecordTraceEventsNonSampledSpanBenchmark.java
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  /**
+   * This benchmark attempts to measure performance of {@link Span#addNetworkEvent(NetworkEvent)}.
+   */
+  @Benchmark
+  @BenchmarkMode(Mode.SampleTime)
+  @OutputTimeUnit(TimeUnit.NANOSECONDS)
+  public Span addNetworkEvent() {
+    span.addNetworkEvent(NetworkEvent.builder(NetworkEvent.Type.RECV, 1).setMessageSize(3).build());
```
bogdandrutu: See my previous comment.

benchmarks/src/jmh/java/com/google/instrumentation/trace/RecordTraceEventsNonSampledSpanBenchmark.java
```diff
@@ -0,0 +1,88 @@
...
+    return span;
+  }
+
+  /**
+   * This benchmark attempts to measure performance of {@link Span#addNetworkEvent(NetworkEvent)}.
```
bogdandrutu: Thanks. Done.

benchmarks/src/jmh/java/com/google/instrumentation/trace/RecordTraceEventsSampledSpanBenchmark.java
```diff
@@ -0,0 +1,88 @@
...
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.TearDown;
+
+/** Benchmarks for {@link SpanImpl} to record trace events. */
+@State(Scope.Benchmark)
+public class RecordTraceEventsSampledSpanBenchmark {
```
bogdandrutu: Making a helper class can unfortunately change the "generated" code that we are trying to measure. It is better to make calls directly to avoid a function call (which I agree is few cycles and the noise of these benchmarks is higher) but I feel that benchmarking directly the calls that we expect users to do is clearer.

If you feel strongly we can chat offline.

benchmarks/src/jmh/java/com/google/instrumentation/trace/RecordTraceEventsNonSampledSpanBenchmark.java
sebright: I just asked because it seemed inconsistent with the Javadoc.  Maybe the Javadoc should say something like "This benchmark attempts to measure performance of adding an attribute to a span."

benchmarks/src/jmh/java/com/google/instrumentation/trace/RecordTraceEventsSampledSpanBenchmark.java
sebright: I assumed that the refactoring wouldn't make much difference, because the benchmark methods and fields would all be in the superclass, and the methods could be final.  But using two separate classes will ensure that there is no problem.  My comment was only a suggestion, and I think the code is fine the way it is.

benchmarks/src/jmh/java/com/google/instrumentation/trace/RecordTraceEventsNonSampledSpanBenchmark.java
sebright: The reason that I suggested making it non-static is that it makes it clearer to the reader that the benchmarks don't share any state, whether or not the class actually has state.  I think the code is fine as is, though.

benchmarks/src/jmh/java/com/google/instrumentation/trace/RecordTraceEventsNonSampledSpanBenchmark.java
sebright: This also seemed like an inconsistency between the code and Javadoc.

benchmarks/src/jmh/java/com/google/instrumentation/trace/RecordTraceEventsNonSampledSpanBenchmark.java
bogdandrutu: Changed. Thanks.

benchmarks/src/jmh/java/com/google/instrumentation/trace/RecordTraceEventsNonSampledSpanBenchmark.java
bogdandrutu: Changed. Done.

PR #299, opened 2017-05-12
Implement minimal functionality for StatsContext to interact with gRPC Context.

songy23:
This class is similar to [trace/ContextUtils.java](https://github.com/google/instrumentation-java/blob/master/core/src/main/java/com/google/instrumentation/trace/ContextUtils.java), to provide minimal util methods for StatsContext to interact with gRPC Context.

I made all the util methods package-private, and I plan to expose them via new APIs in StatsContextFactory.

/cc @bogdandrutu @sebright

### DIFF
--- a/core/src/main/java/com/google/instrumentation/stats/ContextUtils.java
+++ b/core/src/main/java/com/google/instrumentation/stats/ContextUtils.java
@@ -1,0 +1,93 @@
+/*
+ * Copyright 2017, Google Inc.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.instrumentation.stats;
+
+import com.google.instrumentation.common.NonThrowingCloseable;
+import io.grpc.Context;
+
+/**
+ * Util methods/functionality to interact with the {@link io.grpc.Context}.
+ *
+ * <p>Users must interact with the current Context via the public APIs in {@link
+ * StatsContextFactory} and avoid usages of the {@link #STATS_CONTEXT_KEY} directly.
+ */
+public final class ContextUtils {
+
+  /**
+   * The {@link io.grpc.Context.Key} used to interact with {@link io.grpc.Context}.
+   */
+  public static final Context.Key<StatsContext> STATS_CONTEXT_KEY = Context.key(
+      "instrumentation-stats-key");
+
+  // Static class.
+  private ContextUtils() {
+  }
+
+  /**
+   * Returns The {@link StatsContext} from the current context.
+   *
+   * @return The {@code StatsContext} from the current context.
+   */
+  static StatsContext getCurrentStatsContext() {
+    return STATS_CONTEXT_KEY.get(Context.current());
+  }
+
+  /**
+   * Construct a new {@link Context} where the given {@link StatsContext} is set, with current
+   * context as its parent.
+   *
+   * @param statsContext The {@code StatsContext} to be set to the current context.
+   * @return a new context where the given {@code StatsContext}, and its parent is current context.
+   */
+  static Context withStatsContext(StatsContext statsContext) {
+    return Context.current().withValue(STATS_CONTEXT_KEY, statsContext);
+  }
+
+  /**
+   * Enters the scope of code where the given {@link StatsContext} is in the current context, and
+   * returns an object that represents that scope. The scope is exited when the returned object
+   * is closed.
+   *
+   * <p>Supports try-with-resource idiom.
+   *
+   * @param statsContext The {@code StatsContext} to be set to the current context.
+   * @return An object that defines a scope where the given {@code StatsContext} is set to the
+   *     current context.
+   */
+  static NonThrowingCloseable withScopedStatsContext(StatsContext statsContext) {
+    return new WithStatsContext(statsContext, STATS_CONTEXT_KEY);
+  }
+
+  // Supports try-with-resources idiom.
+  private static final class WithStatsContext implements NonThrowingCloseable {
+
+    private final Context origContext;
+
+    /**
+     * Constructs a new {@link WithStatsContext}.
+     *
+     * @param statsContext is the {@code StatsContext} to be added to the current {@code Context}.
+     * @param contextKey is the {@code Context.Key} used to set/get {@code StatsContext} from the
+     * {@code Context}.
+     */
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/google/instrumentation/stats/ContextUtils.java
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  /**
+   * Construct a new {@link Context} where the given {@link StatsContext} is set, with current
+   * context as its parent.
+   *
+   * @param statsContext The {@code StatsContext} to be set to the current context.
+   * @return a new context where the given {@code StatsContext}, and its parent is current context.
+   */
+  static Context withStatsContext(StatsContext statsContext) {
```
bogdandrutu: Where do you need this?

songy23: I'm considering exposing this as a public API in `StatsContextFactory`, because compared to `withScopedStatsContext()` (which internally calls `Context.attach()` and `detach()`), this method gives user more flexibility on using the created `Context`.

bogdandrutu: This flexibility is error prone. I would suggest we discuss more and we see a real need before we add it. Please don't do this initially, advance users can always use the key to get the same functionality, but we should not encourage the normal usage to be this.

songy23: Makes sense. At first I added this method because this might be useful to [some of the gRPC logic](https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/internal/CensusStatsModule.java#L325), but I'm fine with excluding it initially. I'll go ahead and remove this.

core/src/main/java/com/google/instrumentation/stats/ContextUtils.java
```diff
@@ -0,0 +1,82 @@
...
+import com.google.instrumentation.common.NonThrowingCloseable;
+import io.grpc.Context;
+
+/**
+ * Util methods/functionality to interact with the {@link io.grpc.Context}.
+ *
+ * <p>Users must interact with the current Context via the public APIs in {@link
+ * StatsContextFactory} and avoid usages of the {@link #STATS_CONTEXT_KEY} directly.
+ */
+public final class ContextUtils {
```
sebright: Since this class should only be used directly by projects like gRPC, should it have a name that discourages its use, such as UnsafeContextUtils?  Another option is to put STATS_CONTEXT_KEY in its own class with a scary name and make this class package-private.

core/src/test/java/com/google/instrumentation/stats/ContextUtilsTest.java
```diff
@@ -0,0 +1,78 @@
...
+    try {
+      assertThat(ContextUtils.getCurrentStatsContext()).isSameAs(statsContext);
+    } finally {
+      scopedStatsCtx.close();
+    }
+    assertThat(ContextUtils.getCurrentStatsContext()).isNull();
+  }
+
+  @Test
+  public void testWithStatsContextUsingWrap() {
```
sebright: It seems like this test is only testing Context.wrap(...), so it might not be necessary.

songy23: I think this test is to verify that we actually have the StatsContext set in current context even after we closed the scopedStatsCtx, so I'd suggest to leave it here.

core/src/main/java/com/google/instrumentation/stats/ContextUtils.java
songy23: I'd prefer to put STATS_CONTEXT_KEY in another class since all the other methods in ContextUtils are package-private, but this might be inconsistent with [trace](https://github.com/google/instrumentation-java/blob/master/core/src/main/java/com/google/instrumentation/trace/ContextUtils.java#L27). Or maybe mark STATS_CONTEXT_KEY as Deprecated?

core/src/test/java/com/google/instrumentation/stats/ContextUtilsTest.java
sebright: Okay

core/src/main/java/com/google/instrumentation/stats/ContextUtils.java
sebright: > I'd prefer to put STATS_CONTEXT_KEY in another class since all the other methods in ContextUtils are package-private, but this might be inconsistent with trace.

Yeah, it would be nice to keep this part of the API consistent, though I think it is less important than keeping the main API consistent.
/cc @bogdandrutu 

> Or maybe mark STATS_CONTEXT_KEY as Deprecated?

I would rather change the name than deprecate the field, because an element can be deprecated for many reasons, but a name can give the reason that the element's use is discouraged.  Also, if someone has a good reason to use the field, I think it would be better if they didn't need to suppress deprecation warnings.

songy23: Yeah using Deprecated may not be a good idea, I'll go ahead and rename this class to GrpcOnlyContextUtils unless others have a better idea on preventing other users using this class.

sebright: I'm not sure I understand gRPC's use case.  Is there any reason that other projects, such as other RPC frameworks, would need to use the key?

songy23: It's in gRPC's [CensusStatsModule](https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/internal/CensusStatsModule.java#L86), I think they're using this key for context propagation on StatsContext.

zhangkun83: Anyone who wants to change the tags of the current StatsContext would need directly or indirectly access to this key.

bogdandrutu: I would suggest to keep it as ContextUtil because:
1) This is how we call it in the trace package as well.
2) Having "grpc" in the name of a class makes users think this is a gRPC specific library.
3) Advance users may need access to this key.

songy23: I see, for now I'll keep stats/ContextUtil consistent with trace/ContextUtil. I'll add a TODO here and see if we would like to change it later.

core/src/main/java/com/google/instrumentation/stats/ContextUtils.java
```diff
@@ -0,0 +1,83 @@
...
+    private final Context origContext;
+
+    /**
+     * Constructs a new {@link WithStatsContext}.
+     *
+     * @param statsContext is the {@code StatsContext} to be added to the current {@code Context}.
+     * @param contextKey is the {@code Context.Key} used to set/get {@code StatsContext} from the
+     * {@code Context}.
+     */
+    WithStatsContext(StatsContext statsContext, Context.Key<StatsContext> contextKey) {
```
bogdandrutu: why is this package protected? Can be private.

songy23: Sure, the class `WithStatsContext` is private, I'll update the constructor as private as well.

PR #303, opened 2017-05-13
Add implementation for Span#addAttributes.

bogdandrutu:


core_impl/src/main/java/com/google/instrumentation/trace/SpanImpl.java
```diff
@@ -303,20 +330,45 @@ public void end(EndSpanOptions options) {
     void onEnd(SpanImpl span);
   }
 
+  private static final class AttributesWithCapacity extends LinkedHashMap<String, AttributeValue> {
+    private int capacity;
+    private int totalRecordedAttributes = 0;
+    private static final long serialVersionUID = 42L;
```
sebright: This field is a little misleading if the class is never serialized.  I think it would help to have a comment.

core_impl/src/test/java/com/google/instrumentation/trace/SpanImplTest.java
```diff
@@ -98,7 +101,9 @@ public void noEventsRecordedAfterEnd() {
             timestampConverter,
             testClock);
     span.end();
-    // Check that adding trace events after Span#end() does not throw any exception.
+    // Check that adding trace events after Span#end() does not throw any exception and are not
+    // recorded.
+    span.addAttributes(attributes);
```
sebright: Should this test check that the attributes are empty now?

core_impl/src/test/java/com/google/instrumentation/trace/SpanImplTest.java
```diff
@@ -221,6 +232,55 @@ public void toSpanData_EndedSpan() {
...
+            parentSpanId,
+            false,
+            traceParams,
+            startEndHandler,
+            timestampConverter,
+            testClock);
+    AttributeValue attributeValue = AttributeValue.stringAttributeValue("MyStringAttributeValue");
+    for (int i = 0; i < 2 * maxNumberOfAttributes; i++) {
+      Map<String, AttributeValue> attributes = new HashMap<String, AttributeValue>();
+      attributes.put("MyStringAttributeKey" + i, attributeValue);
```
sebright: Suggestion: It might help to also update attributeValue with each iteration.

core_impl/src/main/java/com/google/instrumentation/trace/SpanImpl.java
```diff
@@ -54,13 +55,17 @@
   // The start time of the span. Set when the span is created iff the RECORD_EVENTS options is
   // set, otherwise 0.
   private final long startNanoTime;
-  // List of the recorded annotations.
+  // Set of recorded attributes. Eviction is based on the insertion order (note that insertion
+  // order is not affected if a key is re-inserted into the map.)
```
sebright: Why not use access order?  That seems more consistent with the behavior of the fields using TraceEvents.

core_impl/src/main/java/com/google/instrumentation/trace/SpanImpl.java
```diff
@@ -303,20 +330,45 @@ public void end(EndSpanOptions options) {
     void onEnd(SpanImpl span);
   }
 
+  private static final class AttributesWithCapacity extends LinkedHashMap<String, AttributeValue> {
+    private int capacity;
```
sebright: It looks like this field could be final.

core_impl/src/main/java/com/google/instrumentation/trace/SpanImpl.java
```diff
@@ -303,20 +330,45 @@ public void end(EndSpanOptions options) {
...
+      totalRecordedAttributes += attributes.size();
+      putAll(attributes);
+    }
+
+    private int getNumberOfDroppedAttributes() {
+      return totalRecordedAttributes - size();
+    }
+
+    @Override
+    protected boolean removeEldestEntry(Map.Entry<String, AttributeValue> eldest) {
```
sebright: It would also help to have a comment explaining this class, i.e., the reason it extends LinkedHashMap and overrides removeEldestEntry.

core_impl/src/test/java/com/google/instrumentation/trace/SpanImplTest.java
```diff
@@ -221,6 +232,55 @@ public void toSpanData_EndedSpan() {
   }
 
   @Test
+  public void droppingAttributes() {
```
sebright: It would be nice to also test the effect of reinsertion on the dropped attributes.

core_impl/src/main/java/com/google/instrumentation/trace/SpanImpl.java

core_impl/src/main/java/com/google/instrumentation/trace/SpanImpl.java
```diff
@@ -137,21 +142,26 @@ SpanData toSpanData() {
         "Getting SpanData for a Span without RECORD_EVENTS option.");
     synchronized (this) {
       // TODO(bdrutu): Set the attributes in the SpanData when add the support for them.
```
bogdandrutu: Yes, it was.

core_impl/src/main/java/com/google/instrumentation/trace/SpanImpl.java
bogdandrutu: How do you define access order? re-inserted into the map should change the insertion order?

core_impl/src/test/java/com/google/instrumentation/trace/SpanImplTest.java
bogdandrutu: Postpone until we decide on the ordering.

core_impl/src/main/java/com/google/instrumentation/trace/SpanImpl.java
sebright: Yes.  Access order vs insertion order is also a parameter to one of the LinkedHashMap constructors: https://docs.oracle.com/javase/7/docs/api/java/util/LinkedHashMap.html#LinkedHashMap(int,%20float,%20boolean)

core_impl/src/main/java/com/google/instrumentation/trace/SpanImpl.java
sebright: I think that extending a collection or other serializable class isn't very common, so people might expect that a class with a serialVersionUID is supposed to be serialized.  A comment saying that the class shouldn't be serialzed would also help explain why the class has a non-serializable element type.

core_impl/src/main/java/com/google/instrumentation/trace/SpanImpl.java
bogdandrutu: I know but that parameter also change the access order based when we do a get, which should not be the case, but fortunately the only getters are iteration over the whole map which should not change the access order.

core_impl/src/main/java/com/google/instrumentation/trace/SpanImpl.java

bogdandrutu:
PTAL

core_impl/src/test/java/com/google/instrumentation/trace/SpanImplTest.java
bogdandrutu: Done.

core_impl/src/main/java/com/google/instrumentation/trace/SpanImpl.java
```diff
@@ -329,15 +329,20 @@ private AttributesWithCapacity getInitializedAttributes() {
...
     private int totalRecordedAttributes = 0;
+    // Here because -Werror complains about this: [serial] serializable class AttributesWithCapacity
+    // has no definition of serialVersionUID. This class shouldn't be serialized.
     private static final long serialVersionUID = 42L;
 
     private AttributesWithCapacity(int capacity) {
-      super(capacity);
+      // Capacity of the map is capacity + 1 to avoid resizing because removeEldestEntry is invoked
+      // by put and putAll after inserting a new entry into the map.
+      // loadFactor = 1 to avoid resizing because.
```

bogdandrutu:
Done. Merging soon.

PR #304, opened 2017-05-13
Support arbitrary distribution views.

sebright:
#### Support arbitrary distribution views.
This commit also updates the unit tests:
- Adds tests that weren't possible when only one view was supported
- Updates the existing tests so that they don't depend on specific RPC view
  constants
- Refactors the existing getView tests to use StatsTestUtil

_________________________________________

/cc @dinooliva @bogdandrutu @a-veitch


core_impl/src/test/java/com/google/instrumentation/stats/StatsManagerImplTest.java
```diff
@@ -14,243 +14,457 @@
...
+            DISTRIBUTION_AGGREGATION_DESCRIPTOR,
+            Arrays.asList(KEY));
     clock.setTime(Timestamp.create(1, 2));
-    statsManager.registerView(RpcViewConstants.RPC_CLIENT_ROUNDTRIP_LATENCY_VIEW);
-    for (double val : Arrays.<Double>asList(10.0, 20.0, 30.0, 40.0)) {
+    statsManager.registerView(viewDescr);
+    for (double val : Arrays.asList(10.0, 20.0, 30.0, 40.0)) {
       statsManager.record(
-          oneTag, MeasurementMap.of(RpcMeasurementConstants.RPC_CLIENT_ROUNDTRIP_LATENCY, val));
+          createContext(factory, KEY, VALUE), MeasurementMap.of(MEASUREMENT_DESCRIPTOR, val));
```
songy23: Minor: create context once and store it in a local variable or constant.

core_impl/src/test/java/com/google/instrumentation/stats/StatsManagerImplTest.java
```diff
@@ -14,243 +14,457 @@
...
+        DistributionViewDescriptor.create(
+            VIEW_NAME,
+            VIEW_DESCRIPTION,
+            MEASUREMENT_DESCRIPTOR,
+            DISTRIBUTION_AGGREGATION_DESCRIPTOR,
+            Arrays.asList(KEY));
+    clock.setTime(Timestamp.create(10, 0));
+    statsManager.registerView(viewDescr);
+    statsManager.record(
+        createContext(factory, KEY, VALUE), MeasurementMap.of(MEASUREMENT_DESCRIPTOR, 0.1));
```
songy23: Same

core_impl/src/test/java/com/google/instrumentation/stats/StatsManagerImplTest.java
```diff
@@ -14,243 +14,457 @@
...
-      DistributionAggregation distributionAggregation,
-      int count, double sum, double mean, double min, double max, int tagsSize) {
-    assertThat(distributionAggregation.getCount()).isEqualTo(count);
-    assertThat(distributionAggregation.getSum()).isWithin(TOLERANCE).of(sum);
-    assertThat(distributionAggregation.getMean()).isWithin(TOLERANCE).of(mean);
-    assertThat(distributionAggregation.getRange().getMin()).isWithin(TOLERANCE).of(min);
-    assertThat(distributionAggregation.getRange().getMax()).isWithin(TOLERANCE).of(max);
-    assertThat(distributionAggregation.getTags().size()).isEqualTo(tagsSize);
+  @Test
+  public void allowRecordingWithoutRegisteringMatchingView() {
```
songy23: I think a unit test needs to either assert on something or expect an exception. Could we verify something after recording?

sebright: I left this test, even though it no longer throws an exception, because I thought it would be useful to check that calling `record` without registering a view doesn't throw an exception.  There is nothing to assert, because there are no views to check.  I added a comment to explain the purpose of the test.

core_impl/src/test/java/com/google/instrumentation/stats/StatsManagerImplTest.java
```diff
@@ -14,243 +14,457 @@
...
+  private final StatsContextFactoryImpl factory = new StatsContextFactoryImpl(statsManager);
 
   @Test
-  public void testRegisterAndGetView() throws Exception {
-    statsManager.registerView(RpcViewConstants.RPC_CLIENT_ROUNDTRIP_LATENCY_VIEW);
-    View actual = statsManager.getView(RpcViewConstants.RPC_CLIENT_ROUNDTRIP_LATENCY_VIEW);
-    assertThat(actual.getViewDescriptor()).isEqualTo(
-        RpcViewConstants.RPC_CLIENT_ROUNDTRIP_LATENCY_VIEW);
+  public void testRegisterAndGetView() {
+    DistributionViewDescriptor viewDescr =
```
songy23: That totally makes sense, but since code like `DistributionViewDescriptor.create(VIEW_NAME, VIEW_DESCRIPTION, MEASUREMENT_DESCRIPTOR, DISTRIBUTION_AGGREGATION_DESCRIPTOR, Arrays.asList(KEY));` appears many times, I think it would still be better to refactor it. How about refactoring using overload? Something like:
`private static DistributionViewDescriptor create(ViewDescriptor.Name name, String description, ....) { ... }`  // Takes all parameter
`private static DistributionViewDescriptor create(ViewDescriptor.Name name) { ... (call the first one) }` // Takes name as parameter and use default values for other field
`private static DistributionViewDescriptor create() { ... (call the first one) }`  // Takes no parameter and uses all the default values

core_impl/src/test/java/com/google/instrumentation/stats/StatsTestUtil.java
```diff
@@ -0,0 +1,139 @@
...
+   * {@code DistributionAggregation}s has no effect. The expected parameter is last, because it is
+   * likely to be a larger expression.
+   *
+   * @param tolerance the tolerance used for {@code double} comparison.
+   * @param actual the actual test result.
+   * @param expected the expected value.
+   * @throws AssertionError if the {@code DistributionAggregation}s don't match.
+   */
+  static void assertDistributionAggregationsEquivalent(
+      double tolerance,
```
songy23: Minor: I think we could define and use `TOLERANCE` in `StatsTestUtil` so that `assertDistributionAggregationsEquivalent()` has fewer arguments and is easier to use. I don't think it's quite necessary to "customize" the tolerance when calling `assertDistributionAggregationsEquivalent()`, and even it's necessary, we can overload this method (one with the parameter `tolerance` and one without).

core_impl/src/test/java/com/google/instrumentation/stats/StatsManagerImplTest.java
sebright: I added two helper methods for the two main ways that the tests use DistributionViewDescriptors.  Two tests don't use any attributes of the view, and most others use the name, MeasurementDescriptor, AggregationDescriptor, and tag keys.

sebright:
Thanks for the review!

PR #305, opened 2017-05-15
Add getCurrentStatsContext() and withStatsContext() API to StatsContextFactory.

songy23:
Expose the util methods in PR https://github.com/google/instrumentation-java/pull/299 to public APIs in StatsContextFactory. This is similar to [Tracer](https://github.com/google/instrumentation-java/blob/master/core/src/main/java/com/google/instrumentation/trace/Tracer.java#L91), except that currently we don't have no-op objects for Stats (Issue https://github.com/google/instrumentation-java/issues/215), and `getCurrentStatsContext()` may return null (vesus `getCurrentSpan()` returns a no-op). I may need to update `getCurrentStatsContext()` after we came to an agreement on that issue.

/cc @zhangkun83

### DIFF
--- a/core/src/main/java/com/google/instrumentation/stats/StatsContextFactory.java
+++ b/core/src/main/java/com/google/instrumentation/stats/StatsContextFactory.java
@@ -18,7 +18,6 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import com.google.instrumentation.common.NonThrowingCloseable;
 import java.io.IOException;
 import java.io.InputStream;
-import javax.annotation.Nullable;
 
 /**
  * Factory class for {@link StatsContext}.
@@ -44,14 +43,13 @@ public abstract class StatsContextFactory {
   /**
    * Get current StatsContext from current gRPC Context.
    *
-   * @return the current {@code StatsContext} from {@code io.grpc.Context}, or {@code null} if
-   *     there's no {@code StatsContext} associated with current {@code io.grpc.Context}.
+   * @return the current {@code StatsContext} from {@code io.grpc.Context}, or the default
+   * {@code StatsContext} if there's no {@code StatsContext} associated with current
+   * {@code io.grpc.Context}.
    */
-  @Nullable
   public final StatsContext getCurrentStatsContext() {
-    // TODO(songya): return a no-op StatsContext (or default) when getCurrentStatsContext() returns
-    // null?
-    return ContextUtils.getCurrentStatsContext();
+    StatsContext statsContext = ContextUtils.getCurrentStatsContext();
+    return statsContext != null ? statsContext : getDefault();
   }
 
   /**
@@ -64,12 +62,11 @@ public abstract class StatsContextFactory {
    * <p>Example of usage:
    *
    * <pre>{@code
-   * // initialized by constructor or by calling Stats.getStatsContextFactory()
-   * private final StatsContextFactory statsCtxFactory;
-   * void doWork {
+   * private final StatsContextFactory statsCtxFactory = Stats.getStatsContextFactory();
+   * void doWork() {
    *   checkNotNull(statsCtxFactory, "statsCtxFactory");
    *   // Construct a new StatsContext with required tags to be set into current context.
-   *   StatsContext statsCtx = statsCtxFactory.getDefault().with(tagKey, tagValue);
+   *   StatsContext statsCtx = statsCtxFactory.getCurrentStatsContext().with(tagKey, tagValue);
    *   try (NonThrowingCloseable scopedStatsCtx = statsCtxFactory.withStatsContext(statsCtx)) {
    *     doSomeOtherWork();  // Here "scopedStatsCtx" is the current StatsContext.
    *   }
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/google/instrumentation/stats/StatsContextFactory.java
```diff
@@ -35,4 +40,33 @@
...
+   * closed.
+   *
+   * <p>Supports try-with-resource idiom.
+   *
+   * @param statsContext The {@link StatsContext} to be set to the current Context.
+   * @return an object that defines a scope where the given {@link StatsContext} will be set to the
+   *     current Context.
+   * @throws NullPointerException if statsContext is null.
+   */
+  public final NonThrowingCloseable withStatsContext(StatsContext statsContext) {
```
bogdandrutu: It will be good to add an example of usage with java6 and java7.

core/src/main/java/com/google/instrumentation/stats/StatsContextFactory.java
```diff
@@ -35,4 +40,33 @@
...
+    return ContextUtils.withStatsContext(checkNotNull(statsContext, "StatsContext"));
```
bogdandrutu: For consistency with trace we should have the error message exactly the same as the variable name. In this case "statsContext"

core/src/main/java/com/google/instrumentation/stats/StatsContextFactory.java
songy23: Sure, documented.

core/src/main/java/com/google/instrumentation/stats/StatsContextFactory.java
songy23: Updated.

core/src/main/java/com/google/instrumentation/stats/StatsContextFactory.java
```diff
@@ -35,4 +40,69 @@
...
+
+  /**
+   * Get current StatsContext from current gRPC Context.
+   *
+   * @return the current {@code StatsContext} from {@code io.grpc.Context}, or {@code null} if
+   *     there's no {@code StatsContext} associated with current {@code io.grpc.Context}.
+   */
+  @Nullable
+  public final StatsContext getCurrentStatsContext() {
+    // TODO(songya): return a no-op StatsContext (or default) when getCurrentStatsContext() returns
```
bogdandrutu: I think we should do something here because returning null will make a lot of users to write:

StatsContext current = ...
if (current != null) {
  ...
}

So probably we should return the default. @dinooliva do you have any preference here?

core/src/main/java/com/google/instrumentation/stats/StatsContextFactory.java
```diff
@@ -35,4 +40,69 @@
...
+   * closed.
+   *
+   * <p>Supports try-with-resource idiom.
+   *
+   * <p>Example of usage:
+   *
+   * <pre>{@code
+   * // initialized by constructor or by calling Stats.getStatsContextFactory()
+   * private final StatsContextFactory statsCtxFactory;
+   * void doWork {
```
sebright: This should be "doWork()".

core/src/main/java/com/google/instrumentation/stats/StatsContextFactory.java
```diff
@@ -35,4 +40,69 @@
...
+   * returns an object that represents that scope. The scope is exited when the returned object is
+   * closed.
+   *
+   * <p>Supports try-with-resource idiom.
+   *
+   * <p>Example of usage:
+   *
+   * <pre>{@code
+   * // initialized by constructor or by calling Stats.getStatsContextFactory()
+   * private final StatsContextFactory statsCtxFactory;
```
sebright: Since this is just one example of how the API could be used, I think it would be better to include the initialization in the code, i.e., `private final StatsContextFactory statsCtxFactory = Stats.getStatsContextFactory()`.  Then you could remove the null check farther down.

core/src/main/java/com/google/instrumentation/stats/StatsContextFactory.java
```diff
@@ -35,4 +40,69 @@
...
+   *
+   * <p>Prior to Java SE 7, you can use a finally block to ensure that a resource is closed
+   * regardless of whether the try statement completes normally or abruptly.
+   *
+   * <p>Example of usage prior to Java SE7:
+   *
+   * <pre>{@code
+   * // initialized by constructor or by calling Stats.getStatsContextFactory()
+   * private final StatsContextFactory statsCtxFactory;
+   * void doWork {
```
sebright: The last two comments apply here as well.

core/src/main/java/com/google/instrumentation/stats/StatsContextFactory.java
songy23: Yeah initially I implemented this method to return default `StatsContext` when `ContextUtils.getCurrentStatsContext()` returns null, because I think if this method returned null, user need to call `getDefault()` again to get a non-null `StatsContext`. I'll update this.

core/src/main/java/com/google/instrumentation/stats/StatsContextFactory.java
songy23: Done.

core/src/main/java/com/google/instrumentation/stats/StatsContextFactory.java
songy23: Fixed.

core/src/main/java/com/google/instrumentation/stats/StatsContextFactory.java
songy23: Updated, though we still need to check if `statsCtxFactory` is null or not, since [Stats.getStatsContextFactory()](https://github.com/google/instrumentation-java/blob/master/core/src/main/java/com/google/instrumentation/stats/Stats.java#L29) is Nullable.

PR #306, opened 2017-05-15
Add probability-based sampler to API

a-veitch:


core/src/main/java/com/google/instrumentation/trace/ProbabilitySampler.java
```diff
@@ -0,0 +1,92 @@
...
+    // and dynamic increases to sampling probabilities for debugging purposes.
+    // Note use of '<' for comparison. This ensures that we never sample for probability == 0.0,
+    // while allowing for a (very) small chance of *not* sampling if the id == Long.MAX_VALUE.
+    // This is considered a reasonable tradeoff for the simplicity/performance requirements (this
+    // code is executed in-line for every Span creation).
+    if (Math.abs(traceIdNum.longValue()) < idUpperBound) {
+      return true;
+    }
+    // Also enable sampling if parent was sampled.
+    return (parentContext != null && parentContext.getTraceOptions().isSampled());
```
bogdandrutu: We can check this first because this will always return true if the parent is sampled and we save cycles.

core/src/main/java/com/google/instrumentation/trace/ProbabilitySampler.java
```diff
@@ -0,0 +1,92 @@
...
+  private final double probability;
+
+  /**
+   * Create a new {@link ProbabilitySampler}. The probability of sampling a trace is equal to that
+   * of the specified probability.
+   *
+   * @param probability The desired probability of sampling. Must be within [0.0, 1.0].
+   * @throws IllegalArgumentException if {@code probability} is out of range
+   */
+  public ProbabilitySampler(double probability) {
```
bogdandrutu: Please add a factory method in the Samplers:

public static Sampler getProbabilitySampler(double probability) {
}

And make the ctor and the class package protected. Consider also to move this into Sampler as a private class.

a-veitch: Done

core/src/main/java/com/google/instrumentation/trace/ProbabilitySampler.java
a-veitch: Done

a-veitch: PTAL

core/src/main/java/com/google/instrumentation/trace/Samplers.java
```diff
@@ -83,4 +97,70 @@ public String toString() {
...
+    /**
+     * Create a new {@link ProbabilitySampler}. The probability of sampling a trace is equal to that
+     * of the specified probability.
+     *
+     * @param probability The desired probability of sampling. Must be within [0.0, 1.0].
+     * @throws IllegalArgumentException if {@code probability} is out of range
+     */
+    private ProbabilitySampler(double probability) {
+      this.probability = probability;
+      if (probability < 0.0 || probability > 1.0) {
```
bogdandrutu: Use checkArgument
http://google.github.io/guava/releases/snapshot/api/docs/com/google/common/base/Preconditions.html#checkArgument(boolean)

core/src/main/java/com/google/instrumentation/trace/Samplers.java
```diff
@@ -83,4 +97,70 @@ public String toString() {
...
+        String name,
+        @Nullable List<Span> parentLinks) {
+      checkNotNull(traceId, "traceId");
+      checkNotNull(name, "name");
+      // Always enable sampling if parent was sampled.
+      if (parentContext != null && parentContext.getTraceOptions().isSampled()) {
+        return true;
+      }
+      // TODO(bdrutu): Fix this to use TraceId#compareTo.
+      BigInteger traceIdNum = new BigInteger(traceId.getBytes());
```
bogdandrutu: I think the TODO is about avoiding an extra copy of the traceId bytes. In order to do that you may add a package protected method in the TraceId.getLowerLong() { return new BigInteger(bytes).longValue(); }. This will avoid one extra copy of the bytes in getBytes().

If you do this we can remove the TODO.

core/src/main/java/com/google/instrumentation/trace/Samplers.java
a-veitch: Done

core/src/main/java/com/google/instrumentation/trace/Samplers.java
a-veitch: Done. Also avoided the path through BigInteger.

core/src/main/java/com/google/instrumentation/trace/TraceId.java
```diff
@@ -133,6 +133,19 @@ public boolean isValid() {
     return !Arrays.equals(bytes, INVALID.bytes);
   }
 
+  // Return the lower 8 bytes of the trace-id as a long value. This is used in ProbabilitySampler.
+  protected long getLowerLong() {
```
bogdandrutu: Why protected? Just no keyword.

core/src/main/java/com/google/instrumentation/trace/TraceId.java
```diff
@@ -133,6 +133,19 @@ public boolean isValid() {
     return !Arrays.equals(bytes, INVALID.bytes);
   }
 
+  // Return the lower 8 bytes of the trace-id as a long value. This is used in ProbabilitySampler.
```
bogdandrutu: Please mention that this is little endian order.

core/src/test/java/com/google/instrumentation/trace/SamplersTest.java
```diff
@@ -94,4 +94,75 @@ public void neverSampleSampler_AlwaysReturnFalse() {
...
+  public void probabilitySampler_outOfRangeHighProbability() {
+    Samplers.probabilitySampler(1.01);
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void probabilitySampler_outOfRangeLowProbability() {
+    Samplers.probabilitySampler(-0.00001);
+  }
+
+  private void probabilitySampler_AlwaysReturnTrueForSampled(Sampler sampler) {
```
bogdandrutu: This can be final. Same for the next one.

bogdandrutu:
Please use rebase to pickup all the changes from the master not merge.

core/src/main/java/com/google/instrumentation/trace/TraceId.java
a-veitch: Done, though don't think it matters much.

core/src/main/java/com/google/instrumentation/trace/TraceId.java
a-veitch: Still learning java... :-(

core/src/test/java/com/google/instrumentation/trace/SamplersTest.java
a-veitch: Done.

a-veitch:
Rebased, merging.

PR #307, opened 2017-05-16
Fix a few style issues.

songy23:


core/src/test/java/com/google/instrumentation/stats/DistributionAggregationTest.java
```diff
@@ -28,16 +28,19 @@
...
 
     assertThat(aggr.getCount()).isEqualTo(10);
-    assertThat(aggr.getMean()).isEqualTo(5.0);
-    assertThat(aggr.getSum()).isEqualTo(30.0);
-    assertThat(aggr.getRange().getMin()).isEqualTo(1.0);
-    assertThat(aggr.getRange().getMax()).isEqualTo(5.0);
+    assertThat(aggr.getMean()).isWithin(TOLERANCE).of(5.0);
+    assertThat(aggr.getSum()).isWithin(TOLERANCE).of(30.0);
+    assertThat(aggr.getRange().getMin()).isWithin(TOLERANCE).of(1.0);
+    assertThat(aggr.getRange().getMax()).isWithin(TOLERANCE).of(5.0);
```
sebright: This test is just checking that the DistributionAggregation's fields are set correctly and isn't doing any double math, so I think the exact comparison is better.

core/src/test/java/com/google/instrumentation/stats/DistributionAggregationTest.java
```diff
@@ -52,10 +55,10 @@ public void testDistributionAggregationWithBuckets() {
...
```
sebright: Same here

PR #308, opened 2017-05-16
Move VarInt to core_impl and change the package name.

bogdandrutu:


core_impl/src/main/java/com/google/instrumentation/common/VarInt.java
```diff
@@ -18,41 +18,42 @@
 import java.io.OutputStream;
 import java.nio.ByteBuffer;
 
-/**
- * Common methods to encode and decode varints and varlongs into ByteBuffers and
- * arrays.
- */
+/** Common methods to encode and decode varints and varlongs into ByteBuffers and arrays. */
 public class VarInt {
```
sebright: I think it would be better to suppress warnings in this file than refactor it, so that it is easier to upgrade and we don't need to add tests.  I was able to suppress CheckStyle warnings with a comment in https://github.com/sebright/instrumentation-java/commit/572fb4e2907540ea73d15e3123df9700fc5c16dc , and it's also possible to exclude lists of files or directories.

core_impl/src/main/java/com/google/instrumentation/common/VarInt.java
```diff
@@ -11,7 +11,7 @@
  * limitations under the License.
  */
 
-package com.google.io.base;
+package com.google.instrumentation.common;
```
bogdandrutu: Done.

core_impl/src/main/java/com/google/instrumentation/common/VarInt.java
bogdandrutu: Done.

bogdandrutu:
PTAL

PR #314, opened 2017-05-17
Disable Checkstyle's MissingSwitchDefault warning.

sebright:
This rule conflicts with Error Prone's exhaustiveness checking.  When all enum
cases are handled in a switch block, Error Prone doesn't allow a "default" case,
but Checkstyle always requires a "default" case.

______________________

I fixed the conflict this way because I thought the Error Prone warning was more useful.  Is that okay?

/cc @dinooliva @a-veitch 

EDIT: I needed this for #276.

### DIFF
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -76,7 +76,8 @@
         <module name="OneStatementPerLine"/>
         <module name="MultipleVariableDeclarations"/>
         <module name="ArrayTypeStyle"/>
-        <module name="MissingSwitchDefault"/>
+        <!-- This rule conflicts with Error Prone's exhaustiveness checking. -->
+	    <!-- <module name="MissingSwitchDefault"/> -->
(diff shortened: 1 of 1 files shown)


checkstyle.xml
```diff
@@ -76,7 +76,8 @@
         <module name="OneStatementPerLine"/>
         <module name="MultipleVariableDeclarations"/>
         <module name="ArrayTypeStyle"/>
-        <module name="MissingSwitchDefault"/>
+        <!-- This rule conflicts with Error Prone's exhaustiveness checking.
+	    <module name="MissingSwitchDefault"/> -->
```
bogdandrutu: Please don't split the comment on 2 lines, instead make it 2 comments.

sebright: Done.  I'll merge when the build is green.

PR #319, opened 2017-05-24
Change the default sampler.

bogdandrutu:
Make the probability sampler with a default probability of 1/10000 the default sampler for instrumentation library.

### DIFF
--- a/core/src/main/java/com/google/instrumentation/trace/Samplers.java
+++ b/core/src/main/java/com/google/instrumentation/trace/Samplers.java
@@ -14,9 +14,8 @@
 package com.google.instrumentation.trace;
 
 import static com.google.common.base.Preconditions.checkArgument;
-import static com.google.common.base.Preconditions.checkNotNull;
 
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/google/instrumentation/trace/Samplers.java
```diff
@@ -98,6 +102,7 @@ public String toString() {
     }
   }
 
+  @Immutable
   private static final class ProbabilitySampler extends Sampler {
```
sebright: It looks like this class could use AutoValue.

bogdandrutu: Gave a try and it looks decent. So I will keep it for the moment, even if I feel that the class is not a value class.

sebright: It's up to you.  I just suggested AutoValue as a way to avoid boilerplate.  I didn't think of it as documenting the purpose of the class.

PR #320, opened 2017-05-24
Resolve all the Distribution TODO's

a-veitch:
Changes the following:
* Allows for zero-length bucketBoundaries (which creates a single bucket, both under and overflow, as documented in the original description).
* Adds a sum-of-squares (for computing variance etc.)
* Simplifies Range by not allowing null for initial values.

### DIFF
--- a/core/src/main/java/com/google/instrumentation/stats/MutableDistribution.java
+++ b/core/src/main/java/com/google/instrumentation/stats/MutableDistribution.java
@@ -24,6 +24,7 @@ import javax.annotation.Nullable;
 final class MutableDistribution {
   private long count = 0; // The number of values in the population.
   private double sum = 0.0; // The sum of the values in the population.
+  private double sumOfSquares = 0.0; // The sum of the sqaures of the values in the population
   private Range range = Range.create(); // Range of values in the population.
 
   @Nullable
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/google/instrumentation/stats/MutableDistribution.java
```diff
@@ -95,6 +96,16 @@ long getCount() {
   }
 
   /**
+   * The sum of the squares of values in the population. If {@link #getCount()} is zero then this
+   * value will also be zero.
+   *
+   * @return The sum of squares of values in the population.
+   */
+  double getSumSquares() {
```
songy23: Nit: `getSumOfSqurares()`?  `getSumSquares()` sounds a little ambiguous since it could also be the square of the sum of all values ((v1 + v2 + v3 ...) ^ 2).

core/src/main/java/com/google/instrumentation/stats/MutableDistribution.java
```diff
@@ -165,18 +177,17 @@ void add(double value) {
 
   /** Mutable version of {@code Distribution.Range}. */
   static final class Range {
-    // Maintain the invariant that max should always be greater than or equal to min.
-    // TODO(songya): needs to determine how we would want to initialize min and max.
-    private Double min = null;
-    private Double max = null;
+    // Initial "impossible" values, that will get reset as soon as first value is added.
+    private double min = Double.POSITIVE_INFINITY;
```
songy23: Minor: add tests on `getMin` and `getMax` for initial values?

core/src/main/java/com/google/instrumentation/stats/MutableDistribution.java
a-veitch: Ack: I've actually just decided to remove this from this PR, since I realized it can't actually be used the way I want. I'll do a separate PR for variance-related stuff.

a-veitch:
I'm also getting an error from a previous merge commit; going to recreate this PR without that.

PR #321, opened 2017-05-24
Fix Distribution TODOs

a-veitch:
Resolves some TODO's in the original Distribution code:
* Allows for zero-length bucketBoundaries (creates single bucket for over and underflow bucket, matching documentation allowing N=1 total buckets)
* Simplifies Range by not using null for initial min/max.

### DIFF
--- a/core/src/main/java/com/google/instrumentation/stats/MutableDistribution.java
+++ b/core/src/main/java/com/google/instrumentation/stats/MutableDistribution.java
@@ -96,16 +96,6 @@ final class MutableDistribution {
   }
 
   /**
-   * The sum of the squares of values in the population. If {@link #getCount()} is zero then this
-   * value will also be zero.
-   *
-   * @return The sum of squares of values in the population.
-   */
-  double getSumSquares() {
-    return sumOfSquares;
-  }
-
-  /**
    * The range of the population values. If {@link #getCount()} is zero then this returned range is
    * implementation-dependent.
    *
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/google/instrumentation/stats/MutableDistribution.java
```diff
@@ -95,6 +96,16 @@ long getCount() {
   }
 
   /**
+   * The sum of the squares of values in the population. If {@link #getCount()} is zero then this
+   * value will also be zero.
+   *
+   * @return The sum of squares of values in the population.
+   */
+  double getSumSquares() {
```
songy23: Remove this?

a-veitch: D'oh. Thanks, and done.

core/src/main/java/com/google/instrumentation/stats/MutableDistribution.java
```diff
@@ -24,6 +24,7 @@
 final class MutableDistribution {
   private long count = 0; // The number of values in the population.
   private double sum = 0.0; // The sum of the values in the population.
+  private double sumOfSquares = 0.0; // The sum of the sqaures of the values in the population
```
songy23: Also remove this? Didn't see we use it anywhere

core/src/main/java/com/google/instrumentation/stats/MutableDistribution.java
```diff
@@ -157,6 +158,7 @@ private void putIntoBucket(double value) {
   void add(double value) {
     count++;
     sum += value;
+    sumOfSquares += value * value;
```
songy23: Same here

core/src/main/java/com/google/instrumentation/stats/MutableDistribution.java
a-veitch: double and triple d'oh. Done

core/src/main/java/com/google/instrumentation/stats/MutableDistribution.java
a-veitch: and done...

PR #323, opened 2017-05-26
Refactor Tag classes.

sebright:
This PR builds on top of #276.  Currently, only the last commit is new.  I made some changes to #276 after I tried modifying the rest of the code to use the new tags package.  I realized that while the type parameter in `TagKey` is convenient for setting values in a `TagMap`, it doesn't work well for specifying or querying views.  `ViewDescriptor`s/`View`s require a list of `TagKey`s or tags, so the type needed to be something like `List<TagKey<?>>`.  I think it is better to avoid using wildcards anywhere in the API, besides method parameters.

I also added a `Tag` class to represent a paired `TagKey` and value.  It can be used to specify tag values in the view classes.  `Tag` and `TagKey` both need to handle three different types of values (long, boolean, and String).  However, I realized that they may need to handle the polymorphism differently, because they are used in different parts of the API.  `TagKey` is used for setting values.  For example, it is used to set tag values in a `TagMap`.  It is also used to specify views.  In constrast, `Tag` is only used for reading values.  It is returned as part of the `View` that is the result of stats collection, and it could also be used as the element type when iterating through a `TagMap`.

I had a hard time coming up with a clean, simple API for `TagKey` and `Tag`.  I couldn't think of a good way to optimize both classes' APIs for both setting and reading values, so I just tried to optimize `TagKey` for setting values and `Tag` for reading values.

`TagKey` exposes three specialized key types as subclasses, because it is useful to keep the static value type information when setting values in a `TagMap`:

``` java
TagKeyString stringKey = TagKey.createString("key 1");
TagKeyLong longKey = TagKey.createLong("key 2");
TagMap tags =
    TagMap.newBuilder()
        .set(stringKey, "value")  // must be a String
        .set(longKey, 10L)        // must be a long
        .build();
```

Then the keys can be upcast to `TagKey` and put into a list for use in a `View`:

``` java
List<TagKey> keys = Arrays.asList(stringKey, longKey);
```

`Tag` also has three specialized classes, but it doesn't expose them as subclasses, because it is only used to return results.  The subclass relationship doesn't help with reading values, unless we encourage users to downcast the values.  Here is an example of reading tags from a `DistributionAggregation` using this API:

``` java
List<Tag> tags = aggregation.getTags();
StringBuilder sb = new StringBuilder();
for (Tag tag : tags) {
  // Do three different things with the three different types of Tag:
  String message = tag.match(
      strTag -> strTag.getValue().asString(),
      longTag -> longTag.getKey().getName(),
      booleanTag -> "boolean");
  sb.append(message);
}
```

Let me know if you have any ideas for simplifying this API.

/cc @bogdandrutu @dinooliva @songy23


sebright:
I rebased onto master.

sebright:
I rebased this on top of the package changes.

sebright:
I added "match" methods to TagKey and made TagString, TagLong, and TagBoolean subclasses of Tag, so that the Tag and TagKey classes are more similar.

core/src/main/java/io/opencensus/tags/Tag.java
```diff
@@ -0,0 +1,202 @@
...
+    return TagBoolean.create(key, value);
+  }
+
+  /** A tag with a {@code String} key and value. */
+  @Immutable
+  @AutoValue
+  public abstract static class TagString extends Tag {
+    TagString() {}
+
+    static TagString create(TagKeyString key, @Nullable TagValueString value) {
```
bogdandrutu: Why is this package protected? If just for tests use @VisibleForTesting.

core/src/main/java/io/opencensus/tags/Tag.java
```diff
@@ -0,0 +1,202 @@
...
+    }
+  }
+
+  /** A tag with a {@code Long} key and value. */
+  @Immutable
+  @AutoValue
+  public abstract static class TagLong extends Tag {
+    TagLong() {}
+
+    static TagLong create(TagKeyLong key, @Nullable Long value) {
```
bogdandrutu: same here.

core/src/main/java/io/opencensus/tags/TagContext.java
```diff
@@ -31,17 +32,17 @@
  */
 @Immutable
 public final class TagContext {
-  /** The maximum length for a string tag value. */
-  public static final int MAX_STRING_LENGTH = StringUtil.MAX_LENGTH;
+
+  private static final TagContext EMPTY = newBuilder().build();
```
bogdandrutu: Personally I like more the pattern where we make this public instead of having an empty() method. Up to you, in trace package I used this way.

core/src/main/java/io/opencensus/tags/TagContext.java
```diff
@@ -135,7 +145,7 @@ public Builder clear(TagKey<?> key) {
      * @return a {@code TagContext} with the same tags as this builder.
      */
     public TagContext build() {
-      return new TagContext(new HashMap<TagKey<?>, Object>(tags));
+      return new TagContext(new HashMap<TagKey, Object>(tags));
```
bogdandrutu: You copy the tags at least one extra time here + in TagContext ctor.

Also another optimization may be to never copy the tags map in the TagContext and make builder be invalid after called build (don't know if that worths the complication).

core/src/main/java/io/opencensus/tags/TagKey.java
```diff
@@ -14,36 +14,22 @@
...
-  }
-
   TagKey() {}
 
   /**
-   * Constructs a {@code TagKey<String>} with the given name.
+   * Constructs a {@code TagKeyString} with the given name.
    *
    * <p>The name must meet the following requirements:
    *
```
bogdandrutu: Why do you have an extra line here? Do you use https://github.com/google/google-java-format, it does a great job formatting comments.

core/src/main/java/io/opencensus/tags/TagKey.java
```diff
@@ -94,12 +80,112 @@
...
+   * Applies a function to the {@code TagKey} subclass. The function that is called depends on the
+   * type of the tag key. This is similar to the visitor pattern.
+   *
+   * @param stringFunction the function to call when the {@code TagKey} is a {@code TagKeyString}.
+   * @param longFunction the function to call when the {@code TagKey} is a {@code TagKeyLong}.
+   * @param booleanFunction the function to call when the {@code TagKey} is a {@code TagKeyBoolean}.
+   * @param <T> The result type of the function.
+   * @return The result of calling the function that matches the tag key's type.
+   */
+  // TODO(sebright): Should we make this public in the first release?
```
bogdandrutu: If you want to go with this pattern (visitor) then you kind of have to do it from the beginning and probably support all types from the beginning.

core/src/main/java/io/opencensus/tags/TagKey.java
```diff
@@ -94,12 +80,112 @@
...
+   * @param defaultFunction the function to call when the tag key has a type other than
+   *     {@code String}, {@code long}, or {@code boolean}.
+   * @param <T> The result type of the function.
+   * @return The result of calling the function that matches the tag key's type.
+   */
+  // TODO(sebright): Should we make this public in the first release?
+  // TODO(sebright): How should we deal with the possibility of adding more tag types in the future?
+  //                 Will this method work? What type of parameter should we use for the default
+  //                 case? Should we remove the non-backwards compatible "match" method?
+  public abstract <T> T matchWithDefault(
```
bogdandrutu: same question here. i don't see how this will be used.

core/src/main/java/io/opencensus/tags/Tag.java
```diff
@@ -0,0 +1,202 @@
...
+  /**
+   * Returns the tag's key.
+   *
+   * @return the tag's key.
+   */
+  public abstract TagKey getKey();
+
+  /**
+   * Applies a function to the tag's key and value. The function that is called depends on the type
+   * of the tag. This is similar to the visitor pattern.
```
sebright: I added examples to the Javadocs for the four `match` methods.

core/src/main/java/io/opencensus/tags/TagKey.java
sebright: Do you mean that we need to support all types from the beginning if we provide the `match` method without a default, or are you also talking about `matchWithDefault`?

core/src/main/java/io/opencensus/tags/TagKey.java
sebright: I added an example that shows how a user could handle different tag types differently, if they had a variable of type `TagKey`.  I'm still not sure why a user would need to call this method, since the user creates the tag keys and could assign them to variables of type `TagKeyString`, `TagKeyLong`, etc., which still have the tag type information.  I added the `TagKey.match` methods for consistency with the `Tag` class.

sebright:
Currently, this PR has two "match" methods for each class, `match` and `matchWithDefault`.  I'm not sure which one we should keep, or if we should keep both of them.  I wanted to describe how each method would deal with a new type of tag, for comparison.  Say version 1 of the API supports the tag types String, boolean, and long, and version 2 supports an additional type, `List<String>`.

### `match`

#### version 1
`Tag.match` has this signature:
```java
  public abstract <T> T match(
      Function<? super TagString, T> stringFunction,
      Function<? super TagLong, T> longFunction,
      Function<? super TagBoolean, T> booleanFunction)
```
The method could be called like this:
```java
    byte[] serializedValue =
        tag.match(
            stringTag -> serializeString(stringTag.getValue().asString()),
            longTag -> serializeLong(longTag.getValue()),
            booleanTag -> serializeBoolean(booleanTag.getValue()));
```

#### version 2
We add a parameter to handle the new list type, which is a breaking change:
```java
  public abstract <T> T match(
      Function<? super TagString, T> stringFunction,
      Function<? super TagLong, T> longFunction,
      Function<? super TagBoolean, T> booleanFunction,
      Function<? super TagList, T> listFunction);
```
The calling code would need to be updated to handle the new type:

```java
    byte[] serializedValue =
        tag.match(
            stringTag -> serializeString(stringTag.getValue().asString()),
            longTag -> serializeLong(longTag.getValue()),
            booleanTag -> serializeBoolean(booleanTag.getValue()),
            listTag -> serializeList(listTag.getValue()));
```

### `matchWithDefault`

#### version 1
`Tag.matchWithDefault` has this signature:
```java
  public abstract <T> T matchWithDefault(
      Function<? super TagString, T> stringFunction,
      Function<? super TagLong, T> longFunction,
      Function<? super TagBoolean, T> booleanFunction,
      Function<? super Tag, T> defaultFunction);
```
It could be called like this.  The last case isn't used in version 1, because all types are handled by the first three cases.
```java
        byte[] serializedValue =
            tag.matchWithDefault(
                stringTag -> serializeString(stringTag.getValue().asString()),
                longTag -> serializeLong(longTag.getValue()),
                booleanTag -> serializeBoolean(booleanTag.getValue()),
                unknownTag -> serializeString(unknownTag.toString()));
```

#### version 2
We keep the old method and add a new overload that handles lists:
```java
  public abstract <T> T matchWithDefault(
      Function<? super TagString, T> stringFunction,
      Function<? super TagLong, T> longFunction,
      Function<? super TagBoolean, T> booleanFunction,
      Function<? super Tag, T> defaultFunction);

  public abstract <T> T matchWithDefault(
      Function<? super TagString, T> stringFunction,
      Function<? super TagLong, T> longFunction,
      Function<? super TagBoolean, T> booleanFunction,
      Function<? super TagList, T> listFunction,
      Function<? super Tag, T> defaultFunction);
```
The calling code continues to compile, and any `List<String>` tags are serialized using `toString()`.  Alternatively, the user could update the code to call the new overload and handle the list type separately:

```java
        byte[] serializedValue =
            tag.matchWithDefault(
                stringTag -> serializeString(stringTag.getValue().asString()),
                longTag -> serializeLong(longTag.getValue()),
                booleanTag -> serializeBoolean(booleanTag.getValue()),
                listTag -> serializeList(listTag.getValue()),
                unknownTag -> serializeString(unknownTag.toString()));
```

sebright:
I updated the `Tag` class to use non-null Strings, longs, and booleans, so it is ready for review.

core/src/main/java/io/opencensus/tags/Tag.java
```diff
@@ -0,0 +1,205 @@
...
+import com.google.auto.value.AutoValue;
+import io.opencensus.common.Function;
+import io.opencensus.tags.TagKey.TagKeyBoolean;
+import io.opencensus.tags.TagKey.TagKeyLong;
+import io.opencensus.tags.TagKey.TagKeyString;
+import javax.annotation.concurrent.Immutable;
+
+/** {@link TagKey} paired with a value. */
+@Immutable
+public abstract class Tag {
```
dinooliva: In my reference code (https://github.com/census-instrumentation/opencensus-java/pull/375/files?diff=split#diff-debe21dd6f7dd73fdfe20d37abb6b6d3), I parameterize the types for the TagKey and the TagValue, which allows us to have abstract getKey() and getValue() methods - it seems to me that both are of somewhat limited value at the base type (Tag) - thoughts?

The other direction would be to remove getKey() from the base type. Are there any advantages to having it here.

core/src/main/java/io/opencensus/tags/Tag.java
```diff
@@ -0,0 +1,202 @@
...
+   * @param booleanFunction the function to call when the tag has a {@code Boolean} value.
+   * @param defaultFunction the function to call when the tag has a value other than {@code String},
+   *     {@code Long}, or {@code Boolean}.
+   * @param <T> The result type of the function.
+   * @return The result of calling the function that matches the tag's type.
+   */
+  // TODO(sebright): How should we deal with the possibility of adding more tag types in the future?
+  //                 Will this method work? What type of parameter should we use for the default
+  //                 case? Should we remove the non-backwards compatible "match" method?
+  public abstract <T> T matchWithDefault(
```
dinooliva: Can you copy the discussion about the matchWithDefault() to the 'Tagged Unions in Java' design doc? I'd like to capture it there too and it would make it easier to discuss.

One of the advantages of just having the match() method (w/o default) is that when extending the type (which shouldn't happen often) you will get compile time errors for any code that you haven't updated to handle the new cases instead of potentially having them handled by the matchWithDefault() method - I think this is actually preferable.

core/src/main/java/io/opencensus/tags/Tag.java
sebright: > In my reference code (https://github.com/census-instrumentation/opencensus-java/pull/375/files?diff=split#diff-debe21dd6f7dd73fdfe20d37abb6b6d3), I parameterize the types for the TagKey and the TagValue, which allows us to have abstract getKey() and getValue() methods - it seems to me that both are of somewhat limited value at the base type (Tag) - thoughts?

I removed the type variable from TagKey because it is hard to put the keys in a collection when they have different types.  That also applies to Tag.  For example, we would need to cast a `Tag<TagKeyString, String>` and a `Tag<TagKeyLong, Long>` to `Tag<TagKey, Object>`s in order to put them in a list in a View.  One way to handle generic types is to have a three-level class hierarchy:

1. abstract class or interface `Tag`
1. abstract class `TypedTag<TagKeyT, TagValueT> extends (or implements) Tag`
1. concrete classes `TagString extends TypedTag<TagKeyString, String>`, `TagLong extends TypedTag<TagKeyLong, Long>`, etc.

Then any tag could go into a `List<Tag>` without being cast.  I'm not sure of the use cases for the middle class, though.  Do you think users will want to perform generic operations on Tags, without calling `match`?

> The other direction would be to remove getKey() from the base type. Are there any advantages to having it here.

The one advantage I can think of is that the user can get the TagKey and then get the name of the key.

core/src/main/java/io/opencensus/tags/Tag.java
sebright: Which discussion?  Do you mean this thread of comments?

core/src/main/java/io/opencensus/tags/TagContext.java
bogdandrutu: Correct. Collections.emptyList() has to do a cast so that is one of the reason to not have it directly as value.

core/src/main/java/io/opencensus/tags/Tag.java
```diff
@@ -0,0 +1,205 @@
...
+   * @return the tag's key.
+   */
+  public abstract TagKey getKey();
+
+  /**
+   * Applies a function to the tag's key and value. The function that is called depends on the type
+   * of the tag. This is similar to the visitor pattern. For example, this code serializes a {@code
+   * Tag}.
+   *
+   * <pre>{@code
```
bogdandrutu: Should you have a pre-java8 example without lambdas?

core/src/main/java/io/opencensus/tags/Tag.java
```diff
@@ -0,0 +1,205 @@
...
+    TagString() {}
+
+    /**
+     * Creates a {@code TagString} from the given {@code String} key and value.
+     *
+     * @param key the tag key.
+     * @param value the tag value.
+     * @return a {@code TagString} with the given key and value.
+     */
+    public static Tag create(TagKeyString key, TagValueString value) {
```
bogdandrutu: If you return Tag here:
1) what is the point of having TagString public?
2) Can we move the factory methods into Tag? createStringTag, etc.?

Same questions for all instances.

core/src/main/java/io/opencensus/tags/TagContext.java
```diff
@@ -31,17 +32,17 @@
...
+
+  private static final TagContext EMPTY = newBuilder().build();
 
   // The types of the TagKey and value must match for each entry.
-  private final Map<TagKey<?>, Object> tags;
+  private final Map<TagKey, Object> tags;
 
-  TagContext(Map<TagKey<?>, Object> tags) {
-    this.tags = Collections.unmodifiableMap(new HashMap<TagKey<?>, Object>(tags));
+  TagContext(Map<? extends TagKey, ?> tags) {
```
bogdandrutu: Why package protected? If for test only use the annotation.

core/src/main/java/io/opencensus/tags/TagContext.java
bogdandrutu: But this is just useless. If you simply `return new TagContext(tags);` then the result is the same. I don't understand why do we need this extra copy now?

core/src/main/java/io/opencensus/tags/TagKey.java
```diff
@@ -14,92 +14,189 @@
...
+     * <ol>
+     *   <li>It cannot be longer than {@link #MAX_LENGTH}.
+     *   <li>It can only contain printable ASCII characters.
+     * </ol>
+     *
+     * @param name the name of the key.
+     * @return a {@code TagKeyString} with the given name.
+     * @throws IllegalArgumentException if the name is not valid.
+     */
+    public static TagKeyString create(String name) {
```
bogdandrutu: I think in this case we want to expose TagKeyString to the users to make them aware of the type.  Should we do then the same in Tag (then return a TagString)?

core/src/main/java/io/opencensus/tags/TagValueString.java
```diff
@@ -0,0 +1,54 @@
...
+import com.google.auto.value.AutoValue;
+import com.google.common.base.Preconditions;
+import io.opencensus.internal.StringUtil;
+import io.opencensus.tags.TagKey.TagKeyString;
+import javax.annotation.concurrent.Immutable;
+
+/** A validated tag value associated with a {@link TagKeyString}. */
+@Immutable
+@AutoValue
+public abstract class TagValueString {
```
bogdandrutu: For consistency should we do the same for TagValue -> TagValueString as we do for Key and Tag? also match will become:

public <T extends TagValue> T match(
    Function<? super TagString, TagStringValue> stringFunction,
    Function<? super TagLong, TagLongValue> longFunction,
    Function<? super TagBoolean, TagBooleanValue> booleanFunction);

core/src/main/java/io/opencensus/tags/Tag.java
dinooliva: Agree - this should return a TagString

core/src/main/java/io/opencensus/tags/TagValueString.java
dinooliva: We really don't need (so far) a TagValue class that is TagValueString | TagValueLong | TagValueBoolean (the Tag class obviates that).

This particular class is meant to be restriction on String values - it may make more sense to rename it something like CensusString for restrictions on String's that are enforced Census-wide (e.g. printable ascii characters only) but we need to agree Census-wide, what such restrictions are.

core/src/main/java/io/opencensus/tags/Tag.java
sebright: I added the same examples without lambdas.

core/src/main/java/io/opencensus/tags/TagKey.java
```diff
@@ -51,6 +51,40 @@
...
+   *             }
+   *           },
+   *           new Function&lt;TagKeyBoolean, Tag&gt;() {
+   *             &#064;Override
+   *             public Tag apply(TagKeyBoolean booleanKey) {
+   *               return TagBoolean.create(booleanKey, true);
+   *             }
+   *           },
+   *           new Function&lt;TagKey, Tag&gt;() {
+   *             &#064;Override
```
bogdandrutu: Use {@literal @}

sebright: Done

sebright:
I made the `match` methods final, which is consistent with `AttributeValue.match`.

sebright:
Thanks!

PR #324, opened 2017-05-26
Exclude javadoc for internal package.

bogdandrutu:
Add Internal annotation to the internal package.


core/src/main/java/com/google/instrumentation/internal/Internal.java
```diff
@@ -0,0 +1,37 @@
...
+@Retention(RetentionPolicy.SOURCE)
+@Target({
+    ElementType.ANNOTATION_TYPE,
+    ElementType.CONSTRUCTOR,
+    ElementType.FIELD,
+    ElementType.METHOD,
+    ElementType.PACKAGE,
+    ElementType.TYPE})
+@Documented
+public @interface Internal {
```
sebright: What is the purpose of this annotation?  Is it meant to be used as documentation, to warn users not to depend on public elements that aren't considered part of the API, even if they aren't in the `internal` package?  If that is the case, then I think we should expose it by putting it in the `common` package, and it shouldn't be annotated with `@Internal`.

bogdandrutu: Ups, you are correct.

bogdandrutu:
PTAL

PR #328, opened 2017-05-31
Fix NullPointerException in getView() without BucketBoundaries.

songy23:


core_impl/src/test/java/com/google/instrumentation/stats/StatsTestUtil.java
```diff
@@ -124,9 +147,16 @@ private static void assertDistributionAggregationValuesEquivalent(
...
     Truth.assertWithMessage(msg + " sum").that(agg1.getSum()).isWithin(tolerance).of(agg2.getSum());
     Truth.assertWithMessage(msg + " range").that(agg1.getRange()).isEqualTo(agg2.getRange());
-    Truth.assertWithMessage(msg + " bucket counts")
-        .that(removeTrailingZeros(agg1.getBucketCounts()))
-        .isEqualTo(removeTrailingZeros(agg2.getBucketCounts()));
+
+    if (agg1.getBucketCounts() == null && agg2.getBucketCounts() == null) {
+      return;
+    } else if (agg1.getBucketCounts() == null || agg2.getBucketCounts() == null) {
+      throw new AssertionError(msg + " bucket counts");
```
sebright: I think the error message should say how the bucket counts differ, for example, something like "bucket count nullness".  Alternatively, you could use only the else block and modify it to deal with null, to allow `isEqualTo` to generate the error message.

songy23: Updated, thanks.

PR #332, opened 2017-06-05
Split TraceExporter class into ExportComponent and all the sub-components in an export package.

bogdandrutu:
PR got bigger, but the main changes are:

* Create a new package "export" where all the code related to exporting data was moved.
* Rename TraceExporter to ExportComponent.
* Extract inner classes SpanExporter and InProcessDebuggingHandler in normal classes to reduce complexity.


api/src/main/java/io/opencensus/trace/TraceExporter.java
```diff
@@ -49,34 +50,66 @@ static TraceExporter getNoopTraceExporter() {
...
+       * immutable representation {@link SpanData}.
+       *
+       * <p>This may be called from a different thread than the one that called {@link Span#end()}.
+       *
+       * <p>Implementation SHOULD not block the calling thread. It should execute the export on a
+       * different thread if possible.
+       *
+       * @param spanDataList a list of {@code SpanData} objects to be exported.
+       */
+      public abstract void export(Collection<SpanData> spanDataList);
```
adriancole: This is similar to zipkin's Reporter which is often async on this method.

api/src/main/java/io/opencensus/trace/TraceExporter.java
```diff
@@ -49,34 +50,66 @@ static TraceExporter getNoopTraceExporter() {
...
 
   /**
-   * Unregisters the service handler with the provided name.
-   *
-   * @param name the name of the service handler that will be unregistered.
+   * A service that is used by the library to export {@code SpanData} for sampled spans (see
+   * {@link TraceOptions#isSampled()}).
    */
-  public abstract void unregisterServiceHandler(String name);
+  public abstract static class SampledSpansServiceExporter {
```
adriancole: this name seems a bit long, and it isn't clear why the word Service is here at all.

api/src/main/java/io/opencensus/trace/TraceExporter.java
adriancole: maybe SpanExporter (if for some reason SpanReporter isn't acceptable)

bogdandrutu: Picked SpanExporter.

api/src/main/java/io/opencensus/trace/TraceExporter.java
bogdandrutu: Fixed.

PR #334, opened 2017-06-05
Add StatsTester class for creating test instances of StatsManager.

sebright:
This commit changes the way that StatsManagers are created.  Instead of
instantiating one StatsManager with reflection, the Stats class now instantiates
a StatsManagerFactory, which can provide the default StatsManager instance
and parameterized test StatsManager instances.  The test instances are exposed
through a new class, StatsTester.  I added an example program, StatsTesting,
that demonstrates how to use a test StatsManager.

This commit makes several API changes:
- StatsManager.getStatsContextFactory() is now public.
- Stats implementations must extend StatsManagerFactory instead of StatsManager.
- It adds StatsTester.
~~- It renames Stats.getStatsManager() to Stats.getDefaultStatsManager().~~

_____________________________________

/cc @dinooliva @songy23 

EDIT: Here is an example of the test utilities in use: https://github.com/census-instrumentation/instrumentation-java/pull/334/files#diff-b5bd8381d56f10b3e53baa24771eb840R30

### DIFF
--- a/core/src/main/java/io/opencensus/stats/StatsManager.java
+++ b/core/src/main/java/io/opencensus/stats/StatsManager.java
@@ -32,5 +32,5 @@ public abstract class StatsManager {
   /**
    * Returns the default {@link StatsContextFactory}.
    */
-  abstract StatsContextFactory getStatsContextFactory();
+  public abstract StatsContextFactory getStatsContextFactory();
 }
--- a/core/src/main/java/io/opencensus/stats/StatsManagerFactory.java
+++ b/core/src/main/java/io/opencensus/stats/StatsManagerFactory.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2017, Google Inc.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.stats;
+
+import io.opencensus.common.Clock;
+
+/**
+ * Object that determines which implementation of the {@link StatsManager} to use.
+ */
+public abstract class StatsManagerFactory {
+
+  /**
+   * Returns the default {@link StatsManager}.
+   */
+  public abstract StatsManager getDefaultStatsManager();
+
+  /**
+   * Creates a new {@link StatsManager} for testing.
+   */
+  public abstract StatsManager createTestStatsManager();
+
+  /**
+   * Creates a new {@link StatsManager} for testing.
+   *
+   * @param clock The clock that the {@code StatsManager} should use.
+   */
+  public abstract StatsManager createTestStatsManager(Clock clock);
+}
--- a/core/src/main/java/io/opencensus/stats/Stats.java
+++ b/core/src/main/java/io/opencensus/stats/Stats.java
@@ -13,18 +13,12 @@
 
 package io.opencensus.stats;
 
-import com.google.common.annotations.VisibleForTesting;
-import io.opencensus.internal.Provider;
-import java.util.logging.Level;
-import java.util.logging.Logger;
+import io.opencensus.stats.internal.StatsInternal;
 import javax.annotation.Nullable;
 
 /** {@link Stats}. */
 public final class Stats {
-  private static final Logger logger = Logger.getLogger(Stats.class.getName());
-
-  private static final StatsManager statsManager =
-      loadStatsManager(Provider.getCorrectClassLoader(StatsManager.class));
+  private static final StatsManager statsManager = StatsInternal.getStatsManager();
 
   /** Returns the default {@link StatsContextFactory}. */
   @Nullable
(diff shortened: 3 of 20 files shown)


sebright:
I rebased this on top of the package name changes.

sebright:
I moved the StatsTester class into a new package, `io.opencensus.testing.stats`.  StatsTester now has two methods, one that creates a StatsManager in the same way that the singleton is created and one that takes a test clock.

/cc @zhangkun83 @carl-mastrangelo

core/src/main/java/io/opencensus/stats/StatsManagerFactory.java
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+   * Returns the default {@link StatsManager}.
+   */
+  public abstract StatsManager getDefaultStatsManager();
+
+  /**
+   * Creates a new {@link StatsManager} for testing.
+   */
+  public abstract StatsManager createTestStatsManager();
```
zhangkun83: It doesn't feel right to include any for-test methods on a core interface.
I thought you would provide a `TestingStatsContextFactory`, which has a `getEventQueue()`, in the testing package. Users could create `TestingStatsContextFactory` objects as needed.

core/src/main/java/io/opencensus/stats/StatsManager.java
```diff
@@ -32,5 +32,5 @@
   /**
    * Returns the default {@link StatsContextFactory}.
    */
-  abstract StatsContextFactory getStatsContextFactory();
+  public abstract StatsContextFactory getStatsContextFactory();
```
zhangkun83: The users of the views and the users of `StatsContextFactory` are usually different. There isn't much benefit to bundle these two functionalities. I would keep this method as package-private so that `StatsManager` is only a view provider, so that the output component (`StatsManager`) and the input component (`StatsContextFactory`) can be faked and injected separately.

core/src/main/java/io/opencensus/stats/StatsManagerFactory.java
sebright: I agree that it would be better to move all testing-related methods out of the API library, but I wasn't sure how to do it.  `createTestStatsManager` is for the integration test use case.  I wanted to provide a way for users to create a new instance of the StatsManager that is exactly like the singleton, so that they can create a new one for each test but know that it has the same behavior as the singleton.

I'm not sure I understand your suggestion.  Do you mean TestingStatsManagerFactory?  How would the class work?  Would it be able to reuse the implementation of the non-test class?  What would users do with the EventQueue?

core/src/main/java/io/opencensus/stats/StatsManager.java
sebright: I only exposed this method to make the example in this PR work.  We are redesigning the StatsContext class so that it is contained in the API library and doesn't have to be so tightly coupled to the StatsManager (#323).  StatsManager will still have "record" methods, though:

explicit context: `void record(TagContext tags, MeasureMap values)`
implicit context: `void record(MeasureMap values)`

Do you think that registering/querying views and recording stats should also be separate?  I'm not sure how we can separate those parts, since they need to work with shared state.  If people inject them separately, the objects would still need to come from the same StatsManagerFactory instance (or some root object with a better name) that can ensure that the objects are using the same state.

One idea is to call the singleton StatsComponent and give it methods getViewManager() and getStatsRecorder().  Then you could get the ViewManager and StatsRecorder from the singleton and pass them around separately.  Would that address the problem with bundling unrelated functionalities?

zhangkun83: > Do you think that registering/querying views and recording stats should also be separate? I'm not sure how we can separate those parts, since they need to work with shared state. If people inject them separately, the objects would still need to come from the same StatsManagerFactory instance (or some root object with a better name) that can ensure that the objects are using the same state.

I understand in production recording and querying must share state. I think `Stats.java` is taking care of that. For the sake of unit testing, recording and querying do need to be separately injected, because one application component only deal with one of them.

> One idea is to call the singleton StatsComponent and give it methods getViewManager() and getStatsRecorder(). Then you could get the ViewManager and StatsRecorder from the singleton and pass them around separately. Would that address the problem with bundling unrelated functionalities?

That would work. Since you will have `StatsRecorder()`, will you also deprecate the `record()` on `StatsContext`?

core/src/main/java/io/opencensus/stats/StatsManagerFactory.java
sebright: > The two ends already suffice for me, therefore I don't find this testing singleton very useful.

Do you think we shouldn't include the functionality, or is it just not necessary for gRPC?

> I meant TestingStatsContextFactory because I assumed the reporting and querying APIs to be separately injected. It would be a fake implementation that allows user to verify that stats are reported to Census correctly (by looking at EventQueue), while doesn't do any aggregation thus doesn't not involve the view APIs. The idea is similar to Mockito, but easier for users because mocking StatsContextFactory is a pain.

Now I understand.  The new design won't put the record method on StatsContext, so I don't think there will be any need to mock it (and the class will probably be final).  If we had an abstract StatsRecorder class that only had "record" methods on it, do you think that would be simple enough that user could mock it directly, and we wouldn't need to provide a testing implementation like you described?

core/src/main/java/io/opencensus/stats/StatsManager.java
sebright: > I understand in production recording and querying must share state. I think Stats.java is taking care of that. For the sake of unit testing, recording and querying do need to be separately injected, because one application component only deal with one of them.

That makes sense.  We just need to split the classes so that client code can pass around only the part that it needs.

> That would work. Since you will have StatsRecorder(), will you also deprecate the record() on StatsContext?

Yes, in the new design, StatsContext is a simple value class, and it doesn't have a `record` method.

core/src/main/java/io/opencensus/stats/StatsManagerFactory.java
zhangkun83: > Do you think we shouldn't include the functionality, or is it just not necessary for gRPC?

I don't find it useful for gRPC, and I can't think of a potential use case. I would suggest not include this until a real use case comes up. It's always easier to add something than removing or changing it.

>  If we had an abstract StatsRecorder class that only had "record" methods on it, do you think that would be simple enough that user could mock it directly, and we wouldn't need to provide a testing implementation like you described?

Yes, as long as `StatsContext` loses its `record()` and becomes a pure data object.

sebright: Thanks!  I made a PR that splits out a StatsRecorder class: #387

sebright:
I'm closing this for now, because it probably isn't needed, and it would at least need to be updated for #387.

PR #338, opened 2017-06-06
Add a link to JIRA bug for releasing instruction.

songy23:


songy23:
@bogdandrutu I just updated the instruction.
>  soon we will have 2:
does this mean in the future we'll both release to com.google.instrumentation and io.opencensus?

PR #342, opened 2017-06-07
Move all base types into a different package called base.

bogdandrutu:


adriancole:
Completely arbitrary, but I sometimes use packages named api common or
model for this stuff. Dont read too much into it.. though ;)

PR #343, opened 2017-06-07
Move SpanData to export package. Move implementation of export package class to export package.

bogdandrutu:
More cleanups included:

* Add internal annotation to all implementation classes.
* Fix javadoc export for implementation (avoid exporting io.opencensus.internal).

### DIFF
--- a/api/src/test/java/io/opencensus/trace/export/SpanDataTest.java
+++ b/api/src/test/java/io/opencensus/trace/export/SpanDataTest.java
@@ -11,25 +11,26 @@
  * limitations under the License.
  */
 
-package io.opencensus.trace;
+package io.opencensus.trace.export;
 
 import static com.google.common.truth.Truth.assertThat;
 
 import com.google.common.testing.EqualsTester;
 import io.opencensus.common.Timestamp;
-import io.opencensus.trace.base.Link;
-import io.opencensus.trace.base.Link.Type;
-import io.opencensus.trace.SpanData.Attributes;
-import io.opencensus.trace.SpanData.Links;
-import io.opencensus.trace.SpanData.TimedEvent;
-import io.opencensus.trace.SpanData.TimedEvents;
+import io.opencensus.trace.SpanContext;
 import io.opencensus.trace.base.Annotation;
 import io.opencensus.trace.base.AttributeValue;
+import io.opencensus.trace.base.Link;
+import io.opencensus.trace.base.Link.Type;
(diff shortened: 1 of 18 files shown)


api/src/test/java/io/opencensus/trace/export/SpanDataTest.java
```diff
@@ -35,7 +45,8 @@
 
 /** Unit tests for {@link SpanData}. */
 @RunWith(JUnit4.class)
-public class SpanDataTest {
+public class
```
adriancole: format glitch

core_impl/src/main/java/io/opencensus/package-info.java
```diff
@@ -0,0 +1,21 @@
...
+ * limitations under the License.
+ */
+
+/**
+ * Implementations that are internal to OpenCensus.
+ *
+ * <p>All the content under this package and its subpackages are considered annotated with {@link
+ * io.opencensus.common.Internal}.
+ */
+@io.opencensus.common.Internal
```
adriancole: hmm so sharing packages across jars can cause modular systems to hiccup (ex OSGi makes you use fragments). might be worth reconsidering

core_impl/src/main/java/io/opencensus/trace/SpanImpl.java
```diff
@@ -34,7 +36,7 @@
 
 /** Implementation for the {@link Span} class. */
 @ThreadSafe
-final class SpanImpl extends Span {
+public final class SpanImpl extends Span {
```
adriancole: instead of doing this, you could make a Internal package accessor type. like `InternalExport.startSpan(...)` which returns span. This makes it blatantly obvious what's for internal use and not.

core_impl/src/main/java/io/opencensus/package-info.java
bogdandrutu: Can you give us clarifications on this?

adriancole: sorry. yeah in OSGi, the default is that each package is sealed within a jar. You can use fragments to contribute other classes to a jar https://stackoverflow.com/questions/25670577/should-osgi-fragments-export-packages-contributed-to-their-host

There are other things about classloading which are simpler when everything in a package is in the same jar. If memory serves me right, package-private access is more reliable if in the same jar/classloader.

PR #348, opened 2017-06-08
Move more classes into internal packages to be clearer that they are internal.

bogdandrutu:
Move EventQueue and all implementations to internal package from common. 
Move TimestampConverter to internal implementation (may be useful for other packages).
Move RandomHandler to trace/internal.


core_impl_java/src/main/java/io/opencensus/internal/package-info.java
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+/**
+ * Interfaces and implementations that are internal to OpenCensus.
+ *
+ * <p>All the content under this package and its subpackages are considered annotated with {@link
+ * io.opencensus.common.Internal}.
+ */
+@io.opencensus.common.Internal
+package io.opencensus.internal;
```
sebright: Isn't the core_impl_java directory already treated as internal?

bogdandrutu: Removed.

PR #351, opened 2017-06-10
Split InProcessDebuggingHandler into ActiveSpans and SampleStore.

bogdandrutu:
This PR tries to split the two different functionality that were supported by the InProcessDebuggingHandler. With this change it is a bit more clear which class and API is responsible for any of the library capabilities.

### DIFF
--- a/api/src/main/java/io/opencensus/trace/export/ActiveSpans.java
+++ b/api/src/main/java/io/opencensus/trace/export/ActiveSpans.java
@@ -1,0 +1,120 @@
+/*
+ * Copyright 2017, Google Inc.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.trace.export;
+
+import static com.google.common.base.Preconditions.checkArgument;
+
+import com.google.auto.value.AutoValue;
+import io.opencensus.trace.Span;
+import java.util.Collection;
+import java.util.Map;
+import javax.annotation.concurrent.Immutable;
+import javax.annotation.concurrent.ThreadSafe;
+
+/**
+ * This class allows users to access in-process information about all active spans.
+ *
+ * <p>The active spans tracking is available for all the spans with the option {@link
+ * Span.Options#RECORD_EVENTS}. This functionality allows users to debug stuck operations or long
+ * living operations.
+ */
+@ThreadSafe
+public abstract class ActiveSpans {
+
+  ActiveSpans() {}
+
+  /**
+   * Returns the number of active spans for every different span name.
+   *
+   * @return the number of active spans for every different span name.
+   */
+  public abstract Map<String, PerSpanNameSummary> getPerSpanNameSummary();
+
+  /**
+   * Returns a list of active spans that match the {@code Filter}.
+   *
+   * @param filter used to filter the returned spans.
+   * @return a list of active spans that match the {@code Filter}.
+   */
+  public abstract Collection<SpanData> getActiveSpans(Filter filter);
+
+  /** Summary of all available data for a span name. */
+  @AutoValue
+  @Immutable
+  public abstract static class PerSpanNameSummary {
+
+    PerSpanNameSummary() {}
+
+    /**
+     * Returns a new instance of {@code PerSpanNameSummary}.
+     *
+     * @param numActiveSpans the number of sampled spans.
+     * @return a new instance of {@code PerSpanNameSummary}.
+     * @throws IllegalArgumentException if {@code numActiveSpans} is negative.
+     */
+    public static PerSpanNameSummary create(int numActiveSpans) {
+      checkArgument(numActiveSpans >= 0, "Negative numActiveSpans.");
+      return new AutoValue_ActiveSpans_PerSpanNameSummary(numActiveSpans);
+    }
+
+    /**
+     * Returns the number of active spans.
+     *
+     * @return the number of active spans.
+     */
+    public abstract int getNumActiveSpans();
+  }
+
+  /**
+   * Filter for active spans. Used to filter results returned by the {@link #getActiveSpans(Filter)}
+   * request.
+   */
+  @AutoValue
+  @Immutable
+  public abstract static class Filter {
+
+    Filter() {}
+
+    /**
+     * Returns a new instance of {@code Filter}.
+     *
+     * <p>Filters all the spans based on {@code spanName} and returns a maximum of {@code
+     * maxSpansToReturn}.
+     *
+     * @param spanName the name of the span.
+     * @param maxSpansToReturn the maximum number of results to be returned. {@code 0} means all.
+     * @return a new instance of {@code Filter}.
+     * @throws NullPointerException if {@code spanName} is {@code null}.
+     * @throws IllegalArgumentException if {@code maxSpansToReturn} is negative.
+     */
+    public static Filter create(String spanName, int maxSpansToReturn) {
+      checkArgument(maxSpansToReturn >= 0, "Negative maxSpansToReturn.");
+      return new AutoValue_ActiveSpans_Filter(spanName, maxSpansToReturn);
+    }
+
+    /**
+     * Returns the span name.
+     *
+     * @return the span name.
+     */
+    public abstract String getSpanName();
+
+    /**
+     * Returns the maximum number of spans to be returned. {@code 0} means all.
+     *
+     * @return the maximum number of spans to be returned.
+     */
+    public abstract int getMaxSpansToReturn();
+  }
+}
--- a/api/src/main/java/io/opencensus/trace/export/ActiveSpans.java
+++ b/api/src/main/java/io/opencensus/trace/export/ActiveSpans.java
@@ -32,14 +32,14 @@ import javax.annotation.concurrent.ThreadSafe;
 @ThreadSafe
 public abstract class ActiveSpans {
 
-  ActiveSpans() {}
+  protected ActiveSpans() {}
 
   /**
    * Returns the number of active spans for every different span name.
    *
    * @return the number of active spans for every different span name.
    */
-  public abstract Map<String, PerSpanNameSummary> getPerSpanNameSummary();
+  public abstract Map<String, Integer> getNumberOfActiveSpans();
 
   /**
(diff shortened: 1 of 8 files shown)


api/src/main/java/io/opencensus/trace/export/ExportComponent.java
```diff
@@ -45,14 +44,22 @@ public static ExportComponent getNoopExportComponent() {
...
 
   /**
-   * Returns the {@link InProcessDebuggingHandler} that can be used to get useful debugging
-   * information such as (active spans, latency based sampled spans, error based sampled spans).
+   * Returns the {@link ActiveSpans} that can be used to get useful debugging information such as
+   * (active spans, latency based sampled spans, error based sampled spans).
    *
-   * @return the {@code InProcessDebuggingHandler} or {@code null} if in-process debugging is not
-   *     supported.
+   * @return the {@code SampleStore} or {@code null} if in-process debugging is not supported.
```
sebright: s/SampleStore/ActiveSpans

api/src/main/java/io/opencensus/trace/export/SampleStore.java
```diff
@@ -14,58 +14,51 @@
...
-public abstract class InProcessDebuggingHandler {
+public abstract class SampleStore {
 
-  InProcessDebuggingHandler() {}
+  protected SampleStore() {}
 
   /**
-   * Returns the summary of all available in-process debugging data such as number of active spans,
-   * number of sampled spans in the latency based samples or error based samples.
+   * Returns the number of sampled spans in all the latency buckets for every different span name.
```
sebright: Same here, I think "different" is redundant.

api/src/main/java/io/opencensus/trace/export/SampleStore.java
```diff
@@ -14,58 +14,51 @@
...
-   * only for span names registered using {@link #registerSpanNamesForCollection(Collection)}.
+   * <p>Data available only for span names registered using {@link
+   * #registerSpanNamesForCollection(Collection)}.
    *
-   * @return the summary of all available in-process debugging data.
+   * @return the number of sampled spans in all the latency buckets for every different span name.
    */
-  public abstract Summary getSummary();
+  public abstract Map<String, Map<LatencyBucketBoundaries, Integer>>
+      getNumberOfLatencySampledSpans();
```
sebright: > public abstract Map<String, Map<LatencyBucketBoundaries, Integer>> getNumberOfLatencySampledSpans();

Would it be better to make a class for this return type?  Then you could change it to add more data in the future, if necessary.  The name would also be more descriptive.

Another improvement would be adding a `SpanName` class for the first key type, to improve readability and type safety.  This suggestion also applies to `ActiveSpans.getNumberOfActiveSpans()`.

api/src/main/java/io/opencensus/trace/export/ActiveSpans.java
```diff
@@ -0,0 +1,93 @@
...
+   */
+  public abstract Collection<SpanData> getActiveSpans(Filter filter);
+
+  /**
+   * Filter for active spans. Used to filter results returned by the {@link #getActiveSpans(Filter)}
+   * request.
+   */
+  @AutoValue
+  @Immutable
+  public abstract static class Filter {
```
sebright: SpanNameFilter might be clearer.

api/src/main/java/io/opencensus/trace/export/ActiveSpans.java
```diff
@@ -0,0 +1,120 @@
...
+ */
+@ThreadSafe
+public abstract class ActiveSpans {
+
+  ActiveSpans() {}
+
+  /**
+   * Returns the number of active spans for every different span name.
+   *
+   * @return the number of active spans for every different span name.
```
bogdandrutu: Done.

api/src/main/java/io/opencensus/trace/export/ExportComponent.java
```diff
@@ -45,14 +44,22 @@ public static ExportComponent getNoopExportComponent() {
   public abstract SpanExporter getSpanExporter();
 
   /**
-   * Returns the {@link InProcessDebuggingHandler} that can be used to get useful debugging
-   * information such as (active spans, latency based sampled spans, error based sampled spans).
+   * Returns the {@link ActiveSpans} that can be used to get useful debugging information such as
+   * (active spans, latency based sampled spans, error based sampled spans).
```
bogdandrutu: It was not. Thanks.

api/src/main/java/io/opencensus/trace/export/ExportComponent.java
bogdandrutu: Done.

api/src/main/java/io/opencensus/trace/export/ExportComponent.java
```diff
@@ -45,14 +44,22 @@ public static ExportComponent getNoopExportComponent() {
...
    *
-   * @return the {@code InProcessDebuggingHandler} or {@code null} if in-process debugging is not
-   *     supported.
+   * @return the {@code SampleStore} or {@code null} if in-process debugging is not supported.
    */
   @Nullable
-  public abstract InProcessDebuggingHandler getInProcessDebuggingHandler();
+  public abstract ActiveSpans getActiveSpans();
```
bogdandrutu: Yes it is, usually the "component" pattern is used to return implementation for specific interfaces. And in this case it is. See the comment on the top of the ExportComponent class.

api/src/main/java/io/opencensus/trace/export/ExportComponent.java
```diff
@@ -45,14 +44,22 @@ public static ExportComponent getNoopExportComponent() {
...
+
+  /**
+   * Returns the {@link SampleStore} that can be used to get useful debugging information such as
+   * (active spans, latency based sampled spans, error based sampled spans).
```
bogdandrutu: Done.

api/src/main/java/io/opencensus/trace/export/SampleStore.java
```diff
@@ -29,43 +29,29 @@
...
+ * and error based sampled spans.
  *
  * <p>For all completed spans with the option {@link Span.Options#RECORD_EVENTS} the library can
  * store samples based on latency for succeeded operations or based on error code for failed
  * operations. To activate this, users MUST manually configure all the span names for which samples
  * will be collected (see {@link #registerSpanNamesForCollection(Collection)}).
  */
 @ThreadSafe
-public abstract class InProcessDebuggingHandler {
+public abstract class SampleStore {
```
bogdandrutu: Done. Changed to what you proposed, Is SampledSpan[s]Store better?

api/src/main/java/io/opencensus/trace/export/SampleStore.java
bogdandrutu: Done.

api/src/main/java/io/opencensus/trace/export/ActiveSpans.java
bogdandrutu: For the same reasons as you pointed in the usage of classes for the return types I think it would be good to keep the name generic in case we want to add more conditions for filtering.

api/src/main/java/io/opencensus/trace/export/SampleStore.java
bogdandrutu: What name do you think of?

public abstract Summary getSummary();

class Summary {
  public abstract Map<SpanName, Map<CanonicalCode, Integer>> numberOfErrorSampledSpans();
  public abstract Map<SpanName, Map<LatencyBucketBoundaries, Integer>> numberOfLatencySampledSpans();
}

OR something like

public abstract Map<SpanName, PerSpanNameSummary> getPerSpanNameSummary();

class PerSpanNameSummary {
  public abstract Map<LatencyBucketBoundaries, Integer> numberOfLatencySampledSpans();
  public abstract Map<CanonicalCode, Integer> numberOfErrorSampledSpans();
}

api/src/main/java/io/opencensus/trace/export/ExportComponent.java
sebright: I still think it would help to give the class a more descriptive name, such as "ActiveSpansExporter" or "ActiveSpansReader".

api/src/main/java/io/opencensus/trace/export/ExportComponent.java
```diff
@@ -44,22 +45,22 @@ public static ExportComponent getNoopExportComponent() {
...
   public abstract ActiveSpans getActiveSpans();
 
   /**
-   * Returns the {@link SampleStore} that can be used to get useful debugging information such as
-   * (active spans, latency based sampled spans, error based sampled spans).
+   * Returns the {@link SampledSpanStore} that can be used to get useful debugging information, such
+   * as latency based sampled spans, error based sampled spans.
    *
-   * @return the {@code SampleStore} or {@code null} if in-process debugging is not supported.
+   * @return the {@code SampleStore} or {@code null} if not supported.
```
bogdandrutu: Done.

bogdandrutu:
PTAL

api/src/main/java/io/opencensus/trace/export/ExportComponent.java
```diff
@@ -45,19 +45,19 @@ public static ExportComponent getNoopExportComponent() {
...
-   * the current active spans.
+   * Returns the {@link ActiveSpansExporter} that can be used to get useful debugging information
+   * about all the current active spans.
    *
-   * @return the {@code ActiveSpans} or {@code null} if not supported.
+   * @return the {@code ActiveSpansExporter} or {@code null} if not supported.
    */
   @Nullable
-  public abstract ActiveSpans getActiveSpans();
+  public abstract ActiveSpansExporter getActiveSpansExporter();
```
sebright: I just noticed that the pluralization of "span" is inconsistent; the classes are SpanExporter, SampledSpanStore, and ActiveSpansExporter.  I don't know if that's a problem, and I don't think it needs to be changed in this PR.

api/src/main/java/io/opencensus/trace/export/SampledSpanStore.java
```diff
@@ -105,6 +98,84 @@ protected SampledSpanStore() {}
...
+    PerSpanNameSummary() {}
+
+    /**
+     * Returns a new instance of {@code PerSpanNameSummary}.
+     *
+     * @param numberOfLatencySampledSpans the summary for the latency buckets.
+     * @param numberOfErrorSampledSpans the summary for the error buckets.
+     * @return a new instance of {@code PerSpanNameSummary}.
+     * @throws NullPointerException if {@code latencyBucketSummaries} or {@code
+     *     errorBucketSummaries} are {@code null}.
```
sebright: >     * @throws NullPointerException if {@code latencyBucketSummaries} or {@code
>     *     errorBucketSummaries} are {@code null}.

The two variable names are out of date.

adriancole:
Noticed in another PR Active Span doesnt seem defined clearly enough. Maybe
say on the top level type that it is in flight (started but not stopped or
simply not stopped)

PR #353, opened 2017-06-10
Add an IntrusiveList implementation to be used to implement ActiveSpans.

bogdandrutu:


core_impl/src/main/java/io/opencensus/trace/internal/ConcurrentIntrusiveList.java
```diff
@@ -0,0 +1,185 @@
...
+
+import io.opencensus.trace.internal.ConcurrentIntrusiveList.Element;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import javax.annotation.concurrent.ThreadSafe;
+
+/**
+ * An {@code ConcurrentIntrusiveList<T>} is a doubly-linked list where the link pointers are
+ * embedded in the elements. This makes insertion and removal into a known position constant time.
```
adriancole: I'd prefer things like this introduced when they are used as it helps better tell the story of what it is doing. For example, benchmarks could show if it works well for the use case.

One thing interesting about this is mention of a "known position", though the operations on the list aren't positional in nature (or at least not explicitly so). indeed Element is positional, so this hints at how the list is used.. which is to my point about it being clearer if this is raised when it is used :P

core_impl/src/main/java/io/opencensus/trace/internal/ConcurrentIntrusiveList.java
```diff
@@ -0,0 +1,185 @@
...
+    return all;
+  }
+
+  /**
+   * This is an interface that must be implemented by any element that uses {@link
+   * ConcurrentIntrusiveList}.
+   *
+   * @param <T> the element that will be used for the list.
+   */
+  public interface Element<T extends Element<T>> {
```
adriancole: I guess long story short.. unless this is to be used by multiple types.. maybe just make activespan list as the api would still be testable and benchmarkable, but much more relevant to the codebase.

core_impl/src/main/java/io/opencensus/trace/internal/ConcurrentIntrusiveList.java
bogdandrutu: I need to merge https://github.com/census-instrumentation/instrumentation-java/pull/351 before I can have the usage example. Added the SpanImpl changes only for the moment.

core_impl/src/main/java/io/opencensus/trace/internal/ConcurrentIntrusiveList.java
```diff
@@ -0,0 +1,185 @@
...
+import javax.annotation.concurrent.ThreadSafe;
+
+/**
+ * An {@code ConcurrentIntrusiveList<T>} is a doubly-linked list where the link pointers are
+ * embedded in the elements. This makes insertion and removal into a known position constant time.
+ *
+ * <p>Elements must derive from the {@code Element<T extends Element<T>>} base class:
+ *
+ * <pre>{@code
+ * class MyClass implements Element<MyClass> {
```
caiwaifung: Wouldn't it be nicer to implement {get,set}{Next,Prev} in Element instead of forcing the users to implement them?

caiwaifung: A simple example would be:

class Element<T extends Element<T>> {
  private T prev = null;
  private T next = null;
  public T getPrev() { return prev;}
  public void setPrev(T prev) { this.prev = prev; }
  public T getNext() { return next; }
  public void setNext(T next) { this.next = next; }
}


class IntrusiveList<T extends Element<T>> {
  private T head = null;
  public void add(T e) {
    assert (e.getNext() == null && e.getPrev() == null);
    if (head == null) {
      head = e;
    } else {
      e.setNext(head);
      head.setPrev(e);
      head = e;
    }
  }

  public T pop() {
    T e = head;
    head = head.getNext();
    e.setNext(null);
    head.setPrev(null);
    return e;
  }

  public void print() {
    for (T e = head; e != null; e = e.getNext()) {
      System.out.print(e + " ");
    }
    System.out.println("");
  }
}


class Int extends Element<Int> {
  public int x;

  public Int(int x) {
    this.x = x;
  }

  @Override
  public String toString() {
    return "" + x;
  }
}


public class Main {
  public static void main(String[] args) {
    IntrusiveList<Int> a = new IntrusiveList<Int>();
    a.print();
    a.add(new Int(5));
    a.add(new Int(6));
    a.print();
    System.out.println(a.pop());
    a.add(new Int(7));
    a.add(new Int(8));
    a.print();
  }
}

core_impl/src/main/java/io/opencensus/trace/internal/ConcurrentIntrusiveList.java
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  /**
+   * Returns all the elements from this list.
+   *
+   * @return all the elements from this list.
+   */
+  public synchronized Collection<T> getAll() {
+    List<T> all = new ArrayList<T>(size);
+    T it = head;
```
caiwaifung: Does a for look better?

for (T e = head; e != null; e = e.getNext()) {
    all.add(e);
}

core_impl/src/main/java/io/opencensus/trace/internal/ConcurrentIntrusiveList.java
```diff
@@ -0,0 +1,185 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import javax.annotation.concurrent.ThreadSafe;
+
+/**
+ * An {@code ConcurrentIntrusiveList<T>} is a doubly-linked list where the link pointers are
+ * embedded in the elements. This makes insertion and removal into a known position constant time.
+ *
+ * <p>Elements must derive from the {@code Element<T extends Element<T>>} base class:
```
sebright: "interface" instead of "base class"?

core_impl/src/main/java/io/opencensus/trace/internal/ConcurrentIntrusiveList.java
```diff
@@ -0,0 +1,185 @@
...
+ *
+ *   {@literal @}Override
+ *   void setPrev(MyClass element) {
+ *     prev = element;
+ *   }
+ * }
+ * }</pre>
+ */
+@ThreadSafe
+public final class ConcurrentIntrusiveList<T extends Element<T>> {
```
sebright: Will this list only be used with internal classes as the element type, so that there is no way for users to break the invariants by calling setNext or setPrev on the elements?

core_impl/src/main/java/io/opencensus/trace/internal/ConcurrentIntrusiveList.java
sebright: In my opinion, the generic type is clearer, because it shows that this is a general-purpose collection, and it isn't making use of any other part of the span.

core_impl/src/test/java/io/opencensus/trace/internal/ConcurrentIntrusiveListTest.java
```diff
@@ -0,0 +1,128 @@
...
+    FakeElement element1 = new FakeElement();
+    FakeElement element2 = new FakeElement();
+    FakeElement element3 = new FakeElement();
+    intrusiveList.addElement(element1);
+    intrusiveList.addElement(element2);
+    intrusiveList.addElement(element3);
+    assertThat(intrusiveList.size()).isEqualTo(3);
+    assertThat(intrusiveList.getAll().contains(element1)).isTrue();
+    assertThat(intrusiveList.getAll().contains(element2)).isTrue();
+    assertThat(intrusiveList.getAll().contains(element3)).isTrue();
```
sebright: Suggestion: You could use containsExactly to also check the order, i.e.,

    assertThat(intrusiveList.getAll()).containsExactly(element1, element2, element3).inOrder();

core_impl/src/main/java/io/opencensus/trace/internal/ConcurrentIntrusiveList.java
adriancole: fair enough.. maybe hide the type until it has more than one use? The term generic implies it would, but so far I'm not seeing any other use case.

PR #356, opened 2017-06-15
Add initial implementation of the ActiveSpansExporter.

bogdandrutu:


adriancole:
Looks like this implementation gives a breakdown of spans by name and also
a way to query them. Notably, these are in-flight spans.

The "active span" here is different than another library which uses active
to indicate it the current implicit span in context. There's no strict
relationship to that here. Eventhough no need to be vocab parity with
others seems in-flight or running would work (start run end!). Just
thoughts not important ones.

bogdandrutu:
Other possible names (using in flight) are:
* InFlightSpanStore
* InFlightSpansExporter
* InFlightSpanTracker

What do you prefer?

adriancole:
>
>
>    - InFlightSpanStore
>
> works best in my brain. We are tracking and querying and somehow this
seems to be ok for both!

bogdandrutu:
I will do the change in a separate PR.

core_impl/src/test/java/io/opencensus/trace/export/ActiveSpansExporterTest.java
```diff
@@ -0,0 +1,160 @@
...
+
+  @Test
+  public void getActiveSpans_SpansWithSameName() {
+    SpanImpl span1 = createSpan(SPAN_NAME_1);
+    SpanImpl span2 = createSpan(SPAN_NAME_1);
+    SpanImpl span3 = createSpan(SPAN_NAME_1);
+    assertThat(activeSpansExporter.getActiveSpans(Filter.create(SPAN_NAME_1, 0)))
+        .containsExactly(span1.toSpanData(), span2.toSpanData(), span3.toSpanData());
+    assertThat(activeSpansExporter.getActiveSpans(Filter.create(SPAN_NAME_1, 2)))
+        .containsExactly(span2.toSpanData(), span3.toSpanData());
```
sebright: Can users rely on getting only the latest spans when they set a maximum?  I think it would be better to only test that any two are returned, or document the behavior in the Filter class.

core_impl/src/main/java/io/opencensus/trace/StartEndHandlerImpl.java
```diff
@@ -14,51 +14,103 @@
...
+import io.opencensus.trace.export.SampledSpanStore;
 import io.opencensus.trace.export.SpanData;
 import io.opencensus.trace.export.SpanExporterImpl;
+import javax.annotation.Nullable;
 
 /**
  * Uses the provided {@link EventQueue} to defer processing/exporting of the {@link SpanData} to
  * avoid impacting the critical path.
  */
 public final class StartEndHandlerImpl implements StartEndHandler {
```
bogdandrutu: Done.

core_impl/src/main/java/io/opencensus/trace/StartEndHandlerImpl.java
```diff
@@ -14,51 +14,103 @@
...
+  private final SpanExporterImpl spanExporter;
+  private final ActiveSpansExporterImpl activeSpansExporter;
+  private final SampledSpanStore sampledSpanStore;
   private final EventQueue eventQueue;
-  private final SpanExporterImpl sampledSpansServiceExporter;
+  // true if any of (activeSpansExporter OR sampledSpanStore) are different than null, which
+  // means the spans with RECORD_EVENTS should be enqueued in the queue.
+  private final boolean enqueEventForNonSampledSpans;
```
bogdandrutu: Done.

core_impl/src/main/java/io/opencensus/trace/StartEndHandlerImpl.java
```diff
@@ -14,51 +14,103 @@
...
+   * @param spanExporter the {@code SpanExporter} implementation.
+   * @param activeSpansExporter the {@code ActiveSpansExporter} implementation.
+   * @param sampledSpanStore the {@code SampledSpanStore} implementation.
+   * @param eventQueue the event queue where all the events are enqueued.
+   */
   public StartEndHandlerImpl(
-      SpanExporterImpl sampledSpansServiceExporter, EventQueue eventQueue) {
-    this.sampledSpansServiceExporter = sampledSpansServiceExporter;
+      SpanExporterImpl spanExporter,
+      @Nullable ActiveSpansExporterImpl activeSpansExporter,
```
bogdandrutu: Forgot to add that code. On the lite version we do not want to support active and sampled.

core_impl/src/test/java/io/opencensus/trace/export/ExportComponentImplTest.java
```diff
@@ -32,7 +32,8 @@ public void implementationOfSpanExporter() {
   @Test
   public void implementationOfActiveSpans() {
     // TODO(bdrutu): Change this when implementation is available.
```
bogdandrutu: Done.

core_impl/src/main/java/io/opencensus/trace/SpanImpl.java
```diff
@@ -133,7 +133,7 @@ public static SpanImpl startSpan(
    *
    * @return the name of the {@code Span}.
    */
-  String getName() {
+  public String getName() {
     return name;
   }
```
bogdandrutu: Very good catch.

core_impl/src/test/java/io/opencensus/trace/export/ActiveSpansExporterTest.java
bogdandrutu: done.

bogdandrutu:
PTAL

impl_core/src/test/java/io/opencensus/trace/export/ExportComponentImplTest.java
```diff
@@ -22,23 +22,25 @@
 /** Unit tests for {@link ExportComponentImpl}. */
 @RunWith(JUnit4.class)
 public class ExportComponentImplTest {
-  private final ExportComponent exportComponent = new ExportComponentImpl();
+  private final ExportComponent exportComponentWithInProcess = new ExportComponentImpl(true);
+  private final ExportComponent exportComponentWitoutInProcess = new ExportComponentImpl(false);
```
sebright: s/Witout/Without

impl_core/src/main/java/io/opencensus/trace/export/ExportComponentImpl.java
```diff
@@ -42,9 +42,14 @@ public SampledSpanStore getSampledSpanStore() {
...
 
-  /** Constructs a new {@code ExportComponentImpl}. */
-  public ExportComponentImpl() {
+  /**
+   * Constructs a new {@code ExportComponentImpl}.
+   *
+   * @param supportInProcessStores {@code true} to instantiate {@link ActiveSpansExporter} and
+   *     {@link SampledSpanStore}.
+   */
+  public ExportComponentImpl(boolean supportInProcessStores) {
```
sebright: Suggestion: Using separate factory methods for full and lite functionality (or using an enum instead of boolean) would make the calling code easier to read.

bogdandrutu: Done.

impl_core/src/test/java/io/opencensus/trace/export/ExportComponentImplTest.java
bogdandrutu: Done.

PR #366, opened 2017-06-16
Add a skeleton Java agent

ubschmidt2:
Add a skeleton Java agent, which hooks up Bytebuddy, but doesn't do anything yet besides logging. Also add it to the build process, bundling the required classes and dependencies in a single agent JAR file.

### DIFF
--- a/contrib/agent/build.gradle
+++ b/contrib/agent/build.gradle
@@ -1,0 +1,40 @@
+plugins {
+  id 'com.github.johnrengelman.shadow' version '2.0.0'
+}
+
+description = 'OpenCensus Agent'
+
+sourceCompatibility = 1.7
+targetCompatibility = 1.7
+
+def agentPackage = 'io.opencensus.contrib.agent'
+def agentMainClass = "${agentPackage}.AgentMain"
+
+dependencies {
+  compile group: 'net.bytebuddy', name: 'byte-buddy', version: '1.7.1'
+
+  signature 'org.codehaus.mojo.signature:java17:+@signature'
+}
+
+jar {
+  // Set required manifest attributes, cf.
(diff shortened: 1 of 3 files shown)


agent/src/main/java/io/opencensus/agent/AgentBuilderListener.java
```diff
@@ -0,0 +1,73 @@
...
+import net.bytebuddy.agent.builder.AgentBuilder;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.dynamic.DynamicType;
+import net.bytebuddy.utility.JavaModule;
+
+/**
+ * An {@link AgentBuilder.Listener} which uses {@link java.util.logging} for logging events of
+ * interest.
+ */
+final class AgentBuilderListener implements AgentBuilder.Listener {
```
adriancole: I don't recall where, but I have seen bytebuddy unit tests for agent installation. If you can't find one lemme know and I'll dig. I think basic infra should start with a basic test as usually this encourages future ones by lowering the barrier to entry

ubschmidt2: Makes sense. For this no-op skeleton agent, I don't have meaningful tests. I have integration tests for the actual instrumentation that I'll add in separate PRs, sure. FWIW, I've seen https://stackoverflow.com/questions/35577485/junit-test-framework-for-javaagent-instrumentation-framework/35597908#35597908. Please LMK if there's anything better/easier.

ubschmidt2:
Rebased.

bsideup:
FYI it has ByteBuddy branch:
https://github.com/bsideup/javaagent-boilerplate/tree/bytebuddy

ubschmidt2:
@bsideup: Thanks!

@adriancole: I had some thoughts about 1.6. I'll add some comments to the build script itself.

PR #372, opened 2017-06-19
Clean minor things.

bogdandrutu:


api/src/main/java/io/opencensus/trace/Span.java
```diff
@@ -52,7 +51,7 @@
   public enum Options {
     /**
      * This option is set if the Span is part of a sampled distributed trace OR the {@link
-     * StartSpanOptions#getRecordEvents()} is true.
+     * SpanBuilder#setRecordEvents(boolean)} is true.
```
sebright: Should this be "`OR {@link SpanBuilder#setRecordEvents(boolean)} was called with true.`"?

bogdandrutu: Done

PR #375, opened 2017-06-20
For Reference Only: new tags package

dinooliva:


bogdandrutu:
If you use this plugin https://github.com/google/google-java-format it will fix all issues like this (unused imports). I would highly recommend to use it.

PR #387, opened 2017-06-23
Replace StatsManager with StatsComponent class.  (Fixes #376)

### LINKED ISSUE #376: Rename StatsManager to match naming in the trace package.
The corresponding class for tracing is `TraceComponent`, so `StatsComponent` would be more consistent.

sebright:
StatsComponent is the class that is instantiated with reflection, to determine
which stats library implementation to use.  It contains a ViewManager,
StatsRecorder, and StatsContextFactory.  (The StatsContextFactory will be
removed when we replace StatsContext with TagContext.)

Breaking up the stats library functionality into multiple classes allows client
code to pass around only the parts that it needs.  Then when users mock the
stats library, they only need to mock the parts that their code uses.  For
example, a user can write unit tests for code that should record specific stats
by mocking StatsRecorder and asserting that StatsRecorder.record was called
with the correct arguments.

This change also makes the stats package more consistent with the trace package,
which has a TraceComponent that holds all other objects that provide tracing
functionality.

/cc @zhangkun83 @carl-mastrangelo

EDIT: This is related to the discussion in #334.


core/src/main/java/io/opencensus/stats/StatsComponent.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+public abstract class StatsComponent {
+  /** Returns the default {@link ViewManager}. */
+  public abstract ViewManager getViewManager();
+
+  /** Returns the default {@link StatsRecorder}. */
+  public abstract StatsRecorder getStatsRecorder();
+
+  /** Returns the default {@link StatsContextFactory}. */
+  abstract StatsContextFactory getStatsContextFactory();
```
bogdandrutu: As you mentioned this will be removed and if there will be an equivalent it will be in tags/Tagger.java or something similar, correct?

core/src/main/java/io/opencensus/stats/Stats.java
```diff
@@ -19,51 +19,57 @@
...
-  private static final StatsManager statsManager =
-      loadStatsManager(Provider.getCorrectClassLoader(StatsManager.class));
+  private static final StatsComponent statsComponent =
+      loadStatsComponent(Provider.getCorrectClassLoader(StatsComponent.class));
 
   /** Returns the default {@link StatsContextFactory}. */
   @Nullable
   public static StatsContextFactory getStatsContextFactory() {
-    return statsManager == null ? null : statsManager.getStatsContextFactory();
+    return statsComponent == null ? null : statsComponent.getStatsContextFactory();
```
bogdandrutu: Instead of having all this if conditions probably easier to have a NoopStatsComponent which returns the expected value if no-op implementation.

Maybe make sense in a different PR.

core/src/main/java/io/opencensus/stats/StatsRecorder.java
```diff
@@ -0,0 +1,25 @@
...
+
+/** Provides methods to record stats against tags. */
+public abstract class StatsRecorder {
+  /**
+   * Records a set of measurements with a set of tags.
+   *
+   * @param tags the tags associated with the measurements.
+   * @param measurementValues the measurements to record.
+   */
+  abstract void record(StatsContext tags, MeasurementMap measurementValues);
```
bogdandrutu: Probably a TODO to add an:
void record(MeasurementMap measurementValues) {
  record(getCurrentTagContext(), measurementValues);
}

To help users.

core/src/main/java/io/opencensus/stats/StatsComponent.java
sebright: Yes, there probably won't be a factory for TagContexts, but if we need one, it will go in the `tags` package.

Issue #379, opened 2017-06-21
Avoid sharing Java packages between API and implementation libraries.


sebright:
Using different packages for implementation classes will make it easier to avoid accessing non-public elements in the API library and help maintain backwards compatibility.

sebright:
#483 moved all `tags` implementation classes into the package `io.opencensus.impl.tags` to separate them from the API.  Do you think there is any advantage to also separating the classes from the three implementation libraries, so that we have something like this?

```
opencensus-api:       io.opencensus.tags
opencensus-impl-core: io.opencensus.core.tags
opencensus-impl:      io.opencensus.impl.tags
opencensus-impl-lite: io.opencensus.impl.tags
```
or

```
opencensus-api:       io.opencensus.tags
opencensus-impl-core: io.opencensus.implcore.tags
opencensus-impl:      io.opencensus.impl.tags
opencensus-impl-lite: io.opencensus.impllite.tags
```

bogdandrutu:
I don't think that is needed (separation between packages, the implementation packages are not intended to be shared with other non-internal implementations).

sebright:
My second suggestion was related to https://github.com/census-instrumentation/opencensus-java/pull/343#discussion_r120812070.

PR #390, opened 2017-06-26
Make :jmh work by removing duplicate package-info.

g-easy:
The warning I get is: Encountered duplicate path "io/opencensus/internal/package-info.class" during copy operation configured with DuplicatesStrategy.WARN

I guess a collision between
```
./impl_core/src/main/java/io/opencensus/internal/package-info.java
./api/src/main/java/io/opencensus/internal/package-info.java
```


g-easy:
Done, thanks!

sebright:
#379 would also help us avoid duplicates.

PR #396, opened 2017-06-27
Refactor MeasurementDescriptor.

songy23:
Based on our ongoing discussion on API design, a few updates on MeasurementDescripor:
1. replace all occurrence of MeasurementDescripor with Measure;
2. make Measure a hierarchical class and add LongMeasure and DoubleMeasure subclasses;
3. remove Unit class and use a structured string to represent unit. The format refers to [StackDriver documentation](https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.metricDescriptors).


core/src/main/java/io/opencensus/stats/Measure.java
```diff
@@ -0,0 +1,166 @@
...
+
+import com.google.auto.value.AutoValue;
+import io.opencensus.common.Function;
+import io.opencensus.internal.StringUtil;
+import javax.annotation.concurrent.Immutable;
+
+/**
+ * Measure.
+ *
+ * <p>Note: Measure names are {@link String}s with enforced restrictions.
```
dinooliva: At this point, we don't have any restrictions on measure names.

core/src/main/java/io/opencensus/stats/Measure.java
```diff
@@ -0,0 +1,166 @@
...
+   */
+  public abstract String getUnit();
+
+  /**
+   * The name of a {@code Measure}.
+   */
+  // This type should be used as the key when associating data with Measures.
+  @Immutable
+  @AutoValue
+  public abstract static class Name {
```
dinooliva: I don't think we need this class, unless I'm missing something.

core/src/main/java/io/opencensus/stats/MeasurementMap.java
```diff
@@ -18,34 +18,34 @@
 import java.util.NoSuchElementException;
 
 /**
- * A map from {@link MeasurementDescriptor}'s to measured values.
+ * A map from {@link Measure}'s to measured values.
  */
 public final class MeasurementMap implements Iterable<MeasurementValue> {
```
dinooliva: Consider naming this class MeasureMap

dinooliva: Similarly MeasurementValue should probably be MeasureValue. As discussed, StatsEvent or Measurement may also be reasonable alternatives.

core/src/main/java/io/opencensus/stats/Measure.java
songy23: This is from PR: https://github.com/census-instrumentation/opencensus-java/pull/192. From my understanding the reason we added the `Name` class is mainly for comparing `Measure`. @sebright is that correct?

core/src/main/java/io/opencensus/stats/Measure.java
songy23: Done, removed this TODO

core/src/main/java/io/opencensus/stats/MeasurementMap.java
songy23: This will be done in another PR on refactoring `MeasurementValue`, and that PR will be built on this one. I didn't put all changes together so that I can keep each PR relatively small :p

songy23: Same, will post another PR for this change. For now I'm using `Measurement` as the new name.

core/src/main/java/io/opencensus/stats/Measure.java
sebright: It would be useful to have a type that is like a key to a Measure.  Then Views could specify Measures, but MeasureMaps could contain just the `Measure.Name`s paired with values.  However, now that Measure has a type argument, Name would probably also need a type argument, and I don't know if it would be worth keeping the Name type if it also had to have subclasses for the different measurement value types.

sebright: > This is from PR: #192. From my understanding the reason we added the Name class is mainly for comparing Measure. @sebright is that correct?

Yes, the class contains only the fields that should be compared when comparing Measures.

songy23: I think if we would allow different types of `Measure` to have the same name, we'll need to have typed subclasses on `Name`, but this is not determined yet. Anyway for now I'll go ahead and replace the `Name` class with plain string, we might add it back later if it's necessary.

songy23:
PTAL

core/src/main/java/io/opencensus/stats/Measure.java
```diff
@@ -0,0 +1,114 @@
...
+import io.opencensus.internal.StringUtil;
+import javax.annotation.concurrent.Immutable;
+
+/**
+ * Measure.
+ */
+@Immutable
+public abstract class Measure {
+
+  public static final int MAX_LENGTH = StringUtil.MAX_LENGTH;
```
dinooliva: I don't think we need this any longer.

core/src/main/java/io/opencensus/stats/Measure.java
```diff
@@ -0,0 +1,114 @@
...
+   *     Expression = Component { "." Component } { "/" Component } ;
+   *     Component = [ PREFIX ] UNIT [ Annotation ] | Annotation | "1" ;
+   *     Annotation = "{" NAME "}" ;
+   * For example, string “MBy{transmitted}/ms” stands for megabytes per milliseconds, and the
+   * annotation transmitted inside {} is just a comment of the unit.
+   */
+  public abstract String getUnit();
+
+  @Immutable
+  @AutoValue
```
dinooliva: Without AutoValue, DoubleMesaure would be final - is there any way to make the class final with AutoValue?

core/src/main/java/io/opencensus/stats/RpcMeasurementConstants.java
```diff
@@ -28,125 +23,125 @@
...
+  // Constants used to define the following Measures.
+  private static final String BYTE = "By";
+  private static final String COUNT = "1";
+  private static final String MILLISECOND = "ms";
 
-  // RPC client {@link MeasurementDescriptor}s.
-  public static final MeasurementDescriptor RPC_CLIENT_ERROR_COUNT =
-      MeasurementDescriptor.create(
+  // RPC client {@link Measure}s.
+  public static final Measure RPC_CLIENT_ERROR_COUNT =
```
dinooliva: You should specify that this is a DoubleMeasure (here and below).

core/src/main/java/io/opencensus/stats/Measure.java
songy23: AFAIK classes with AutoValue needs to be abstract and cannot be final, and the best we can do is to make the constructor package-private to prevent it from being extended anywhere else. For examples [DistributionView](https://github.com/census-instrumentation/opencensus-java/blob/master/core/src/main/java/io/opencensus/stats/View.java#L52). @sebright do you know a better way to fix this?

core/src/main/java/io/opencensus/stats/RpcMeasurementConstants.java
songy23: Updated, though I think if we decided to have `LongMeasure` in v0.1, it may make more sense to define *_COUNT constants as `LongMeasure`.

core/src/main/java/io/opencensus/stats/Measure.java
dinooliva: Having the constructor package-private is probably sufficient. We should update the 'Tagged Unions' doc to show the example using autovalue.

core/src/main/java/io/opencensus/stats/RpcMeasurementConstants.java
dinooliva: Agreed - add a TODO() noting this.

core/src/main/java/io/opencensus/stats/Measure.java
songy23: Sure, I'll document an example on AutoValue in the doc.

core/src/main/java/io/opencensus/stats/RpcMeasurementConstants.java
songy23: Done.

dinooliva: You don't need to add a comment for all of them - just make a file-level comment to consider changing *_COUNT DoubleMeasures to LongMeasures

songy23: That's better - just updated.

core/src/main/java/io/opencensus/stats/Measure.java
```diff
@@ -0,0 +1,114 @@
...
+   * Name of measure, as a {@code String}.
+   */
+  public abstract String getName();
+
+  /**
+   * Detailed description of the measure, used in documentation.
+   */
+  public abstract String getDescription();
+
+  /**
```
dinooliva: Is there a link to document where grammar for unit is defined?

songy23: Yes, here's the link to StackDriver documentation (refer to `unit` specification): https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.metricDescriptors#resource-metricdescriptor 

Here's a complete style guide: http://unitsofmeasure.org/ucum.html

core/src/main/java/io/opencensus/stats/Measure.java
sebright: > AFAIK classes with AutoValue needs to be abstract and cannot be final, and the best we can do is to make the constructor package-private to prevent it from being extended anywhere else. For examples DistributionView. @sebright do you know a better way to fix this?

I think that is the best that we can do.

core/src/main/java/io/opencensus/stats/Measure.java
```diff
@@ -0,0 +1,119 @@
+/*
+ * Copyright 2016, Google Inc.
```
songy23: Done.

core/src/main/java/io/opencensus/stats/Measure.java
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * Measure.
+ */
+@Immutable
+public abstract class Measure {
+
+  /**
+   * Applies the given match function to the underlying data type.
+   */
+  public abstract <T> T match(Function<DoubleMeasure, T> p0, Function<LongMeasure, T> p1);
```
songy23: Fixed. That's more flexible.

core/src/main/java/io/opencensus/stats/Measure.java
```diff
@@ -0,0 +1,119 @@
...
+
+  /**
+   * Detailed description of the measure, used in documentation.
+   */
+  public abstract String getDescription();
+
+  /**
+   * The units in which {@link Measure} values are measured.
+   *
+   * <p>The grammar for a unit is as follows:
```
songy23: I've added a TODO here, since for now we are not sure whether we want to perform check on its format.

core/src/test/java/io/opencensus/stats/MeasureTest.java
```diff
@@ -0,0 +1,116 @@
...
+        .addEqualityGroup(
+            Measure.DoubleMeasure.create(
+                "name",
+                "description 2",
+                "bit/s"))
+        .testEquals();
+  }
+
+  @Test
+  public void testLongMeasureEquals() {
```
songy23: For now I added one more test to test DoubleMeasure is not equal to LongMeasure. Will add more tests once we determine whether names should be unique for all types of measures.

sebright:
Thanks

PR #397, opened 2017-06-28
Change RpcMeasurementConstants.RPC_STATUS to "canonical_status".

sebright:
/cc @zhangkun83 @dinooliva

### DIFF
--- a/core/src/main/java/io/opencensus/stats/RpcMeasurementConstants.java
+++ b/core/src/main/java/io/opencensus/stats/RpcMeasurementConstants.java
@@ -24,7 +24,7 @@ import java.util.List;
 public final class RpcMeasurementConstants {
 
   // Rpc tag keys.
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/opencensus/stats/RpcMeasurementConstants.java
```diff
@@ -24,7 +24,7 @@
 public final class RpcMeasurementConstants {
 
   // Rpc tag keys.
-  public static final TagKey RPC_STATUS = TagKey.create("OpStatus");
+  public static final TagKey RPC_STATUS = TagKey.create("canonical_status");
```
adriancole: nag: start javadoc'ing these

sebright: I created an issue: #453

PR #400, opened 2017-06-29
Package the classes that need to be loaded by the bootstrap classloader i…

ubschmidt2:
…n a separate JAR file, bootstrap.jar, and bundle that with the agent's JAR file.

### DIFF
--- a/contrib/agent/build.gradle
+++ b/contrib/agent/build.gradle
@@ -9,6 +9,16 @@ targetCompatibility = 1.7
 
 def agentPackage = 'io.opencensus.contrib.agent'
 def agentMainClass = "${agentPackage}.AgentMain"
+
+// The package containing the classes that need to be loaded by the bootstrap classloader because
+// they are used from classes loaded by the bootstrap classloader.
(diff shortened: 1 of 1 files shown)


contrib/agent/build.gradle
```diff
@@ -43,21 +66,37 @@ shadowJar {
 
   // Exclude cruft which still snuck in.
   exclude 'META-INF/maven/**'
+  exclude agentBootstrapClasses
 
   // Relocate third party packages to avoid any conflicts of the agent's classes with the app's
   // classes, which are loaded by the same classloader (the system classloader).
   relocate 'net.bytebuddy', agentRepackaged + '.bytebuddy'
```
adriancole: bytebuddy has no deps or we assume its deps will not conflict w/ app's classes?

ubschmidt2: Byte Buddy itself has a dependency on org.objectweb.asm, which is relocated by Byte Buddy's build process and bundled with the bytebuddy JAR. Also see https://github.com/raphw/byte-buddy/blob/3736a10c170a415c1027d87a5424aca589169710/byte-buddy/pom.xml#L48

PR #402, opened 2017-06-30
Added experimental support for legacy Java 1.6 JVMs through Retrolambda.

ubschmidt2:


contrib/agent/build.gradle
```diff
@@ -1,11 +1,17 @@
...
 }
 
 description = 'OpenCensus Agent'
 
-sourceCompatibility = 1.7
-targetCompatibility = 1.7
+sourceCompatibility = JavaVersion.VERSION_1_7
+targetCompatibility = JavaVersion.VERSION_1_7
+
+// Experimental support for legacy Java 1.6 JVMs.
```
adriancole: might want to consider adding animal sniffer to ensure later apis are not used

ubschmidt2: IIUIC, the root project takes care of this (animalsniffer) already. Signature for the agent is 
signature 'org.codehaus.mojo.signature:java16:+@signature'

ubschmidt2:
Now fails to build with JDK 7 because Retrolambda wants JDK 8. Investigating...

adriancole:
ah maybe worth a comment saying this.

ubschmidt2:
So Retrolambda itself needs a JDK 8 for doing its magic. It uses $JAVA8_HOME, which I've added to the Travis config.

PR #403, opened 2017-06-30
Append bootstrap.jar to the bootstrap classloader search path.

ubschmidt2:


adriancole:
hmm. while nice to see things in pieces, there's no value in adding guava by itself. Maybe this can be deferred until guava is used? That way, the value taken by special-casing guava can be paid in part by whatever code that shows it necessary.

ubschmidt2:
@adriancole: sounds good. I'm adding a few more bits to this PR, which use Guava, though in a trivial way (Preconditions, ByteStreams).

contrib/agent/src/main/java/io/opencensus/contrib/agent/AgentMain.java
```diff
@@ -51,8 +53,15 @@ private AgentMain() {
    * @see java.lang.instrument
    */
   public static void premain(String agentArgs, Instrumentation inst) throws Exception {
+    checkNotNull(inst);
```
bogdandrutu: checkNotNull usually is better to use it with an error message because the log entry or exception may not be clear.

s/inst/instrumentation
s/checkNotNull(inst)/checkNotNull(instrumentation, "instrumentation")

contrib/agent/src/main/java/io/opencensus/contrib/agent/AgentMain.java
```diff
@@ -51,8 +53,15 @@ private AgentMain() {
    * @see java.lang.instrument
    */
   public static void premain(String agentArgs, Instrumentation inst) throws Exception {
```
bogdandrutu: Maybe preMain is better?

contrib/agent/src/main/java/io/opencensus/contrib/agent/Resources.java
```diff
@@ -0,0 +1,75 @@
...
+  /**
+   * Returns a resource of the given name as a temporary file.
+   *
+   * @param resourceName name of the resource
+   * @return a temporary {@link File} containing a copy of the resource
+   * @throws FileNotFoundException if no resource of the given name is found
+   * @throws IOException if an I/O error occurs
+   */
+  static File getResourceAsTempFile(String resourceName) throws IOException {
+    checkArgument(!Strings.isNullOrEmpty(resourceName));
```
bogdandrutu: Please also add an error message.

contrib/agent/src/main/java/io/opencensus/contrib/agent/AgentMain.java
ubschmidt2: It must be "premain", cf. https://docs.oracle.com/javase/8/docs/api/java/lang/instrument/package-summary.html.

contrib/agent/src/main/java/io/opencensus/contrib/agent/AgentMain.java
ubschmidt2: Done.

contrib/agent/src/main/java/io/opencensus/contrib/agent/Resources.java
ubschmidt2: Done.

PR #404, opened 2017-06-30
Refactor View and ViewDescriptor.

songy23:
Based on our ongoing discussion on API design, a few updates on View and ViewDescriptor:

- Rename ViewDescriptor to View, rename View to ViewData
- Rename ViewDescriptor.TagKeys to View.Dimensions


songy23:
The diff looks too huge now, I'll try to break it down to individual commits.

songy23:
Diff is too huge and difficult to review, will split this PR.

PR #405, opened 2017-06-30
Refactor MeasurementValue and MeasurementMap

songy23:
Based on our ongoing discussion on API design, a few updates on MeasurementValue and MeasurementMap:
- Rename MeasurementValue to Measurement, rename MeasurementMap to MeasureMap.
- Now since we have two types of Measures (Double and Long), Measurement needs to have two typed subclasses too.
- MeasureMap also has different entries to add DoubleMeasure and LongMeasure.


core/src/test/java/io/opencensus/stats/MeasureMapTest.java
```diff
@@ -0,0 +1,120 @@
...
+/**
+ * Tests for {@link MeasureMap}
+ */
+@RunWith(JUnit4.class)
+public class MeasureMapTest {
+
+  @Test
+  public void testOf1() {
+    ImmutableList<Measurement> expected = ImmutableList.of(
+        (Measurement) Measurement.DoubleMeasurement.create(M1, 44.4));
```
sebright: Is there any way that we can avoid the cast when adding Measurements to a collection?

core/src/test/java/io/opencensus/stats/MeasureMapTest.java
```diff
@@ -0,0 +1,120 @@
...
+    MeasureMap.Builder builder = MeasureMap.builder();
+    for (int i = 1; i <= 10; i++) {
+      builder.put(makeSimpleDoubleMeasure("m" + i), i * 11.1);
+      assertThat(builder.build()).hasSize(i);
+    }
+  }
+
+  private static final DoubleMeasure M1 = makeSimpleDoubleMeasure("m1");
+  private static final DoubleMeasure M2 = makeSimpleDoubleMeasure("m2");
+  private static final DoubleMeasure M3 = makeSimpleDoubleMeasure("m3");
```
sebright: Should one of these be a LongMeasure, to test the class more thoroughly?

core_impl/src/test/java/io/opencensus/stats/StatsContextTest.java
```diff
@@ -132,8 +136,8 @@ public Void apply(IntervalView view) {
     StatsContext context =
         defaultStatsContext.with(
             RpcMeasurementConstants.RPC_CLIENT_METHOD, TagValue.create("myMethod"));
-    MeasurementMap measurements =
-        MeasurementMap.of(RpcMeasurementConstants.RPC_CLIENT_ROUNDTRIP_LATENCY, 5.1);
+    MeasureMap measurements =
+        MeasureMap.of((DoubleMeasure) RpcMeasurementConstants.RPC_CLIENT_ROUNDTRIP_LATENCY, 5.1);
```
sebright: The RPC Measure constants should use the subclass types so that this cast isn't necessary, though that should be a separate change.

core_impl/src/main/java/io/opencensus/stats/MeasureToViewMap.java
```diff
@@ -90,12 +90,19 @@ synchronized void registerView(ViewDescriptor viewDescriptor, Clock clock) {
...
       Collection<MutableView> views =
-          mutableMap.get(mv.getMeasurement().getName());
+          mutableMap.get(measurement.getMeasure().getName());
       for (MutableView view : views) {
-        view.record(tags, mv.getValue());
+        if (measurement.getValue() instanceof Double) {
+          view.record(tags, (Double) measurement.getValue());
+        } else if (measurement.getValue() instanceof Long) {
+          // TODO: determine if we want to support LongMeasure in v0.1
+          throw new UnsupportedOperationException("Long measurements not supported.");
```
sebright: I think we should have a test for this case.

core/src/main/java/io/opencensus/stats/MeasureMap.java
```diff
@@ -0,0 +1,190 @@
...
+  public static MeasureMap of(DoubleMeasure measure, double value) {
+    return builder().put(measure, value).build();
+  }
+
+  /**
+   * Constructs a {@link MeasureMap} from the given {@link DoubleMeasure}'s
+   * and associated values.
+   */
+  public static MeasureMap of(DoubleMeasure measure1, double value1,
+      DoubleMeasure measure2, double value2) {
```
songy23: I agree, and I think now that we have different Measure types, maybe it's better to replace the static method `of()` with `with()`, similar to [StatsContext.with()](https://github.com/census-instrumentation/opencensus-java/blob/master/core/src/main/java/io/opencensus/stats/StatsContext.java#L31).

core/src/main/java/io/opencensus/stats/MeasureMap.java
```diff
@@ -0,0 +1,127 @@
...
+  public static class Builder {
+    /**
+     * Associates the {@link DoubleMeasure} with the given value. Subsequent updates to the
+     * same {@link DoubleMeasure} are ignored.
+     *
+     * @param measure the {@link DoubleMeasure}
+     * @param value the value to be associated with {@code measure}
+     * @return this
+     */
+    public Builder put(DoubleMeasure measure, double value) {
```
sebright: Should we name this method `set`, for consistency with TagContext's builder?

core/src/test/java/io/opencensus/stats/MeasureMapTest.java
```diff
@@ -0,0 +1,163 @@
...
+  private static final LongMeasure makeSimpleLongMeasure(String measure) {
+    return Measure.LongMeasure.create(
+        measure, measure + " description", "1");
+  }
+
+  private static void assertEquals(
+      Iterator<Measurement> expected, Iterator<Measurement> actual) {
+    while (expected.hasNext() && actual.hasNext()) {
+      Measurement expectedMeasurement = expected.next();
+      Measurement actualMeasurement = actual.next();
```
sebright: I don't think we should compare the element order, since MeasureMap could be implemented with a map. (`containsExactly` would also avoid testing the order.)

core/src/test/java/io/opencensus/stats/MeasureMapTest.java
```diff
@@ -0,0 +1,163 @@
...
+          .add(Measurement.DoubleMeasurement.create(makeSimpleDoubleMeasure("m" + i), i * 11.1));
+      builder.put(makeSimpleDoubleMeasure("m" + i), i * 11.1);
+      assertEquals(expected.iterator(), builder.build().iterator());
+    }
+  }
+
+  @Test
+  public void testDuplicateMeasures() {
+    // DoubleMeasure
+    MeasureMap metrics1 = MeasureMap.builder().put(M1, 1.0).build();
```
sebright: Suggestion: You could split this test into three methods, and then the comments wouldn't be necessary.

core/src/main/java/io/opencensus/stats/MeasureMap.java
songy23: I'll prefer to keep the method name consistent with TagContext. @dinooliva WDYT?

core/src/test/java/io/opencensus/stats/MeasureMapTest.java
```diff
@@ -0,0 +1,163 @@
...
+ */
+@RunWith(JUnit4.class)
+public class MeasureMapTest {
+
+  @Test
+  public void testPutDouble() {
+    ImmutableList<Measurement> expected = ImmutableList.of(
+        (Measurement) Measurement.DoubleMeasurement.create(M1, 44.4));
+    MeasureMap metrics = MeasureMap.builder().put(M1, 44.4).build();
+    assertEquals(expected.iterator(), metrics.iterator());
```
songy23: That's cleaner, I've updated these tests. Thanks.

core/src/test/java/io/opencensus/stats/MeasureMapTest.java
songy23: Done.

core/src/test/java/io/opencensus/stats/MeasureMapTest.java
songy23: That's true. I just realized that after I push the commits, and I've removed this test.

core_impl/src/main/java/io/opencensus/stats/MeasureToViewMap.java
```diff
@@ -90,18 +93,30 @@ synchronized void registerView(ViewDescriptor viewDescriptor, Clock clock) {
...
+    while (iterator.hasNext()) {
+      Measurement measurement = iterator.next();
       Collection<MutableView> views =
-          mutableMap.get(mv.getMeasurement().getName());
+          mutableMap.get(measurement.getMeasure().getName());
+      Object value = measurement.match(new GetDoubleValueFunc(), new GetLongValueFunc());
+
       for (MutableView view : views) {
-        view.record(tags, mv.getValue());
+        if (value instanceof Double) {
```
songy23: Updated.

core/src/test/java/io/opencensus/stats/MeasureMapTest.java
```diff
@@ -77,87 +66,63 @@ public void testBuilder() {
       expected
           .add(Measurement.DoubleMeasurement.create(makeSimpleDoubleMeasure("m" + i), i * 11.1));
       builder.put(makeSimpleDoubleMeasure("m" + i), i * 11.1);
-      assertEquals(expected.iterator(), builder.build().iterator());
+      assertThat(Lists.newArrayList(builder.build().iterator()))
+          .containsExactlyElementsIn(expected);
+      assertEquals(builder.build(), expected.toArray(new Measurement[i]));
```
sebright: Is one of these assertions redundant?

core/src/test/java/io/opencensus/stats/MeasureMapTest.java
```diff
@@ -77,87 +66,63 @@ public void testBuilder() {
...
   private static final DoubleMeasure M1 = makeSimpleDoubleMeasure("m1");
   private static final DoubleMeasure M2 = makeSimpleDoubleMeasure("m2");
   private static final LongMeasure M3 = makeSimpleLongMeasure("m3");
   private static final LongMeasure M4 = makeSimpleLongMeasure("m4");
 
-  private static final DoubleMeasure makeSimpleDoubleMeasure(String measure) {
+  private static final Measurement measurement1 = Measurement.DoubleMeasurement.create(M1, 1.0);
+  private static final Measurement measurement2 = Measurement.DoubleMeasurement.create(M2, 2.0);
+  private static final Measurement measurement3 = Measurement.LongMeasurement.create(M3, 100L);
+  private static final Measurement measurement4 = Measurement.LongMeasurement.create(M4, 200L);
```
sebright: I think these fields should use constant case.

core/src/test/java/io/opencensus/stats/MeasureMapTest.java
```diff
@@ -77,87 +66,63 @@ public void testBuilder() {
...
-    }
-  }
-
-  private static final class GetLongValue implements Function<LongMeasurement, Object> {
-
-    @Override
-    public Long apply(LongMeasurement arg) {
-      return arg.getValue();
-    }
+  private static void assertEquals(MeasureMap metrics, Measurement... measurements) {
```
sebright: I think a name like `assertContains` would be more descriptive now that it isn't comparing for equality.

core/src/main/java/io/opencensus/stats/Measurement.java
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * Immutable representation of a Measure.
+ */
+@Immutable
+public abstract class Measurement {
+  /**
+   * Applies the given match function to the underlying data type.
+   */
+  public abstract <T> T match(Function<DoubleMeasurement, T> p0, Function<LongMeasurement, T> p1);
```
sebright: This match method could also use "? super".

songy23: Correct, fixed.

core/src/main/java/io/opencensus/stats/MeasureMap.java
dinooliva: I originally went with 'put' to be consistent w/ImmutableMap.Builder but I believe being consistent with TagContext's Builder is the right way to go so, for now, we should go with set.

core/src/main/java/io/opencensus/stats/Measurement.java
dinooliva: Agreed.

core/src/main/java/io/opencensus/stats/MeasureMap.java
```diff
@@ -0,0 +1,127 @@
...
+     */
+    public MeasureMap build() {
+      // Note: this makes adding measurements quadratic but is fastest for the sizes of
+      // MeasureMaps that we should see. We may want to go to a strategy of sort/eliminate
+      // for larger MeasureMaps.
+      for (int i = 0; i < measurements.size(); i++) {
+        String current =
+            measurements.get(i).getMeasure().getName();
+        for (int j = i + 1; j < measurements.size(); j++) {
+          if (current.equals(measurements.get(j).getMeasure().getName())) {
```
dinooliva: I believe Measure's now overload equality (they originally did not) so you can simply use the equality operator here.

songy23: Done.

songy23:
Thanks!

PR #406, opened 2017-07-01
Fix sereral issues on Measure.

songy23:


core/src/test/java/io/opencensus/stats/MeasureTest.java
```diff
@@ -110,6 +110,12 @@ public void testLongMeasureEquals() {
         .testEquals();
   }
 
+  @Test
+  public void testDoubleMeasureIsNotEqualToLongMeasure() {
+    assertThat(Measure.DoubleMeasure.create("name", "description", "bit/s"))
+        .isNotEqualTo(Measure.LongMeasure.create("name", "description 2", "bit/s"));
```
sebright: I think it would be better to compare two Measures with the same description.  Even if we decide that recording Measures with the same name is equivalent in `StatsRecorder.record`, `Measure.equal` should compare all fields and the class.

songy23: Sure, I've updated the test and now the measures have exactly same fields.

songy23:
Thanks!

PR #380, opened 2017-06-21
Change API to remove the "Option" logic and keep only Builder.

bogdandrutu:

PR #407, opened 2017-07-01
Change tracer API names to create a SpanBuilder.

bogdandrutu:
This is a followup after https://github.com/census-instrumentation/opencensus-java/pull/380.

* Rollback the change to not accept null for remote parent. It is better for our users to pass null there in case no context received or unable to parse.

* Rename the spanBuilder method that accepts an explicit parent to spanBuilderWithParent, this is consistent with spanBuilderWithRemoteParent.

* NoopSpanBuilder changed to have factory methods.

* Update SpanBuilderImpl factory methods to be consistent with NoopSpanBuilder.

### DIFF
--- a/api/src/main/java/io/opencensus/trace/SpanBuilder.java
+++ b/api/src/main/java/io/opencensus/trace/SpanBuilder.java
@@ -87,7 +87,7 @@ import javax.annotation.Nullable;
  * class MyClass {
  *   private static final Tracer tracer = Tracing.getTracer();
  *   void DoWork(Span parent) {
- *     Span childSpan = tracer.spanBuilderWithParent("MyRootSpan", parent).startSpan();
+ *     Span childSpan = tracer.spanBuilderWithExplicitParent("MyChildSpan", parent).startSpan();
  *     childSpan.addAnnotation("my annotation");
  *     try {
  *       doSomeWork(childSpan); // Manually propagate the new span down the stack.
@@ -146,7 +146,7 @@ public abstract class SpanBuilder {
    * class MyClass {
    *   private static final Tracer tracer = Tracing.getTracer();
    *   void DoWork(Span parent) {
-   *     Span childSpan = tracer.spanBuilderWithParent("MyRootSpan", parent).startSpan();
+   *     Span childSpan = tracer.spanBuilderWithExplicitParent("MyChildSpan", parent).startSpan();
    *     childSpan.addAnnotation("my annotation");
    *     try {
    *       doSomeWork(childSpan); // Manually propagate the new span down the stack.
--- a/api/src/main/java/io/opencensus/trace/SpanBuilder.java
+++ b/api/src/main/java/io/opencensus/trace/SpanBuilder.java
@@ -87,7 +87,7 @@ import javax.annotation.Nullable;
  * class MyClass {
  *   private static final Tracer tracer = Tracing.getTracer();
  *   void DoWork() {
- *     Span span = tracer.spanBuilder(null, "MyRootSpan").startSpan();
+ *     Span span = tracer.spanBuilderWithParent(null, "MyRootSpan").startSpan();
  *     span.addAnnotation("my annotation");
  *     try {
  *       doSomeWork(span); // Manually propagate the new span down the stack.
@@ -146,7 +146,7 @@ public abstract class SpanBuilder {
    * class MyClass {
    *   private static final Tracer tracer = Tracing.getTracer();
    *   void DoWork() {
-   *     Span span = tracer.spanBuilder(null, "MyRootSpan").startSpan();
+   *     Span span = tracer.spanBuilderWithParent(null, "MyRootSpan").startSpan();
    *     span.addAnnotation("my annotation");
    *     try {
    *       doSomeWork(span); // Manually propagate the new span down the stack.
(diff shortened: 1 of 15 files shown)


api/src/main/java/io/opencensus/trace/SpanBuilder.java
```diff
@@ -146,7 +146,7 @@
    * class MyClass {
    *   private static final Tracer tracer = Tracing.getTracer();
    *   void DoWork() {
-   *     Span span = tracer.spanBuilder(null, "MyRootSpan").startSpan();
+   *     Span span = tracer.spanBuilderWithParent(null, "MyRootSpan").startSpan();
```
adriancole: since we are changing apis, some look at style and say declare parameters like a sentence. so if method says "span with parent" then parameters spanName, parent

I think this was from an old book "clean code"

api/src/main/java/io/opencensus/trace/Tracer.java
```diff
@@ -169,51 +173,50 @@ public final SpanBuilder spanBuilder(String name) {
...
+   * when creating a root {@code Span} with a {@code null} parent.
    *
-   * @param parent The parent of the returned Span. If null the {@code SpanBuilder} will build a
-   *     root {@code Span}.
+   * @param parent The parent of the returned Span. If {@code null} the {@code SpanBuilder} will
+   *     build a root {@code Span}.
    * @param name The name of the returned Span.
    * @return a {@code SpanBuilder} to create and start a new {@code Span}.
-   * @throws NullPointerException if {@code name} is null.
+   * @throws NullPointerException if {@code name} is {@code null}.
```
adriancole: generally, we don't need to add docs like this if using nullable annotation anything not nullable and passed null can be assumed to throw a NPE

api/src/main/java/io/opencensus/trace/SpanBuilder.java
bogdandrutu: Done.

api/src/main/java/io/opencensus/trace/Tracer.java
bogdandrutu: Maybe in a future PR. This change does not break compatibility or anything.

bogdandrutu:
PTAL

api/src/main/java/io/opencensus/trace/Tracer.java
```diff
@@ -169,51 +175,50 @@ public final SpanBuilder spanBuilder(String name) {
...
    * <p>This <b>must</b> be used to create a {@code Span} when the parent is in a different process.
    * This is only intended for use by RPC systems or similar.
    *
-   * <p>If no {@link SpanContext} OR fail to parse the {@link SpanContext} on the server side,
-   * users must call this method with an {@link SpanContext#INVALID} remote parent {@code
-   * SpanContext}.
+   * <p>If no {@link SpanContext} OR fail to parse the {@link SpanContext} on the server side, users
+   * must call this method with a {@code null} remote parent {@code SpanContext}.
    *
+   * @param spanName The spanName of the returned Span.
```
sebright: Should this be `@param spanName The name of the returned Span.`?

api/src/main/java/io/opencensus/trace/SpanBuilder.java
```diff
@@ -86,14 +86,14 @@
  * <pre>{@code
  * class MyClass {
  *   private static final Tracer tracer = Tracing.getTracer();
- *   void DoWork() {
- *     Span span = tracer.spanBuilder(null, "MyRootSpan").startSpan();
- *     span.addAnnotation("my annotation");
+ *   void DoWork(Span parent) {
+ *     Span childSpan = tracer.spanBuilderWithParent("MyRootSpan", parent).startSpan();
```
sebright: >  Span childSpan = tracer.spanBuilderWithParent("MyRootSpan", parent).startSpan();

Should the name be changed, now that the span has a parent?

api/src/main/java/io/opencensus/trace/Tracer.java
```diff
@@ -169,51 +175,50 @@ public final SpanBuilder spanBuilder(String name) {
...
-   * @param name The name of the returned Span.
+   * @param spanName The spanName of the returned Span.
+   * @param parent The parent of the returned Span. If {@code null} the {@code SpanBuilder} will
+   *     build a root {@code Span}.
    * @return a {@code SpanBuilder} to create and start a new {@code Span}.
-   * @throws NullPointerException if {@code name} is null.
+   * @throws NullPointerException if {@code spanName} is {@code null}.
    */
-  public abstract SpanBuilder spanBuilder(@Nullable Span parent, String name);
+  public abstract SpanBuilder spanBuilderWithParent(String spanName, @Nullable Span parent);
```
sebright: Shouldn't this be something like `spanBuilderWithExplicitParent`, since `spanBuilder` also creates the SpanBuilder with a parent, if there is one in the Context?

api/src/main/java/io/opencensus/trace/SpanBuilder.java
bogdandrutu: Done.

api/src/main/java/io/opencensus/trace/Tracer.java
bogdandrutu: Done.

api/src/main/java/io/opencensus/trace/Tracer.java
bogdandrutu: Done.

PR #371, opened 2017-06-18
Add fake implementations for Tracer and Span to allow testing.

bogdandrutu:

PR #410, opened 2017-07-04
Add fake Span and SpanBuilder to simplify testing.

bogdandrutu:


adriancole:
Have you considered WannaBeSpan or PoserSpan?

testing/src/main/java/io/opencensus/testing/trace/FakeSpanBuilder.java
```diff
@@ -0,0 +1,85 @@
...
+ *
+ *       // Here can verify the end of the Span:
+ *       // verify(spySpan).end(EndSpanOptions.builder().setStatus(Status.OK).build());
+ *     }
+ *   }
+ * }
+ * </code></pre>
+ */
+@VisibleForTesting
+public class FakeSpanBuilder extends SpanBuilder {
```
adriancole: style, but I often make classes like this static inner types

ex. 
```java
public final class FakeSpan extends Span {
...
  public static final class Builder extends SpanBuilder {
...
}

Comment (bogdandrutu) [inline testing/src/main/java/io/opencensus/testing/trace/FakeSpanBuilder.java]: Done. Not final for easy mock.

File: testing/src/main/java/io/opencensus/testing/trace/MockableSpan.java [hunk 2]
```diff
@@ -0,0 +1,176 @@
...
+ *
+ *       // Here can verify the end of the Span:
+ *       // verify(spySpan).end(EndSpanOptions.builder().setStatus(Status.OK).build());
+ *     }
+ *   }
+ * }
+ * </code></pre>
+ */
+@VisibleForTesting
+public class MockableSpan extends Span {
```
carl-mastrangelo: FYI, all final methods cannot be mocked, and iirc mockito won't even complain about this.  Span exposes 

addAnnotation()
getContext()
getOptions()
and()

Will users want to mock these methods?

sebright:
Does this PR replace #371?

testing/src/main/java/io/opencensus/testing/trace/MockableSpan.java
bogdandrutu: So not necessary because:
end() is just a simplified version of end(EndSpanOptions.DEFAULT) so users can verify on the end(EndSpanOptions) that the right call was made.

addAnnotation(String) is just a simplified version of addAnnotation(String, Collections.emptyMap()) so similar with end users can verify on this method the right String description.

For getOptions and getContext people can mock them by returning the right "MockableSpan" see the generate Span versions.

bogdandrutu:
Correct @sebright

adriancole:
Yeah mocks are a pain (you can use powermock to mock callsite of a final
method, but yeah...) if mock is desired as opposed to an in-memory test
tracing component, likely you may need to unfinal a lot perhaps using a
base type with a package private ctor to seal the class within its package
instead of final.

HailongWen:
Friendly ping.
Should we merge this PR?

testing/src/main/java/io/opencensus/testing/trace/MockableSpan.java
HailongWen: I guess it is fine if since `end()` and `addAnnotation()` are just convenience methods and do not contain details of the implementation.

sebright:
@bogdandrutu Is this PR still needed?

sebright:
Closing for now, since this PR isn't active.  I opened an issue instead: #1195

PR #411, opened 2017-07-04
Add automatic context propagation for Executors.

ubschmidt2:
This change adds the required plumbing for accessing/manipulating the
context from the instrumented code. This completes the underlying infrastructure
for the executor instrumentation, which is also included in this change.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -105,6 +105,7 @@ subprojects {
 
         libraries = [
                 auto_value     : "com.google.auto.value:auto-value:${autoValueVersion}",
+                byte_buddy   : 'net.bytebuddy:byte-buddy:1.7.1',
                 disruptor    : 'com.lmax:disruptor:3.3.6',
                 errorprone   : 'com.google.errorprone:error_prone_annotations:2.0.11',
                 findbugs_annotations : "com.google.code.findbugs:annotations:${findBugsVersion}",
--- a/build.gradle
+++ b/build.gradle
@@ -107,6 +107,7 @@ subprojects {
                 auto_value     : "com.google.auto.value:auto-value:${autoValueVersion}",
                 disruptor    : 'com.lmax:disruptor:3.3.6',
                 errorprone   : 'com.google.errorprone:error_prone_annotations:2.0.11',
+                findbugs_annotations : "com.google.code.findbugs:annotations:${findBugsVersion}",
                 grpc_context : "io.grpc:grpc-context:${grpcContextVersion}",
                 guava        : "com.google.guava:guava:${guavaVersion}",
                 jsr305       : "com.google.code.findbugs:jsr305:${findBugsVersion}",
--- a/contrib/agent/build.gradle
+++ b/contrib/agent/build.gradle
@@ -30,6 +30,9 @@ def agentBootstrapClasses = agentBootstrapPackageDir + '**'
 def agentRepackaged = "${agentPackage}.deps"
 
 dependencies {
+  compileOnly libraries.grpc_context
+  testCompile libraries.grpc_context
+  compile group: 'com.google.code.findbugs', name: 'annotations', version: '3.0.1u2'
   compile libraries.guava
   compile group: 'net.bytebuddy', name: 'byte-buddy', version: '1.7.1'
 
@@ -119,3 +122,9 @@ shadowJar {
 jar.finalizedBy shadowJar
 
 // TODO(stschmidt): Proguard-shrink the agent JAR.
+
+// TODO(stschmidt): Move the integration tests to a separate SourceSet once
+// https://github.com/johnrengelman/shadow/issues/297 is fixed.
+test {
+  jvmArgs "-javaagent:${shadowJar.archivePath}"
+}
--- a/contrib/agent/build.gradle
+++ b/contrib/agent/build.gradle
@@ -34,7 +34,7 @@ dependencies {
   testCompile libraries.grpc_context
   compile libraries.findbugs_annotations
   compile libraries.guava
-  compile group: 'net.bytebuddy', name: 'byte-buddy', version: '1.7.1'
+  compile libraries.byte_buddy
 
   signature 'org.codehaus.mojo.signature:java16:+@signature'
 }
--- a/contrib/agent/build.gradle
+++ b/contrib/agent/build.gradle
@@ -32,7 +32,7 @@ def agentRepackaged = "${agentPackage}.deps"
 dependencies {
   compileOnly libraries.grpc_context
   testCompile libraries.grpc_context
-  compile group: 'com.google.code.findbugs', name: 'annotations', version: '3.0.1u2'
+  compile libraries.findbugs_annotations
   compile libraries.guava
   compile group: 'net.bytebuddy', name: 'byte-buddy', version: '1.7.1'
 
--- a/contrib/agent/src/main/java/io/opencensus/contrib/agent/AgentMain.java
+++ b/contrib/agent/src/main/java/io/opencensus/contrib/agent/AgentMain.java
@@ -94,7 +94,7 @@ public final class AgentMain {
     static AgentBuilder addGrpcContextPropagation(AgentBuilder agentBuilder) {
       // TODO(stschmidt): Gracefully handle the case of missing io.grpc.Context at runtime.
 
-      // Initialize the ContextManager singleton with the concrete GrpcContextStrategy.
+      // Initialize the ContextManager with the concrete GrpcContextStrategy.
       ContextManager.setContextStrategy(new GrpcContextStrategy());
 
       // Add automatic context propagation to Executor#execute.
--- a/contrib/agent/src/main/java/io/opencensus/contrib/agent/AgentMain.java
+++ b/contrib/agent/src/main/java/io/opencensus/contrib/agent/AgentMain.java
@@ -14,8 +14,11 @@
 package io.opencensus.contrib.agent;
 
 import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
 import static net.bytebuddy.matcher.ElementMatchers.none;
 
+import io.opencensus.contrib.agent.bootstrap.ContextManager;
+import io.opencensus.contrib.agent.bootstrap.ContextStrategy;
 import java.lang.instrument.Instrumentation;
 import java.util.jar.JarFile;
 import java.util.logging.Logger;
@@ -62,14 +65,48 @@ public final class AgentMain {
     instrumentation.appendToBootstrapClassLoaderSearch(
             new JarFile(Resources.getResourceAsTempFile("bootstrap.jar")));
 
+    assertIsLoadedByBootstrapClassloader(ContextManager.class);
+    assertIsLoadedByBootstrapClassloader(ContextStrategy.class);
+
     AgentBuilder agentBuilder = new AgentBuilder.Default()
             .disableClassFormatChanges()
             .with(AgentBuilder.RedefinitionStrategy.RETRANSFORMATION)
             .with(new AgentBuilderListener())
             .ignore(none());
-    // TODO(stschmidt): Add instrumentation for context propagation.
+    agentBuilder = LazyLoaded.addGrpcContextPropagation(agentBuilder);
+    // TODO: Add more instrumentation, potentially introducing a plugin mechanism.
     agentBuilder.installOn(instrumentation);
 
     logger.info("Initialized.");
+  }
+
+  private static void assertIsLoadedByBootstrapClassloader(Class<?> clazz) {
+    checkNotNull(clazz, "clazz");
+
+    checkState(clazz.getClassLoader() == null,
+            "%s must be loaded by the bootstrap classloader",
+            clazz);
+  }
+
+  private static class LazyLoaded {
+
+    /**
+     * Adds automatic context propagation.
+     */
+    static AgentBuilder addGrpcContextPropagation(AgentBuilder agentBuilder) {
+      checkNotNull(agentBuilder, "agentBuilder");
+
+      // TODO(stschmidt): Gracefully handle the case of missing io.grpc.Context at runtime.
+
+      ContextManager.setContextHandler(new GrpcContextStrategy());
+
+      agentBuilder = agentBuilder
+              .type(ExecutorInstrumentation.matcher())
+              .transform(ExecutorInstrumentation.transformer());
+
+      // TODO(stschmidt): Add instrumentation for {@link Thread}.
+
+      return agentBuilder;
+    }
   }
 }
--- a/contrib/agent/src/main/java/io/opencensus/contrib/agent/AgentMain.java
+++ b/contrib/agent/src/main/java/io/opencensus/contrib/agent/AgentMain.java
@@ -60,8 +60,9 @@ public final class AgentMain {
 
     logger.info("Initializing.");
 
-    // The classes in bootstrap.jar will be referenced from classes loaded by the bootstrap
-    // classloader. Thus, these classes have to be loaded by the bootstrap classloader, too.
+    // The classes in bootstrap.jar, such as ContextManger and ContextStrategy, will be referenced
+    // from classes loaded by the bootstrap classloader. Thus, these classes have to be loaded by
+    // the bootstrap classloader, too.
     instrumentation.appendToBootstrapClassLoaderSearch(
             new JarFile(Resources.getResourceAsTempFile("bootstrap.jar")));
 
--- a/contrib/agent/src/main/java/io/opencensus/contrib/agent/AgentMain.java
+++ b/contrib/agent/src/main/java/io/opencensus/contrib/agent/AgentMain.java
@@ -94,8 +94,8 @@ public final class AgentMain {
     static AgentBuilder addContextPropagation(AgentBuilder agentBuilder) {
       // TODO(stschmidt): Gracefully handle the case of missing io.grpc.Context at runtime.
 
-      // Initialize the ContextManager with the concrete GrpcContextStrategy.
-      ContextManager.setContextStrategy(new GrpcContextStrategy());
+      // Initialize the ContextManager with the concrete ContextStrategy.
+      ContextManager.setContextStrategy(new ContextStrategyImpl());
 
       // Add automatic context propagation to Executor#execute.
       agentBuilder = agentBuilder
--- a/contrib/agent/src/main/java/io/opencensus/contrib/agent/AgentMain.java
+++ b/contrib/agent/src/main/java/io/opencensus/contrib/agent/AgentMain.java
@@ -99,8 +99,8 @@ public final class AgentMain {
 
       // Add automatic context propagation to Executor#execute.
       agentBuilder = agentBuilder
-              .type(ExecutorInstrumentation.matcher())
-              .transform(ExecutorInstrumentation.transformer());
+              .type(ExecutorInstrumentation.createMatcher())
+              .transform(ExecutorInstrumentation.createTransformer());
 
       // TODO(stschmidt): Add automatic context propagation to Thread#start.
 
--- a/contrib/agent/src/main/java/io/opencensus/contrib/agent/AgentMain.java
+++ b/contrib/agent/src/main/java/io/opencensus/contrib/agent/AgentMain.java
@@ -73,7 +73,7 @@ public final class AgentMain {
             .with(AgentBuilder.RedefinitionStrategy.RETRANSFORMATION)
             .with(new AgentBuilderListener())
             .ignore(none());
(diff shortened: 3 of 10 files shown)


contrib/agent/src/main/java/io/opencensus/contrib/agent/AgentMain.java
```diff
@@ -62,14 +65,48 @@ public static void premain(String agentArgs, Instrumentation instrumentation) th
...
+     * Adds automatic context propagation.
+     */
+    static AgentBuilder addGrpcContextPropagation(AgentBuilder agentBuilder) {
+      checkNotNull(agentBuilder, "agentBuilder");
+
+      // TODO(stschmidt): Gracefully handle the case of missing io.grpc.Context at runtime.
+
+      ContextManager.setContextHandler(new GrpcContextStrategy());
+
+      agentBuilder = agentBuilder
```
adriancole: maybe comment what this does (as other readers may not be aware of the voodoo :) )

contrib/agent/src/main/java/io/opencensus/contrib/agent/ExecutorInstrumentation.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @Override
+    public DynamicType.Builder<?> transform(DynamicType.Builder<?> builder,
+            TypeDescription typeDescription, ClassLoader classLoader, JavaModule module) {
+      return builder.visit(Advice.to(Execute.class).on(named("execute")));
+    }
+  }
+
+  static ElementMatcher.Junction<TypeDescription> matcher() {
+    // TODO(stschmidt): Exclude known call sites that already propagate the context.
```
adriancole: maybe also exclude types if grpc context has a publicly visible type, otherwise your comment is fine

contrib/agent/src/main/java/io/opencensus/contrib/agent/ExecutorInstrumentation.java
```diff
@@ -0,0 +1,66 @@
...
+
+    return isSubTypeOf(Executor.class).and(not(isAbstract()));
+  }
+
+  static AgentBuilder.Transformer transformer() {
```
adriancole: nit: create?

contrib/agent/src/main/java/io/opencensus/contrib/agent/ExecutorInstrumentation.java
```diff
@@ -0,0 +1,66 @@
...
+    return new Transformer();
+  }
+
+  private static class Execute {
+
+    @Advice.OnMethodEnter
+    @SuppressWarnings(value = "UnusedAssignment")
```
adriancole: this requires intimate knowledge of why you can re-assign a parameter ref to something else, not return it, and have it actually used :) Is there a short comment that can give a jist of why (or maybe a javadoc or http link)

contrib/agent/src/main/java/io/opencensus/contrib/agent/AgentMain.java
```diff
@@ -62,14 +65,48 @@ public static void premain(String agentArgs, Instrumentation instrumentation) th
...
+
+    /**
+     * Adds automatic context propagation.
+     */
+    static AgentBuilder addGrpcContextPropagation(AgentBuilder agentBuilder) {
+      checkNotNull(agentBuilder, "agentBuilder");
+
+      // TODO(stschmidt): Gracefully handle the case of missing io.grpc.Context at runtime.
+
+      ContextManager.setContextHandler(new GrpcContextStrategy());
```
adriancole: any way to not use static state here?

contrib/agent/src/main/java/io/opencensus/contrib/agent/bootstrap/ContextManager.java
```diff
@@ -0,0 +1,54 @@
...
+ * {@link ContextStrategy} interface.
+ *
+ * <p>Both {@link ContextManager} and {@link ContextStrategy} are loaded by the bootstrap
+ * classloader so that they can be used from classes loaded by the bootstrap classloader.
+ * A concrete implementations of {@link ContextStrategy} will be loaded by the system classloader.
+ * This allows for using the same context implementation as the instrumented application.
+ */
+public final class ContextManager {
+
+  private static ContextStrategy contextStrategy;
```
adriancole: comment why no synchronization needed? (plus earlier comment on why static ref is needed)

contrib/agent/src/test/java/io/opencensus/contrib/agent/ExecutorInstrumentationTest.java
```diff
@@ -0,0 +1,148 @@
...
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/**
+ * Integration tests for {@link ExecutorInstrumentation}.
+ */
+@RunWith(JUnit4.class)
```
adriancole: maybe comment how the agent is attached here (if not already commented in package javadoc, or maybe even if it is commented in package javadoc). might eliminate questions later if test doesn't pass

contrib/agent/src/main/java/io/opencensus/contrib/agent/AgentMain.java
```diff
@@ -62,14 +65,48 @@ public static void premain(String agentArgs, Instrumentation instrumentation) th
...
             .ignore(none());
-    // TODO(stschmidt): Add instrumentation for context propagation.
+    agentBuilder = LazyLoaded.addGrpcContextPropagation(agentBuilder);
+    // TODO: Add more instrumentation, potentially introducing a plugin mechanism.
     agentBuilder.installOn(instrumentation);
 
     logger.info("Initialized.");
   }
+
+  private static void assertIsLoadedByBootstrapClassloader(Class<?> clazz) {
```
ubschmidt2: SGTM. Done.

contrib/agent/src/main/java/io/opencensus/contrib/agent/AgentMain.java
```diff
@@ -62,14 +65,48 @@ public static void premain(String agentArgs, Instrumentation instrumentation) th
...
+    checkNotNull(clazz, "clazz");
```
ubschmidt2: Done.

contrib/agent/src/main/java/io/opencensus/contrib/agent/AgentMain.java
```diff
@@ -62,14 +65,48 @@ public static void premain(String agentArgs, Instrumentation instrumentation) th
...
+            clazz);
+  }
+
+  private static class LazyLoaded {
+
+    /**
+     * Adds automatic context propagation.
+     */
+    static AgentBuilder addGrpcContextPropagation(AgentBuilder agentBuilder) {
+      checkNotNull(agentBuilder, "agentBuilder");
```
ubschmidt2: Done.

contrib/agent/src/main/java/io/opencensus/contrib/agent/AgentMain.java
ubschmidt2: Going to add sth. to the javadoc of ContextManager to explain why it is implemented as a singleton.

contrib/agent/src/main/java/io/opencensus/contrib/agent/AgentMain.java
ubschmidt2: Done. (high level only, not going into details here)

contrib/agent/src/main/java/io/opencensus/contrib/agent/ExecutorInstrumentation.java
ubschmidt2: Hmm. I'm not sure what you mean. Leaving as is for now.

contrib/agent/src/main/java/io/opencensus/contrib/agent/ExecutorInstrumentation.java
ubschmidt2: Not sure. Leaving this as is for now. Happy to change it if there are strong feelings about this.

contrib/agent/src/main/java/io/opencensus/contrib/agent/ExecutorInstrumentation.java
ubschmidt2: Added some Javadoc.

contrib/agent/src/main/java/io/opencensus/contrib/agent/ExecutorInstrumentation.java
adriancole: I guess I'm wondering if there are cases where the match expression will double-instrument and you could know by looking at it. In many cases, you can't (ex often the instrumented executor is a hidden type)

contrib/agent/src/main/java/io/opencensus/contrib/agent/bootstrap/ContextManager.java
ubschmidt2: Added more comments. Let me know if it doesn't make sense ;-)

adriancole:
>
> Added more comments. Let me know if it doesn't make sense ;-)
>
did you push a commit?

contrib/agent/src/test/java/io/opencensus/contrib/agent/ExecutorInstrumentationTest.java
ubschmidt2: Done.

contrib/agent/src/main/java/io/opencensus/contrib/agent/AgentMain.java
ubschmidt2: (it's not a singleton, of course)

ubschmidt2:
Pushed additional commits now.

contrib/agent/src/main/java/io/opencensus/contrib/agent/ExecutorInstrumentation.java
ubschmidt2: Yes, exactly, the hidden types in io.grpc.Context#currentContextExecutor and io.grpc.Context#fixedContextExecutor are examples. I think it should be possible to match those, but for now I'm leaving any attempts to do so to a separate change.

contrib/agent/src/main/java/io/opencensus/contrib/agent/GrpcContextStrategy.java
```diff
@@ -0,0 +1,29 @@
...
+package io.opencensus.contrib.agent;
+
+import io.grpc.Context;
+import io.opencensus.contrib.agent.bootstrap.ContextStrategy;
+
+/**
+ * Implementation of {@link ContextStrategy} for accessing and manipulating the
+ * {@link io.grpc.Context}.
+ */
+final class GrpcContextStrategy implements ContextStrategy {
```
bogdandrutu: Is the intention to support other Context implementations? I would not use the name gRPC here. Probably make ContextStrategy an implementation that is this class.

contrib/agent/src/main/java/io/opencensus/contrib/agent/AgentMain.java
```diff
@@ -62,14 +65,46 @@ public static void premain(String agentArgs, Instrumentation instrumentation) th
...
+            "%s must be loaded by the bootstrap classloader",
+            clazz);
+  }
+
+  private static class LazyLoaded {
+
+    /**
+     * Adds automatic context propagation.
+     */
+    static AgentBuilder addGrpcContextPropagation(AgentBuilder agentBuilder) {
```
bogdandrutu: I think we should not name it grpc, also an open issue to move context into a separate package which does not contain gRPC is in progress (discussions around that subject for the moment).

contrib/agent/src/main/java/io/opencensus/contrib/agent/ExecutorInstrumentation.java
bogdandrutu: I think create is a better name.

contrib/agent/src/main/java/io/opencensus/contrib/agent/AgentMain.java
ubschmidt2: Done.

contrib/agent/src/main/java/io/opencensus/contrib/agent/ExecutorInstrumentation.java
ubschmidt2: Done.

contrib/agent/src/main/java/io/opencensus/contrib/agent/GrpcContextStrategy.java
ubschmidt2: Main reason for the separation is to not ship a copy of the context class with agent (and load it from the bootstrap classloader), but instead use the context class that is loaded from the system classloader (the one that comes with the to-be-instrumented application).

The separation would allow for other implementations of Context, but this is more a side effect of the separation, not the primary intention. I'm not aware of a good example of other (resusable) implementations of a context class.

I've renamed the class to ContextStrategyImpl.

ubschmidt2:
PTAL.

contrib/agent/src/main/java/io/opencensus/contrib/agent/AgentMain.java
```diff
@@ -62,14 +65,46 @@ public static void premain(String agentArgs, Instrumentation instrumentation) th
     instrumentation.appendToBootstrapClassLoaderSearch(
             new JarFile(Resources.getResourceAsTempFile("bootstrap.jar")));
 
+    checkLoadedByBootstrapClassloader(ContextManager.class);
```
adriancole: Maybe comment why these need to be on boot classloader

ubschmidt2: Done. Mentioned the classes in the existing comment a few lines above.

contrib/agent/build.gradle
```diff
@@ -30,6 +30,9 @@ def agentBootstrapClasses = agentBootstrapPackageDir + '**'
 def agentRepackaged = "${agentPackage}.deps"
 
 dependencies {
+  compileOnly libraries.grpc_context
+  testCompile libraries.grpc_context
+  compile group: 'com.google.code.findbugs', name: 'annotations', version: '3.0.1u2'
```
bogdandrutu: We use 'com.google.code.findbugs:jsr305:3.0.0' so probably we should use 3.0.0 here as well and be consistent. In order to keep consistency an variable for the findbugs_version in the main build file may help (similar with grpcContextVersion).

Also probably better to define the these in the libraries in the main build. Same probably for byte-buddy.

contrib/agent/src/main/java/io/opencensus/contrib/agent/AgentMain.java
```diff
@@ -57,19 +60,52 @@ public static void premain(String agentArgs, Instrumentation instrumentation) th
...
+            "%s must be loaded by the bootstrap classloader",
+            clazz);
+  }
+
+  private static class LazyLoaded {
+
+    /**
+     * Adds automatic context propagation.
+     */
+    static AgentBuilder addContextPropagation(AgentBuilder agentBuilder) {
```
bogdandrutu: This can be private because the class is private anyway.

contrib/agent/src/main/java/io/opencensus/contrib/agent/AgentMain.java
```diff
@@ -57,19 +60,52 @@ public static void premain(String agentArgs, Instrumentation instrumentation) th
...
+  private static void checkLoadedByBootstrapClassloader(Class<?> clazz) {
+    checkState(clazz.getClassLoader() == null,
+            "%s must be loaded by the bootstrap classloader",
+            clazz);
+  }
+
+  private static class LazyLoaded {
+
+    /**
+     * Adds automatic context propagation.
```
bogdandrutu: We try to use normal comments if needed for private methods/classes. No need to have incomplete javadoc (missing param and return).

contrib/agent/src/main/java/io/opencensus/contrib/agent/ExecutorInstrumentation.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @Override
+    public DynamicType.Builder<?> transform(DynamicType.Builder<?> builder,
+            TypeDescription typeDescription, ClassLoader classLoader, JavaModule module) {
+      return builder.visit(Advice.to(Execute.class).on(named("execute")));
+    }
+  }
+
+  static ElementMatcher.Junction<TypeDescription> createMatcher() {
+    // TODO(stschmidt): Exclude known call sites that already propagate the context.
```
bogdandrutu: You can add as an example the Executor implementation from gRPC.

contrib/agent/src/main/java/io/opencensus/contrib/agent/bootstrap/ContextManager.java
```diff
@@ -0,0 +1,73 @@
...
+   * Sets the concrete strategy for accessing and manipulating the context.
+   *
+   * <p>NB: The agent is responsible for setting the context strategy once before any other method
+   * of this class is called.
+   *
+   * @param contextStrategy the concrete strategy for accessing and manipulating the context
+   */
+  public static void setContextStrategy(ContextStrategy contextStrategy) {
+    if (ContextManager.contextStrategy != null) {
+      throw new IllegalStateException("contextStrategy was already set");
```
bogdandrutu: This seems to be the way how we handle the missing grpc context dependency at runtime. Probably document in the TODO.

contrib/agent/src/main/java/io/opencensus/contrib/agent/AgentMain.java
ubschmidt2: Done.

contrib/agent/src/main/java/io/opencensus/contrib/agent/AgentMain.java
ubschmidt2: Done.

contrib/agent/src/main/java/io/opencensus/contrib/agent/ExecutorInstrumentation.java
ubschmidt2: Done.

contrib/agent/src/main/java/io/opencensus/contrib/agent/bootstrap/ContextManager.java
ubschmidt2: No, if the io.grpc.Context.class is missing at runtime, then the agent will not instrument the bytecode for automatic context propagation at all. At least that's the plan. There's a TODO about that in AgentMain.LazyLoaded#addContextPropagation: // TODO(stschmidt): Gracefully handle the case of missing io.grpc.Context at runtime.

contrib/agent/build.gradle
ubschmidt2: I'll make a separate PR for the findbugs version. The root buildscript uses two artifacts, the jsr305 one (3.0.0) and another one for the tool (3.0.1). I guess there's no reason for using a different version.

The jar referenced here is yet another one, for the SuppressFBWarnings annotation, which then should also use the same version.

ubschmidt2: Done. Now both libraries are defined in the root buildscript.

build.gradle
```diff
@@ -105,8 +105,10 @@ subprojects {
 
         libraries = [
                 auto_value     : "com.google.auto.value:auto-value:${autoValueVersion}",
+                byte_buddy   : 'net.bytebuddy:byte-buddy:1.7.1',
                 disruptor    : 'com.lmax:disruptor:3.3.6',
                 errorprone   : 'com.google.errorprone:error_prone_annotations:2.0.11',
+                findbugs_annotations : "com.google.code.findbugs:annotations:${findBugsVersion}",
```
bogdandrutu: Do we really need 3.0.1? Can we use 3.0.0, because gRPC uses 3.0.0 as well.

ubschmidt2: Good question. In the grpc-java repo, I see that findbugs jsr305 is used (version 3.0.0), https://github.com/grpc/grpc-java/search?q=findbugs, but no other reference to findbugs (the tool or the annotations for suppressing FB warnings), this puzzles me a bit, but well....

WRT JSR305, there don't seem to be any API-relevant changes between 3.0.0 and 3.0.1 (https://abi-laboratory.pro/java/tracker/timeline/jsr305/). Since opencensus-java is using tool version 3.0.1 (the current version), I figured that it would be best to use the same version for jsr305 and annotations. I'm leaving this as is for now, but happy to send a follow-up PR. PLKM.

PR #413, opened 2017-07-05
Clarify the nature of the Link.Type enum

bogdandrutu:


adriancole:
I guess the otherwise wordy sounding child-linked-span on a span's links
hints at there is a non span type of link happening? Also is it needed to
add the word linked? Isnt the type already telling you that? If not, maybe
link isnt the right type name...

PR #416, opened 2017-07-05
Move logging exporter into an independent package.

bogdandrutu:
Main reason is that we do not want to ship this with the API package and also we understood that is probably better to define a common text format of the data between languages for logs, this way users may parse logs and construct traces even if the data are generated by different languages.

This package will not be shipped in 0.5.0 because the text format is not defined.


exporters/logging/build.gradle
```diff
@@ -0,0 +1,10 @@
+description = 'OpenCensus Logging Exporter'
```
sebright: Is this library specific to tracing, or do you think we should also provide logging for stats?

bogdandrutu: probably all the exporters will be specific to trace or stats. Maybe name directories trace_logging and the package opencensus-trace-logging-exporter?

PR #417, opened 2017-07-06
Get rid of base directory and move everything on level up.

bogdandrutu:


api/src/test/java/io/opencensus/trace/AnnotationTest.java
```diff
@@ -11,11 +11,14 @@
  * limitations under the License.
  */
 
-package io.opencensus.trace.base;
+package io.opencensus.trace;
 
 import static com.google.common.truth.Truth.assertThat;
 
 import com.google.common.testing.EqualsTester;
+import io.opencensus.trace.Annotation;
```
g-easy: How come these weren't needed before but are needed now that AnnotationTest is in the same package?

bogdandrutu: Done. I need to understand why errorprone does not work with tests.

PR #418, opened 2017-07-06
Rename View to ViewData.

songy23:
Rename: View -> ViewData, MutableView -> MutableViewData

This PR only contains renaming stuff, so I believe it doesn't conflict with our ongoing discussion on View refactoring.

### DIFF
--- a/core/src/main/java/io/opencensus/stats/ViewData.java
+++ b/core/src/main/java/io/opencensus/stats/ViewData.java
@@ -27,9 +27,9 @@ import javax.annotation.concurrent.Immutable;
  * The aggregated data for a particular {@link ViewDescriptor}.
  */
 @Immutable
-public abstract class View {
+public abstract class ViewData {
   /**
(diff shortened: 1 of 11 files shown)


songy23:
Rebased, PTAL

core/src/main/java/io/opencensus/stats/ViewData.java
```diff
@@ -27,36 +27,36 @@
...
    */
   public abstract ViewDescriptor getViewDescriptor();
 
   /**
    * Applies the given match function to the underlying data type.
    */
   public abstract <T> T match(
-      Function<DistributionView, T> p0,
-      Function<IntervalView, T> p1);
+      Function<DistributionViewData, T> p0,
```
dinooliva: Probably worth changing to ? super ... too

songy23: Nice catch, updated.

PR #421, opened 2017-07-06
Move ContextUtils to io.opencensus.trace.unsafe package.

sebright:
/cc @songy23 @dinooliva

### DIFF
--- a/api/src/main/java/io/opencensus/trace/unsafe/ContextUtils.java
+++ b/api/src/main/java/io/opencensus/trace/unsafe/ContextUtils.java
@@ -14,7 +14,6 @@
 package io.opencensus.trace.unsafe;
 
 import io.grpc.Context;
-import io.opencensus.common.NonThrowingCloseable;
 import io.opencensus.trace.Span;
 import io.opencensus.trace.Tracer;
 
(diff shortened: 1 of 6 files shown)


api/src/main/java/io/opencensus/trace/unsafe/ContextUtils.java
```diff
@@ -11,10 +11,12 @@
  * limitations under the License.
  */
 
-package io.opencensus.trace;
+package io.opencensus.trace.unsafe;
 
 import io.grpc.Context;
 import io.opencensus.common.NonThrowingCloseable;
+import io.opencensus.trace.Span;
+import io.opencensus.trace.Tracer;
```
bogdandrutu: This is here just because of the comments. Maybe use io.opencensus.Tracer in the comments.

bogdandrutu:
You have to make methods public. It is fine because this is in an unsafe package.

api/src/main/java/io/opencensus/trace/unsafe/ContextUtils.java
sebright: Done

sebright:
Sorry, I forgot to build the branch.  I split ContextUtils into two classes to avoid exposing the internal methods.  Is that okay, or would you rather also expose the other utility methods?  I added this commit:

#### Put CONTEXT_SPAN_KEY in a separate class, to avoid exposing methods in ContextUtils.

After this commit, ContextUtils is split into two classes,
io.opencensus.trace.unsafe.ContextUtils and
io.opencensus.trace.CurrentSpanUtils.  ContextUtils has one public member,
CONTEXT_SPAN_KEY.  CurrentSpanUtils has internal methods for working with the
current Span, and it is package-private.

api/src/main/java/io/opencensus/trace/unsafe/ContextUtils.java
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * <p>Users must interact with the current Context via the public APIs in {@link
+ * io.opencensus.trace.Tracer} and avoid usages of the {@link #CONTEXT_SPAN_KEY} directly.
+ */
+public final class ContextUtils {
+  // No instance of this class.
+  private ContextUtils() {}
+
+  /** The {@link io.grpc.Context.Key} used to interact with {@link io.grpc.Context}. */
+  public static final Context.Key<Span> CONTEXT_SPAN_KEY = Context.key("instrumentation-trace-key");
```
bogdandrutu: Maybe rename the key to:
opencensus-trace-key

bogdandrutu: maybe: opencensus-trace-span-key?

sebright: I renamed it to the second one.

PR #425, opened 2017-07-06
Add attributes to the Link to match with the data model.

bogdandrutu:


api/src/main/java/io/opencensus/trace/Link.java
```diff
@@ -44,7 +49,24 @@
    * @return a new {@code Link}.
    */
   public static Link fromSpanContext(SpanContext context, Type type) {
-    return new AutoValue_Link(context.getTraceId(), context.getSpanId(), type);
+    return new AutoValue_Link(context.getTraceId(), context.getSpanId(), type, EMPTY_ATTRIBUTES);
```
songy23: s/`return fromSpanContext(context, type, EMPTY_ATTRIBUTES);` ?

bogdandrutu: Unnecessary copy of the empty map.

api/src/main/java/io/opencensus/trace/Link.java
```diff
@@ -44,7 +49,24 @@
...
+   * @param attributes the attributes of the {@code Link}.
+   * @return a new {@code Link}.
+   */
+  public static Link fromSpanContext(
+      SpanContext context, Type type, Map<String, AttributeValue> attributes) {
+    return new AutoValue_Link(
+        context.getTraceId(),
+        context.getSpanId(),
+        type,
+        Collections.unmodifiableMap(new HashMap<String, AttributeValue>(attributes)));
```
adriancole: always use LinkedHashMap as iteration order differences are an unnecessary distraction

api/src/main/java/io/opencensus/trace/Link.java
```diff
@@ -68,5 +90,12 @@ public static Link fromSpanContext(SpanContext context, Type type) {
    */
   public abstract Type getType();
 
+  /**
+   * Returns the set of attributes.
```
adriancole: why you'd put attributes on a link isn't obvious. What's the use case? Can an example end up in javadoc?

PR #426, opened 2017-07-06
Add an optional number of child spans that were generated while this span was running.

bogdandrutu:


adriancole:
Nice response to the 10k span problem :) ramble below which is just fyi

we are likely to do similar in some zipkin projects, to help with bugs or
otherwise that create too many spans, though implementation requires a bit
of state management not currently in place.

Ex in our Brave model, we do control span creation (similar to the
SpanFactory here). The likely idea is to hide a variable in a real span
(possibly also in noop) and increment that on new span. Another way is to
keep a map of count or child id by trace context and touch that (which
allows externally generated spans to be visible). Regardless it is only
viable for spans local to the process unless the backend can do an atomic
update. Ex the doc here is pretty clear, but for example a remote child
that is dropped is not visible.

PR #431, opened 2017-07-07
Rename Aggregation to Aggregate.

songy23:
More renaming changes... Aggregation -> Aggregate


songy23:
Based on our discussion with @maxim-l I think the renaming stuff for Java is correct:
AggregationDescriptor -> Aggregation 

> ... specifies how to aggregate;

Aggregation -> Aggregate

> ... be used later to call the value of sum/count/etc in the view data consumption interface.

Merging soon, will post another PR to rename AggregationDescriptor to Aggregation

PR #433, opened 2017-07-07
Remove View 'getName' and factory methods that use Strings for view name.

sebright:
The View `getName` and factory methods had different versions for handling names as type `String` and `View.Name`.  This PR removes the String versions, since they are no longer needed for backwards compatibility.  View.Name can be used to look up views when only the name is known.

/cc @bogdandrutu


sebright:
We might need to change how we handle view names later, but I'll merge this for now so that we don't have duplicate methods.

PR #436, opened 2017-07-07
github: add an issue template

bogdandrutu:


.github/ISSUE_TEMPLATE
```diff
@@ -0,0 +1,16 @@
+Please answer these questions before submitting your issue.
```
sebright: It might be better to say something like "Please answer these questions before submitting a bug report.", since these questions don't apply to feature requests.

bogdandrutu: In github is more common to use issue than bug-report. I would keep it as it is.

sebright: I suggested the change because it makes it clearer that the information is only required for a subset of issues.

bogdandrutu: Ok. Please make a PR to update that 👍

PR #438, opened 2017-07-07
Add a section of known issues with the release.

bogdandrutu:


RELEASING.md
```diff
@@ -167,3 +167,14 @@ the repository. If this completes successfully, the repository can then be
...
+## Known Issues
+
+### Deployment for branch v0.5.x
+For all releases on the branch v0.5.x to deploy to maven central use command
+```bash
+$ ./gradlew clean build && ./gradlew uploadArchives
+```
+
+If option `-Dorg.gradle.parallel=false` is used, you will hit [this bug](https://issues.sonatype.org/browse/OSSRH-19485)
+caused by [this bug](https://github.com/gradle/gradle/issues/1827) in gradle 3.5.
```
sebright: Can we cherry-pick 7d91f470803115e988e59f4dd875309899101209 onto the release branch to avoid this problem?

bogdandrutu: We can do that, but if we ever have to rebuild 0.5.0 tag we should use this command. So we need to have this documented.

We can do it later (the cherry-pick).

PR #441, opened 2017-07-10
Rename AggregationDescriptor to Aggregation.

songy23:


core_impl/src/test/java/io/opencensus/stats/ViewManagerImplTest.java
```diff
@@ -87,7 +87,7 @@ private static DistributionView createDistributionView() {
   private static DistributionView createDistributionView(
       View.Name name,
       Measure measure,
-      DistributionAggregationDescriptor aggDescr,
+      DistributionAggregation aggDescr,
```
sebright: This variable still uses "Descr".

songy23: Done.

PR #446, opened 2017-07-11
Add Scope as interface for Closeables that work with current context.

sebright:
This interface allows users to write a shorter, more descriptive type name when
using scoped tags or spans.

### DIFF
--- a/api/src/main/java/io/opencensus/common/Scope.java
+++ b/api/src/main/java/io/opencensus/common/Scope.java
@@ -1,0 +1,20 @@
+/*
+ * Copyright 2017, Google Inc.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.common;
+
+/**
+ * A {@link NonThrowingCloseable} that represents a change to the current context over a scope of
+ * code.
+ */
+public interface Scope extends NonThrowingCloseable {}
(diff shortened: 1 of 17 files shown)


sebright:
Thanks.  I deprecated NonThrowingCloseable.

api/src/main/java/io/opencensus/common/Scope.java
```diff
@@ -0,0 +1,32 @@
...
+ * <pre>
+ *   try (Scope ctx = tryEnter()) {
+ *     ...
+ *   }
+ * </pre>
+ */
+@SuppressWarnings("deprecation")
+public interface Scope extends NonThrowingCloseable {
+  @Override
+  void close();
```
bogdandrutu: Do you need this?

sebright: The non-throwing `close` method will be needed if NonThrowingCloseable is removed.

An alternative design is to make Scope extend Closeable directly and then make ScopedSpanHandle and WithSpan implement both Scope and NonThrowingCloseable.

bogdandrutu: I was saying that we don't really need it right now. Yes if we remove that we will need, but right now we don't 👍 

Keeping it is fine with me.

sebright: I removed it.  The build will fail if we remove NonThrowingCloseable but don't add the close method to Scope.

sebright: I just realized that removing the `close` method caused a deprecation warning for all uses of Scope.close(), because the SuppressWarnings annotation no longer applied to it.  I'll add it back.

api/src/main/java/io/opencensus/common/Scope.java
```diff
@@ -0,0 +1,29 @@
...
+ * <p>Example of usage:
+ *
+ * <pre>
+ *   try (Scope ctx = tryEnter()) {
+ *     ...
+ *   }
+ * </pre>
+ */
+@SuppressWarnings("deprecation")
+public interface Scope extends NonThrowingCloseable {}
```
adriancole: maybe note if close is idempotent or undefined. for example, if you close a scope twice (due to auto-resource management etc), does it mess up things?

sebright: Thanks.  I created an issue for this: #452

PR #447, opened 2017-07-11
Move stats Context key to io.opencensus.tags.unsafe package.

sebright:
This commit starts refactoring the stats io.grpc.Context key so that it can be
used to access a TagContext.  It also organizes the classes that interact with
the Context to be more similar to the tracing package.

The commit renames the constant from STATS_CONTEXT_KEY to TAG_CONTEXT_KEY and
moves it from io.opencensus.stats.ContextUtils to
io.opencensus.tags.unsafe.ContextUtils.  It also renames the
io.opencensus.stats.ContextUtils class to CurrentTagsUtils and makes it
package-private, since the only public member was the Context key.  The Context
key still has type Context.Key<StatsContext>, for now.

### DIFF
--- a/core/src/main/java/io/opencensus/stats/CurrentTagsUtils.java
+++ b/core/src/main/java/io/opencensus/stats/CurrentTagsUtils.java
@@ -21,6 +21,7 @@ import io.opencensus.tags.unsafe.ContextUtils;
 /**
  * Util methods/functionality to interact with the {@link io.grpc.Context}.
  */
+// TODO(sebright): Move this into the tags package.
 final class CurrentTagsUtils {
 
   // Static class.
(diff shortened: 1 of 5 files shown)


core/src/main/java/io/opencensus/stats/CurrentTagsUtils.java
```diff
@@ -14,25 +14,17 @@
...
  */
-public final class ContextUtils {
-
-  /**
-   * The {@link io.grpc.Context.Key} used to interact with {@link io.grpc.Context}.
-   */
-  // TODO(songya): Discourage the usage of STATS_CONTEXT_KEY for normal users if needed.
-  public static final Context.Key<StatsContext> STATS_CONTEXT_KEY = Context.key(
-      "instrumentation-stats-key");
+final class CurrentTagsUtils {
```
bogdandrutu: Probably a TODO to move this into tags package?

sebright: Done

PR #448, opened 2017-07-11
Add APIs for working with the current TagContext.

sebright:
The main commit is ["Add APIs for working with the current TagContext."](https://github.com/sebright/opencensus-java/commit/dd3d60af61f07b222974a96cd346e2c430aebc9e).

I put the three main methods for interacting with the context in TagContext, so that they are easier to find and it is harder for users to accidentally use a new TagContext when they mean to use the current context.  Does that make sense, or would it be better to put the context interaction methods in a separate class?

I also tried to match the structure of the context interaction methods in tracing: Users can decide whether to use the current TagContext or a new TagContext when they create the builder, and they can decide whether to put the new TagContext into the current Context when they call the build method.

### DIFF
--- a/core/src/main/java/io/opencensus/stats/CurrentStatsContextUtils.java
+++ b/core/src/main/java/io/opencensus/stats/CurrentStatsContextUtils.java
@@ -15,17 +15,23 @@ package io.opencensus.stats;
 
 import io.grpc.Context;
 import io.opencensus.common.Scope;
-import io.opencensus.tags.unsafe.ContextUtils;
 
 /**
  * Util methods/functionality to interact with the {@link io.grpc.Context}.
  */
 // TODO(sebright): Move this into the tags package.
-final class CurrentTagsUtils {
+final class CurrentStatsContextUtils {
 
   // Static class.
-  private CurrentTagsUtils() {
+  private CurrentStatsContextUtils() {
   }
+
+  /**
+   * The {@link io.grpc.Context.Key} used to interact with the {@code StatsContext} contained in the
+   * {@link io.grpc.Context}.
+   */
+  public static final Context.Key<StatsContext> STATS_CONTEXT_KEY =
+      Context.key("opencensus-stats-context-key");
 
   /**
    * Returns The {@link StatsContext} from the current context.
@@ -33,7 +39,7 @@ final class CurrentTagsUtils {
    * @return The {@code StatsContext} from the current context.
    */
   static StatsContext getCurrentStatsContext() {
-    return ContextUtils.TAG_CONTEXT_KEY.get(Context.current());
+    return STATS_CONTEXT_KEY.get(Context.current());
   }
 
   /**
@@ -48,7 +54,7 @@ final class CurrentTagsUtils {
    *     current context.
    */
   static Scope withStatsContext(StatsContext statsContext) {
-    return new WithStatsContext(statsContext, ContextUtils.TAG_CONTEXT_KEY);
+    return new WithStatsContext(statsContext, STATS_CONTEXT_KEY);
   }
 
   // Supports try-with-resources idiom.
--- a/core/src/main/java/io/opencensus/stats/CurrentStatsContextUtils.java
+++ b/core/src/main/java/io/opencensus/stats/CurrentStatsContextUtils.java
@@ -19,7 +19,6 @@ import io.opencensus.common.Scope;
 /**
  * Util methods/functionality to interact with the {@link io.grpc.Context}.
  */
-// TODO(sebright): Move this into the tags package.
 final class CurrentStatsContextUtils {
 
   // Static class.
--- a/core/src/main/java/io/opencensus/tags/CurrentTagsUtils.java
+++ b/core/src/main/java/io/opencensus/tags/CurrentTagsUtils.java
@@ -14,7 +14,7 @@
 package io.opencensus.tags;
 
 import io.grpc.Context;
-import io.opencensus.common.NonThrowingCloseable;
+import io.opencensus.common.Scope;
 import io.opencensus.tags.unsafe.ContextUtils;
 
 /**
@@ -42,11 +42,11 @@ final class CurrentTagsUtils {
    * @return an object that defines a scope where the given {@code TagContext} is set to the current
    *     context.
    */
-  static NonThrowingCloseable withTagContext(TagContext tags) {
+  static Scope withTagContext(TagContext tags) {
     return new WithTagContext(tags);
   }
 
-  private static final class WithTagContext implements NonThrowingCloseable {
+  private static final class WithTagContext implements Scope {
 
     private final Context orig;
 
--- a/core/src/main/java/io/opencensus/tags/CurrentTagsUtils.java
+++ b/core/src/main/java/io/opencensus/tags/CurrentTagsUtils.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2017, Google Inc.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.tags;
+
+import io.grpc.Context;
+import io.opencensus.common.NonThrowingCloseable;
+import io.opencensus.tags.unsafe.ContextUtils;
+
+/**
+ * Utility methods for accessing the {@link TagContext} contained in the {@link io.grpc.Context}.
+ */
+final class CurrentTagsUtils {
+
+  private CurrentTagsUtils() {}
+
+  /**
+   * Returns the {@link TagContext} from the current context.
+   *
+   * @return the {@code TagContext} from the current context.
+   */
+  static TagContext getCurrentTagContext() {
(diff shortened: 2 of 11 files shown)


core/src/main/java/io/opencensus/stats/CurrentStatsContextUtils.java
```diff
@@ -14,27 +14,32 @@
...
-
 import io.opencensus.common.NonThrowingCloseable;
-import io.opencensus.tags.unsafe.ContextUtils;
 
 /**
  * Util methods/functionality to interact with the {@link io.grpc.Context}.
  */
 // TODO(sebright): Move this into the tags package.
-final class CurrentTagsUtils {
+final class CurrentStatsContextUtils {
```
bogdandrutu: I don't understand. Do we still use StatsContext?

core/src/main/java/io/opencensus/tags/TagContext.java
```diff
@@ -47,30 +49,39 @@
...
   /**
-   * Returns an empty {@code TagContext}.
+   * Returns a new empty {@code Builder}.
    *
-   * @return an empty {@code TagContext}.
+   * @return a new empty {@code Builder}.
    */
-  public static TagContext empty() {
-    return EMPTY;
+  public static Builder emptyBuilder() {
```
bogdandrutu: I would name this just builder() usually it means empty.

core/src/main/java/io/opencensus/tags/TagContext.java
```diff
@@ -47,30 +49,39 @@
...
   /**
-   * Returns a builder based on this {@code TagContext}.
+   * Returns the current {@code TagContext}.
    *
-   * @return a builder based on this {@code TagContext}.
+   * @return the current {@code TagContext}.
    */
-  public Builder toBuilder() {
-    return new Builder(getTags());
+  public static TagContext getCurrentTags() {
```
bogdandrutu: To be more consistent with trace API, do we want to have a Tagger class with all these static methods? That will also allow us to have the TagContext just as abstract class and have the implementation in the impl package if needed.

core/src/main/java/io/opencensus/stats/CurrentStatsContextUtils.java
```diff
@@ -14,27 +14,32 @@
...
 
 import io.grpc.Context;
-
 import io.opencensus.common.NonThrowingCloseable;
-import io.opencensus.tags.unsafe.ContextUtils;
 
 /**
  * Util methods/functionality to interact with the {@link io.grpc.Context}.
  */
 // TODO(sebright): Move this into the tags package.
```
bogdandrutu: Is the intention to remove this soon? Then please update the TODO.

core/src/main/java/io/opencensus/tags/TagContext.java
dinooliva: This should not be part of TagContext.

core/src/main/java/io/opencensus/tags/CurrentTagsUtils.java
```diff
@@ -0,0 +1,67 @@
...
+  /**
+   * Enters the scope of code where the given {@link TagContext} is in the current context and
+   * returns an object that represents that scope. The scope is exited when the returned object is
+   * closed.
+   *
+   * @param tags the {@code TagContext} to be set to the current context.
+   * @return an object that defines a scope where the given {@code TagContext} is set to the current
+   *     context.
+   */
+  static NonThrowingCloseable withTagContext(TagContext tags) {
```
dinooliva: Scope, once that's merged in

core/src/main/java/io/opencensus/tags/CurrentTagsUtils.java
```diff
@@ -0,0 +1,67 @@
...
+package io.opencensus.tags;
+
+import io.grpc.Context;
+import io.opencensus.common.NonThrowingCloseable;
+import io.opencensus.tags.unsafe.ContextUtils;
+
+/**
+ * Utility methods for accessing the {@link TagContext} contained in the {@link io.grpc.Context}.
+ */
+final class CurrentTagsUtils {
```
dinooliva: Rather than Tags, we should use TagContext so this class could either be CurrentTagContextUtils or just CurrentTagContext.

core/src/main/java/io/opencensus/tags/TagContext.java
```diff
@@ -47,30 +49,39 @@
...
   /**
-   * Returns an empty {@code Builder}.
+   * Returns a new {@code Builder} created from the current {@code TagContext}.
    *
-   * @return an empty {@code Builder}.
+   * @return a new {@code Builder} created from the current {@code TagContext}.
    */
-  public static Builder newBuilder() {
-    return new Builder();
+  public static Builder currentBuilder() {
```
dinooliva: The builder logic shouldn't change at all. Overall, this class shouldn't need to be changed at all.

core/src/main/java/io/opencensus/stats/CurrentStatsContextUtils.java
```diff
@@ -14,27 +14,32 @@
 package io.opencensus.stats;
```
dinooliva: This class should go away, but that may need to happen later.

core/src/main/java/io/opencensus/tags/TagContext.java
```diff
@@ -147,5 +158,17 @@ public Builder clear(TagKey key) {
...
+
+    /**
+     * Enters the scope of code where the {@link TagContext} created from this builder is in the
+     * current context and returns an object that represents that scope. The scope is exited when
+     * the returned object is closed.
+     *
+     * @return an object that defines a scope where the {@code TagContext} created from this builder
+     *     is set to the current context.
+     */
+    public NonThrowingCloseable buildScoped() {
```
songy23: Is this consistent with tracing? I thought the usage would be something like `Tagger.withTagContext(builder.build())`

core/src/main/java/io/opencensus/tags/TagContext.java
sebright: I changed the name to `emptyBuilder` to distinguish it from the method that creates a builder from the current tags.  @dinooliva thought that adding tags to the current context would be the most common use case and that we should make it clear when methods ignore the current context.

core/src/main/java/io/opencensus/tags/TagContext.java
sebright: I'm not sure.  On one hand, adding the indirection would give us more flexibility to change TagContext in the future, for example, by adding error reporting or validation of tag metadata.  On the other hand, it would be nice to keep the API simple by making TagContext a final class that users don't need to mock.

I would rather use the simpler design for now and add a factory after we decide we need it.

bogdandrutu: So testing is a minor thing in my opinion compared with the benefit that an indirection gives us. One extra reason to have an indirection and implement it as a no-op is that we can enable this by default in gRPC with 0 cost unless users are linking the implementation. With the current proposal users will pay the cost of tagging even if they do not use it which is in my opinion bad.

core/src/main/java/io/opencensus/tags/TagContext.java
bogdandrutu: So I agree and I did not say to remove it, it is just a proposal, we have the same issues in tracing and we want to a different approach where we do not have an explicit startSpanRoot. In my opinion an API like this:

```
public abstract class Tagger {
  // Returns the current TagContext from the Context or TagContext.Empty if nothing available in the Context.
  public TagContext getCurrentTagContext();

  // Starts a builder from the current tag context.
  public TagContextBuilder buildTagContext() {
    return buildTagContext(getCurrentTagContext());
  }

  // Starts a builder from the given tag context. If null then empty builder.
  public abstract TagContextBuilder buildTagContext(TagContext parent);
}
```

This way we make the API look very close to trace API + default implementation of the Tagger should return a NoopTagContextBuilder which does nothing (0 cost for not having the library installed).

Then you will need a class with a global state that allows access to the implementation: 
```
public final class Tagging {
  // loads with reflection the TagsComponent or TaggerImpl if we are sure we do not need to add more things. 
  private final Tagger tagger = Provider(...);

  public Tagger getTagger() {
    return tagger;
  }
}
```

Users will use this similar with Logger and Tracer in their class:

```
class MyClass {
  private static final Tagger tagger = Tagging.getTagger();

   ....
}
```

core/src/main/java/io/opencensus/tags/TagContext.java
sebright: That's a great idea.  I didn't consider the benefit of avoiding building the map of tags when it's not needed.  Now I'm convinced that we should use an abstract factory for TagContexts, too.   I'll update this branch and make a new PR.

sebright: I'm going to try a different design that uses a TagContext factory in a new PR.

core/src/main/java/io/opencensus/tags/TagContext.java
sebright: Same here

core/src/main/java/io/opencensus/tags/TagContext.java
sebright: I'll try that design.  What about `TagsComponent` or `TaggingComponent` for the abstract class?  We may also need to add a factory for tag keys, if we need to enforce uniqueness of names.  Then TagsComponent would contain the TagContext and TagKey factories.

sebright:
Thanks!  I'll merge this and continue working on a new branch.

adriancole:
I think somewhere we need to make it crystal clear the relationship between tags context and stats context.

sebright:
@adriancole We are replacing StatsContext with TagContext.  The difference is that TagContext doesn't depend on the stats package, so it could be used to label anything associated with an operation, and it can have multiple types of tag values (String, long, boolean).

PR #450, opened 2017-07-11
Add basic Aggregation and Aggregate

songy23:
Based on my understanding on our new View/ViewData/Aggregation/Aggregate, our current view-related data structures will need to have some huge changes. I'd like to share my thoughts before I move forward:

## What we currently have
Currently we have the following classes related to view and aggregation:

- [Distribution](https://github.com/census-instrumentation/opencensus-java/blob/master/core/src/main/java/io/opencensus/stats/Distribution.java) contains summary statistics for a population of values, including mandatory members sum, count, mean, range, sum of squared deviations, and an optional histogram representing the distribution of values across specified buckets;
-  [MutableDistribution](https://github.com/census-instrumentation/opencensus-java/blob/master/core/src/main/java/io/opencensus/stats/MutableDistribution.java) is the class that actually records and holds the aggregated stats. It's a mutable class where new values will be added and summary stats (sum, count, etc.) will be updated;
- [DistributionAggregation](https://github.com/census-instrumentation/opencensus-java/blob/master/core/src/main/java/io/opencensus/stats/DistributionAggregation.java) describes data aggregations based on distributions. Currently the only usage of this class is to specify if there's a histogram for the aggregation, which makes this class a little redundant. All the other aggregation statistics (sum, count, etc.) will be recorded by default, and cannot be configured;
- [DistributionAggregate](https://github.com/census-instrumentation/opencensus-java/blob/master/core/src/main/java/io/opencensus/stats/DistributionAggregate.java) is the result (or snapshot) of applying aggregation to measured values. It contains statistics on sum, count, range and mean by default, and contains an optional histogram;
- [IntervalAggregation](https://github.com/census-instrumentation/opencensus-java/blob/master/core/src/main/java/io/opencensus/stats/IntervalAggregation.java) describes data aggregations based on time intervals. It contains number of sub-intervals, and interval sizes represented by [Duration](https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/common/Duration.java)s. However, in our new definition, there will be exactly one `Duration` for each `IntervalView`, so this class is now redundant;
- [IntervalAggregate](https://github.com/census-instrumentation/opencensus-java/blob/master/core/src/main/java/io/opencensus/stats/IntervalAggregate.java) contains summary stats over various time intervals. Currently `Interval` only contains sum and count. For the same reason above, there will be exactly one interval for each interval aggregate now. Moreover, in the new design, `Interval` will need to support all the stats, including histogram;
- [DistributionView](https://github.com/census-instrumentation/opencensus-java/blob/master/core/src/main/java/io/opencensus/stats/View.java#L93) is for distribution-base aggregations. Currently it has one DistributionAggregation;
- [IntervalView](https://github.com/census-instrumentation/opencensus-java/blob/master/core/src/main/java/io/opencensus/stats/View.java#L130) is for interval-based aggregations. Currently it has one IntervalAggregation, which only supports recording sum and count;
- [DistributionViewData](https://github.com/census-instrumentation/opencensus-java/blob/master/core/src/main/java/io/opencensus/stats/ViewData.java#L52) contains a list of `DistributionAggregate`s;
- [IntervalViewData](https://github.com/census-instrumentation/opencensus-java/blob/master/core/src/main/java/io/opencensus/stats/ViewData.java#L100) contains a list of `IntervalAggregate`s.

## The proposed view-related class structure

- In our new design, each `View` will have a sequence of `Aggregation`s and a sequence of `TagKey`s. Here `Aggregation` refers to basic aggregation type, such as sum, count or histogram. Now views should be able to record any combination of aggregation types. For example, if we are only interested in recording count and don't need stats on sum, mean or anything else, we should be able to specify a view that only aggregates on count. Also, both DistributionView and IntervalView should support that.
- As stated above, `IntervalView` will now only support one duration. To record stats on different durations, users need to specify multiple `IntervalView`s.
- `Aggregation` and `View` are the description about how we want to aggregate on values/tagkeys, and they should be kept immutable. `Aggregate` and `ViewData` are conceptually snapshots of the former two, they should be also immutable. `MutableAggregation` will process values and update summary stats, and it's mutable.
- Similarly, `ViewData` will contain any combination of `Aggregate`s, corresponding to its definition in `View`.

## Changes to be made

- Split `Distribution` into independent basic `Aggregation`s, e.g. sum, count, histogram, etc.
- Split `MutableDistribution` into independent basic `MutableAggregation`s.
- Split `DistributionAggregate` into independent basic `Aggregate`s.
- Basic `Aggregation` types should expose factory methods so that users could define them in a `View`, while `Aggregate` should not be created by users, but should expose methods to get its internal value.
- Distribution related stuffs ([Distribution](https://github.com/census-instrumentation/opencensus-java/blob/master/core/src/main/java/io/opencensus/stats/Distribution.java), [MutableDistribution](https://github.com/census-instrumentation/opencensus-java/blob/master/core/src/main/java/io/opencensus/stats/MutableDistribution.java),  [DistributionAggregation](https://github.com/census-instrumentation/opencensus-java/blob/master/core/src/main/java/io/opencensus/stats/DistributionAggregation.java), [DistributionAggregate](https://github.com/census-instrumentation/opencensus-java/blob/master/core/src/main/java/io/opencensus/stats/DistributionAggregate.java)) and Interval related stuffs ([IntervalAggregate](https://github.com/census-instrumentation/opencensus-java/blob/master/core/src/main/java/io/opencensus/stats/IntervalAggregate.java), [IntervalAggregation](https://github.com/census-instrumentation/opencensus-java/blob/master/core/src/main/java/io/opencensus/stats/IntervalAggregation.java)) will all be removed, in favor of more generic `Aggregation` and `Aggregate` in this PR.
- Both `DistributionView` and `IntervalView` will contain a list of basic `Aggregation`s, instead of `DistributionAggregation` or `IntervalAggregation`. In addition, `IntervalView` will contain exactly one time window, which is its `Duration`.
- Both `DistributionViewData` and `IntervalViewData` will contain a list of basic `Aggregate`s, instead of `DistributionAggregate` or `IntervalAggregate`. In addition, `IntervalViewData` will contain exactly one time window, which is its `Duration`.

## Open questions

- How to store tag values in `ViewData`? One possible way is to store a `Map<List<TagValue>, List<Aggregate>>`


sebright:
I have some questions:

- Why are Aggregation and View mutable?  I thought that they only specified what data should be collected, and the implementation held the stats for each view or aggregation in separate, internal data structures.
- Are IntervalView and DistributionView the same, except that IntervalView has a duration?  Should we combine them into one class that has an optional duration field?

songy23:
Talked offline:
- `View` and `Aggregation` should be kept immutable. `Aggregation` should just specify how we want to aggregate stats, and the implementation or some class similar to `MutableDistribution` will actually record and hold stats.
- The only difference between IntervalView and DistributionView will be whether it has a Duration. Whether to combine them to one class is TBD.

Also add details on [MutableDistribution](https://github.com/census-instrumentation/opencensus-java/blob/master/core/src/main/java/io/opencensus/stats/MutableDistribution.java)

sebright:
@songy23 Is the code ready for review now?

songy23:
@sebright sure, PTAL (I'll add tests later).

core/src/main/java/io/opencensus/stats/Aggregate.java
```diff
@@ -0,0 +1,282 @@
...
+
+  /** The histogram of aggregated {@code MeasureValue}s. */
+  @Immutable
+  @AutoValue
+  public abstract static class AggregateHistogram extends Aggregate {
+
+    AggregateHistogram() {
+    }
+
+    static AggregateHistogram create(long[] bucketCounts) {
```
dinooliva: Consider long... bucketCounts

core/src/main/java/io/opencensus/stats/Aggregate.java
```diff
@@ -0,0 +1,282 @@
...
+     * @return the aggregated sum.
+     */
+    public abstract double get();
+
+    @Override
+    public final <T> T match(
+        Function<? super AggregateSum, T> p0,
+        Function<? super AggregateCount, T> p1,
+        Function<? super AggregateHistogram, T> p2,
+        Function<? super AggregateRange, T> p3,
```
dinooliva: We only need to support the first 3 for V0.1

core/src/main/java/io/opencensus/stats/Aggregate.java
```diff
@@ -0,0 +1,282 @@
...
+        Function<? super AggregateMean, T> p4,
+        Function<? super AggregateStdDev, T> p5) {
+      return p5.apply(this);
+    }
+  }
+
+  /** The range of a population's values. */
+  @Immutable
+  @AutoValue
+  public abstract static class Range {
```
dinooliva: We should probably split this out into it's own class - maybe move it into common.

songy23: If we would allow end users to create their own Range, then it should probably go into common. However from my understanding Range should only be created internally, and users should only be allowed to get its value (max and min), and that's why I made its factory method package private.

core/src/main/java/io/opencensus/stats/Aggregate.java
songy23: Just thought it's nice to have :) and is more consistent with what we currently record: https://github.com/census-instrumentation/opencensus-java/blob/master/core/src/main/java/io/opencensus/stats/MutableDistribution.java#L27

core/src/main/java/io/opencensus/stats/Aggregate.java
songy23: Done.

core/src/main/java/io/opencensus/stats/Aggregate.java
sebright: If we implement all of the known Aggregation types in the first release, and only hide their factory methods, then any client code will be prepared to handle all of the cases when we make them available.

songy23: That's a good idea, I've made these methods package private and mark them as not v0.1.

core/src/main/java/io/opencensus/stats/Aggregate.java
```diff
@@ -0,0 +1,254 @@
...
+      Function<? super AggregateCount, T> p1,
+      Function<? super AggregateHistogram, T> p2,
+      Function<? super AggregateRange, T> p3,
+      Function<? super AggregateMean, T> p4,
+      Function<? super AggregateStdDev, T> p5);
+
+  /** The sum value of aggregated {@code MeasureValue}s. */
+  @Immutable
+  @AutoValue
+  public abstract static class AggregateSum extends Aggregate {
```
dinooliva: I think we can drop the Aggregate prefix on these class names.

core/src/main/java/io/opencensus/stats/Aggregate.java
```diff
@@ -0,0 +1,254 @@
...
+@Immutable
+public abstract class Aggregate {
+
+  private Aggregate() {
+  }
+
+  /**
+   * Applies the given match function to the underlying data type.
+   */
+  public abstract <T> T match(
```
dinooliva: You should also add the default option to Match.

core/src/main/java/io/opencensus/stats/Aggregate.java
```diff
@@ -0,0 +1,254 @@
...
+    static AggregateCount create(long count) {
+      return new AutoValue_Aggregate_AggregateCount(count);
+    }
+
+    /**
+     * Returns the aggregated count.
+     *
+     * @return the aggregated count.
+     */
+    public abstract long get();
```
dinooliva: You've gone with a generic 'get()' method for all of the Aggregates but what if an Aggregate holds multiple values. For example, what if AggregateRange were to have the fields 'min' and 'max' with getters 'min()' and 'max()' (or 'getMin()', getMax()).

Is there any reason not to name these getters something more specific?

core/src/main/java/io/opencensus/stats/Aggregate.java
```diff
@@ -0,0 +1,254 @@
...
+        Function<? super AggregateMean, T> p4,
+        Function<? super AggregateStdDev, T> p5) {
+      return p2.apply(this);
+    }
+  }
+
+  /** The range of aggregated {@code MeasureValue}s. */
+  @Immutable
+  @AutoValue
+  public abstract static class AggregateRange extends Aggregate {
```
dinooliva: As per previous comment, why not just inline Range by adding fields 'min' and 'max'

core/src/main/java/io/opencensus/stats/Aggregate.java
songy23: Sure, added

core/src/main/java/io/opencensus/stats/Aggregate.java
songy23: I just thought it would be shorter and cleaner to just name these methods `get()`. For example, personally I'll perfer `Sum.get()` than `Sum.getSum()`. It makes sense for `Range` to have getters `getMin()` and `getMax` though.

core/src/main/java/io/opencensus/stats/Aggregate.java
songy23: Done. I dropped prefix for `Aggregate-*` but kept it for `Aggregation-*`, to avoid confusions.

core/src/main/java/io/opencensus/stats/Aggregate.java
songy23: This is from our previous implementation for [DistributionAggregate](https://github.com/census-instrumentation/opencensus-java/blob/master/core/src/main/java/io/opencensus/stats/DistributionAggregate.java#L115), its usage is like `distributionAggregate.getRange().getMin()` or `.getMax()`. I've updated the new `Aggregate` so that `Range` will be just a subclass of `Aggregate`, and has fields min and max.

core/src/main/java/io/opencensus/stats/Aggregate.java
```diff
@@ -0,0 +1,273 @@
...
+      return new AutoValue_Aggregate_Range(min, max);
+    }
+
+    /**
+     * Returns the minimum of the population values.
+     *
+     * @return the minimum of the population values.
+     */
+    // TODO(songya): not v0.1, expose this method later.
+    public abstract double getMin();
```
sebright: Should this TODO be removed, since the method is already public?  I think we will need it to be public in v0 for users to access the data in the `match` method.

core/src/main/java/io/opencensus/stats/Aggregate.java
```diff
@@ -0,0 +1,273 @@
...
+      return new AutoValue_Aggregate_Mean(mean);
+    }
+
+    /**
+     * Returns the aggregated mean.
+     *
+     * @return the aggregated mean.
+     */
+    // TODO(songya): not v0.1, expose this method later.
+    abstract double get();
```
sebright: Same here.  Should this be public?

core/src/main/java/io/opencensus/stats/Aggregate.java
```diff
@@ -0,0 +1,273 @@
...
+
+  /** The standard deviation value of aggregated {@code MeasureValue}s. */
+  @Immutable
+  @AutoValue
+  public abstract static class StdDev extends Aggregate {
+
+    StdDev() {
+    }
+
+    static StdDev create(double stdDev) {
```
sebright: Same here

core/src/main/java/io/opencensus/stats/Aggregation.java
```diff
@@ -0,0 +1,242 @@
...
+     * Construct a {@code AggregationHistogram}.
+     *
+     * @return a new {@code AggregationHistogram}.
+     */
+    public static AggregationHistogram create(BucketBoundaries bucketBoundaries) {
+      checkNotNull(bucketBoundaries, "bucketBoundaries should not be null.");
+      return new AutoValue_Aggregation_AggregationHistogram(bucketBoundaries);
+    }
+
+    abstract BucketBoundaries getBucketBoundaries();
```
sebright: Should this be public?

core/src/main/java/io/opencensus/stats/Aggregate.java
songy23: Make sense, updated

core/src/main/java/io/opencensus/stats/Aggregate.java
songy23: Agree, now that we have `match` I think it makes more sense to just expose these values.

core/src/main/java/io/opencensus/stats/Aggregate.java
songy23: Done

core/src/main/java/io/opencensus/stats/Aggregate.java
songy23: I don't think `Aggregate.create()` needs to be public, since `Aggregate`s should all be created internally, and all users need is to read stats from them, i.e expose `get()` methods.

core/src/main/java/io/opencensus/stats/Aggregation.java
songy23: Nice catch, this is needed when exporting histograms.

core/src/main/java/io/opencensus/stats/MutableAggregation.java
```diff
@@ -0,0 +1,382 @@
...
+ */
+
+package io.opencensus.stats;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import io.opencensus.common.Function;
+
+/** Mutable version of {@link Aggregation} that supports adding values. */
+abstract class MutableAggregation {
```
songy23: Sure, done

core/src/main/java/io/opencensus/stats/Aggregate.java
songy23: Yeah that's my original thought. I feel it's better to keep the `Aggregate` naming consistent with `Tag`/`TagKey`.

dinooliva: Actually, I feel that TagKey is more of the anomalous case since the variants are named after types, which can lead to confusion. For the general case, I believe that prefixing the variant by the superclass name is redundant and unnecessary - e.g. for our List example, I don't think there's a need to call the variant's ListNil and ListCons - if client code really wants the prefix, they can use List.Nil and List.Cons.

We could similarly consider TagKey.String rather than TagKeyString but I'm wary of doing that.

songy23: Per our discussion, I've renamed Aggregate to AggregationData, and all its subclasses to *Data (SumData, CountData, etc.). Also renamed subclasses of Aggregation to not have the prefix.

core/src/main/java/io/opencensus/stats/AggregationData.java
```diff
@@ -0,0 +1,269 @@
...
+import java.util.List;
+import javax.annotation.concurrent.Immutable;
+
+/**
+ * {@link AggregationData} is the result of applying a given {@link Aggregation} to a set of
+ * {@code MeasureValue}s.
+ *
+ * <p>{@link AggregationData} currently supports 6 types of basic aggregation:
+ * <ul>
+ *   <li>SUM
```
dinooliva: Think this comment needs to be updated.

songy23: Sure, updated

core/src/test/java/io/opencensus/stats/AggregationTest.java
```diff
@@ -0,0 +1,146 @@
...
+    thrown.expect(NullPointerException.class);
+    thrown.expectMessage("bucketBoundaries should not be null.");
+    Histogram.create(null);
+  }
+
+  @Test
+  public void testUnsortedBoundaries() throws Exception {
+    List<Double> buckets = Arrays.asList(0.0, 1.0, 1.0);
+    thrown.expect(IllegalArgumentException.class);
+    Histogram.create(BucketBoundaries.create(buckets));
```
sebright: I think the BucketBoundaries class should enforce this invariant, so we shouldn't need to test it in the Aggregation tests.

core/src/test/java/io/opencensus/stats/AggregationDataTest.java
```diff
@@ -0,0 +1,168 @@
...
+public class AggregationDataTest {
+
+  @Rule
+  public ExpectedException thrown = ExpectedException.none();
+
+  @Test
+  public void testCreateHistogramData() {
+    HistogramData histogram = HistogramData.create(0, 0, 0);
+    assertThat(new ArrayList<Long>(histogram.getBucketCounts())).isEqualTo(
+        Arrays.asList(0L, 0L, 0L));
```
sebright: Is the List copy necessary?  I think this would also work:

    assertThat(histogram.getBucketCounts()).containsExactly(0L, 0L, 0L).inOrder();

core/src/main/java/io/opencensus/stats/AggregationData.java
```diff
@@ -0,0 +1,269 @@
...
+  /** The range of aggregated {@code MeasureValue}s. */
+  @Immutable
+  @AutoValue
+  public abstract static class RangeData extends AggregationData {
+
+    RangeData() {
+    }
+
+    static RangeData create(double min, double max) {
+      if (min != Double.POSITIVE_INFINITY && max != Double.NEGATIVE_INFINITY) {
```
songy23: Done, nice catch

core/src/main/java/io/opencensus/stats/View.java
```diff
@@ -42,13 +42,13 @@
   public abstract Measure getMeasure();
 
   /**
-   * Dimensions (a.k.a Tag Keys) to match with the associated {@link Measure}.
-   * If no dimensions are specified, then all stats are recorded. Dimensions must be unique.
+   * Columns (a.k.a Tag Keys) to match with the associated {@link Measure}.
+   * If no dimensions are specified, then all stats are recorded. Columns must be unique.
```
songy23: Sure, I'll add this to Javadoc in the View PR https://github.com/census-instrumentation/opencensus-java/pull/462.

core/src/test/java/io/opencensus/stats/AggregationDataTest.java
songy23: That's better, updated

core/src/test/java/io/opencensus/stats/AggregationTest.java
songy23: Right, this test is a little redundant. I've removed it.

songy23:
Thanks for review! merging soon

PR #456, opened 2017-07-17
Run the test with MockitoJUnitRunner.

ubschmidt2:


contrib/agent/src/test/java/io/opencensus/contrib/agent/ResourcesTest.java
```diff
@@ -27,20 +26,23 @@
...
 import org.junit.runner.RunWith;
-import org.junit.runners.JUnit4;
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
 
 /**
  * Unit tests for {@link Resources}.
  */
-@RunWith(JUnit4.class)
+@RunWith(MockitoJUnitRunner.class)
```
bogdandrutu: Why do you need this PR (change)?

ubschmidt2: It's not strictly needed, but has some advantages. I couldn't remember details, but now found https://stackoverflow.com/questions/10806345/runwithmockitojunitrunner-class-vs-mockitoannotations-initmocksthis and https://stackoverflow.com/questions/44794403/why-we-use-mockitojunitrunner-class-in-our-junit-test. 

I see that this is different from the other tests in this repo, so I don't mind sticking with the existing pattern. Alternatively, if the advantages are compelling enough, we could switch all tests to MockitoJUnitRunner.

adriancole: I'd caveat switch all mockito tests to MockitoJUnitRunner :)

PR #461, opened 2017-07-19
Add automatic context propagation to Thread#start.

ubschmidt2:


ubschmidt2:
Not urgent. Sometime next week would be great.

ubschmidt2:
PTAL, if you have a chance.

dinooliva:
Sure, I'll take a look tomorrow morning.

contrib/agent/src/main/java/io/opencensus/contrib/agent/ContextStrategyImpl.java
```diff
@@ -22,8 +24,27 @@
  */
 final class ContextStrategyImpl implements ContextStrategy {
 
+  private final Cache<Thread, Context> savedContexts
+          = CacheBuilder.newBuilder().weakKeys().build();
```
dinooliva: So, iiuc, you use weakKeys() so that threads are not kept around when they're done. Is there any danger of new thread being created that reuses a thread id and causes a false positive here?

dinooliva:
Apologies for the delay on review but this is really cool - thanks!

contrib/agent/src/main/java/io/opencensus/contrib/agent/ContextStrategyImpl.java
adriancole: comment around this should make it to the field javadoc

contrib/agent/src/main/java/io/opencensus/contrib/agent/ThreadInstrumentation.java
```diff
@@ -0,0 +1,91 @@
...
+import net.bytebuddy.agent.builder.AgentBuilder;
+import net.bytebuddy.asm.Advice;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.dynamic.DynamicType;
+import net.bytebuddy.matcher.ElementMatcher;
+import net.bytebuddy.utility.JavaModule;
+
+/**
+ * Propagates the context of the caller of {@link Thread#start} to the new thread, just like the
+ * Microsoft .Net Framework propagates the <a
```
adriancole: would a comparison to InheritableThreadLocal be reasonable?

contrib/agent/src/main/java/io/opencensus/contrib/agent/ContextStrategyImpl.java
ubschmidt2: Added a comment to #savedContexts. Re reuse of thread IDs: no - thread IDs don't matter in this implementation. If using thread IDs as keys (not doing that), it should still be safe as thread IDs generated by incrementing a long in Thread.

contrib/agent/src/main/java/io/opencensus/contrib/agent/ThreadInstrumentation.java
ubschmidt2: Yes, added this to the class Javadoc.

ubschmidt2:
> guessing that the double-instrumentation risks are here and similar to InheritableThreadLocal

> ex an executor backed by a thread pool that grows during a request might have the wrong context attached? (ex instead of root the one accidentally causing the pool to grow?)

> maybe tests can show this.

Yes, this is a good example. Going to add a test for this scenario. The expected behavior is that the context that is (automatically) propagated by Executor#execute becomes the effective context, of course.

ubschmidt2:
>> guessing that the double-instrumentation risks are here and similar to InheritableThreadLocal
>>
>> ex an executor backed by a thread pool that grows during a request might have the wrong context attached? (ex instead of root the one accidentally causing the pool to grow?)
>>
>> maybe tests can show this.
>
> Yes, this is a good example. Going to add a test for this scenario. The expected behavior is that the context that is (automatically) propagated by Executor#execute becomes the effective context, of course.

Added a simple test for this scenario.

contrib/agent/src/integration-test/java/io/opencensus/contrib/agent/ThreadInstrumentationTest.java
```diff
@@ -0,0 +1,129 @@
...
+    final Thread callerThread = Thread.currentThread();
+    final Context context = Context.current().withValue(KEY, "myvalue");
+    previousContext = context.attach();
+
+    final AtomicBoolean tested = new AtomicBoolean(false);
+
+    Thread thread = new Thread(new Runnable() {
+      @Override
+      public void run() {
+        assertThat(Thread.currentThread()).isNotSameAs(callerThread);
```
adriancole: nit: I usually add a trailing comment // sanity check for things that are sanity checks.

contrib/agent/src/integration-test/java/io/opencensus/contrib/agent/ThreadInstrumentationTest.java
```diff
@@ -0,0 +1,129 @@
...
+
+  @Test(timeout = 5000)
+  public void start_Subclass() throws Exception {
+    final Thread callerThread = Thread.currentThread();
+    final Context context = Context.current().withValue(KEY, "myvalue");
+    previousContext = context.attach();
+
+    final AtomicBoolean tested = new AtomicBoolean(false);
+
+    Thread thread = new Thread() {
```
adriancole: nit I'd make a method local class as it is easy to miss that this is a subclass when scrolling down (anonymous subclass looks very like a thread with runnable)

contrib/agent/src/integration-test/java/io/opencensus/contrib/agent/ThreadInstrumentationTest.java
```diff
@@ -0,0 +1,129 @@
...
+    };
+
+    thread.start();
+    thread.join();
+
+    assertThat(tested.get()).isTrue();
+  }
+
+  @Test(timeout = 5000)
+  public void start_wrappedRunnable() throws Exception {
```
adriancole: start_wrapsRunnable? I was looking in this method to see use of context.wrap :) I think the key here is that we are testing that things are implicitly wrapped.

contrib/agent/src/integration-test/java/io/opencensus/contrib/agent/ThreadInstrumentationTest.java
```diff
@@ -0,0 +1,129 @@
...
+        try {
+          thread.join();
+        } catch (InterruptedException ex) {
+          Thread.currentThread().interrupt();
+        }
+        context2.detach(context);
+      }
+    };
+
+    newThreadExecutor.execute(new Runnable() {
```
adriancole: add a comment about why we are doing a secondary check

contrib/agent/src/integration-test/java/io/opencensus/contrib/agent/ThreadInstrumentationTest.java
ubschmidt2: Renamed to "start_automaticallyWrappedRunnable" and added a comment.

contrib/agent/src/integration-test/java/io/opencensus/contrib/agent/ThreadInstrumentationTest.java
ubschmidt2: Done.

contrib/agent/src/integration-test/java/io/opencensus/contrib/agent/ThreadInstrumentationTest.java
ubschmidt2: Sounds good, though in this specific test it occurs to me that this sanity check doesn't add much value, so I'm just removing it. The sanity check made more sense in ExecutorInstrumentTest.

contrib/agent/src/integration-test/java/io/opencensus/contrib/agent/ThreadInstrumentationTest.java
ubschmidt2: Done.

PR #462, opened 2017-07-19
Refactor to use more generic View and ViewData.

songy23:
This PR is the API level changes.
- Updated `View` and `ViewData` based on our new data model. 
- Removed subclasses `DistributionView` and `IntervalView`, `DistributionViewData` and `IntervalViewData`, in favor of more generic `View` and `ViewData`.
- Comment out affected implementation and tests. Changes on implementations will be made after API changes.

### DIFF
--- a/core/src/main/java/io/opencensus/stats/RpcViewConstants.java
+++ b/core/src/main/java/io/opencensus/stats/RpcViewConstants.java
@@ -42,9 +42,6 @@ import static io.opencensus.stats.RpcMeasureConstants.RPC_STATUS;
 import io.opencensus.common.Duration;
 import io.opencensus.stats.Aggregation.Count;
 import io.opencensus.stats.Aggregation.Histogram;
-import io.opencensus.stats.Aggregation.Mean;
-import io.opencensus.stats.Aggregation.Range;
-import io.opencensus.stats.Aggregation.StdDev;
 import io.opencensus.stats.Aggregation.Sum;
 import io.opencensus.stats.View.Window;
 import io.opencensus.stats.View.Window.CumulativeWindow;
@@ -75,22 +72,19 @@ public final class RpcViewConstants {
           4096.0, 8192.0, 16384.0, 32768.0, 65536.0));
 
   static final List<Aggregation> AGGREGATION_NO_HISTOGRAM = Collections.unmodifiableList(
-      Arrays.asList(Sum.create(), Count.create(), Range.create(), Mean.create(), StdDev.create()));
+          Arrays.asList(Sum.create(), Count.create()));
 
   static final List<Aggregation> AGGREGATION_WITH_BYTES_HISTOGRAM = Collections.unmodifiableList(
-      Arrays.asList(Sum.create(), Count.create(), Range.create(),
-          Histogram.create(BucketBoundaries.create(RPC_BYTES_BUCKET_BOUNDARIES)),
-          Mean.create(), StdDev.create()));
+      Arrays.asList(Sum.create(), Count.create(),
+          Histogram.create(BucketBoundaries.create(RPC_BYTES_BUCKET_BOUNDARIES))));
 
   static final List<Aggregation> AGGREGATION_WITH_MILLIS_HISTOGRAM = Collections.unmodifiableList(
-      Arrays.asList(Sum.create(), Count.create(), Range.create(),
-          Histogram.create(BucketBoundaries.create(RPC_MILLIS_BUCKET_BOUNDARIES)),
-          Mean.create(), StdDev.create()));
+      Arrays.asList(Sum.create(), Count.create(),
+          Histogram.create(BucketBoundaries.create(RPC_MILLIS_BUCKET_BOUNDARIES))));
 
   static final List<Aggregation> AGGREGATION_WITH_COUNT_HISTOGRAM = Collections.unmodifiableList(
-      Arrays.asList(Sum.create(), Count.create(), Range.create(),
-          Histogram.create(BucketBoundaries.create(RPC_COUNT_BUCKET_BOUNDARIES)),
-          Mean.create(), StdDev.create()));
+      Arrays.asList(Sum.create(), Count.create(),
+          Histogram.create(BucketBoundaries.create(RPC_COUNT_BUCKET_BOUNDARIES))));
 
   static final Duration MINUTE = Duration.create(60, 0);
   static final Duration HOUR = Duration.create(60 * 60, 0);
@@ -98,7 +92,7 @@ public final class RpcViewConstants {
   static final Window INTERVAL_MINUTE = IntervalWindow.create(MINUTE);
   static final Window INTERVAL_HOUR = IntervalWindow.create(HOUR);
 
-  // Rpc client {@link View}s.
+  // Rpc client cumulative {@link View}s.
   public static final View RPC_CLIENT_ERROR_COUNT_VIEW =
       View.create(
           View.Name.create("grpc.io/client/error_count/distribution_cumulative"),
@@ -173,7 +167,7 @@ public final class RpcViewConstants {
           CUMULATIVE);
 
 
-  // Rpc server {@link View}s.
+  // Rpc server cumulative {@link View}s.
   public static final View RPC_SERVER_ERROR_COUNT_VIEW =
       View.create(
           View.Name.create("grpc.io/server/error_count/distribution_cumulative"),
@@ -249,11 +243,11 @@ public final class RpcViewConstants {
 
   // Interval Stats
 
-  // RPC client {@link View}s.
+  // RPC client interval {@link View}s.
   public static final View RPC_CLIENT_ROUNDTRIP_LATENCY_MINUTE_VIEW =
       View.create(
           View.Name.create("grpc.io/client/roundtrip_latency/interval"),
-          "Minute and Hour stats for latency in msecs",
+          "Minute stats for latency in msecs",
           RPC_CLIENT_ROUNDTRIP_LATENCY,
           AGGREGATION_NO_HISTOGRAM,
           Arrays.asList(RPC_CLIENT_METHOD),
@@ -262,7 +256,7 @@ public final class RpcViewConstants {
   public static final View RPC_CLIENT_REQUEST_BYTES_MINUTE_VIEW =
       View.create(
           View.Name.create("grpc.io/client/request_bytes/interval"),
-          "Minute and Hour stats for request size in bytes",
+          "Minute stats for request size in bytes",
           RPC_CLIENT_REQUEST_BYTES,
           AGGREGATION_NO_HISTOGRAM,
           Arrays.asList(RPC_CLIENT_METHOD),
@@ -271,7 +265,7 @@ public final class RpcViewConstants {
   public static final View RPC_CLIENT_RESPONSE_BYTES_MINUTE_VIEW =
       View.create(
           View.Name.create("grpc.io/client/response_bytes/interval"),
-          "Minute and Hour stats for response size in bytes",
+          "Minute stats for response size in bytes",
           RPC_CLIENT_RESPONSE_BYTES,
           AGGREGATION_NO_HISTOGRAM,
           Arrays.asList(RPC_CLIENT_METHOD),
@@ -280,7 +274,7 @@ public final class RpcViewConstants {
   public static final View RPC_CLIENT_ERROR_COUNT_MINUTE_VIEW =
       View.create(
           View.Name.create("grpc.io/client/error_count/interval"),
-          "Minute and Hour stats for rpc errors",
+          "Minute stats for rpc errors",
           RPC_CLIENT_ERROR_COUNT,
           AGGREGATION_NO_HISTOGRAM,
           Arrays.asList(RPC_CLIENT_METHOD),
@@ -289,7 +283,7 @@ public final class RpcViewConstants {
   public static final View RPC_CLIENT_UNCOMPRESSED_REQUEST_BYTES_MINUTE_VIEW =
       View.create(
           View.Name.create("grpc.io/client/uncompressed_request_bytes/interval"),
-          "Minute and Hour stats for uncompressed request size in bytes",
+          "Minute stats for uncompressed request size in bytes",
           RPC_CLIENT_UNCOMPRESSED_REQUEST_BYTES,
           AGGREGATION_NO_HISTOGRAM,
           Arrays.asList(RPC_CLIENT_METHOD),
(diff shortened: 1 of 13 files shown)


songy23:
I've rebased the branch, should be ready for review now.

core/src/main/java/io/opencensus/stats/View.java
```diff
@@ -14,18 +14,24 @@
...
 @Immutable
+@AutoValue
 public abstract class View {
+
+  View() {
+  }
+
   /**
    * Name of view. Must be unique.
    */
```
sebright: I don't think we should use a String here or in other places where the view name appears in the API.  The `View.Name` class is more type-safe and also helps document that the type returned by View.getName() is the same type as the name accepted by ViewManager.getView(...), View.create(...), etc.  Also, the wrapper class should have little performance overhead because only one instance needs to be created per View.

dinooliva: The view name should never be used separate from the view itself.

sebright: I think we will need a way to look up data for a View by name, or at least a subset of the View fields.  For example, a request coming from the network shouldn't need to specify the view's description.  Even if we don't need the functionality in the first release, I think we should use the Name class here to avoid needing to change the signature after it's released.

songy23: Per our discussion today, I'll hold off changing `View.Name` until we have more details on naming constrains.

core/src/main/java/io/opencensus/stats/View.java
```diff
@@ -42,20 +48,48 @@
...
 
   /**
-   * Applies the given match function to the underlying data type.
+   * Returns the time window for this {@code View}.
+   *
+   * @return the time window. Time window is null for cumulative views.
+   */
+  @Nullable
+  public abstract Duration getWindow();
+
```
songy23: Makes sense, I've added a hierarchical `Window` class.

core/src/main/java/io/opencensus/stats/ViewData.java
```diff
@@ -14,116 +14,85 @@
...
-    public static DistributionViewData create(DistributionView distributionView,
-        List<DistributionAggregate> distributionAggregates, Timestamp start, Timestamp end) {
-      return new AutoValue_ViewData_DistributionViewData(
-          distributionView,
-          Collections.unmodifiableList(
-              new ArrayList<DistributionAggregate>(distributionAggregates)),
-          start,
-          end);
-    }
+  public abstract static class WindowTimestamp {
```
songy23: Agree, I feel `WindowData` is a better name, align with other fields in ViewData. I've renamed this class to `WindowData`.

core/src/main/java/io/opencensus/stats/ViewData.java
```diff
@@ -14,116 +14,85 @@
...
 
-    @Override
-    public abstract DistributionView getView();
+    WindowTimestamp() {
+    }
 
     /**
-     * The {@link DistributionAggregate}s associated with this {@link DistributionViewData}.
+     * Returns the start {@code Timestamp} for a {@link ViewData}. For interval-based views, start
```
songy23: Updated both of them.

core/src/main/java/io/opencensus/stats/RpcViewConstants.java
```diff
@@ -51,329 +55,546 @@
...
+  // RPC client {@link View}s.
+  public static final View RPC_CLIENT_ROUNDTRIP_LATENCY_MINUTE_VIEW =
+      View.create(
           View.Name.create("grpc.io/client/roundtrip_latency/interval"),
           "Minute and Hour stats for latency in msecs",
           RPC_CLIENT_ROUNDTRIP_LATENCY,
-          IntervalAggregation.create(Arrays.asList(MINUTE, HOUR)),
-          Arrays.asList(RPC_CLIENT_METHOD));
+          AGGREGATION_NO_HISTOGRAM,
+          Arrays.asList(RPC_CLIENT_METHOD), MINUTE);
```
songy23: Sure, updated

core/src/test/java/io/opencensus/stats/ViewDataTest.java
```diff
@@ -36,118 +47,99 @@
...
-            tagKeys);
-    List<IntervalAggregate> iAggregates =
-        Arrays.asList(
-            IntervalAggregate.create(
-                tags1, Arrays.asList(Interval.create(Duration.fromMillis(111), 10, 100))));
+    View dView = 
+        View.create(name, description, measure, AGGREGATIONS, tagKeys, RpcViewConstants.CUMULATIVE);
+    View iView =
+        View.create(
+            name, description, measure, AGGREGATIONS, tagKeys, RpcViewConstants.INTERVAL_HOUR);
```
sebright: Same here

core/src/main/java/io/opencensus/stats/ViewData.java
```diff
@@ -16,114 +16,134 @@
...
+   *
+   * @return the {@code WindowData}.
+   */
+  public abstract WindowData getWindowData();
+
+  /**
+   * Constructs a new {@link ViewData}.
+   */
+  public static ViewData create(View view, Map<List<TagValue>, List<AggregationData>> map,
+      WindowData windowData) {
```
sebright: Should this method do any validation, such as checking that the Window and WindowData types match?

core/src/test/java/io/opencensus/stats/ViewTest.java
```diff
@@ -13,118 +13,78 @@
...
-          @Override public Boolean apply(DistributionView dView) {
-            return dView == view;
-          }
-        },
-        new Function<IntervalView, Boolean> () {
-          @Override public Boolean apply(IntervalView iView) {
-            return false;
-          }
-        }));
+    assertThat(view.getWindow()).isEqualTo(RpcViewConstants.CUMULATIVE);
```
sebright: Same here, this test shouldn't depend on the RPC constants, especially if changing their values would cause the test to fail.

core/src/test/java/io/opencensus/stats/ViewTest.java
```diff
@@ -13,118 +13,78 @@
...
-          @Override public Boolean apply(DistributionView dView) {
-            return false;
-          }
-        },
-        new Function<IntervalView, Boolean> () {
-          @Override public Boolean apply(IntervalView iView) {
-            return iView == view;
-          }
-        }));
+    assertThat(view.getWindow()).isEqualTo(RpcViewConstants.INTERVAL_MINUTE);
```
sebright: Same here

core/src/test/java/io/opencensus/stats/ViewTest.java
```diff
@@ -13,118 +13,78 @@
...
-                name, description, measure, iAggr, keys))
+            View.create(
+                name, description, measure, aggregations, keys, IntervalWindow.create(minute)),
+            View.create(
+                name, description, measure, aggregations, keys, IntervalWindow.create(minute)))
         .addEqualityGroup(
-            IntervalView.create(
-                name, description + 2, measure, iAggr, keys))
+            View.create(
+                name, description + 2, measure, aggregations, keys, IntervalWindow.create(minute)))
```
sebright: Suggestion: It might be better to modify the interval size instead of the description, since the equality group above already modifies the description.

core_impl/src/main/java/io/opencensus/stats/StatsManager.java
```diff
@@ -33,12 +32,6 @@
   }
 
   void registerView(View view) {
-    // Only DistributionViews are supported currently.
-    // TODO(sebright): Remove this once all views are supported.
-    if (!(view instanceof DistributionView)) {
-      throw new UnsupportedOperationException(
```
sebright: I think we should still check that the View has a cumulative window here.

core/src/test/java/io/opencensus/stats/RpcViewConstantsTest.java
```diff
@@ -27,6 +36,42 @@
...
+        Sum.create(), Count.create(), Range.create(), Mean.create(), StdDev.create());
+    assertThat(RpcViewConstants.AGGREGATION_WITH_BYTES_HISTOGRAM).containsExactly(
+        Sum.create(), Count.create(), Range.create(), Mean.create(), StdDev.create(),
+        Histogram.create(BucketBoundaries.create(RpcViewConstants.RPC_BYTES_BUCKET_BOUNDARIES)));
+    assertThat(RpcViewConstants.AGGREGATION_WITH_COUNT_HISTOGRAM).containsExactly(
+        Sum.create(), Count.create(), Range.create(), Mean.create(), StdDev.create(),
+        Histogram.create(BucketBoundaries.create(RpcViewConstants.RPC_COUNT_BUCKET_BOUNDARIES)));
+    assertThat(RpcViewConstants.AGGREGATION_WITH_MILLIS_HISTOGRAM).containsExactly(
+        Sum.create(), Count.create(), Range.create(), Mean.create(), StdDev.create(),
+        Histogram.create(BucketBoundaries.create(RpcViewConstants.RPC_MILLIS_BUCKET_BOUNDARIES)));
```
sebright: Should these assertions call `inOrder()` on the results?

core/src/test/java/io/opencensus/stats/ViewDataTest.java
```diff
@@ -36,118 +47,99 @@
...
+            fail("CumulativeWindowData expected.");
+            return null;
           }
-        }));
+        },
+        Functions.<Void>throwIllegalArgumentException());
   }
 
   @Test
   public void testIntervalViewData() {
```
songy23: Makes sense, I've split the tests.

core_impl/src/main/java/io/opencensus/stats/StatsManager.java
songy23: Done.

sebright:
LGTM.  I thought of an idea for avoiding having commented out tests on master.  We could create a branch for the view changes in this repository.  Then we could merge this PR into it and merge the new branch into master once the implementation is updated.

songy23:
@sebright sounds good, I've created a new branch `view` and changed the base to point to it.

core/src/main/java/io/opencensus/stats/View.java
```diff
@@ -42,20 +48,43 @@
   public abstract Measure getMeasure();
 
   /**
+   * The {@link Aggregation}s associated with this {@link View}.
+   */
```
dinooliva: We should document that aggregations should be unique.

songy23: Done

core/src/main/java/io/opencensus/stats/View.java
```diff
@@ -85,77 +114,77 @@ public static Name create(String name) {
...
+
+      CumulativeWindow() {}
+
+      /**
+       * Constructs a cumulative {@code Window} that does not have an explicit {@code Duration}.
+       * Instead, cumulative {@code Window} always has an interval of infinite {@code Duration}.
+       *
+       * @return a cumulative {@code Window}.
+       */
+      public static CumulativeWindow create() {
```
sebright: Could we make the existing `create()` method return a constant in the future, as an optimization, or do you think the name would be misleading?

songy23: Sounds reasonable. I think the naming is fine since having a constant is kind of implementation detail.

core/src/main/java/io/opencensus/stats/View.java
```diff
@@ -42,20 +48,43 @@
...
+  public abstract Window getWindow();
+
+  /**
+   * Constructs a new {@link View}.
+   */
+  public static View create(
+      Name name,
+      String description,
+      Measure measure,
+      List<Aggregation> aggregations,
```
sebright: I think it would be confusing to require users to rely on the order in a LinkedHashSet when matching TagKeys with a list of TagValues or matching Aggregations with a list of AggregationDatas.  Why don't we just check for duplicates in the input TagKey and Aggregation lists?  We only need to check once, because the data structures are immutable.  Lists make it clear that the order matters.

Another option is to allow users to provide unordered Sets of TagKeys and Aggregations in Views.  Then we could return data structures that pair the inputs and outputs, i.e., we could return TagValues as a `List<Tag>` or return AggregationDatas as `Map<Aggregation, AggregationData>`.  Users wouldn't need to match the objects in the View with objects in the ViewData.

songy23: How about changing the parameter Set<> to a SortedSet<> (and also change the implementation to use SortedSet like TreeSet)? SortedSet both implies to users that the order does matter and it cannot contain duplicates.

sebright: A SortedSet sorts the items using a Comparator, so it would be harder for the user to control the order.  Then they would need to match inputs in the SortedSet with outputs in a List.  I think that could also be confusing, since the set and list order their elements in different ways (sorted vs explicit), and users might not expect the order of their elements to match.  Users could also write something like `new TreeSet<Aggregation>(Arrays.asList(Aggregation.StdDev.create(), Aggregation.Count.create()))` and accidentally expect the results to also be in that order.

songy23: Make sense to me. Java does not have such an interface as OrderedSet, and SortedSet does not quite meet our need. I would prefer to keep columns and aggregations as lists and do verification on duplicates when creating views, so that users could rely on their order to match TagKey/TagValue and Aggregation/AggregationData. In addition, ViewData will contain only the "data" part (TagValue and AggregationData), which could be better for its on-the-wire representation.

dinooliva: Keeping it as a list works for me - makes sure to document that duplicates aren't allowed in either case.

songy23: Sure, I've added a java doc for that.

core/src/main/java/io/opencensus/stats/View.java
```diff
@@ -42,20 +51,59 @@
...
+  /**
+   * Constructs a new {@link View}.
+   *
+   * @param name the {@link Name} of view. Must be unique.
+   * @param description the description of view.
+   * @param measure the {@link Measure} to be aggregated by this view.
+   * @param aggregations basic {@link Aggregation}s that this view will support. The aggregation
+   *     list should not contain duplicates.
+   * @param columns the {@link TagKey}s that this view will aggregate on. Columns should not contain
+   *     duplicate.
```
dinooliva: duplicates

songy23: Done.

songy23:
Thanks for review, going to merge soon..

PR #463, opened 2017-07-20
Modify stats-impl to use the new APIs and data models

songy23:
This PR is implementation level changes.

Modify the existing implementation to work with our new stats APIs and data models. As before, interval-based views and long measures are not supported at this time.

### DIFF
--- a/core_impl/src/main/java/io/opencensus/stats/MutableViewData.java
+++ b/core_impl/src/main/java/io/opencensus/stats/MutableViewData.java
@@ -38,6 +38,7 @@ import io.opencensus.stats.MutableAggregation.MutableMean;
 import io.opencensus.stats.MutableAggregation.MutableRange;
 import io.opencensus.stats.MutableAggregation.MutableStdDev;
 import io.opencensus.stats.MutableAggregation.MutableSum;
+import io.opencensus.stats.View.Window.Cumulative;
 import io.opencensus.stats.View.Window.Interval;
 import io.opencensus.stats.ViewData.WindowData.CumulativeData;
 import java.util.ArrayList;
@@ -67,13 +68,27 @@ final class MutableViewData {
 
   /**
    * Constructs a new {@link MutableViewData}.
+   *
+   * @param view the {@code View} linked with this {@code MutableViewData}.
+   * @param start the start {@code Timestamp}. Not needed for {@code Interval} based {@code View}s.
+   * @return a {@code MutableViewData}.
    */
-  static MutableViewData create(View view, Timestamp start) {
-    if (view.getWindow() instanceof Interval) {
-      // TODO(songya): support IntervalView.
-      throw new UnsupportedOperationException("Interval views not supported yet.");
-    }
-    return new MutableViewData(view, start);
+  static MutableViewData create(final View view, @Nullable final Timestamp start) {
+    return view.getWindow().match(
+        new Function<Cumulative, MutableViewData>() {
+          @Override
+          public MutableViewData apply(Cumulative arg) {
+            return new MutableViewData(view, start);
+          }
+        },
+        new Function<Interval, MutableViewData>() {
+          @Override
+          public MutableViewData apply(Interval arg) {
+            // TODO(songya): support IntervalView.
+            throw new UnsupportedOperationException("Interval views not supported yet.");
+          }
+        },
+        Functions.<MutableViewData>throwIllegalArgumentException());
   }
 
   /**
@@ -158,42 +173,12 @@ final class MutableViewData {
   @VisibleForTesting
   static MutableAggregation createMutableAggregation(Aggregation aggregation) {
     return aggregation.match(
-        new Function<Sum, MutableAggregation>() {
-          @Override
-          public MutableAggregation apply(Sum arg) {
-            return MutableSum.create();
-          }
-        },
-        new Function<Count, MutableAggregation>() {
-          @Override
-          public MutableAggregation apply(Count arg) {
-            return MutableCount.create();
-          }
-        },
-        new Function<Histogram, MutableAggregation>() {
-          @Override
-          public MutableAggregation apply(Histogram arg) {
-            return MutableHistogram.create(arg.getBucketBoundaries());
-          }
-        },
-        new Function<Range, MutableAggregation>() {
-          @Override
-          public MutableAggregation apply(Range arg) {
-            return MutableRange.create();
-          }
-        },
-        new Function<Mean, MutableAggregation>() {
-          @Override
-          public MutableAggregation apply(Mean arg) {
-            return MutableMean.create();
-          }
-        },
-        new Function<StdDev, MutableAggregation>() {
-          @Override
-          public MutableAggregation apply(StdDev arg) {
-            return MutableStdDev.create();
-          }
-        },
+        CreateMutableSum.INSTANCE,
+        CreateMutableCount.INSTANCE,
+        CreateMutableHistogram.INSTANCE,
+        CreateMutableRange.INSTANCE,
+        CreateMutableMean.INSTANCE,
+        CreateMutableStdDev.INSTANCE,
         Functions.<MutableAggregation>throwIllegalArgumentException());
   }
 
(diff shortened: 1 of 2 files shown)


songy23:
I've rebased this branch, should be ready for review now.

songy23:
PR still too big, will try to split it to smaller PRs.

songy23:
Please review only the last commit. The first 3 commits are included in https://github.com/census-instrumentation/opencensus-java/pull/480.

core_impl/src/main/java/io/opencensus/stats/MutableViewData.java
```diff
@@ -15,179 +15,232 @@
...
   // TODO(songya): might want to update the default tag value later.
   @VisibleForTesting
   static final TagValue UNKNOWN_TAG_VALUE = TagValue.create("unknown/not set");
 
+  private final View view;
+  private final Map<List<TagValue>, List<MutableAggregation>> tagValueAggregationMap =
+      new HashMap<List<TagValue>, List<MutableAggregation>>();
+  @Nullable private final Timestamp start;
+
+  private MutableViewData(View view, @Nullable Timestamp start) {
```
sebright: Why is the start time nullable?  I thought it was always set to the time that the View was registered.

core_impl/src/main/java/io/opencensus/stats/MutableViewData.java
```diff
@@ -15,179 +15,232 @@
...
+    }
+    return ViewData.create(view, map, CumulativeData.create(start, clock.now()));
+  }
 
-  private MutableViewData() {
+  /**
+   * Returns start timestamp for this aggregation.
+   */
+  @Nullable
+  Timestamp getStart() {
```
sebright: Same here, I'm not sure this should be nullable.

core_impl/src/main/java/io/opencensus/stats/MutableViewData.java
```diff
@@ -15,179 +15,232 @@
...
+  }
+
+  /**
+   * Create an {@link AggregationData} snapshot based on the given {@link MutableAggregation}.
+   *
+   * @param aggregation {@code MutableAggregation}
+   * @return an {@code AggregationData} which is the snapshot of current summary statistics.
+   */
+  @VisibleForTesting
+  static AggregationData createAggregationData(MutableAggregation aggregation) {
```
sebright: Suggestion: Would it be easier to add a `toAggregationData()` method to MutableAggregation?

core_impl/src/main/java/io/opencensus/stats/MutableViewData.java
songy23: I thought start time won't be necessary to IntervalView, since IntervalView only needs end time and duration. Do you think we still need to record start time for IntervalView?

core_impl/src/main/java/io/opencensus/stats/MutableViewData.java
```diff
@@ -15,179 +15,232 @@
...
+    throw new UnsupportedOperationException("Not implemented.");
+  }
 
   /**
    * Convert this {@link MutableViewData} to {@link ViewData}.
    */
-  abstract ViewData toViewData(Clock clock);
+  ViewData toViewData(Clock clock) {
+    Map<List<TagValue>, List<AggregationData>> map =
+        new HashMap<List<TagValue>, List<AggregationData>>();
```
songy23: Looks better, updated.

core_impl/src/main/java/io/opencensus/stats/MutableViewData.java
songy23: Either adding a `toAggregationData()` method or having a `createAggregationData()` method that does the type matching is fine. I slightly prefer to keep using the match-and-create method since it's more consistent with creating `MutableAggregation` from `Aggregation`.

core_impl/src/main/java/io/opencensus/stats/MutableViewData.java
sebright: That makes sense.  I was only considering the existing code, though I wonder if we should use a separate class for interval views if they need different fields and use a significantly different algorithm.

core_impl/src/main/java/io/opencensus/stats/MutableViewData.java
sebright: Okay

core_impl/src/main/java/io/opencensus/stats/MutableViewData.java
songy23: That's true, I think we might need to have a separate MutableAggregation class for interval views later. MutableAggregation stores and updates the actual stats, and its current implementation is based on cumulative values (values that will never expire).

songy23:
I've split the tests to a separate PR #489, this PR only contains changes on implementations.

songy23:
Merged #480, this PR is ready for merge now.

core_impl/src/main/java/io/opencensus/stats/MutableViewData.java
```diff
@@ -14,180 +14,233 @@
...
+  private MutableViewData(View view, @Nullable Timestamp start) {
+    this.view = view;
+    this.start = start;
+  }
+
+  /**
+   * Constructs a new {@link MutableViewData}.
+   */
+  static MutableViewData create(View view, Timestamp start) {
+    if (view.getWindow() instanceof Interval) {
```
dinooliva: Should this be using match?

core_impl/src/main/java/io/opencensus/stats/MutableViewData.java
```diff
@@ -14,180 +14,233 @@
...
   // TODO(songya): might want to update the default tag value later.
   @VisibleForTesting
   static final TagValue UNKNOWN_TAG_VALUE = TagValue.create("unknown/not set");
 
+  private final View view;
+  private final Map<List<TagValue>, List<MutableAggregation>> tagValueAggregationMap =
+      Maps.newHashMap();
+  @Nullable private final Timestamp start;
+
+  private MutableViewData(View view, @Nullable Timestamp start) {
```
dinooliva: In what cases will start be null?

core_impl/src/main/java/io/opencensus/stats/MutableViewData.java
```diff
@@ -14,180 +14,233 @@
...
+  /**
+   * Create an empty {@link MutableAggregation} based on the given {@link Aggregation}.
+   *
+   * @param aggregation {@code Aggregation}.
+   * @return an empty {@code MutableAggregation}.
+   */
+  @VisibleForTesting
+  static MutableAggregation createMutableAggregation(Aggregation aggregation) {
+    return aggregation.match(
+        new Function<Sum, MutableAggregation>() {
```
dinooliva: Should these functions be static so we don't need to create a new instance each time we create a mutable aggregation?

songy23: Make sense to me, I've changed the anonymous Functions to named static classes.

core_impl/src/main/java/io/opencensus/stats/MutableViewData.java
```diff
@@ -206,41 +191,148 @@ public MutableAggregation apply(StdDev arg) {
...
+        CreateCountData.INSTANCE,
+        CreateHistogramData.INSTANCE,
+        CreateRangeData.INSTANCE,
+        CreateMeanData.INSTANCE,
+        CreateStdDevData.INSTANCE);
+  }
+  
+  // static inner Function classes
+  
+  private static final class CreateMutableSum implements Function<Sum, MutableAggregation> {
```
sebright: It would be nice to be able to use the MutableAggregation subclass for the second type argument here.  I wonder if we should change the match method signature to accept more return types, like this:

```java
  public abstract <T> T match(
      Function<? super Sum, ? extends T> p0,
      Function<? super Count, ? extends T> p1,
      ...
);
```

I don't know whether that would cause any problems with type inference, though.  I'm not suggesting it for this PR.

core_impl/src/main/java/io/opencensus/stats/MutableViewData.java
```diff
@@ -206,41 +191,148 @@ public MutableAggregation apply(StdDev arg) {
...
+    @Override
+    public MutableAggregation apply(Sum arg) {
+      return MutableSum.create();
+    }
+    
+    private CreateMutableSum() {}
```
sebright: Suggestion: I don't think the constructors are necessary.

songy23: Correct, these are private static classes. Removed.

core_impl/src/main/java/io/opencensus/stats/MutableViewData.java
dinooliva: As discussed, this approach won't work if/when we extend the WindowData alternatives. Still, since this is package-private, I'm okay with it for now.

core_impl/src/main/java/io/opencensus/stats/MutableViewData.java
songy23: Just saw an argument against this from Effective Java, maybe worth considering:

> Do not use wildcard types as return types. Rather than providing additional flexibility for your users, it would force them to use wildcard types in client code.

> Properly used, wildcard types are nearly invisible to users of a class. They cause methods to accept the parameters they should accept and reject those they should reject. If the user of a class has to think about wildcard types, there is probably something wrong with the class’s API.

Also refer to https://stackoverflow.com/questions/34641441/using-generic-wildcard-types-in-return-parameters.

cc @dinooliva

songy23: Talked offline, it seems the argument above doesn't quite apply to our case here. PECS (Producer Extends Consumer Super) may be a more concrete argument that we should go with `? extends T` as the second argument here. (reference: https://stackoverflow.com/questions/2723397/what-is-pecs-producer-extends-consumer-super)

sebright: Thanks for looking into this.  I tried adding `? extends` to the TagKey `match` method to test that it could accept a wider range of arguments.  I changed the signature to

```java
  public abstract <T> T match(
      Function<? super TagKeyString, ? extends T> stringFunction,
      Function<? super TagKeyLong, ? extends T> longFunction,
      Function<? super TagKeyBoolean, ? extends T> booleanFunction,
      Function<? super TagKey, ? extends T> defaultFunction);
```

 Then I was able to write this code that uses it:

```java
  static Number method() {
    TagKey key = TagKeyString.create("key");
    Function<TagKeyString, ? extends Number> f =
        new Function<TagKey.TagKeyString, Double>() {

          @Override
          public Double apply(TagKeyString arg) {
            return 1.0;
          }
        };
    return key.match(
        f,
        Functions.<Number>returnConstant(2),
        Functions.<Long>returnConstant(3L),
        Functions.<Integer>returnConstant(4));
  }
```

It compiled with Java 6 and 8.  The method call shows that `match` can take functions returning the super class (`Number`), subclasses (`Long` and `Integer`), and a bounded wildcard (`? extends Number`).

PR #464, opened 2017-07-20
WIP: Add TagsComponent and TagContextFactory, and make TagContext abstract.

sebright:
/cc @dinooliva @songy23 

I still need to clean this up and add more tests.

### DIFF
--- a/core/src/main/java/io/opencensus/tags/TagContextFactory.java
+++ b/core/src/main/java/io/opencensus/tags/TagContextFactory.java
@@ -13,10 +13,121 @@
 
 package io.opencensus.tags;
 
+import io.opencensus.common.Scope;
+import io.opencensus.tags.TagKey.TagKeyBoolean;
+import io.opencensus.tags.TagKey.TagKeyLong;
+import io.opencensus.tags.TagKey.TagKeyString;
+
 /**
  * Factory for new {@link TagContext}s and {@code TagContext}s based on the current context.
+ *
+ * <p>This class returns {@link Builder builders} that can be used to create the
+ * implementation-dependent {@link TagContext}s.
  */
 // TODO(sebright): Pick a more descriptive name for this class.
 public abstract class TagContextFactory {
-  // TODO(sebright): Add TagContext related methods to this class.
+
+  /**
+   * Returns an empty {@code TagContext}.
+   *
+   * @return an empty {@code TagContext}.
+   */
+  public abstract TagContext empty();
+
+  /**
+   * Returns the current {@code TagContext}.
+   *
+   * @return the current {@code TagContext}.
+   */
+  // TODO(sebright): Should we let the implementation override this method?
+  public final TagContext getCurrentTagContext() {
+    TagContext tags = CurrentTagContextUtils.getCurrentTagContext();
+    return tags == null ? empty() : tags;
+  }
+
+  /**
+   * Returns a new empty {@code Builder}.
+   *
+   * @return a new empty {@code Builder}.
+   */
+  public abstract Builder emptyBuilder();
+
+  /**
+   * Returns a builder based on this {@code TagContext}.
+   *
+   * @return a builder based on this {@code TagContext}.
+   */
+  public abstract Builder toBuilder(TagContext tags);
+
+  /**
+   * Returns a new {@code Builder} created from the current {@code TagContext}.
+   *
+   * @return a new {@code Builder} created from the current {@code TagContext}.
+   */
+  public final Builder currentBuilder() {
+    return toBuilder(getCurrentTagContext());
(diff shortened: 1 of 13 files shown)


core/src/main/java/io/opencensus/tags/TagContextFactory.java
```diff
@@ -0,0 +1,127 @@
...
+ */
+
+package io.opencensus.tags;
+
+import io.opencensus.common.Scope;
+import io.opencensus.tags.TagKey.TagKeyBoolean;
+import io.opencensus.tags.TagKey.TagKeyLong;
+import io.opencensus.tags.TagKey.TagKeyString;
+
+/** Factory for new {@link TagContext}s and {@code TagContext}s based on the current context. */
```
songy23: Guess should be `"Factory for new {@link TagContext}s and {@code Builder}s..."` ?

core/src/main/java/io/opencensus/tags/TagsComponent.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package io.opencensus.tags;
+
+/**
+ * Class that holds the implementations for {@link TagContextFactory}.
+ *
+ * <p>All objects returned by methods on {@code TagsComponent} are cacheable.
+ */
+public abstract class TagsComponent {
```
songy23: Nit: should we name this `TagComponent`? It feels strange to have `TagsComponent` (plural) and `TagContextFactory`,

core/src/main/java/io/opencensus/tags/TagContextFactory.java
```diff
@@ -0,0 +1,127 @@
...
+   * @return a new empty {@code Builder}.
+   */
+  public abstract Builder emptyBuilder();
+
+  /**
+   * Returns a builder based on this {@code TagContext}.
+   *
+   * @return a builder based on this {@code TagContext}.
+   */
+  public abstract Builder toBuilder(TagContext tags);
```
songy23: Shall we make this package private? It seems to me `emptyBuilder()` and `currentBuilder()` have covered all use cases.

sebright: This method is abstract, so it needs to be public to allow classes outside the package to subclass TagContextFactory.

adriancole:
needs an intro (maybe top-level readme or package.java doc)

Particularly, if nothing but in the PR description I'm curious of the aim of this (more than the fact it is happening). I'd guess this is to allow tag propagation to work independent of api like tracer or stats, but want to not guess :)

sebright:
@adriancole This PR addresses the comments about making TagContext abstract in #448.  I still need to finish it and write a complete description in the commit message.  I can also add a package-info.java for the tags package.

adriancole:
Thanks.. I lose track and saw new files which made me think this was new. You answered me on the other one wrt overall:

>We are replacing StatsContext with TagContext. The difference is that TagContext doesn't depend on the stats package, so it could be used to label anything associated with an operation, and it can have multiple types of tag values (String, long, boolean).

core/src/main/java/io/opencensus/tags/TagUtils.java
```diff
@@ -0,0 +1,41 @@
...
+        }
+      };
+
+  /**
+   * A {@code Function} that always ignores its argument and throws an {@link AssertionError}.
+   *
+   * @return a {@code Function} that always ignores its argument and throws an {@link
+   *     AssertionError}.
+   */
+  static <T> Function<Object, T> throwAssertionError() {
```
songy23: Should this method be under io.opencensus.common.Functions instead?

sebright: I'm not sure.  `Functions` already has `throwIllegalArgumentException`.  I figured that it is better to hide the method for now, and expose it later if we need to.

Do you think users will want to assert that all `match` cases are handled?  It's safe for this library to make the assertion, because there is no possibility of a version mismatch.

songy23: Hiding this method sounds good to me, since I'm not quite sure about the use case either. Maybe having `throwIllegalArgumentException` and `throwAssertionError` together at `Functions` will cause confusion at user side, since it will be unclear which one is preferable for the default/undesired match case.

sebright: Thanks.  I'll leave it in the internal class for now.

sebright:
This PR is too large, so I'll break it up.  The first part is #465.

sebright:
I rebased onto #465.

sebright:
Subsumed by #485.

PR #465, opened 2017-07-21
Add Tags and TagsComponent classes to the 'tags' package.

sebright:
TagsComponent is an abstract class that holds the implementations for different
parts of the 'tags' package.  The class currently has one field, which is a
skeleton for a new TagContextFactory class.  It allows the implementation
library to determine the implementation of TagContext, which could be important
for performance.

This commit also adds a Tags class, which instantiates the TagsComponent with
reflection.  The design is very similar to the 'stats' package.  Tags
corresponds to the Stats class, and TagsComponent corresponds to StatsComponent.

___________________________________________________________________

/cc @adriancole

This PR is part of #464.

### DIFF
--- a/core/src/main/java/io/opencensus/tags/TagContextFactory.java
+++ b/core/src/main/java/io/opencensus/tags/TagContextFactory.java
@@ -16,6 +16,7 @@ package io.opencensus.tags;
 /**
  * Factory for new {@link TagContext}s and {@code TagContext}s based on the current context.
  */
+// TODO(sebright): Pick a more descriptive name for this class.
 public abstract class TagContextFactory {
   // TODO(sebright): Add TagContext related methods to this class.
 }
--- a/core/src/main/java/io/opencensus/tags/TagContextFactory.java
+++ b/core/src/main/java/io/opencensus/tags/TagContextFactory.java
@@ -17,4 +17,5 @@ package io.opencensus.tags;
  * Factory for new {@link TagContext}s and {@code TagContext}s based on the current context.
  */
 public abstract class TagContextFactory {
+  // TODO(sebright): Add TagContext related methods to this class.
 }
--- a/core/src/main/java/io/opencensus/tags/TagContextFactory.java
+++ b/core/src/main/java/io/opencensus/tags/TagContextFactory.java
@@ -1,0 +1,20 @@
+/*
+ * Copyright 2017, Google Inc.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.tags;
+
+/**
+ * Factory for new {@link TagContext}s and {@code TagContext}s based on the current context.
+ */
+public abstract class TagContextFactory {
+}
--- a/core/src/main/java/io/opencensus/tags/Tags.java
+++ b/core/src/main/java/io/opencensus/tags/Tags.java
@@ -34,7 +34,7 @@ public final class Tags {
    * @return the default {@code TagContextFactory}.
    */
   public static TagContextFactory getTagContextFactory() {
-    return tagsComponent.getTagContextFactory();
+    return tagsComponent == null ? null : tagsComponent.getTagContextFactory();
   }
 
   // Any provider that may be used for TagsComponent can be added here.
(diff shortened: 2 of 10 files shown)


core/src/main/java/io/opencensus/tags/TagContextFactory.java
```diff
@@ -0,0 +1,20 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.tags;
+
+/**
+ * Factory for new {@link TagContext}s and {@code TagContext}s based on the current context.
+ */
+public abstract class TagContextFactory {
```
bogdandrutu: What about TagContextBuilder?

Also where are you going to have a method like:
Scope withTagContext(TagContext)?

Does it make sense to have a Tagger as well (similar with Tracer)?

sebright: > What about TagContextBuilder?

The builder is a nested class in TagContextFactory, in #464.

> Also where are you going to have a method like:
Scope withTagContext(TagContext)?

I forgot about that.  I think it could also belong in TagContextFactory.

> Does it make sense to have a Tagger as well (similar with Tracer)?

I think that TagContextFactory corresponds to Tracer in the tracing package.  Tagger might be a more descriptive name, though.

bogdandrutu: I would like to name it Tagger for more consistency if possible.

sebright: I renamed TagContextFactory to Tagger.  I don't like the new name as much as Tracer, but it is more accurate than TagContextFactory, and I don't have any other ideas.

core_impl/src/main/java/io/opencensus/tags/TagsComponentImplBase.java
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.tags;
+
+/** Base implementation of {@link TagsComponent}. */
+public abstract class TagsComponentImplBase extends TagsComponent {
+  private final Tagger tagContextFactory = new TaggerImpl();
```
songy23: Minor: tagContextFactory -> tagger

core/src/main/java/io/opencensus/tags/TagContextFactory.java
sebright: @dinooliva also wanted to change the class name.  Since the main purpose of adding TagContextFactory in this PR is to test that the TagsComponent is instantiated correctly (in TagsTest), I removed the commit that renamed TagContextFactory and added a TODO to come up with a better name.  I'll rename the class and add more methods to it in a later PR.

core_impl/src/main/java/io/opencensus/tags/TagsComponentImplBase.java
sebright: I renamed the class back to TagContextFactory for now.

core/src/main/java/io/opencensus/tags/TagContextFactory.java
```diff
@@ -0,0 +1,22 @@
...
+ * limitations under the License.
+ */
+
+package io.opencensus.tags;
+
+/**
+ * Factory for new {@link TagContext}s and {@code TagContext}s based on the current context.
+ */
+// TODO(sebright): Pick a more descriptive name for this class.
+public abstract class TagContextFactory {
```
dinooliva: I think this is okay. You could consider something like TagContextManager.

core/src/main/java/io/opencensus/tags/TagsComponent.java
```diff
@@ -0,0 +1,25 @@
...
+package io.opencensus.tags;
+
+/**
+ * Class that holds the implementation for {@link TagContextFactory}.
+ *
+ * <p>All objects returned by methods on {@code TagsComponent} are cacheable.
+ */
+public abstract class TagsComponent {
+
+  /** Returns the default {@link TagContextFactory}. */
```
dinooliva: What do you mean by 'default TagContextFactory'?

core/src/main/java/io/opencensus/tags/Tags.java
```diff
@@ -0,0 +1,66 @@
...
+  private Tags() {}
+
+  public static TagContextFactory getTagContextFactory() {
+    return tagsComponent.getTagContextFactory();
+  }
+
+  // Any provider that may be used for TagsComponent can be added here.
+  @VisibleForTesting
+  @Nullable
+  static TagsComponent loadTagsComponent(ClassLoader classLoader) {
```
adriancole: possible to have a generic utility extracted if this is copy/pasted in places?

sebright: I tried refactoring this method, but it didn't work very well.  I think that the most repetitive part is handling the ClassNotFoundException that can be thrown by `Class.forName`.  However, `Class.forName` needs to be called with a literal String so that it can be modified by shading tools, so it can't be moved to a helper method.  I don't think that there is enough remaining code to split out.

sebright:
I rebased onto master in order to get #474.

sebright:
Thanks!

PR #468, opened 2017-07-24
Add a no-op StatsComponent to return when no implementation is available.

sebright:
The no-op StatsComponent returns a no-op StatsRecorder and a null ViewManager.
The StatsRecorder is non-null so that code that only records stats doesn't need
to check for null.  NoopStatsRecorder.record(...) simply ignores the stats.
However, the ViewManager is null to avoid returning misleading empty results
from NoopStatsComponent.getViewManager().getView(...).

### DIFF
--- a/core/src/main/java/io/opencensus/stats/StatsComponent.java
+++ b/core/src/main/java/io/opencensus/stats/StatsComponent.java
@@ -52,6 +52,7 @@ public abstract class StatsComponent {
     @Override
     @Nullable
     public ViewManager getViewManager() {
+      // TODO(sebright): Decide whether to also provide a no-op implementation for the ViewManager.
       return null;
     }
 
--- a/core/src/main/java/io/opencensus/stats/StatsComponent.java
+++ b/core/src/main/java/io/opencensus/stats/StatsComponent.java
@@ -13,15 +13,20 @@
 
 package io.opencensus.stats;
 
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.Immutable;
+
 /**
  * Class that holds the implementations for {@link ViewManager}, {@link StatsRecorder}, and {@link
  * StatsContextFactory}.
  *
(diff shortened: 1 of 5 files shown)


core/src/main/java/io/opencensus/stats/StatsComponent.java
```diff
@@ -13,21 +13,57 @@
...
+    return NOOP_STATS_COMPONENT;
+  }
+
+  @Immutable
+  private static final class NoopStatsComponent extends StatsComponent {
+
+    @Override
+    @Nullable
+    public ViewManager getViewManager() {
+      return null;
```
bogdandrutu: Probably we should add a todo to also provide a no-op for this.

sebright: I'm not sure if we should create a no-op ViewManager, because it could give the user incorrect results, but I added a TODO for us to decide whether to add one.

core/src/main/java/io/opencensus/stats/StatsComponent.java
```diff
@@ -13,21 +13,58 @@
...
+
+    @Override
+    public StatsRecorder getStatsRecorder() {
+      return StatsRecorder.getNoopStatsRecorder();
+    }
+
+    @Override
+    @Nullable
+    StatsContextFactory getStatsContextFactory() {
+      return null;
```
songy23: Should StatsContextFactory also have a no-op?

sebright: getStatsContextFactory() is going to be removed when we replace StatsContext with TagContext, so I would rather leave it null until then.  We will need something similar in the `tags` package after #465 is merged, though.

sebright: See #476.

PR #469, opened 2017-07-24
Add a function that always throws an AssertionError.

sebright:
The function is internal for now, in io.opencensus.internal.InternalFunctions.
It can be used for all default cases in calls to "match" functions in the API
library, since a version mismatch isn't possible, and the default case cannot
occur in the "match" function that takes all subclasses.

### DIFF
--- a/api/src/main/java/io/opencensus/internal/InternalFunctions.java
+++ b/api/src/main/java/io/opencensus/internal/InternalFunctions.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2017, Google Inc.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


api/src/main/java/io/opencensus/internal/InternalFunctions.java
```diff
@@ -0,0 +1,41 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.internal;
+
+import io.opencensus.common.Function;
+
```
songy23: Documentation?

api/src/main/java/io/opencensus/internal/InternalFunctions.java
```diff
@@ -0,0 +1,41 @@
...
+public final class InternalFunctions {
```
songy23: minor: Shall we remove the prefix `Internal` like other classes under internal package?

sebright:
I made a new PR that adds throwAssertionError() to the existing Functions class.  I think it will be simpler to have all of the convenience Function instances in one place.

PR #470, opened 2017-07-24
Add a function that always throws an AssertionError.

sebright:
The function can be used for all default cases in calls to "match" functions in
the API library, since a version mismatch isn't possible, and the default case
cannot occur in "match" functions that take all subclasses.

### DIFF
--- a/api/src/main/java/io/opencensus/common/Functions.java
+++ b/api/src/main/java/io/opencensus/common/Functions.java
@@ -33,6 +33,14 @@ public final class Functions {
         }
       };
 
+  private static final Function<Object, Void> THROW_ASSERTION_ERROR =
+      new Function<Object, Void>() {
+        @Override
+        public Void apply(Object ignored) {
+          throw new AssertionError();
+        }
+      };
+
   /**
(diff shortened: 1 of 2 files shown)


api/src/main/java/io/opencensus/common/Functions.java
```diff
@@ -72,4 +80,17 @@ public T apply(Object ignored) {
...
   }
+
+  /**
+   * A {@code Function} that always ignores its argument and throws an {@link AssertionError}.
+   *
+   * @return a {@code Function} that always ignores its argument and throws an {@code
+   *     AssertionError}.
+   */
+  public static <T> Function<Object, T> throwAssertionError() {
+    // It is safe to cast a producer of Void to anything, because Void is always null.
```
adriancole: I like explaining, but isn't this focusing on the wrong thing? ex you could have object result. the fact that assertion error is raised always negates the need to explain return type choices, right?

sebright: I was trying to limit the scope of the warning suppression and the code that I explained to the one line.  I see that it's a little misleading, though, since the function never returns null!  I reworded it.

PR #471, opened 2017-07-24
Avoid double duty context propagation for Executors.

ubschmidt2:


adriancole:
Looks good. Would be neat if we could test by inspecting the state of context or stack trace from within a grpc instrumented executor

ubschmidt2:
Added a test. Note that I had to leave a TODO as it turned out that there's another unneeded context propagation that could be removed. I'd like to postpone this to a later PR as I first want to change the interface of the *Instrumentation classes to take an AgentBuilder.

PR #476, opened 2017-07-26
Add a no-op TagsComponent to return when no implementation is available.

sebright:
The no-op TagsComponent will allow users to record stats without checking for
null.  If no implementation is available, the tagging operations should be very
fast, because the no-op TagsComponent contains a no-op TagContextFactory, which
will only ever return an empty TagContext.

__________________________________________

~~This PR contains #465, so only the last commit is new.~~


sebright:
@songy23 Yes.  #464 has the rough changes that I'm planning on making, but I need to break it up into smaller PRs.

sebright:
I rebased this onto master, since #465 was merged.

PR #480, opened 2017-07-27
Add MutableAggregation and tests.

songy23:
This PR is implementation level changes.

MutableAggregation is the implementation class that does the actual recording and updates the corresponding summary stats (count, sum, etc.), similar to [MutableDistribution](https://github.com/census-instrumentation/opencensus-java/blob/master/core/src/main/java/io/opencensus/stats/MutableDistribution.java) in our previous design.

### DIFF
--- a/core_impl/src/main/java/io/opencensus/stats/MutableAggregation.java
+++ b/core_impl/src/main/java/io/opencensus/stats/MutableAggregation.java
@@ -17,6 +17,8 @@ import static com.google.common.base.Preconditions.checkNotNull;
 
 import io.opencensus.common.Function;
 
+import java.util.Arrays;
+
 /** Mutable version of {@link Aggregation} that supports adding values. */
 abstract class MutableAggregation {
 
@@ -39,8 +41,7 @@ abstract class MutableAggregation {
       Function<? super MutableHistogram, T> p2,
       Function<? super MutableRange, T> p3,
       Function<? super MutableMean, T> p4,
-      Function<? super MutableStdDev, T> p5,
-      Function<? super MutableAggregation, T> defaultFunction);
+      Function<? super MutableStdDev, T> p5);
 
   /** Calculate sum on aggregated {@code MeasureValue}s. */
   static final class MutableSum extends MutableAggregation {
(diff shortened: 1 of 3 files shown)


core_impl/src/main/java/io/opencensus/stats/MutableAggregation.java
```diff
@@ -0,0 +1,348 @@
...
+   * Applies the given match function to the underlying data type.
+   */
+  abstract <T> T match(
+      Function<? super MutableSum, T> p0,
+      Function<? super MutableCount, T> p1,
+      Function<? super MutableHistogram, T> p2,
+      Function<? super MutableRange, T> p3,
+      Function<? super MutableMean, T> p4,
+      Function<? super MutableStdDev, T> p5,
+      Function<? super MutableAggregation, T> defaultFunction);
```
sebright: I don't think we need a default case, because this method is internal, and it can be updated when the subclasses are updated.

core_impl/src/main/java/io/opencensus/stats/MutableAggregation.java
```diff
@@ -0,0 +1,348 @@
...
+      bucketCounts[bucketCounts.length - 1]++;
+    }
+
+    /**
+     * Returns the aggregated bucket count.
+     *
+     * @return the aggregated bucket count.
+     */
+    long[] getBucketCounts() {
+      return bucketCounts;
```
sebright: I think we should copy the bucket counts to be safe.

core_impl/src/main/java/io/opencensus/stats/MutableAggregation.java
```diff
@@ -0,0 +1,348 @@
...
+      return new MutableStdDev();
+    }
+
+    @Override
+    void add(double value) {
+      count++;
+      double deltaFromMean = value - mean;
+      mean += deltaFromMean / count;
+      double deltaFromMean2 = value - mean;
+      sumOfSquaredDeviations += deltaFromMean * deltaFromMean2;
```
sebright: Could you please add the note about the algorithm from MutableDistribution.java?

core_impl/src/main/java/io/opencensus/stats/MutableAggregation.java
songy23: Make sense, removed

core_impl/src/main/java/io/opencensus/stats/MutableAggregation.java
songy23: Done.

core_impl/src/main/java/io/opencensus/stats/MutableAggregation.java
songy23: Sure, I've copied the explanation here.

PR #483, opened 2017-07-27
Move tagging implementation classes into io.opencensus.impl.tags. (part of #379)

sebright:
This change will ensure that the 'tags' implementation only accesses the API
classes through their public API.  That will help us maintain backwards
compatibility.

EDIT: We should probably also put the core, Java 7/8, and Android implementation classes in different packages before the release.


sebright:
This change only affects the package names of the implementation classes.  It doesn't affect the API or top-level directory structure.

PR #484, opened 2017-07-27
Add initial package-info.java to the 'tags' package.

sebright:


core/src/main/java/io/opencensus/tags/package-info.java
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * API for associating tags with scoped operations.
+ *
+ * <p>This package manages a set of tags in the {@code io.grpc.Context}. The tags can be used to
+ * label anything that is associated with a specific operation. For example, the {@link
+ * io.opencensus.stats} package labels all measurements with the current tags.
+ *
+ * <p>{@link io.opencensus.tags.Tag Tags} are key-value pairs. The {@link io.opencensus.tags.TagKey
+ * keys} are wrapped {@code String}s, but the values can have multiple types, such as {@code
```
songy23: Shall we mention any restriction on TagKey/TagValue in this doc?

sebright: I'm not sure how much detail we should give, since this links to the docs for the main tag classes.  We should at least add that information to the class-level Javadocs for `TagKey` and `TagValueString`, though.

sebright: I added more documentation to TagKey and TagValueString in #486.

core/src/main/java/io/opencensus/tags/package-info.java
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * API for associating tags with scoped operations.
+ *
+ * <p>This package manages a set of tags in the {@code io.grpc.Context}. The tags can be used to
+ * label anything that is associated with a specific operation. For example, the {@link
+ * io.opencensus.stats} package labels all measurements with the current tags.
```
songy23: Nitpick: measurements -> {@code io.opencensus.stats.Measurement}s?

core/src/main/java/io/opencensus/tags/package-info.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * <p>{@link io.opencensus.tags.Tag Tags} are key-value pairs. The {@link io.opencensus.tags.TagKey
+ * keys} are wrapped {@code String}s, but the values can have multiple types, such as {@code
+ * String}, {@code long}, and {@code boolean}. They are stored as a map in a {@link
+ * io.opencensus.tags.TagContext}.
+ */
+// TODO(sebright): Add code examples after the API is updated to use a TagContext factory.
```
adriancole: maybe note that trace IDs (or encoded trace context) are not considered tags? (and anything else that we might be propagating that aren't tags that I'm not aware of)

core/src/main/java/io/opencensus/tags/package-info.java
sebright: I wanted to give a general example of a use case for labeling things with tags, so I wasn't referring to a specific class.  I changed it to "stats" to make it more general.

core/src/main/java/io/opencensus/tags/package-info.java
sebright: I added a note.  SpanContext and TagContext are the only things that this library propagates in the gRPC context.

Issue #224, opened 2017-04-14
Fix a flaky test in StatsContextTest.java


songy23:
https://github.com/google/instrumentation-java/blob/master/core_impl/src/test/java/com/google/instrumentation/stats/StatsContextTest.java#L126

This test is flaky because we are storing the tags in a Map, and the order of tags could be random on different machines. Yet we're comparing the expected and actual serialized tags under the assumption that they have the same order. Needs to figure out a better way for this test.

PR #485, opened 2017-07-27
Add TagContextFactory and TagContextBuilder, and make TagContext abstract.

sebright:
This commit splits TagContext into an abstract class in the API library and an
implementation (TagContextImpl) in the implementation library. TagContextFactory
has methods for creating TagContexts and working with the current TagContext.
(TagContextFactory should probably be renamed, because it has more functionality
than a factory.)

Advantages of making TagContext abstract and moving its implementation to the
core implementation library:

  - The `stats` and `tags` implementation classes are in the same library.  That
    means that the `stats` package is free to access TagContextImpl's internal
    state without any risk of compatibility issues.  There are no restrictions
    on optimizations to TagContextImpl's internal state or the `tags` package's
    use of that state.

  - The `tags` package can provide a no-op TagContext implementation when no
    `tags` implementation library is available.  The no-op implementation could
    reduce the overhead of setting tags and recording stats when the stats
    aren't needed.

  - The factory class gives us more flexibility to change TagContext in the
    future.

Changes to the API:

  - TagContexts must be created through builders provided by the
    TagContextFactory, which is contained in the TagsComponent.  Previously,
    TagContext had a method for creating builders.

  - TagContextBuilder is a top-level class, instead of a nested class within
    TagContext.

  - TagContext implements `Iterable<Tag>`.  `TagContext.iterator()` is necessary
    because the `stats` and `tags` packages have some methods that take values
    of type TagContext.  It is possible that the TagContext is not a
    TagContextImpl, and, in that case, the code can't use implementation-specific
    methods to get the tags.  TagContext must provide a public accessor.

_____________________________________________________________

/cc @adriancole


sebright:
I rebased onto master after #483.  I also simplified the PR by removing all of the public methods that work with the current TagContext.

core/src/main/java/io/opencensus/tags/TagContextBuilder.java
```diff
@@ -0,0 +1,109 @@
...
+public abstract class TagContextBuilder {
+  private static final TagContextBuilder NOOP_TAG_CONTEXT_BUILDER = new NoopTagContextBuilder();
+
+  /**
+   * Adds the key/value pair regardless of whether the key is present.
+   *
+   * @param key the {@code TagKey} which will be set.
+   * @param value the value to set for the given key.
+   * @return this
+   * @throws IllegalArgumentException if either argument is null.
```
adriancole: Probably comment made elsewhere but NPE is the usual exception for null

core/src/main/java/io/opencensus/tags/TagContextBuilder.java
```diff
@@ -0,0 +1,109 @@
...
+ */
+
+package io.opencensus.tags;
+
+import io.opencensus.tags.TagKey.TagKeyBoolean;
+import io.opencensus.tags.TagKey.TagKeyLong;
+import io.opencensus.tags.TagKey.TagKeyString;
+import javax.annotation.concurrent.Immutable;
+
+/** Builder for the {@link TagContext} class. */
```
adriancole: Maybe make a note that the keys are non simple types because X and intended to be reused. This might help tackle concerns around perceived complexity here. That or say the key types are explaine here (link tag)

core/src/main/java/io/opencensus/tags/TagContextBuilder.java
```diff
@@ -0,0 +1,109 @@
...
+   * @param key the {@code TagKey} which will be set.
+   * @param value the value to set for the given key.
+   * @return this
+   * @throws IllegalArgumentException if the key is null.
+   */
+  // TODO(sebright): Make this public once we support types other than String.
+  protected abstract TagContextBuilder set(TagKeyBoolean key, boolean value);
+
+  /**
+   * Removes the key if it exists.
```
adriancole: Nit: This style of javadoc is burdensome to read as it just restates the method in english. Id prefer we make docs lines and comments count, as useless comments draw the same attention as useful ones.

core/src/main/java/io/opencensus/tags/TagContextBuilder.java
```diff
@@ -0,0 +1,109 @@
...
+
+  /**
+   * Adds the key/value pair regardless of whether the key is present.
+   *
+   * @param key the {@code TagKey} which will be set.
+   * @param value the value to set for the given key.
+   * @return this
+   * @throws IllegalArgumentException if either argument is null.
+   */
+  public abstract TagContextBuilder set(TagKeyString key, TagValueString value);
```
adriancole: Why is this public while others protected? Maybe make a comment if intentional

core/src/main/java/io/opencensus/tags/TagContextFactory.java
```diff
@@ -16,13 +16,35 @@
...
 
 /**
- * Factory for new {@link TagContext}s and {@code TagContext}s based on the current context.
+ * Factory for new {@link TagContext}s.
+ *
+ * <p>This class returns {@link TagContextBuilder builders} that can be used to create the
+ * implementation-dependent {@link TagContext}s.
  */
 // TODO(sebright): Pick a more descriptive name for this class.
 public abstract class TagContextFactory {
```
adriancole: TagContexts (if a utility bag)

sebright:
I also rebased onto master.

adriancole:
>
> /** Removes the key if it exists and returns the {@code this} reference.
> */
> public abstract TagContextBuilder clear(TagKey key);

I guess I feel like things that are surprising should go into docs. ex
builders usually return this (vs a copy of this). I'm not sure what else
clear would do besides removing the key if it exists. I'm not sure I've
ever encountered someone asking what a builder method does, but maybe
others have had questions and that's why we are a bit explicit.

core_impl/src/test/java/io/opencensus/tags/UnreleasedApiAccessor.java
```diff
@@ -11,9 +11,23 @@
...
-import io.opencensus.tags.TagContextFactory;
+import io.opencensus.tags.TagKey.TagKeyBoolean;
+import io.opencensus.tags.TagKey.TagKeyLong;
 
-final class TagContextFactoryImpl extends TagContextFactory {
+/**
+ * Class for accessing methods in the {@code io.opencensus.tags} package that haven't been made
+ * public yet.
+ */
+public final class UnreleasedApiAccessor {
```
songy23: Put this class in unsafe?

core_impl/src/test/java/io/opencensus/impl/tags/TagContextImplTest.java
```diff
@@ -0,0 +1,99 @@
...
+  @Test
+  public void testClear() {
+    TagContext tags = tagContexts.emptyBuilder().set(KS1, V1).build();
+    assertThat(asList(tagContexts.toBuilder(tags).clear(KS1).build())).isEmpty();
+    assertThat(asList(tagContexts.toBuilder(tags).clear(KS2).build()))
+        .containsExactly(TagString.create(KS1, V1));
+  }
+
+  @Test
+  public void disallowCallingRemoveOnIterator() {
```
songy23: Add tests on next() and hasNext() also?

core/src/main/java/io/opencensus/tags/Tag.java
```diff
@@ -142,8 +142,7 @@ public static TagString create(TagKeyString key, TagValueString value) {
      * @param value the tag value.
      * @return a {@code TagLong} with the given key and value.
      */
-    // TODO(sebright): Make this public once we support types other than String.
-    static TagLong create(TagKeyLong key, long value) {
+    public static TagLong create(TagKeyLong key, long value) {
```
sebright: I realized that there is no reason to hide any of the Tag factory methods, because we already hide the TagKey factory methods for the unsupported types, which means that it still isn't possible to create a `TagLong` or `TagBoolean`.

Since our documentation already explains that tags can have different types, I found that it was easier to expose these methods so that the documentation can reference them.  Another reason to expose all Tag factory methods is that they are needed by the `tags` implementation, which is in a different package.

core/src/main/java/io/opencensus/tags/TagContext.java
```diff
@@ -30,145 +23,35 @@
...
-  private final Map<TagKey, Object> tags;
-
-  TagContext(Map<? extends TagKey, ?> tags) {
-    this.tags = Collections.unmodifiableMap(new HashMap<TagKey, Object>(tags));
-  }
-
-  Map<TagKey, Object> getTags() {
-    return tags;
-  }
+public abstract class TagContext implements Iterable<Tag> {
```
sebright: I changed it to `unsafeGetIterator()` for now and left the TODOs about possibly removing the ability to iterate over the tags.

core/src/main/java/io/opencensus/tags/TagContexts.java
```diff
@@ -0,0 +1,75 @@
...
+   * @return an empty {@code TagContext}.
+   */
+  public abstract TagContext empty();
+
+  /**
+   * Returns a new empty {@code Builder}.
+   *
+   * @return a new empty {@code Builder}.
+   */
+  public abstract TagContextBuilder emptyBuilder();
```
sebright: I added the default implementation but also left the override of `emptyBuilder` in TagContextsImpl.  `TagContextsImpl.emptyBuiler()` just calls the constructor of the builder and doesn't need to do an `instanceof` check.

core_impl/src/test/java/io/opencensus/tags/UnreleasedApiAccessor.java
sebright: This class needs to be in the same package as the tags API to access the non-public methods.  Also, I don't think it needs a warning, since it is just a test class and it won't be used outside of this project.

core_impl/src/test/java/io/opencensus/impl/tags/TagContextImplTest.java
sebright: I added a test.

sebright:
@adriancole I opened an issue (#494) to discuss changing the Javadoc policy to not require all tags, which I thought would be a good first step to making the Javadocs more concise.  If enough people agree to change it, then I'll update these Javadocs and also try to shorten the summary fragments.

core_impl/src/test/java/io/opencensus/impl/tags/TagContextImplTest.java
```diff
@@ -86,6 +89,19 @@ public void testClear() {
   }
 
   @Test
+  public void testIterator() {
+    TagContext tags = tagContexts.emptyBuilder().set(KS1, V1).set(KS2, V2).build();
+    Iterator<Tag> i = tags.unsafeGetIterator();
+    assertTrue(i.hasNext());
+    assertThat(i.next()).isEqualTo(TagString.create(KS1, V1));
```
songy23: The order of elements to be iterated is non-deterministic, similar to https://github.com/census-instrumentation/opencensus-java/issues/224.

sebright: Oops!  Thanks for catching that.

adriancole:
@adriancole <https://github.com/adriancole> I opened an issue (#494
<https://github.com/census-instrumentation/opencensus-java/issues/494>) to
discuss changing the Javadoc policy to not require all tags, which I
thought would be a good first step to making the Javadocs more concise. If
enough people agree to change it, then I'll update these Javadocs and also
try to shorten the summary fragments.

Thanks for the followup

core/src/main/java/io/opencensus/tags/TagContexts.java
```diff
@@ -0,0 +1,72 @@
...
+   * @return an empty {@code TagContext}.
+   */
+  public abstract TagContext empty();
+
+  /**
+   * Returns a new empty {@code Builder}.
+   *
+   * @return a new empty {@code Builder}.
+   */
+  public TagContextBuilder emptyBuilder() {
```
sebright: Making this method final would prevent us from overriding it in TagContextsImpl.  Or do you think we shouldn't override it with the simpler implementation?

core_impl/src/test/java/io/opencensus/impl/tags/TagContextImplTest.java
```diff
@@ -0,0 +1,119 @@
...
+  private static final TagKeyString KS1 = TagKeyString.create("k1");
+  private static final TagKeyString KS2 = TagKeyString.create("k2");
+
+  private static final TagValueString V1 = TagValueString.create("v1");
+  private static final TagValueString V2 = TagValueString.create("v2");
+
+  @Rule public final ExpectedException thrown = ExpectedException.none();
+
+  @Test
+  public void allowMutlipleKeysWithSameNameButDifferentTypes() {
```
sebright: I added a TODO to TagContextBuilder.

core/src/main/java/io/opencensus/tags/TagContexts.java
dinooliva: That's fine - I just didn't think there was a simpler implementation :)

PR #490, opened 2017-08-01
Implementation for IntervalView

songy23:
Implementation to support `IntervalView`. Tests are in PR #495.

The basic idea is, for IntervalView we will store a queue of values associated with the timestamp when they get processed (after going through the async `DisruptorQueue`). Every time when getView() is called, we dequeue those values that have expired (i.e. the time when they are processed + specified duration of interval view < current time), and update the summary stats by eliminating expired values.


songy23:
Rebased, should be ready to merge now.

songy23:
Since the PR is relatively small, I think it's better to review implementation along with tests. I'll close this PR, in favor of PR https://github.com/census-instrumentation/opencensus-java/pull/495, which contains changes on both implementation and tests.

PR #492, opened 2017-08-01
Slightly refactor the index check of the agent JAR in anticipation of future additions to the list of expected entries.

ubschmidt2:


ubschmidt2:
PTAL, if you have a chance.

PR #489, opened 2017-07-31
Update tests on implementations.

songy23:
Change the impl tests based on the refactored stats implementation.

PR #493, opened 2017-08-01
Update ViewManagerImplTest.

songy23:
Pulled out from PR #489 since ViewManagerImplTest contains too many changes.

### DIFF
--- a/core_impl/src/test/java/io/opencensus/stats/StatsTestUtil.java
+++ b/core_impl/src/test/java/io/opencensus/stats/StatsTestUtil.java
@@ -87,7 +87,6 @@ final class StatsTestUtil {
    * @param actual the actual mapping from {@code List<TagValue>} to
    *     {@code List<AggregationData>}.
    * @param tolerance the tolerance used for {@code double} comparison.
-   * @throws AssertionError if the AggregationMap does not contain exactly the keys and values.
    */
   static void assertAggregationMapEquals(
       Map<List<TagValue>, List<AggregationData>> actual,
@@ -105,61 +104,61 @@ final class StatsTestUtil {
    * @param expectedValue the expected list of {@code AggregationData}.
    * @param actualValue the actual list of {@code AggregationData}.
    * @param tolerance the tolerance used for {@code double} comparison.
-   * @throws AssertionError if the AggregationMap does not contain exactly the keys and values.
    */
   static void assertAggregationDataListEquals(
       List<AggregationData> expectedValue, List<AggregationData> actualValue,
       final double tolerance) {
     assertThat(expectedValue.size()).isEqualTo(actualValue.size());
     for (int i = 0; i < expectedValue.size(); i++) {
-      AggregationData actual = actualValue.get(i);
+      final AggregationData actual = actualValue.get(i);
       final AggregationData expected = expectedValue.get(i);
-      actual.match(
+      expected.match(
           new Function<SumData, Void>() {
             @Override
             public Void apply(SumData arg) {
-              assertAggregationDataEquals(
-                  expected, arg.getSum(), null, null, null, null, null, null, tolerance);
+              assertThat(actual).isInstanceOf(SumData.class);
+              assertThat(((SumData) actual).getSum()).isWithin(tolerance).of(arg.getSum());
               return null;
             }
           },
           new Function<CountData, Void>() {
             @Override
             public Void apply(CountData arg) {
-              assertAggregationDataEquals(
-                  expected, null, arg.getCount(), null, null, null, null, null, tolerance);
+              assertThat(actual).isInstanceOf(CountData.class);
+              assertThat(((CountData) actual).getCount()).isEqualTo(arg.getCount());
               return null;
             }
           },
           new Function<HistogramData, Void>() {
             @Override
             public Void apply(HistogramData arg) {
-              assertAggregationDataEquals(
-                  expected, null, null, arg.getBucketCounts(), null, null, null, null, tolerance);
+              assertThat(actual).isInstanceOf(HistogramData.class);
+              assertThat(removeTrailingZeros(((HistogramData) actual).getBucketCounts()))
+                  .isEqualTo(removeTrailingZeros(arg.getBucketCounts()));
               return null;
             }
           },
           new Function<RangeData, Void>() {
             @Override
             public Void apply(RangeData arg) {
-              assertAggregationDataEquals(
-                  expected, null, null, null, arg.getMin(), arg.getMax(), null, null, tolerance);
+              assertThat(actual).isInstanceOf(RangeData.class);
+              assertRangeDataEquals((RangeData) actual, arg, tolerance);
               return null;
             }
           },
           new Function<MeanData, Void>() {
             @Override
             public Void apply(MeanData arg) {
-              assertAggregationDataEquals(
-                  expected, null, null, null, null, null, arg.getMean(), null, tolerance);
+              assertThat(actual).isInstanceOf(MeanData.class);
+              assertThat(((MeanData) actual).getMean()).isWithin(tolerance).of(arg.getMean());
               return null;
             }
           },
           new Function<StdDevData, Void>() {
             @Override
             public Void apply(StdDevData arg) {
-              assertAggregationDataEquals(
-                  expected, null, null, null, null, null, null, arg.getStdDev(), tolerance);
+              assertThat(actual).isInstanceOf(StdDevData.class);
+              assertThat(((StdDevData) actual).getStdDev()).isWithin(tolerance).of(arg.getStdDev());
               return null;
             }
           },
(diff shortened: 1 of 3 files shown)


songy23:
@sebright  I just took a look again at `MapSubject.containsExactly()`, and I don't quite see its disadvantage for comparing the `AggregationMap`. For example one of the tests is: 
```
assertThat(viewData.getAggregationMap()).containsExactly(
        Arrays.asList(TagValue.create("VALUE")), 
        StatsTestUtil.createAggregationData(AGGREGATIONS, values));
```
Both key and value are deep-copied (new list with newly-created values). Since `AggregationData` is autovalued, its `equals()` has been overridden to compare its internal value, and `containsExactly` will do "deep" comparison on keys and values. Do you think we still need another method to compare this map?

sebright:
@songy23 I think we still need a custom comparison method like [`assertDistributionAggregatesEquivalent`](https://github.com/census-instrumentation/opencensus-java/blob/7c67f5b0e38622e3b4017baf39b4ae7fe737550c/core_impl/src/test/java/io/opencensus/stats/StatsTestUtil.java#L114-117) for two reasons:  Some of the AggregationData subclasses have double fields that are the result of arithmetic, so we should compare them with a tolerance instead of `==`.    See the comment comparing `isEqualTo` and `isWithin` in http://google.github.io/truth/api/latest/com/google/common/truth/DoubleSubject.html#isEqualTo-java.lang.Double- .  We also shouldn't compare the histogram bucket counts for equality, since they can be truncated to remove trailing zeros.

songy23:
@sebright makes sense, I've added another commit to add such a method.

sebright:
It looks like you rebased this onto the view branch.  Should I remove the "not ready for merge" label?

songy23:
Oh please do, I forgot to remove that label. Thanks.

core_impl/src/test/java/io/opencensus/stats/StatsTestUtil.java
```diff
@@ -80,6 +81,93 @@ private static StatsContextImpl createContext(
...
+   * @throws AssertionError if the AggregationMap does not contain exactly the keys and values.
+   */
+  static void assertAggregationDataListEquals(
+      List<AggregationData> expectedValue, List<AggregationData> actualValue,
+      final double tolerance) {
+    assertThat(expectedValue.size()).isEqualTo(actualValue.size());
+    for (int i = 0; i < expectedValue.size(); i++) {
+      AggregationData actual = actualValue.get(i);
+      final AggregationData expected = expectedValue.get(i);
+      actual.match(
```
sebright: I think you could compare the AggregationDatas more directly by calling `match` on the expected value instead of the actual value.  Then you could assert that the actual value has the right type inside of the `apply` methods.  For example, you could write this for the SumData case:

```java
          new Function<SumData, Void>() {
            @Override
            public Void apply(SumData expectedSubclass) {
              assertThat(actual).isInstanceOf(SumData.class);
              assertThat(((SumData) actual).getSum())
                  .isWithin(tolerance)
                  .of(expectedSubclass.getSum());
              return null;
            }
          },
```

Then you wouldn't need to call assertAggregationDataEquals.

core_impl/src/test/java/io/opencensus/stats/ViewManagerImplTest.java
```diff
@@ -71,391 +92,368 @@
...
-//    assertThat(viewData2.getEnd()).isEqualTo(Timestamp.create(4, 0));
-//    assertDistributionAggregatesEquivalent(
-//        viewData2.getDistributionAggregates(),
-//        Arrays.asList(
-//            StatsTestUtil.createDistributionAggregate(
-//                Arrays.asList(Tag.create(KEY, VALUE)), BUCKET_BOUNDARIES, Arrays.asList(2.2))));
-//  }
-//
-//  @Test
-//  public void testGetDistributionViewDataWithoutBucketBoundaries() {
```
sebright: Do you think it would be useful to test a view that has only a subset of the possible Aggregations?

core_impl/src/test/java/io/opencensus/stats/ViewManagerImplTest.java
```diff
@@ -71,391 +92,368 @@
...
+            StatsTestUtil.createAggregationData(AGGREGATIONS, 1.1, 4.4),
+            Arrays.asList(TagValue.create("v1"), TagValue.create("v20")),
+            StatsTestUtil.createAggregationData(AGGREGATIONS, 2.2),
+            Arrays.asList(TagValue.create("v2"), TagValue.create("v10")),
+            StatsTestUtil.createAggregationData(AGGREGATIONS, 3.3)),
+        EPSILON);
+  }
+
+  @Test
+  public void testMultipleViewDataSameMeasure() {
```
sebright: I think `testMultipleViewsSameMeasure` would be more accurate.

core_impl/src/test/java/io/opencensus/stats/StatsTestUtil.java
songy23: Correct, since `assertAggregationDataEquals` is not being used anywhere else, I've removed that method and update the validation here.

core_impl/src/test/java/io/opencensus/stats/StatsTestUtil.java
```diff
@@ -80,6 +81,93 @@ private static StatsContextImpl createContext(
...
+    }
+  }
+
+  /**
+   * Compare the expected and actual list of {@code AggregationData} within the given tolerance.
+   *
+   * @param expectedValue the expected list of {@code AggregationData}.
+   * @param actualValue the actual list of {@code AggregationData}.
+   * @param tolerance the tolerance used for {@code double} comparison.
+   * @throws AssertionError if the AggregationMap does not contain exactly the keys and values.
```
songy23: Removed.

sebright:
@songy23 Do you want to make a PR to merge the view branch into master after this is merged, since all of the tests are updated?

songy23:
Sure, I'll post another PR to merge the changes to mainline after this one is merged.

core_impl/src/test/java/io/opencensus/stats/StatsTestUtil.java
```diff
@@ -80,91 +81,102 @@ private static StatsContextImpl createContext(
...
-  static List<Long> removeTrailingZeros(long... longs) {
-    if (longs == null) {
-      return null;
+  static void assertAggregationDataListEquals(
+      List<AggregationData> expectedValue, List<AggregationData> actualValue,
+      final double tolerance) {
+    assertThat(expectedValue.size()).isEqualTo(actualValue.size());
+    for (int i = 0; i < expectedValue.size(); i++) {
+      final AggregationData actual = actualValue.get(i);
+      final AggregationData expected = expectedValue.get(i);
```
dinooliva: nit - this doesn't need to be final.

core_impl/src/test/java/io/opencensus/stats/ViewManagerImplTest.java
```diff
@@ -71,391 +92,394 @@
...
-//    DistributionViewData viewData = (DistributionViewData) viewManager.getView(VIEW_NAME);
-//    assertThat(viewData.getStart()).isEqualTo(Timestamp.create(1, 0));
-//    assertThat(viewData.getEnd()).isEqualTo(Timestamp.create(3, 0));
-//    assertDistributionAggregatesEquivalent(
-//        viewData.getDistributionAggregates(),
-//        Arrays.asList(
-//            StatsTestUtil.createDistributionAggregate(
-//                Arrays.asList(Tag.create(KEY, VALUE)), Arrays.asList(1.1))));
-//  }
+  private static View createDistributionView() {
```
dinooliva: Consider renaming Distribution to Cumulative.

core_impl/src/test/java/io/opencensus/stats/ViewManagerImplTest.java
```diff
@@ -71,391 +92,394 @@
...
+    StatsTestUtil.assertAggregationMapEquals(
+        viewData2.getAggregationMap(),
+        ImmutableMap.of(
+            Arrays.asList(VALUE),
+            StatsTestUtil.createAggregationData(AGGREGATIONS, 2.2)),
+        EPSILON);
+  }
+
+  @Test
+  public void testGetDistributionViewDataWithoutBucketBoundaries() {
```
dinooliva: This name is fairly specific to the old views (i.e. Distributions), consider renamining.

core_impl/src/test/java/io/opencensus/stats/StatsTestUtil.java
songy23: Done.

core_impl/src/test/java/io/opencensus/stats/ViewManagerImplTest.java
songy23: Updated.

core_impl/src/test/java/io/opencensus/stats/ViewManagerImplTest.java
songy23: Updated.

songy23:
@dinooliva sorry I did a rebase and a force push this morning, and that overwrote some latter changes. Now the diff should be ready.

PR #495, opened 2017-08-09
Implementation and tests for IntervalView.

songy23:
Implementation and tests to support IntervalView.

The basic idea is, for IntervalView we will store a queue of values associated with the timestamp when they get processed (after going through the async DisruptorQueue). Every time when getView() is called, we dequeue those values that have expired (i.e. the time when they are processed + specified duration of interval view < current time), and update the summary stats by eliminating expired values.


core_impl/src/main/java/io/opencensus/stats/MutableViewData.java
```diff
@@ -136,9 +140,45 @@ ViewData toViewData(Clock clock) {
       }
       map.put(entry.getKey(), aggregates);
     }
-    return ViewData.create(view, map, CumulativeData.create(start, clock.now()));
+
+    WindowData windowData = view.getWindow().match(
+        new CreateCumulativeData(start, clock.now()),
+        new CreateIntervalData(clock.now()),
+        Functions.<WindowData>throwIllegalArgumentException());
```
sebright: I think AssertionError would be better if we expect the API and implementation to always have the same version.  Alternatively, we could try to handle an unknown window, possibly by treating it as cumulative.

core_impl/src/main/java/io/opencensus/stats/MutableViewData.java
```diff
@@ -61,6 +72,9 @@
       Maps.newHashMap();
   @Nullable private final Timestamp start;
 
+  // The queue of TimestampedValues for interval aggregations.
+  private final Queue<TimestampedValue> valueQueue = new LinkedList<TimestampedValue>();
```
sebright: I think we need to also record the tag values for each value, so we know which aggregations they are part of.  We could either use one queue per tag value combination or add the tag values to the elements of the queue.

core_impl/src/main/java/io/opencensus/stats/MutableViewData.java
```diff
@@ -127,6 +130,7 @@ void record(StatsContextImpl tags, long value) {
    * Convert this {@link MutableViewData} to {@link ViewData}.
    */
   ViewData toViewData(Clock clock) {
+    removeExpiredValues(clock.now());
```
sebright: We should also remove expired values when we add new values, to limit the queue size in the case where getView is never called.

core_impl/src/main/java/io/opencensus/stats/MeasureToViewMap.java
```diff
@@ -84,47 +84,52 @@ synchronized void registerView(View view, Clock clock) {
   }
 
   // Records stats with a set of tags.
-  synchronized void record(StatsContextImpl tags, MeasureMap stats) {
+  synchronized void record(StatsContextImpl tags, MeasureMap stats, Clock clock) {
```
dinooliva: There's got to be a better way then passing a clock around everywhere. Maybe we need something like Clock in StatsComponent.

core_impl/src/main/java/io/opencensus/stats/MutableAggregation.java
```diff
@@ -33,6 +39,20 @@ private MutableAggregation() {
   abstract void add(double value);
 
   /**
+   * Update the summary stats by removing an expired value.
+   *
+   * @param value the expired value to be removed from population.
+   */
+  abstract void remove(double value);
```
dinooliva: Different window types will likely require different implementation approaches so splitting MutableAggregations along these lines probably makes the most sense. For interval stats, we can go with the straightforward approach and then optimize for different measures later (e.g. keeping a running sum).

sebright: Another idea is to keep the approach used in this PR for sum, count, and histogram, but remove support for stddev, mean, and range.  The current approach should work well for the three exposed aggregations.  When we add the other three, we can decide how to optimize interval views, and maybe start breaking the interval into sub intervals to avoid storing all values.

core_impl/src/main/java/io/opencensus/stats/MeasureToViewMap.java
songy23: The actual `clock` is stored in `StatsManager`: https://github.com/census-instrumentation/opencensus-java/blob/master/core_impl/src/main/java/io/opencensus/stats/StatsManager.java#L25. 
Call stack to this is: 
TagContext.record() -> StatsRecorded.record() -> StatsManager.record() -> EventQueue.enqueue() -> going through event queue -> EventQueue.process() -> MeasureToViewMap.record() -> MutableViewData.record(). 
In the last call we need to record the current `Timestamp` by calling `clock.now()`. Do you think it's better to get `Timestamp` here in MeasureToViewMap.record(), than passing in the `clock` reference and get `Timestamp` at MutableViewData.record()?

core_impl/src/test/java/io/opencensus/stats/ViewManagerImplTest.java
```diff
@@ -264,6 +335,43 @@ public void testRecordMultipleTagValues() {
...
+    // record for TagValue2 at 5th second
+    clock.setTime(Timestamp.fromMillis(5 * MILLIS_PER_SECOND));
+    statsRecorder.record(
+        createContext(factory, KEY, VALUE_2),
+        MeasureMap.builder().set(MEASURE, 30.0).build());
+    statsRecorder.record(
+        createContext(factory, KEY, VALUE_2),
+        MeasureMap.builder().set(MEASURE, 50.0).build());
+
+    // get view at 12th second, stats for TagValue1 should have expired.
```
sebright: It would also be useful to check that there are values from both tag combinations before any expire.

core_impl/src/main/java/io/opencensus/stats/MeasureToViewMap.java
songy23: I've updated `MeasureToViewMap.record()` and `MutableViewData.record()` to take `Timestamp` instead of `Clock`.

core_impl/src/main/java/io/opencensus/stats/MutableAggregation.java
songy23: That sounds good. I've marked that Interval stats are not supported for stddev, mean, and range MutableAggregation.

core_impl/src/main/java/io/opencensus/stats/MutableViewData.java
```diff
@@ -198,7 +181,91 @@ static AggregationData createAggregationData(MutableAggregation aggregation) {
...
+    @Override
+    ViewData toViewData(Clock clock) {
+      removeExpiredValues(clock.now());
+      return ViewData.create(
+          super.view, super.toAggregationMap(), IntervalData.create(clock.now()));
+    }
+
+    // If this MutableViewData has an IntervalWindow, add the value associated with the given
+    // timestamp to valueQueue.
+    private void addValueToQueue(double value, Timestamp now, List<TagValue> tagValues) {
```
songy23: Updated.

core_impl/src/main/java/io/opencensus/stats/MutableViewData.java
```diff
@@ -198,7 +181,91 @@ static AggregationData createAggregationData(MutableAggregation aggregation) {
...
+      checkNotNull(now, "Timestamp");
+      checkNotNull(tagValues, "TagValues");
+      valueQueue.addLast(new TimestampedValue(value, now, tagValues));
+    }
+
+    // If this MutableViewData has an IntervalWindow, dequeue expired values from valueQueue against
```
songy23: Done.

core_impl/src/test/java/io/opencensus/stats/ViewManagerImplTest.java
songy23: Sure, updated.

core_impl/src/main/java/io/opencensus/stats/MutableViewData.java
```diff
@@ -198,7 +181,90 @@ static AggregationData createAggregationData(MutableAggregation aggregation) {
...
+
+    private IntervalMutableViewData(View view) {
+      super(view);
+    }
+
+    @Override
+    void record(StatsContextImpl context, double value, Timestamp timestamp) {
+      removeExpiredValues(timestamp);
+      List<TagValue> tagValues = getTagValues(context.tags, super.view.getColumns());
+      // Add Timestamp to value after the value went through the DisruptorQueue.
```
sebright: Should this comment be moved to the call to now() in StatsManager?

core_impl/src/main/java/io/opencensus/stats/StatsManager.java
```diff
@@ -56,17 +57,17 @@ void record(StatsContextImpl tags, MeasureMap measurementValues) {
...
+    StatsEvent(StatsManager statsManager, StatsContextImpl tags, MeasureMap stats) {
+      this.statsManager = statsManager;
       this.tags = tags;
       this.stats = stats;
     }
 
     @Override
     public void process() {
-      viewManager.measureToViewMap.record(tags, stats);
+      statsManager.measureToViewMap.record(tags, stats, statsManager.clock.now());
```
sebright: :+1:  Now all views use the same time for the same value.

core_impl/src/main/java/io/opencensus/stats/MutableViewData.java
songy23: Sure, updated.

songy23:
Rebased against mainline.

songy23:
Going to close this PR since we've decided to go with the bucket/sub-interval approach.

PR #496, opened 2017-08-10
Minimal API changes to support MeasureLong.

songy23:
From the conversation with @dinooliva, we have summarized the correspondence between different types of `Measure` and `Aggregation`:

                       Sum           Count      Histogram         Range               Mean       StandardDeviation
    MeasureDouble    double          long       List<Long>     (double, double)     double            double
    MeasureLong       long           long       List<Long>      (long, long)        double            double

Only `Sum` and `Range` need to have different output types for different `Measure`s. In order to support `MeasureLong` in v0.1, I think the minimal API changes to make is to add subclasses of `Sum` and `Range` that hold double and long values respectively. I don't think it's necessary for other `Aggregation`s (Count, Mean, etc.) to have such hierarchy on API level, since their output types on double or long are the same.

Update: we decide to have `SumDataDouble`, `SumDataLong`, `RangeDataDouble` and `RangeDataLong` at top level, inline with other `Aggregation` types.

### DIFF
--- a/core/src/main/java/io/opencensus/stats/AggregationData.java
+++ b/core/src/main/java/io/opencensus/stats/AggregationData.java
@@ -27,12 +27,14 @@ import javax.annotation.concurrent.Immutable;
  * {@link AggregationData} is the result of applying a given {@link Aggregation} to a set of
  * {@code MeasureValue}s.
  *
- * <p>{@link AggregationData} currently supports 6 types of basic aggregation values:
+ * <p>{@link AggregationData} currently supports 8 types of basic aggregation values:
  * <ul>
- *   <li>SumData
+ *   <li>SumDataDouble
+ *   <li>SumDataLong
  *   <li>CountData
  *   <li>HistogramData
- *   <li>RangeData
+ *   <li>RangeDataDouble
+ *   <li>RangeDataLong
  *   <li>MeanData
  *   <li>StdDevData (standard deviation)
  * </ul>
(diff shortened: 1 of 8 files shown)


adriancole:
I think your ranges are flipped in the desc.

I suppose the rationale for mean and stddev of long being double is that the fraction is important enough to make the types a bit out if alignment. Wdyt @jkschneider

songy23:
Correct, thanks for catching that.

jkschneider:
I would add that I'm suspicious of standard deviation as a useful measure for application metrics, as many if not most distributions in reality are non-normal.

songy23:
Standard deviation is a requirement for exporting our stats to StackDriver: https://cloud.google.com/monitoring/api/ref_v3/rest/v3/TypedValue

core/src/main/java/io/opencensus/stats/AggregationData.java
```diff
@@ -88,6 +76,68 @@ static SumData create(double sum) {
...
+    /** Applies the given match function to the underlying data type. */
+    public abstract <T> T match(
+        Function<? super SumDataDouble, T> p0,
+        Function<? super SumDataLong, T> p1,
+        Function<? super SumData, T> defaultFunction);
+
+    /** The double sum value on aggregated values of type {@code MeasureDouble}. */
+    @Immutable
+    @AutoValue
+    public abstract static class SumDataDouble extends SumData {
```
dinooliva: As discussed irl, I think it makes sense just to inline these types all in AggregationData - it doesn't increase the overall number of things we need to deal with and removes the need to deal with additional defaultFunctions.

PR #498, opened 2017-08-11
Add methods for interacting with the current TagContext.

sebright:
This PR adds methods for getting the current TagContext and setting the TagContext in a scope.  It also adds convenience methods for creating a builder from the current tags and setting the scoped tags with a builder.

The new methods correspond to methods for working with the current span.  (There are some differences, because the TagContext doesn't need to be started, and it doesn't have a parent-child relationship.)

Tagging                                                                    | Tracing
-------------------------------------------------------------------|------------------------------------------------------
`TagContexts.getCurrentTagContext()`                   | `Tracer.getCurrentSpan()`
`TagContexts.withTagContext(TagContext tags)`    | `Tracer.withSpan(Span span)`
`TagContexts.currentBuilder()`                                | `Tracer.spanBuilder(String spanName)`
`TagContextBuilder.buildScoped()`                          | `SpanBuilder.startScopedSpan()`

Existing methods that don't use the current context:

Tagging                                                      | Tracing
--------------------------------------------------------|------------------------------------------------------
`TagContexts.emptyBuilder()` and `TagContexts.toBuilder(TagContext tags)` | `Tracer.spanBuilderWithExplicitParent(String spanName, @Nullable Span parent)`
`TagContextBuilder.build()`                        | `SpanBuilder.startSpan()`

/cc @adriancole

### DIFF
--- a/core/src/main/java/io/opencensus/tags/TagContextBuilder.java
+++ b/core/src/main/java/io/opencensus/tags/TagContextBuilder.java
@@ -13,6 +13,7 @@
 
 package io.opencensus.tags;
 
+import io.opencensus.common.Scope;
 import io.opencensus.tags.TagKey.TagKeyBoolean;
 import io.opencensus.tags.TagKey.TagKeyLong;
 import io.opencensus.tags.TagKey.TagKeyString;
@@ -65,6 +66,18 @@ public abstract class TagContextBuilder {
    * @return a {@code TagContext} with the same tags as this builder.
    */
   public abstract TagContext build();
+
+  /**
+   * Enters the scope of code where the {@link TagContext} created from this builder is in the
+   * current context and returns an object that represents that scope. The scope is exited when the
+   * returned object is closed.
+   *
+   * @return an object that defines a scope where the {@code TagContext} created from this builder
+   *     is set to the current context.
+   */
+  public final Scope buildScoped() {
+    return CurrentTagContextUtils.withTagContext(build());
+  }
 
   /**
    * Returns a {@code TagContextBuilder} that ignores all calls to {@link #set}.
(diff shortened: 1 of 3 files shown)


adriancole:
Dunno about the currentBuilder and buildScoped.. tag context doesnt have a
lifecycle like a span. Do we really need these methods? How worse would it
be if there was only a scoping method?

sebright:
@adriancole `currentBuilder` and `buildScoped` are just convenience methods.  My goal was to make it no harder to update the current TagContext than to create a new TagContext, since using the current context is the expected use case.  I wanted to shorten the code required to update tags and also make it harder to accidentally clear the existing tags.

With the convenience methods, a user can update the tags with

```java
try (Scope scope = tagContexts.currentBuilder().set(KEY, VALUE).buildScoped()) {
  ...
}
```

instead of

```java
try (Scope scope =
    tagContexts.withTagContext(
        tagContexts.toBuilder(tagContexts.getCurrentTagContext()).set(KEY, VALUE).build())) {
  ...
}
```

adriancole:
Thanks for giving an example of this. I see your point!

core/src/main/java/io/opencensus/tags/TagContextBuilder.java
```diff
@@ -67,6 +68,18 @@
...
 
   /**
+   * Enters the scope of code where the {@link TagContext} created from this builder is in the
+   * current context and returns an object that represents that scope. The scope is exited when the
+   * returned object is closed.
+   *
+   * @return an object that defines a scope where the {@code TagContext} created from this builder
+   *     is set to the current context.
+   */
+  public final Scope buildScoped() {
```
dinooliva: I still have reservations about intermixing CurrentTagContext in with TagContext{Builder}, let's talk about it at our next sync (tomorrow).

sebright:
Thanks!

Issue #487, opened 2017-07-28
Thread contextClassLoader failed to load Trace impl statically


songy23:
Thanks @zhangkun83 for helping finding the root cause of this issue.

In my E2E gRPC code, I've found an issue with loading trace impl statically. Assuming we have dependency on both the core library and implementation on trace, if we call Tracing.getTracer() statically like this:
```
private static Tracer tracer = Tracing.getTracer();
```
Tracing will be unable to load the implementation, and instead a `NoopTracer` will be returned. However if we change to call Tracing.getTracer() within main() or constructor like this:
```
private static Tracer tracer;
public static void main(String[] args) {
   tracer = Tracing.getTracer(); 
   ....
}
```
or
```
private Tracer tracer;
public Foo() {
  this.tracer = Tracing.getTracer(); 
}
public static void main(String[] args) {
   Foo foo = new Foo();
   ....
}
```
The real `TraceImpl` will be correctly loaded.

I think this issue is related to `Thread.currentThread().getContextClassLoader();` in our [Provide.getCorrectClassLoader()](https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/internal/Provider.java#L59). When I changed to always use `superClass.getClassLoader();`, trace impl could be loaded statically.

ejona86:
Avoiding `Thread.getContextClassLoader()` is a good idea. I wouldn't have suggested to use it in conjunction with `Class.forName()`, but in grpc/grpc-java#2375 I decided it was likely a bad idea anywhere but in the standard library.

PR #502, opened 2017-08-11
Stop using context ClassLoader in Provider

bogdandrutu:
This fixes https://github.com/census-instrumentation/opencensus-java/issues/487


dinooliva:
I'm curious about why you were using the the thread-local class loader in the first place.

PR #506, opened 2017-08-11
Introduce a simplistic plug-in mechanism.

ubschmidt2:
This reduces the coupling between AgentMain and the plugins (well, one ATM, but there should be more eventually).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -105,6 +105,7 @@ subprojects {
 
         libraries = [
                 auto_value     : "com.google.auto.value:auto-value:${autoValueVersion}",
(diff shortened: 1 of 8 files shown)


build.gradle
```diff
@@ -105,6 +105,7 @@ subprojects {
 
         libraries = [
                 auto_value     : "com.google.auto.value:auto-value:${autoValueVersion}",
+                auto_service : 'com.google.auto.service:auto-service:1.0-rc3',
```
adriancole: rc3 4evar!

PR #509, opened 2017-08-11
Enable checkstyle for tests in stats package.

sebright:
/cc @dinooliva @bogdandrutu

### DIFF
--- a/core_impl/src/test/java/io/opencensus/stats/ViewManagerImplTest.java
+++ b/core_impl/src/test/java/io/opencensus/stats/ViewManagerImplTest.java
@@ -375,13 +375,13 @@ public class ViewManagerImplTest {
 
   @Test
   public void testMultipleViewSameMeasure() {
-    View view1 =
+    final View view1 =
         createCumulativeView(
             VIEW_NAME,
             MEASURE,
             AGGREGATIONS,
             Arrays.asList(KEY));
-    View view2 =
+    final View view2 =
         createCumulativeView(
             VIEW_NAME_2,
             MEASURE,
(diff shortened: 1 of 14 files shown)


sebright:
Thanks!

core_impl/src/test/java/io/opencensus/stats/ViewManagerImplTest.java
```diff
@@ -375,13 +375,13 @@ public void testViewDataWithMultipleTagKeys() {
 
   @Test
   public void testMultipleViewSameMeasure() {
-    View view1 =
+    final View view1 =
```
dinooliva: Why make these final?

sebright: I made the views final to fix this warning:

```
core_impl/src/test/java/io/opencensus/stats/ViewManagerImplTest.java:384: Distance between variable 'view2' declaration and its first usage is 4, but allowed 3.  Consider to make that variable as final if you still need to store its value in advance (before method calls that might do side effect on original value). [VariableDeclarationUsageDistance]
```

I don't know how useful this warning is, though.  Maybe we should suppress it or increase the allowed distance.

bogdandrutu: You can just move view2 2-lines bottom and you will fix the warning I think.

sebright: There was a second warning in the file where it wasn't easy to shorten the distance, because the test defined 4 constants at the top of the method (The maximum distance between declaration and use is 3).  I fixed both with `final` for consistency.

PR #510, opened 2017-08-11
Move stats to impl.stats, change some visibility.

songy23:
Will remove `StatsContextFactory`, `Tag`, `TagKey`, `TagValue` under stats package, and use new APIs in tags package in the next PR.

### DIFF
--- a/core/src/main/java/io/opencensus/stats/AggregationData.java
+++ b/core/src/main/java/io/opencensus/stats/AggregationData.java
@@ -66,6 +66,12 @@ public abstract class AggregationData {
     SumData() {
     }
 
+    /**
+     * Creates a {@code SumData}.
+     *
+     * @param sum the aggregated sum.
+     * @return a {@code SumData}.
+     */
     public static SumData create(double sum) {
       return new AutoValue_AggregationData_SumData(sum);
     }
@@ -98,6 +104,12 @@ public abstract class AggregationData {
     CountData() {
     }
 
+    /**
+     * Creates a {@code CountData}.
+     *
+     * @param count the aggregated count.
+     * @return a {@code CountData}.
+     */
     public static CountData create(long count) {
       return new AutoValue_AggregationData_CountData(count);
     }
@@ -131,7 +143,7 @@ public abstract class AggregationData {
     }
 
     /**
-     * Create a {@code HistogramData}.
+     * Creates a {@code HistogramData}.
      *
      * @param bucketCounts bucket counts.
      * @return a {@code HistogramData}.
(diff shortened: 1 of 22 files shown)


core/src/main/java/io/opencensus/stats/AggregationData.java
```diff
@@ -130,10 +130,16 @@ static CountData create(long count) {
     HistogramData() {
     }
 
-    static HistogramData create(long... bucketCountDatas) {
-      checkNotNull(bucketCountDatas, "bucket counts should not be null.");
+    /**
+     * Create a {@code HistogramData}.
```
sebright: The summary fragment should be an incomplete sentence, such as "`Creates a {@code HistogramData}.`" (https://google.github.io/styleguide/javaguide.html#s7.2-summary-fragment).  This applies to RangeData, too.

core/src/main/java/io/opencensus/stats/AggregationData.java
```diff
@@ -66,7 +66,7 @@ private AggregationData() {
     SumData() {
     }
 
-    static SumData create(double sum) {
+    public static SumData create(double sum) {
```
sebright: All of the public factory methods in this class should have Javadocs.

songy23: Added.

core/src/main/java/io/opencensus/stats/AggregationData.java
songy23: Updated. Thanks for pointing out.

songy23:
@bogdandrutu `tags` impl has already been moved to `impl/tags`: https://github.com/census-instrumentation/opencensus-java/tree/master/core_impl/src/main/java/io/opencensus. This PR is meant to move `stats` impl package to the same location with `tags`.

bogdandrutu:
It will be incompatible with Trace so @dinooliva can you make a call if that is acceptable?

songy23:
Rebased against mainline.

bogdandrutu:
Trace changed to the same implementation so it is fine to merge this PR.

songy23:
Got it. Rebased again... should be ready to merge now.

core/src/main/java/io/opencensus/stats/StatsRecorder.java
```diff
@@ -33,7 +33,7 @@
    * @param tags the tags associated with the measurements.
    * @param measurementValues the measurements to record.
    */
-  abstract void record(TagContext tags, MeasureMap measurementValues);
+  protected abstract void record(TagContext tags, MeasureMap measurementValues);
```
dinooliva: Should this be public?

sebright: I think this can be public now that we've switched to TagContext.

songy23: Done.

songy23:
Thanks!

PR #511, opened 2017-08-12
Add TagContext serialization.

sebright:
This commit adds a class for serializing and deserializing TagContexts,
TagContextSerializer, to the TagsComponent.  TagContext serialization is very
similar to the existing StatsContext serialization, and it currently only
supports String tag values.

_____________________________________________________________

I tried to break up the commits to make it easier to review, since I copied a lot of the code from the stats package.  The first four commits just copy files and format the code.

### DIFF
--- a/core/src/main/java/io/opencensus/tags/TagContextSerializer.java
+++ b/core/src/main/java/io/opencensus/tags/TagContextSerializer.java
@@ -30,6 +30,7 @@ public abstract class TagContextSerializer {
    *
    * @param tags the {@code TagContext} to serialize.
    * @param output the {@link OutputStream} to write the serialized tags to.
+   * @throws IOException if there is an {@code IOException} while writing to {@code output}.
    */
   public abstract void serialize(TagContext tags, OutputStream output) throws IOException;
 
@@ -40,6 +41,8 @@ public abstract class TagContextSerializer {
    *
    * @param input on-the-wire representation of a {@code TagContext}.
    * @return a {@code TagContext} deserialized from {@code input}.
+   * @throws IOException if there is a parse error or an {@code IOException} while reading from
+   *     {@code input}.
    */
   public abstract TagContext deserialize(InputStream input) throws IOException;
 
--- a/core/src/main/java/io/opencensus/tags/TagContextSerializer.java
+++ b/core/src/main/java/io/opencensus/tags/TagContextSerializer.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2017, Google Inc.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.tags;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import javax.annotation.concurrent.Immutable;
+
+/** Object for serializing and deserializing {@link TagContext}s. */
+public abstract class TagContextSerializer {
+  private static final TagContextSerializer NOOP_TAG_CONTEXT_SERIALIZER =
+      new NoopTagContextSerializer();
+
+  /**
+   * Serializes the {@code TagContext} into the on-the-wire representation.
+   *
+   * <p>This method should be the inverse of {@link #deserialize}.
+   *
+   * @param tags the {@code TagContext} to serialize.
+   * @param output the {@link OutputStream} to write the serialized tags to.
+   */
+  public abstract void serialize(TagContext tags, OutputStream output) throws IOException;
+
+  /**
+   * Creates a {@code TagContext} from the given on-the-wire encoded representation.
+   *
+   * <p>This method should be the inverse of {@link #serialize}.
+   *
+   * @param input on-the-wire representation of a {@code TagContext}.
+   * @return a {@code TagContext} deserialized from {@code input}.
+   */
+  public abstract TagContext deserialize(InputStream input) throws IOException;
+
+  /**
+   * Returns a {@code TagContextSerializer} that serializes all {@code TagContext}s to zero bytes
+   * and deserializes all inputs to empty {@code TagContext}s.
+   */
+  static TagContextSerializer getNoopTagContextSerializer() {
+    return NOOP_TAG_CONTEXT_SERIALIZER;
+  }
+
+  @Immutable
+  private static final class NoopTagContextSerializer extends TagContextSerializer {
+
+    @Override
+    public void serialize(TagContext tags, OutputStream output) throws IOException {}
+
+    @Override
+    public TagContext deserialize(InputStream input) throws IOException {
+      return TagContext.getNoopTagContext();
+    }
+  }
+}
--- a/core/src/main/java/io/opencensus/tags/TagValueString.java
+++ b/core/src/main/java/io/opencensus/tags/TagValueString.java
@@ -46,7 +46,6 @@ public abstract class TagValueString {
    * @throws IllegalArgumentException if the {@code String} is not valid.
    */
   public static TagValueString create(String value) {
-    // TODO(sebright): Should we disallow empty?
     Preconditions.checkArgument(isValid(value));
     return new AutoValue_TagValueString(value);
   }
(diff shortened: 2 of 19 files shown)


core/src/main/java/io/opencensus/tags/TagValueString.java
```diff
@@ -46,6 +46,7 @@
    * @throws IllegalArgumentException if the {@code String} is not valid.
    */
   public static TagValueString create(String value) {
+    // TODO(sebright): Should we disallow empty?
```
adriancole: I think yes on value

core_impl/src/main/java/io/opencensus/impl/tags/SerializationUtils.java
```diff
@@ -0,0 +1,192 @@
...
+ *                     <li>{@code <tag_key_len>} == varint encoded integer
+ *                     <li>{@code <tag_key>} == tag_key_len bytes comprising tag key name
+ *                     <li>{@code <tag_val_len>} == varint encoded integer
+ *                     <li>{@code <tag_val>} == tag_val_len bytes comprising UTF-8 string
+ *                   </ul>
+ *               <li>{@code (tag_field_id == 1) == <tag_key_len><tag_key><int_tag_val>}
+ *                   <ul>
+ *                     <li>{@code <tag_key_len>} == varint encoded integer
+ *                     <li>{@code <tag_key>} == tag_key_len bytes comprising tag key name
+ *                     <li>{@code <int_tag_value>} == 8 bytes, little-endian integer
```
adriancole: Java defaults to big endian for a lot of things. Might want to comment why little endian somewhere

core_impl/src/main/java/io/opencensus/impl/tags/TagContextSerializerImpl.java
```diff
@@ -0,0 +1,34 @@
...
+
+package io.opencensus.impl.tags;
+
+import io.opencensus.tags.TagContext;
+import io.opencensus.tags.TagContextSerializer;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+
+final class TagContextSerializerImpl extends TagContextSerializer {
```
songy23: Is there a specific reason to keep `SerializationUtils` separated from this class?

core_impl/src/test/java/io/opencensus/impl/tags/TagContextDeserializationTest.java
```diff
@@ -15,60 +15,54 @@
...
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
 
-/** Tests for {@link StatsContextFactory}. */
+/**
+ * Tests for deserializing tags with {@link SerializationUtils} and {@link
+ * TagContextSerializerImpl}.
+ */
 @RunWith(JUnit4.class)
 public class TagContextDeserializationTest {
```
songy23: Might be better to put all three test classes together?

core/src/main/java/io/opencensus/tags/TagValueString.java
sebright: Thanks.  Why do you think that values shouldn't be empty?  Do you think that the same should apply to keys?  I'm not sure what we should allow, but I think tag values should be no more restrictive than tag keys, since tag keys are more likely to be known statically.

core_impl/src/main/java/io/opencensus/impl/tags/SerializationUtils.java
sebright: /cc @bogdandrutu

bogdandrutu: There is no a good or bad answer. Some languages default to little-endian some to big-endian, the majority of the CPUs are little-endian. So we only need to be consistent between all the languages and avoid unnecessary overhead.

@adriancole  Why does it matter for the users? I think it is important only for other developers of the library but not for the users.

core_impl/src/main/java/io/opencensus/impl/tags/TagContextSerializerImpl.java
sebright: I thought it would be simpler for SerializationUtils to remain a utility class, with only static methods.

core_impl/src/test/java/io/opencensus/impl/tags/TagContextDeserializationTest.java
sebright: I think the class would be too large if we combined all of the tests, and serialization vs deserialization seemed like a good way to split it.  The test suites will also grow when we add support for the other tag types.

adriancole:
Java stream utils tend to prefer big endian. I ask because we had a lot of
questions around this in zipkin. People will write custom clients or
snoops. I agree it matters less which, but in java big endian is more
typical.

core/src/main/java/io/opencensus/tags/TagContextSerializer.java
```diff
@@ -0,0 +1,65 @@
...
+      new NoopTagContextSerializer();
+
+  /**
+   * Serializes the {@code TagContext} into the on-the-wire representation.
+   *
+   * <p>This method should be the inverse of {@link #deserialize}.
+   *
+   * @param tags the {@code TagContext} to serialize.
+   * @param output the {@link OutputStream} to write the serialized tags to.
+   */
```
songy23: Should this also document exceptions?

core/src/main/java/io/opencensus/tags/TagContextSerializer.java
```diff
@@ -0,0 +1,65 @@
...
+  public abstract void serialize(TagContext tags, OutputStream output) throws IOException;
+
+  /**
+   * Creates a {@code TagContext} from the given on-the-wire encoded representation.
+   *
+   * <p>This method should be the inverse of {@link #serialize}.
+   *
+   * @param input on-the-wire representation of a {@code TagContext}.
+   * @return a {@code TagContext} deserialized from {@code input}.
+   */
```
songy23: Same

core/src/main/java/io/opencensus/tags/TagContextSerializer.java
sebright: Done

core/src/main/java/io/opencensus/tags/TagContextSerializer.java
sebright: Done

core_impl/src/main/java/io/opencensus/impl/tags/SerializationUtils.java
sebright: I opened an issue for this, because I think it would be better to modify the encoding in another PR.

sebright:
> Most likely we will have to support binary and http encodings (for performance reasons).
> 
> Consider a design where we can support multiple types of encodings (serialization). See trace/propagation for that.

I added a TagPropagationComponent class that can reference different serialization formats in the last commit.

core/src/main/java/io/opencensus/tags/TagValueString.java
dinooliva: I don't see a good reason not to allow empty strings.

core_impl/src/main/java/io/opencensus/impl/tags/SerializationUtils.java
dinooliva: I think adding a comment as I suggested in the issue is sufficient.

sebright: I added comments in the last two commits.

adriancole:
Empty values are a way to represent a tag that you search only by key.
Amazon put this to wide use in some of their query apis. For example in
zipkin, we can search for presence of a tag w/o regards to value. Without
allowing empty people will have to hack like adding a boolean value. When
mapping the other direction, theres no way to know that this boolean true
tag was an empty string.

sebright:
@adriancole I'm not sure I understand.  We currently allow empty tag values, and I don't think there's any plan to disallow them.

sebright:
I rebased to fix the build failure.

adriancole:
>
> @adriancole <https://github.com/adriancole> I'm not sure I understand. We
> currently allow empty tag values, and I don't think there's any plan to
> disallow them.
>
> Sorry I responded to the below before coffee and missed the not!

I don't see a good reason not to allow empty strings.

core_impl/src/main/java/io/opencensus/impl/tags/SerializationUtils.java
```diff
@@ -0,0 +1,195 @@
...
+      if (ex != null) {
+        throw ex;
+      }
+    }
+    byteArrayOutputStream.writeTo(output);
+  }
+
+  // Deserializes input to TagContext based on the binary format standard.
+  // The encoded tags are of the form: <version_id><encoded_tags>
+  static TagContextImpl deserialize(InputStream input) throws IOException {
```
bogdandrutu: I know that this is internal but I would name this deserializeBinary or binaryDecode? Probably the reason why you have this class is because you want to share some of the code with the HTTP format?

sebright: I wanted to put the serialization logic in a utility class to simplify TagContextBinarySerializerImpl, but I hadn't really thought about how to share code between the two formats.  I renamed the existing methods to serializeBinary and deserializeBinary.

sebright:
Thanks for the review!

PR #514, opened 2017-08-14
Refactor tag validation code.

sebright:
This commit splits tag validation into separate methods for handling tag keys
and values, since keys and values may need different restrictions on their input
Strings.  There should be no change in behavior.

/cc @dinooliva


sebright:
Thanks!

adriancole:
do tag values need to be constrained to ascii? I understand it is easier to
validate, but this won't suit chinese values for example

PR #515, opened 2017-08-15
Replace original tag classes with classes in the tags package.

sebright:
I removed all of the original tag classes and tests.  There are still a few
things that need to be done after this commit:

- The tags package doesn't have a class to represent a tag value yet, so I used
  Object instead of TagValue in the ViewData API.

- There are no methods for manipulating the current TagContext yet, since #498
  hasn't been merged.  I couldn't update StatsRunner, so I temporarily commented
  most of it out.

- We need to add a StatsRecorder.record method that uses the current context.


core/src/main/java/io/opencensus/stats/ViewData.java
```diff
@@ -53,7 +54,9 @@
 
   /** Constructs a new {@link ViewData}. */
   public static ViewData create(
-      View view, Map<List<TagValue>, List<AggregationData>> map, final WindowData windowData) {
+      View view,
+      Map<? extends List<? extends Object>, List<AggregationData>> map,
```
songy23: Why do we have the first `? extends ` here?

sebright: I don't understand it very well, but I think that `List<String>` is treated as a subtype of `List<? extends Object>` (https://stackoverflow.com/a/22806587), so we also need to allow subtypes of the list.  Without the first `? extends`, the second `? extends` seemed to have no effect, and the method wouldn't accept a `List<TagValueString>`.

Another reason to add the first `? extends` is to allow list subclasses, such as ArrayList.

adriancole:
>
> -      View view, Map<List<TagValue>, List<AggregationData>> map, final WindowData windowData) {
> +      View view,
> +      Map<? extends List<? extends Object>, List<AggregationData>> map,
>
> I don't understand it very well, but I think that List<String> is treated
> as a subtype of List<? extends Object> (https://stackoverflow.com/a/
> 22806587), so we also need to allow subtypes of the list. Without the
> first ? extends, the second ? extends seemed to have no effect, and the
> method wouldn't accept a List<TagValueString>.
>
> Another reason to add the first ? extends is to allow list subclasses,
> such as ArrayList.
>
aint generics fun? :)

sebright:
Thanks!

PR #516, opened 2017-08-15
Move tracing implementation classes to io.opencensus.impl.trace.

sebright:
This is part of #379.  The package name may still need to be changed again, but
this commit includes all of the changes necessary to put the API and
implementation in separate packages.  Other changes in this commit:

- Make TraceComponent's constructor public.
- Leave deprecated copies of the TraceComponent subclasses in the io.opencensus.trace
  package so that they can be loaded by opencensus-api 0.5.
- Move internal classes to io.opencensus.impl.internal.
- Make TraceComponentImplBase and BinaryFormatImpl public.


impl/src/main/java/io/opencensus/impl/trace/TraceComponentImpl.java
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+
+package io.opencensus.impl.trace;
+
+import io.opencensus.impl.common.MillisClock;
+import io.opencensus.impl.internal.DisruptorEventQueue;
+import io.opencensus.impl.trace.TraceComponentImplBase;
+import io.opencensus.trace.TraceComponent;
+import io.opencensus.trace.internal.ThreadLocalRandomHandler;
```
bogdandrutu: Why is this not moved as well?

sebright: Thanks for catching that!  I updated it.

sebright:
Thanks!  I rebased, and I'll merge when the build passes.

PR #518, opened 2017-08-16
Disallow empty tag keys.

sebright:
Do you think we should add this restriction to tag keys?

### DIFF
--- a/api/src/main/java/io/opencensus/tags/TagKey.java
+++ b/api/src/main/java/io/opencensus/tags/TagKey.java
@@ -73,6 +73,6 @@ public abstract class TagKey {
    * @return whether the name is valid.
    */
   private static boolean isValid(String name) {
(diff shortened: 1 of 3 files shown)


dinooliva:
I don't see any particular reason to disallow empty tag keys. Eventually we'll likely have a specified format that tag keys will need to follow but I don't see an advantage to adding this as a special case now.

sebright:
@dinooliva I thought that it would be better to try to enforce all restrictions in the first release, for backwards compatibility.  What format do you think we'll need to enforce later?

dinooliva:
There is no agreed upon format so far, unfortunately - I'll look into it and see if we can resolve this soon.

sebright:
I created an issue: #525

songy23:
Please rebase.

sebright:
I rebased

sebright:
I discussed this with @dinooliva and others, and we are going to disallow empty tag keys in the first release.

PR #523, opened 2017-08-17
Add a naive JMH benchmark for ThreadInstrumentation.

ubschmidt2:
This is pretty naive. I currently don't have a good idea how to improve it, and don't want to sink too much time on it.


ubschmidt2:
From an example run:

```
Benchmark                                   Mode  Cnt   Score   Error  Units
ThreadInstrumentationBenchmark.automatic    avgt   10  68.847 ± 2.628  us/op
ThreadInstrumentationBenchmark.manual       avgt   10  67.598 ± 3.861  us/op
ThreadInstrumentationBenchmark.none         avgt   10  65.625 ± 1.625  us/op
```

PR #526, opened 2017-08-17
Use different package names in each artifact. (fixes #379)

### LINKED ISSUE #379: Avoid sharing Java packages between API and implementation libraries.

sebright:
This commit uses the following package name prefixes in each artifact, in order
to ensure that no package is split across artifacts:

    opencensus-api:       io.opencensus
    opencensus-impl-core: io.opencensus.implcore
    opencensus-impl:      io.opencensus.impl
    opencensus-impl-lite: io.opencensus.impllite

/cc @adriancole


impl_lite/src/main/java/io/opencensus/trace/TraceComponentImplLite.java
```diff
@@ -13,10 +13,10 @@
...
-import io.opencensus.impl.internal.SimpleEventQueue;
-import io.opencensus.impl.trace.TraceComponentImplBase;
-import io.opencensus.impl.trace.internal.RandomHandler.SecureRandomHandler;
+import io.opencensus.implcore.common.MillisClock;
+import io.opencensus.implcore.internal.SimpleEventQueue;
+import io.opencensus.implcore.trace.TraceComponentImplBase;
+import io.opencensus.implcore.trace.internal.RandomHandler.SecureRandomHandler;
 
 /** Android-compatible implementation of the {@link TraceComponent}. */
 // TraceComponentImplLite was moved to io.opencensus.impl.trace. This class exists for backwards
```
songy23: Is it safe to remove this class now? There's an identical class under impllite/trace.

sebright: We still need this class for backwards compatibility with opencensus-api 0.5, though I forgot to update the package name in the comment.

dinooliva:
Not that I particularly object to these changes but 1. what was the impetus behind this change? 2. is there a design doc describing this convention?

sebright:
This change is based on https://github.com/census-instrumentation/opencensus-java/pull/343#discussion_r120812070 .  I think we should avoid any possible problems caused by sharing packages across artifacts by also splitting up the implementation packages.  This is an easy change now that we have split the API and implementation packages.

I just chose the package names to match the artifact names in this PR, but I can change them if you have a better suggestion.

dinooliva:
Thanks for the context. I think artifact name makes sense.

sebright:
Thanks!

PR #531, opened 2017-08-18
Rename integration test classes to *IT.

ubschmidt2:
I'd like to rename integration test classes to `*IT` to avoid potential naming collisions with unit test classes. The `*IT` naming convention is somewhat widely used as it is the default in Maven's Failsafe plugin.

Now, `IT` triggers a checkstyle violation (module AbbreviationAsWordInName) given the current checkstyle rules. It seems to me that adding a `@SuppressWarnings("checkstyle:AbbreviationAsWordInName")` annotation would be a nice, uninstrusive solution. Thus enabling the support for `@SuppressWarnings` in `checkstyle.xml`, too.

### DIFF
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -225,6 +225,8 @@
         </module>
         <module name="CommentsIndentation"/>
         <module name="FileContentsHolder"/>
+        <module name="SuppressWarningsHolder"/>
     </module>
     <module name="SuppressionCommentFilter"/>
+    <module name="SuppressWarningsFilter"/>
 </module>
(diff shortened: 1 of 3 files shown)


sebright:
@ubschmidt2 Would it be better to increase `allowedAbbreviationLength` for the warning, so that the integration tests don't need to use SuppressWarnings?

ubschmidt2:
@sebright - I've considered this, but it may be undesirable to relax the checkstyle checks like this for the whole repo, especially considering https://google.github.io/styleguide/javaguide.html#s5.3-camel-case.

adriancole:
also important, users can skip IT in a more straightforward way (as the
naming convention is a signal)

ex in zipkin we have some docker tests which should have been IT named in
hindsight (in brave we use IT)

ubschmidt2:
@bogdandrutu - ok to merge?

PR #532, opened 2017-08-18
Add documentation on name restrictions.

songy23:
For View and Measure.

### DIFF
--- a/core/src/main/java/io/opencensus/stats/Measure.java
+++ b/core/src/main/java/io/opencensus/stats/Measure.java
@@ -13,8 +13,12 @@
 
 package io.opencensus.stats;
 
+import static com.google.common.base.Preconditions.checkArgument;
+import static io.opencensus.internal.StringUtil.MAX_LENGTH;
+
 import com.google.auto.value.AutoValue;
 import io.opencensus.common.Function;
+import io.opencensus.internal.StringUtil;
 import javax.annotation.concurrent.Immutable;
 
 /**
@@ -76,8 +80,8 @@ public abstract class Measure {
      * @return a {@code MeasureDouble}.
      */
     public static MeasureDouble create(String name, String description, String unit) {
-      // TODO(dpo): ensure that measure names are unique, and consider if there should be any
-      // restricitons on name (e.g. size, characters).
+      checkArgument(StringUtil.isPrintableString(name) && name.length() <= MAX_LENGTH,
+          "Name should be a ASCII string with a length no greater than 255 characters.");
       return new AutoValue_Measure_MeasureDouble(name, description, unit);
     }
 
@@ -112,8 +116,8 @@ public abstract class Measure {
      * @return a {@code MeasureLong}.
      */
     public static MeasureLong create(String name, String description, String unit) {
-      // TODO(dpo): ensure that measure names are unique, and consider if there should be any
-      // restricitons on name (e.g. size, characters).
+      checkArgument(StringUtil.isPrintableString(name) && name.length() <= MAX_LENGTH,
+          "Name should be a ASCII string with a length no greater than 255 characters.");
       return new AutoValue_Measure_MeasureLong(name, description, unit);
     }
 
--- a/core/src/main/java/io/opencensus/stats/Measure.java
+++ b/core/src/main/java/io/opencensus/stats/Measure.java
@@ -31,7 +31,8 @@ public abstract class Measure {
       Function<? super MeasureLong, T> p1);
 
   /**
-   * Name of measure, as a {@code String}. Should be a ASCII string with a length <= 255 characters.
+   * Name of measure, as a {@code String}. Should be a ASCII string with a length no greater than
+   * 255 characters.
    *
    * <p>Suggested format for name: {@code <web_host>/<path>}.
    */
--- a/core/src/main/java/io/opencensus/stats/Measure.java
+++ b/core/src/main/java/io/opencensus/stats/Measure.java
@@ -31,7 +31,9 @@ public abstract class Measure {
       Function<? super MeasureLong, T> p1);
 
   /**
-   * Name of measure, as a {@code String}.
+   * Name of measure, as a {@code String}. Should be a ASCII string with a length <= 255 characters.
+   *
+   * <p>Suggested format for name: {@code <web_host>/<path>}.
    */
   public abstract String getName();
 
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/opencensus/stats/View.java
```diff
@@ -75,7 +77,8 @@
   /**
    * Constructs a new {@link View}.
    *
-   * @param name the {@link Name} of view. Must be unique.
+   * @param name the {@link Name} of view. Must be unique. Suggested format for name:
+   *     {@code <web_host>/<path>}.
```
sebright: I think the suggested format should be specified on Name.create, and Name.create should validate the String.

core/src/main/java/io/opencensus/stats/View.java
```diff
@@ -37,7 +37,10 @@
   }
 
   /**
-   * Name of view. Must be unique.
+   * Name of view. Must be unique. Should be a ASCII string with a length no greater than 255
+   * characters.
+   *
+   * <p>Suggested format for name: {@code <web_host>/<path>}.
```
sebright: See my next comment.  The detail that you added should probably go in the Name class.

core/src/main/java/io/opencensus/stats/Measure.java
```diff
@@ -66,6 +69,11 @@ private Measure() {
...
     /**
      * Constructs a new {@link MeasureDouble}.
+     *
+     * @param name name of {@code Measure}. Suggested format: {@code <web_host>/<path>}.
+     * @param description description of {@code Measure}.
+     * @param unit unit of {@code Measure}.
+     * @return a {@code MeasureDouble}.
      */
     public static MeasureDouble create(String name, String description, String unit) {
       // TODO(dpo): ensure that measure names are unique, and consider if there should be any
```
sebright: This method should validate the String, using StringUtil.  You could also remove part of the TODO.

core/src/main/java/io/opencensus/stats/Measure.java
```diff
@@ -97,6 +105,11 @@ public static MeasureDouble create(String name, String description, String unit)
...
     /**
      * Constructs a new {@link MeasureLong}.
+     *
+     * @param name name of {@code Measure}. Suggested format: {@code <web_host>/<path>}.
+     * @param description description of {@code Measure}.
+     * @param unit unit of {@code Measure}.
+     * @return a {@code MeasureLong}.
      */
     public static MeasureLong create(String name, String description, String unit) {
       // TODO(dpo): ensure that measure names are unique, and consider if there should be any
```
sebright: Same here

core/src/main/java/io/opencensus/stats/Measure.java
songy23: Done.

core/src/main/java/io/opencensus/stats/Measure.java
songy23: Done.

core/src/main/java/io/opencensus/stats/View.java
songy23: Done.

core/src/main/java/io/opencensus/stats/View.java
songy23: Done. Also added a few tests.

core/src/test/java/io/opencensus/stats/MeasureTest.java
```diff
@@ -19,31 +19,30 @@
...
   @Test
-  public void testNameMaxLength() {
-    char[] name = new char[StringUtil.MAX_LENGTH];
-    char[] truncName = new char[StringUtil.MAX_LENGTH + 10];
-    Arrays.fill(name, 'n');
-    Arrays.fill(truncName, 'n');
-    assertThat(makeSimpleMeasure(new String(name)).getName())
-        .isEqualTo(makeSimpleMeasure(new String(truncName)).getName());
+  public void preventTooLongMeasureName() {
+    String longName = new String(new char[StringUtil.MAX_LENGTH + 1]).replace('\0', 'a');
```
songy23: Updated.

core/src/test/java/io/opencensus/stats/ViewTest.java
```diff
@@ -103,6 +110,20 @@ public void preventNullViewName() {
   }
 
   @Test
+  public void preventTooLongViewName() {
+    String longName = new String(new char[StringUtil.MAX_LENGTH + 1]).replace('\0', 'a');
```
songy23: Done.

core/src/main/java/io/opencensus/stats/Measure.java
```diff
@@ -31,7 +35,10 @@
       Function<? super MeasureLong, T> p1);
 
   /**
-   * Name of measure, as a {@code String}.
+   * Name of measure, as a {@code String}. Should be a ASCII string with a length no greater than
+   * 255 characters.
+   *
+   * <p>Suggested format for name: {@code <web_host>/<path>}.
```
adriancole: add rationale for suggested format? For example, this avoids problems with dots (ex in some backends like elasticsearch, dots hint at an object structure and interfere with indexing). Not sure your rationale..

songy23: I think this is following some RPC naming conventions but I'm not sure either. @dinooliva can you answer this question?

core/src/main/java/io/opencensus/stats/Measure.java
```diff
@@ -31,7 +35,10 @@
       Function<? super MeasureLong, T> p1);
 
   /**
-   * Name of measure, as a {@code String}.
+   * Name of measure, as a {@code String}. Should be a ASCII string with a length no greater than
+   * 255 characters.
```
dinooliva: Apologies - it should be 256 characters (here and elsewhere).

core/src/main/java/io/opencensus/stats/Measure.java
dinooliva: Theses are some of our cross-language restrictions - we plan to put a document in place soon with more explanation. Might be worth adding a todo to add a link to the document when it becomes available.

There is still some discussion on this particular point, although details should be resolved soon.

core/src/main/java/io/opencensus/stats/Measure.java
songy23: Updated

core/src/test/java/io/opencensus/stats/ViewTest.java
```diff
@@ -103,6 +113,22 @@ public void preventNullViewName() {
...
+    View.Name.create(longName);
+  }
+
+  @Test
+  public void preventNonPrintableViewName() {
+    thrown.expect(IllegalArgumentException.class);
+    View.Name.create("\2");
+  }
+
+
```
dinooliva: nit - extra line

PR #529, opened 2017-08-18
Implement StatsRecorder.record(MeasureMap).

sebright:
~~Two commits:~~

~~#### Give TAG_CONTEXT_KEY a default value (an empty TagContext).~~

#### Implement StatsRecorder.record(MeasureMap).

This `record` overload uses the current TagContext.  I also split StatsRecorderTest into StatsRecorderTest and NoopStatsRecorderTest.

PR #533, opened 2017-08-19
Give TAG_CONTEXT_KEY a default value (an empty TagContext).

sebright:
This commit also adds tests for io.opencensus.tags.unsafe.ContextUtils.


sebright:
Yes.  #529 adds a convenience method, StatsRecorder.record(MeasureMap), that records stats with the current tags.  The default value in this PR ensures that `record` isn't called with null.  I also made TagContexts.getCurrentContext() return an empty TagContext instead of null in #498, so this PR makes TagContexts.getCurrentContext() and TAG_CONTEXT_KEY.get() behave consistently.

PR #535, opened 2017-08-21
Add Duration.toMillis() method and tests.

songy23:
This method is needed to calculate interval stats based on time buckets, because buckets may be incomplete and we need to calculate the ratio of passed time / bucket duration. E.g, a time bucket starts at timestamp 0 and has a duration of 20 seconds, and now is 14th seconds, the aggregated data of this bucket should be weighted as `toMillis(14 - 0) / toMillis(20)` = 0.7. 

/cc @dinooliva

### DIFF
--- a/core_impl/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
+++ b/core_impl/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
@@ -228,12 +228,7 @@ final class MutableViewData {
     return tagValues;
   }
 
-  /**
-   * Returns the milliseconds representation of this {@code Duration}.
-   *
-   * @param duration {@code Duration}.
-   * @return the milliseconds representation of a {@code Duration}.
-   */
+  // Returns the milliseconds representation of a Duration.
   static long toMillis(Duration duration) {
     return duration.getSeconds() * MILLIS_PER_SECOND + duration.getNanos() / NANOS_PER_MILLI;
   }
--- a/core_impl/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
+++ b/core_impl/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
@@ -20,6 +20,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import io.opencensus.common.Clock;
+import io.opencensus.common.Duration;
 import io.opencensus.common.Function;
 import io.opencensus.common.Functions;
 import io.opencensus.common.Timestamp;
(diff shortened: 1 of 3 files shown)


api/src/main/java/io/opencensus/common/Duration.java
```diff
@@ -71,6 +71,15 @@ public static Duration fromMillis(long millis) {
   }
 
   /**
+   * Returns the milliseconds representation of this {@code Duration}.
+   *
+   * @return the milliseconds representation of this {@code Duration}.
+   */
+  public long toMillis() {
```
bogdandrutu: Why do you need this?

songy23: It's mainly for computing ratio between two Durations. I've moved this method to implcore instead of the api package.

core_impl/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
```diff
@@ -221,6 +225,16 @@ Timestamp getStart() {
   }
 
   /**
+   * Returns the milliseconds representation of this {@code Duration}.
+   *
+   * @param duration {@code Duration}.
+   * @return the milliseconds representation of a {@code Duration}.
+   */
+  static long toMillis(Duration duration) {
```
dinooliva: This is package-private rather than private for testing, correct? So two questions:
1. In open source, do we annotate with @VisibleForTesting or mark this in any way?
2. Should the comments be javadoc? I seem to recall that we were only going to add javadoc for public things but I don't recall if that changed.

bogdandrutu: We use  @VisibleForTesting if that is the case.

songy23: There will be other classes using this method, so I think it's not strictly `VisibleForTesting`. I agree since this method is package-private, using javadoc instead of plain comments might be unnecessary. I've updated the comments here.

dinooliva: If this method will be used across different classes, is there a utility class that it should be part of?

songy23: Currently we don't have such a util class in implcore. It might make sense to put all the static package-private methods together in a util class, though I think that will be in another PR.

dinooliva: Something like this is clearly a utility that should probably be in a util class - that may not be true for all package private methods but agree, we can look into that in a subsequent pr.

PR #536, opened 2017-08-22
Make MutableViewData abstract.

songy23:
Make MutableViewData abstract, so that we can have two separate sub-classes for Cumulative and Interval stats.

### DIFF
--- a/core_impl/src/main/java/io/opencensus/implcore/stats/MeasureToViewMap.java
+++ b/core_impl/src/main/java/io/opencensus/implcore/stats/MeasureToViewMap.java
@@ -20,6 +20,7 @@ import com.google.common.collect.HashMultimap;
 import com.google.common.collect.Multimap;
 import io.opencensus.common.Clock;
 import io.opencensus.common.Function;
+import io.opencensus.common.Timestamp;
 import io.opencensus.stats.MeasureMap;
 import io.opencensus.stats.Measurement;
 import io.opencensus.stats.Measurement.MeasurementDouble;
@@ -92,14 +93,15 @@ final class MeasureToViewMap {
   }
 
   // Records stats with a set of tags.
-  synchronized void record(TagContext tags, MeasureMap stats) {
+  synchronized void record(TagContext tags, MeasureMap stats, Timestamp timestamp) {
     Iterator<Measurement> iterator = stats.iterator();
     while (iterator.hasNext()) {
       Measurement measurement = iterator.next();
       Collection<MutableViewData> views = mutableMap.get(measurement.getMeasure().getName());
       for (MutableViewData view : views) {
         measurement.match(
-            new RecordDoubleValueFunc(tags, view), new RecordLongValueFunc(tags, view));
+            new RecordDoubleValueFunc(tags, view, timestamp),
+            new RecordLongValueFunc(tags, view, timestamp));
       }
     }
   }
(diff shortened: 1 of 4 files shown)


core_impl/src/main/java/io/opencensus/implcore/stats/StatsManager.java
```diff
@@ -34,14 +36,13 @@
...
+    checkNotNull(queue, "EventQueue");
+    checkNotNull(clock, "Clock");
     this.queue = queue;
     this.clock = clock;
   }
 
   void registerView(View view) {
-    if (view.getWindow() instanceof View.Window.Interval) {
-      throw new UnsupportedOperationException("IntervalView not supported yet.");
-    }
```
sebright: I think we still need this check.

core_impl/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
```diff
@@ -266,15 +233,64 @@ static AggregationData createAggregationData(MutableAggregation aggregation) {
         CreateMeanData.INSTANCE,
         CreateStdDevData.INSTANCE);
   }
-  
+
+  // Covert a mapping from TagValues to MutableAggregation, to a mapping from TagValues to
+  // AggregationData.
+  private static Map<List<Object>, List<AggregationData>> createAggregationMap(
+      Map<List<Object>, List<MutableAggregation>> tagValueAggregationMap) {
```
sebright: Suggestion: You could change the signature to

```java
  private static <T> Map<T, List<AggregationData>> createAggregationMap(
      Map<T, List<MutableAggregation>> tagValueAggregationMap) {
```

to show that the key type isn't relevant.

core_impl/src/main/java/io/opencensus/implcore/stats/StatsManager.java
songy23: Updated.

core_impl/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
songy23: Sure, updated.

core_impl/src/main/java/io/opencensus/implcore/stats/MeasureToViewMap.java
```diff
@@ -92,47 +93,52 @@ synchronized void registerView(View view, Clock clock) {
   }
 
   // Records stats with a set of tags.
-  synchronized void record(TagContext tags, MeasureMap stats) {
+  synchronized void record(TagContext tags, MeasureMap stats, Timestamp timestamp) {
```
dinooliva: Why does record() always need a timestamp?

songy23: IntervalView needs timestamps to determine which bucket the stats should be put into, thus we need to pass in the timestamp for each method that calls IntervalView.record(). We could have another record() method that doesn't need timestamp for CumulativeView, but since this is an implementation detail, I just chose to have one generic record() for both Interval and Cumulative views.

dinooliva: As discussed, for all of our current measures, we always record both cumulative and interval stats so the timestamp needs to be created so there is no overhead. Eventually, we may want to optimize cumulative stats but this should be fine for now.

PR #542, opened 2017-08-23
Move checkstyle files to buildscripts directory.

bogdandrutu:
Clean the main directory by moving the checkstyle and future build scripts into a buildscripts directory.


build.gradle
```diff
@@ -31,7 +31,6 @@ subprojects {
     apply plugin: "net.ltgt.apt"
     // Plugins that require java8
     if (JavaVersion.current().isJava8Compatible()) {
-        // TODO(bdrutu): enable all checks.
         apply plugin: "net.ltgt.errorprone"
```
sebright: What other checks needed to be enabled?

bogdandrutu: That was an old one that forgot to remove.

PR #544, opened 2017-08-23
Add combine() method to MutableAggregation.

songy23:
Add `MutableAggregation.combine()` so that we can combine `MutableAggregation`s from different `IntervalBucket`s.

### DIFF
--- a/core_impl/src/main/java/io/opencensus/implcore/stats/MutableAggregation.java
+++ b/core_impl/src/main/java/io/opencensus/implcore/stats/MutableAggregation.java
@@ -16,6 +16,7 @@
 
 package io.opencensus.implcore.stats;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import io.opencensus.common.Function;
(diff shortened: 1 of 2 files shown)


sebright:
It would help to see how `combine` is meant to be used for interval views.  Should I look at any of the other open PRs while reviewing this?

songy23:
Sure, PR https://github.com/census-instrumentation/opencensus-java/pull/545 contains all the changes, since the changes are huge, it might be easier to review by each commit.

core_impl/src/main/java/io/opencensus/implcore/stats/MutableAggregation.java
```diff
@@ -38,6 +39,18 @@ private MutableAggregation() {
...
+   * Creates a new {@code MutableAggregation} by combining the internal values of this {@code 
+   * MutableAggregation} and the given {@code MutableAggregation}, with the given fraction.
+   * 
+   * @param other the other {@code MutableAggregation}. The type of this and other {@code
+   *     MutableAggregation} must match.
+   * @param fraction the fraction that the value in other {@code MutableAggregation} should
+   *     contribute. Must be within [0.0, 1.0].
+   * @return a new {@code MutableAggregation} with the combined value.
+   */
+  abstract MutableAggregation combine(MutableAggregation other, double fraction);
```
sebright: It looks like this method is only used to create a ViewData when the stats are queried.  In that case, it seems like you could take advantage of the class' mutability to avoid returning a new MutableAggregation.  You could create an empty MutableAggregation for each Aggregation/TagValue combination and then call `combine` on the new MutableAggregation to add each time bucket to it.

songy23: Nice suggestion, actually this method is only used to add fractional values to an empty MutableAggregation. I've changed this method to return void and add values to its internal value instead.

core_impl/src/test/java/io/opencensus/implcore/stats/MutableAggregationTest.java
```diff
@@ -196,4 +197,39 @@ public String apply(MutableStdDev arg) {
...
+
+    List<Double> valueList1 = Arrays.asList(-1.0, -5.0);
+    List<Double> valueList2 = Arrays.asList(10.0, 50.0);
+
+    for (int i = 0; i < valueList1.size(); i++) {
+      for (MutableAggregation aggregation : aggregations1) {
+        aggregation.add(valueList1.get(i));
+      }
+      for (MutableAggregation aggregation : aggregations2) {
+        aggregation.add(valueList2.get(i));
```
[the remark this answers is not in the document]
songy23: Done.

songy23:
Thanks!

PR #548, opened 2017-08-24
Rename classes to make room for tracing-related methods.

ubschmidt2:
I want to add tracing-related methods, which are going to be called from classes loaded by the bootstrap classloader, using the exact same mechanism as in ContextManager, ContextStrategy, ContextStrategyImpl. It doesn't make much sense to introduce additional classes for this purpose, but instead I'm going to rename the existing classes to something more meaningful, which covers the extended use case:

*  ContextManager => TraceTrampoline
    (it was a trampoline between classes loaded by different classloaders from the beginning)
*  ContextStrategy => TraceStrategy
*  ContextStrategyImp => TraceStrategyImpl

Otherwise, this change is a no-op. I'll add tracing-related methods in subsequent changes.

### DIFF
--- a/contrib/agent/src/main/java/io/opencensus/contrib/agent/AgentMain.java
+++ b/contrib/agent/src/main/java/io/opencensus/contrib/agent/AgentMain.java
@@ -20,8 +20,8 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Preconditions.checkState;
 import static net.bytebuddy.matcher.ElementMatchers.none;
 
-import io.opencensus.contrib.agent.bootstrap.ContextManager;
-import io.opencensus.contrib.agent.bootstrap.ContextStrategy;
+import io.opencensus.contrib.agent.bootstrap.TraceStrategy;
+import io.opencensus.contrib.agent.bootstrap.TraceTrampoline;
 import io.opencensus.contrib.agent.instrumentation.Instrumenter;
 import java.lang.instrument.Instrumentation;
 import java.util.ServiceLoader;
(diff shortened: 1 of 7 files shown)


adriancole:
I'd prefer separation of concerns. context propagation being below and separate from trace and later metrics, or whatever. Maybe this just hints at the class functions being more like a trampoline etc which is different from what they are carrying

adriancole:
I'd prefer separation of concerns. context propagation being below and separate from trace and later metrics, or whatever. Maybe this just hints at the class functions being more like a trampoline etc which is different from what they are carrying.

Selfishly, I would like to be able to use the context propagation part separately

bogdandrutu:
I agree with @adriancole that it would be nice to have a way to configure what do we want to instrument with the agent (based on a config file or other ways).

@adriancole any recommendation about how the config should be done? I would prefer if possible one JAR in the end.

adriancole:
>
> @adriancole <https://github.com/adriancole> any recommendation about how
> the config should be done? I would prefer if possible one JAR in the end.
>
I think packaging the ultimate census jar into one uber jar is possible
with shade or similar. The config would have to be careful to not miss any
service loader entries which enable features.

ubschmidt2:
Thanks for the comments! I see a couple of different, independent aspects:

1.  Where to put ~2 additional trampoline methods (for creating trace spans and adding span attributes, etc.).
2.  Granular control over which agent features are enabled, e.g. context propagation yes, tracing no.
3.  Packaging of the agent JAR(s).

Re 1.: I'd prefer adding the tracing-relating methods to the existing classes (hence this change), but I don't mind introducing a new set of classes. I see the argument about separation of concerns, which is a valid one. This being said, what about having a **single trampoline class** and a **strategy-interface+impl per concern** (context, tracing), i.e. 5 classes total? PLMK. (Whatever works - I currently don't see a big enough impact from this design decision.)

Re 2.: I want to make the agent adequately configurable. This definitely includes configuration options for enabling/disabling context propagation and various trace instrumentations independent from each other. Plan is to use HOCON with this library: https://github.com/typesafehub/config. 

Re 3.: My plan is to give users a **one single, self-contained agent JAR**, which comes with all bells and whistles, but is configurable (via said HOCON library, with sensible defaults) and extensible (app owner can bring additional instrumentation plugins + custom tracing). Question is if this would fit your intended use case. PLMK if you want to use (or include) the agent in your projects in a different way.

adriancole:
In transit, but early reply is sounds good.

I do think the end packaging in an uber jar doesnt imply we can modularize
prior to that step, or make it possible for others to produce alternate
jars. For example we wont have the universe of all options in this repo.
This is why I thought service loader would be in use. If all options are to
be bundled might be big enough to be separate repo

No comment on hocon have heard it a couple times but no experience with it.

ubschmidt2:
Thanks! Will update this PR (or start a new one).

ubschmidt2:
Closing this for now, didn't find time to get back to this in the last weeks. Going to send a new PR.

PR #550, opened 2017-08-24
Add documentation about binary encoding format that we are using and the trace context.

bogdandrutu:
TODO: Add tags encoding.

### DIFF
--- a/documentation/BinaryEncoding.md
+++ b/documentation/BinaryEncoding.md
@@ -1,0 +1,104 @@
+# BINARY FORMAT
+
+## General Format
+Each encoding will have a 1 byte version followed by the version format encoding:
+
+`<version><version_format>`
+
+This will allow us to, in 1 deprecation cycle to completely switch to a new format if needed.
+
+## Version Format (version_id = 0)
+The version format for the version_id = 0 is based on ideas from proto encoding. The main 
+requirements are to allow adding and removing fields in less than 1 deprecation cycle. It
+contains a list of repeated fields:
+
+`<field><field>...`
+
+### Field
+Each field that we send on the wire will have the following format:
+
+`<field_id><field_format>`
+
+* `field_id` is a single byte.
+
+* `field_format` must be defined for each metadata field separately, that means that for field_id
+ = 0 in trace context the field_value may have a completely different representation than the 
+ field_id = 0 in the server-stats metadata.
+
+Each field is optional and MAY have defined a default value that can be used (if implementation 
+needs one) when the field is missing. Fields can be repeated, e.g. StringTag in the tagging example.
+
+### Serialization Rules
+Because each field has its own format that is not generically defined we are forced to always add
+new field ids at the end. The serialization MUST ensure that fields are serialized in version 
+order (i.e. fields from version (i) must precede fields from version (i+1)). This ordering 
+allows old decoders to ignore any new fields even if they do not know the format for that field.
+Systems that receive extra fields that they cannot decode MAY pass them on when possible (by 
+passing-through the whole opaque tail of bytes starting with the field id that the current 
+binary does not understand).
+
+### Deserialization Rules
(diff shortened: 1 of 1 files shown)


bogdandrutu:
cc @adriancole

documentation/BinaryEncoding.md
```diff
@@ -0,0 +1,99 @@
+# BINARY FORMAT
```
adriancole: maybe put a see-also link somewhere to the trace context repo?

bogdandrutu: Done.

bogdandrutu:
Moved here https://github.com/census-instrumentation/opencensus-specs/pull/1

documentation/BinaryEncoding.md
```diff
@@ -0,0 +1,104 @@
...
+3.  Different load between caller service and callee service might force callee to down sample.
+
+The behavior of other bits is currently undefined.
+
+#### Valid example
+{0, 0, 64, 65, 66, 67, 68, 69, 70, 71, 72, 73, 74, 75, 76, 77, 78, 79, 1, 97, 98, 99, 100, 101, 
+102, 103, 104, 2, 1}
+
+This corresponds to:
+* `traceId` = {64, 65, 66, 67, 68, 69, 70, 71, 72, 73, 74, 75, 76, 77, 78, 79}
```
duggelz: This is very confusing because it uses byte values which happen to be valid ASCII characters, leading one to think that they are perhaps hex digits or something.  It would be equally valid, and more instructive, to have a value like:

`traceId = { 00, 127, 128, 255, 10, 13, 00, 00, 01, 02, 03, 04, 64, 65, 66, 67 }`

documentation/BinaryEncoding.md
```diff
@@ -0,0 +1,104 @@
...
+### Trace Context
+
+#### Trace-id
+
+* `field_id` = 0
+* `len` = 16
+
+Is the ID of the whole trace forest. It is represented as a 16-bytes array,
+e.g. (in hex), `4bf92f3577b34da6a3ce929d0e0e4736`. All bytes 0 is considered invalid.
+
```
duggelz: Can you please explicitly mention that these bytes are (I assume) in big-endian order (== network order), as opposed to little endian?

bogdandrutu: no endianess implied here because trace-id is defined as a 16-byte array and span-id as a 8 byte array.

documentation/BinaryEncoding.md
bogdandrutu: Good suggestion, can you make a quick PR to change this?

https://github.com/census-instrumentation/opencensus-specs

documentation/BinaryEncoding.md
duggelz: To clarify, you're saying that this sequence of 32 hex digits `4bf92f3577b34da6a3ce929d0e0e4736`, is to be understood as a sequence of 16 bytes with values `0x4b`, `0xf9`, `0x2f`, `0x35`, `0x77`, `0xb3`, `0x4d`, `0xa6`, `0xa3`, `0xce`, `0x92`, `0x9d`, `0x0e`, `0x0e`, `0x47`, `0x36`, in that order, but not to be interpreted as the 128 bit number `0x4bf92f3577b34da6a3ce929d0e0e4736` aka `100985939111033328018442752961257817910` in base 10.

This is confusing to me, because we do actually internally treat the span id as a 64 bit integer.  I.e. we treat `spanId` = {97, 98, 99, 100, 101, 102, 103, 104} from your example as equivalent to (converting to hex digits) `spanId` = { 0x61, 0x62, 0x63, 0x64, 0x65, 0x66, 0x67, 0x68 }, and then (concatenating the bytes, treating as big-endian and converting to decimal) `spanId` = 7017280452245743464 as from an HTTP header with value `<traceid>/7017280452245743464;<options>`

PR #552, opened 2017-08-24
Make osx build not required and enable fast_finish.

bogdandrutu:
Because osx builds are very slow scheduled mark osx as not required and enable fast_finish which will mark the build as finished when all the required builds are done.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ dist: precise
 language: java
 
 matrix:
+  fast_finish: true
   include:
   - jdk: oraclejdk7
     env: TASK=BUILD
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -13,12 +14,16 @@ matrix:
...
+
   # Work around https://github.com/travis-ci/travis-ci/issues/2317
   - env: TASK=BUILD
     os: osx
 
-  - env: TASK=CHECK_GIT_HISTORY
-    os: linux
+  allow_failures:
+  - env: TASK=BUILD
+    os: osx
```
sebright: Could you please add a comment explaining that it is only an allowed failure due to the slow builds?

bogdandrutu: Done.

Issue #540, opened 2017-08-22
Occasional integration test failure in Java 7 Travis build


sebright:
Here are two examples:

https://travis-ci.org/census-instrumentation/opencensus-java/jobs/267052610
https://travis-ci.org/census-instrumentation/opencensus-java/jobs/265790708

Error message from the first link:

```
:opencensus-agent:integrationTestAug 22, 2017 5:45:08 AM io.opencensus.contrib.agent.AgentMain premain
INFO: Initializing.
Aug 22, 2017 5:45:09 AM io.opencensus.contrib.agent.AgentMain premain
INFO: Initialized.
io.grpc.Context: Storage override doesn't exist. Using default.
java.lang.ClassNotFoundException: io.grpc.override.ContextStorageOverride
	at java.net.URLClassLoader$1.run(URLClassLoader.java:366)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:355)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:354)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:425)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:308)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:358)
	at java.lang.Class.forName0(Native Method)
	at java.lang.Class.forName(Class.java:191)
	at io.grpc.Context.<clinit>(Context.java:132)
	at io.opencensus.contrib.agent.instrumentation.ContextStrategyImpl.wrapInCurrentContext(ContextStrategyImpl.java:54)
	at io.opencensus.contrib.agent.bootstrap.ContextManager.wrapInCurrentContext(ContextManager.java:71)
	at org.gradle.internal.concurrent.StoppableExecutorImpl.execute(StoppableExecutorImpl.java:38)
	at org.gradle.internal.remote.internal.hub.MessageHub.addHandler(MessageHub.java:131)
	at org.gradle.internal.remote.internal.hub.MessageHubBackedObjectConnection.addIncoming(MessageHubBackedObjectConnection.java:80)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker.startReceivingTests(TestWorker.java:95)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker.execute(TestWorker.java:66)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker.execute(TestWorker.java:45)
[...]

PR #556, opened 2017-08-25
Disable integration tests in the Agent. Always fail.

bogdandrutu:


sebright:
Is this change related to #540?  I think it would be better to just add an `Ignore` annotation to the failing test.

bogdandrutu:
That would be an option if the test would finish and not break everything. The test pauses for a long time (more than a minute) and I don't want to wait that much every time when I build locally.

bogdandrutu:
This is an workaround for https://github.com/census-instrumentation/opencensus-java/issues/557 until a fix lands.

PR #560, opened 2017-08-25
Fix build

sebright:
#### Fix failing current TagContext test.

This commit fixes a test failure caused by the merge of #498 and #533.


sebright:
@bogdandrutu The two commits fix independent build failures, so I'd rather not combine them.

PR #564, opened 2017-08-26
Add a TagValue class.

sebright:
This commit adds TagValue, which can represent a String, long, or boolean tag
value.  It replaces Object as the type for representing tag values in ViewData.

TagValue is an abstract class with three subclasses, similar to TagKey and Tag.
It provides a `match` method for accessing the contained value. TagValueString
is now a subclass of TagValue instead of an independent, top-level class.


core/src/main/java/io/opencensus/tags/TagValue.java
```diff
@@ -0,0 +1,182 @@
...
+   * A validated tag value associated with a {@link TagKeyString}.
+   *
+   * <p>Validation ensures that the {@code String} has a maximum length of {@link #MAX_LENGTH} and
+   * contains only printable ASCII characters.
+   */
+  @Immutable
+  @AutoValue
+  public abstract static class TagValueString extends TagValue {
+    /** The maximum length for a {@code String} tag value. The value is {@value #MAX_LENGTH}. */
+    public static final int MAX_LENGTH = 255;
```
songy23: I think length constraint is 256 characters now.

core/src/main/java/io/opencensus/tags/TagValue.java
```diff
@@ -0,0 +1,182 @@
...
+   * @param defaultFunction the function to call when the tag value has a type other than {@code
+   *     String}, {@code long}, or {@code boolean}.
+   * @param <T> The result type of the function.
+   * @return The result of calling the function that matches the tag value's type.
+   */
+  public abstract <T> T match(
+      Function<? super TagValueString, T> stringFunction,
+      Function<? super TagValueLong, T> longFunction,
+      Function<? super TagValueBoolean, T> booleanFunction,
+      Function<? super Tag, T> defaultFunction);
```
songy23: Tag -> TagValue

core/src/main/java/io/opencensus/tags/TagValue.java
```diff
@@ -0,0 +1,182 @@
...
+      Preconditions.checkArgument(isValid(value));
+      return new AutoValue_TagValue_TagValueString(value);
+    }
+
+    @Override
+    public final <T> T match(
+        Function<? super TagValueString, T> stringFunction,
+        Function<? super TagValueLong, T> longFunction,
+        Function<? super TagValueBoolean, T> booleanFunction,
+        Function<? super Tag, T> defaultFunction) {
```
songy23: Same

core/src/main/java/io/opencensus/tags/TagValue.java
sebright: Done

core/src/main/java/io/opencensus/tags/TagValue.java
sebright: Done

core/src/main/java/io/opencensus/tags/TagValue.java
sebright: Good catch!

core/src/main/java/io/opencensus/tags/TagValue.java
```diff
@@ -0,0 +1,182 @@
...
+        Function<? super TagValue, T> defaultFunction) {
+      return stringFunction.apply(this);
+    }
+
+    /**
+     * Returns the tag value as a {@code String}.
+     *
+     * @return the tag value as a {@code String}.
+     */
+    public abstract String asString();
```
dinooliva: Why 'as' rather then 'get' here and below?

sebright: The only reason I used asString/asLong/asBoolean is that Tag already has a `getValue()` method, so a user would have to call two different `getValue`s to get the unwrapped value from a Tag, i.e., `tag.getValue().getValue()`.

sebright:
Thanks!

PR #569, opened 2017-08-29
Always return implementation classes from TagContextsImpl methods.

sebright:
Now TagContextsImpl.getCurrentTagContext() always returns a TagContextImpl, and
TagContextsImpl.withTagContext(TagContext) always puts a TagContextImpl into
scope, even when they encounter other subclasses of TagContext.

### DIFF
--- a/core/src/main/java/io/opencensus/tags/TagContexts.java
+++ b/core/src/main/java/io/opencensus/tags/TagContexts.java
@@ -40,10 +40,8 @@ public abstract class TagContexts {
    *
    * @return the current {@code TagContext}.
    */
-  // TODO(sebright): Should we let the implementation override this method?
-  public final TagContext getCurrentTagContext() {
-    TagContext tags = CurrentTagContextUtils.getCurrentTagContext();
-    return tags == null ? empty() : tags;
+  public TagContext getCurrentTagContext() {
+    return CurrentTagContextUtils.getCurrentTagContext();
   }
 
   /**
@@ -68,7 +66,7 @@ public abstract class TagContexts {
    * @return a new builder created from the current {@code TagContext}.
    */
   public final TagContextBuilder currentBuilder() {
-    return toBuilder(getCurrentTagContext());
+    return toBuilder(CurrentTagContextUtils.getCurrentTagContext());
   }
 
   /**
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/opencensus/tags/TagContexts.java
```diff
@@ -80,7 +78,7 @@ public final TagContextBuilder currentBuilder() {
    * @return an object that defines a scope where the given {@code TagContext} is set to the current
    *     context.
    */
-  public final Scope withTagContext(TagContext tags) {
```
bogdandrutu: Why not final?

core_impl/src/main/java/io/opencensus/implcore/tags/TagContextsImpl.java
```diff
@@ -16,36 +16,75 @@
...
+  @Override
+  public Scope withTagContext(TagContext tags) {
+    return super.withTagContext(toTagContextImpl(tags));
+  }
+
+  private static TagContextImpl toTagContextImpl(TagContext tags) {
+    if (tags instanceof TagContextImpl) {
+      return (TagContextImpl) tags;
+    } else {
+      Iterator<Tag> i = tags.unsafeGetIterator();
```
bogdandrutu: I like this. :+1:

core/src/main/java/io/opencensus/tags/TagContexts.java
sebright: I overrode this method in TagContextsImpl to ensure that it always puts a TagContextImpl into scope, instead of some other TagContext subclass.

sebright:
I rebased to trigger a rebuild after a failure from #540.

core/src/main/java/io/opencensus/tags/TagContexts.java
adriancole: Why is not putting other subtypes in scope important? (Curious)

sebright: I thought that since `TagContextsImpl` is part of a specific OpenCensus implementation that it might be surprising if instance methods like `getCurrentTagContext` could return `TagContext` instances that are not part of that implementation.  TagContextImpl has some invariants that other subclasses might not follow, such as disallowing duplicate tag keys.  Then I made `withTagContext` convert the type to be consistent with `getCurrentTagContext`.

Do you think it would be better to not convert the `TagContext` type?

adriancole: here's a swipe https://github.com/census-instrumentation/opencensus-java/pull/580

Issue #520, opened 2017-08-16
Add extra API in Span


bogdandrutu:
Add Span.addAttribute(String, AttributeValue) to allow easy usage.

Current API:
```
tracer.getCurrentSpan().addAttributes(
    ImmutableMap.<String, AttributeValue>builder().put(
        "RequestMethod", AttributeValue.stringAttributeValue("GET"))
    .build());
```

Future API:
```
tracer.getCurrentSpan().addAttributes("RequestMethod",AttributeValue.stringAttributeValue("GET"));
```

PR #571, opened 2017-08-30
Add new api that is easier to use for user to add only one Attribute.

bogdandrutu:
Fixes https://github.com/census-instrumentation/opencensus-java/issues/520.

### DIFF
--- a/api/src/main/java/io/opencensus/trace/BlankSpan.java
+++ b/api/src/main/java/io/opencensus/trace/BlankSpan.java
@@ -34,12 +34,13 @@ public final class BlankSpan extends Span {
     super(SpanContext.INVALID, null);
   }
 
-  /** No-op implementation of the {@link Span#addAttribute(String, AttributeValue)} method. */
+  /** No-op implementation of the {@link Span#putAttribute(String, AttributeValue)} method. */
   @Override
-  public void addAttribute(String key, AttributeValue value) {}
+  public void putAttribute(String key, AttributeValue value) {}
 
   /** No-op implementation of the {@link Span#addAttributes(Map)} method. */
   @Override
+  @SuppressWarnings("deprecation")
   public void addAttributes(Map<String, AttributeValue> attributes) {}
 
   /** No-op implementation of the {@link Span#addAnnotation(String, Map)} method. */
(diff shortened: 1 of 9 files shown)


api/src/main/java/io/opencensus/trace/BlankSpan.java
```diff
@@ -34,6 +34,10 @@ private BlankSpan() {
     super(SpanContext.INVALID, null);
   }
 
+  /** No-op implementation of the {@link Span#addAttribute(String, AttributeValue)} method. */
+  @Override
+  public void addAttribute(String key, AttributeValue value) {}
```
adriancole: should this be putAttribute? what happens if you call this twice with the same key

bogdandrutu: Put is good only for Java developer, C# (they have an add method, but fails if key already there), C++ has insert and emplace, Python has "set" (not as a literal function, but they use it in explanation), Go does seem to only support direct access "m["route"] = 66" but they also use "set" to explain the action.

Seems to me that set is the winner form all these languages.

AlekseySanin: I think this should be language specific. putAttribute is the right name for Java and set for python, etc.

adriancole:
I agree. Forcing other language idioms in java is my least favorite thing
about OpenTracing. I hope we dont repeat that here

bogdandrutu:
We fixed the problem. Everything is put (in c++ probably be insert/emplace) and we have for each language what is specific.

AlekseySanin:
LGTM'd the addAttr->putAttr

Aleksey Sanin
work: (415) 736-5250
cell: (650) 450-1321

api/src/main/java/io/opencensus/trace/BlankSpan.java
```diff
@@ -34,8 +34,13 @@ private BlankSpan() {
...
   }
 
+  /** No-op implementation of the {@link Span#putAttribute(String, AttributeValue)} method. */
+  @Override
+  public void putAttribute(String key, AttributeValue value) {}
+
   /** No-op implementation of the {@link Span#addAttributes(Map)} method. */
   @Override
+  @SuppressWarnings("deprecation")
   public void addAttributes(Map<String, AttributeValue> attributes) {}
```
sebright: This method should have a Deprecation annotation instead of SuppressWarnings, if you also want to remove it from BlankSpan.

PR #580, opened 2017-09-04
Clarifies the contract around scoping a trace context

adriancole:
Users may assume that the current trace context is the same as what they
placed into scope. For example, they might wrap a trace context and be
suprised that whatever they decorated it with is lost. This clarifies
the type-level javadoc on `TagContexts` to clarify this isn't supported.

See https://github.com/census-instrumentation/opencensus-java/pull/569#discussion_r136727376

### DIFF
--- a/core/src/main/java/io/opencensus/tags/TagContexts.java
+++ b/core/src/main/java/io/opencensus/tags/TagContexts.java
@@ -25,10 +25,9 @@ import javax.annotation.concurrent.Immutable;
  * <p>This class returns {@link TagContextBuilder builders} that can be used to create the
  * implementation-dependent {@link TagContext}s.
  *
- * <p>Implementations may have different constraints and are free to require their own subtypes of
- * {@linkplain TagContext}. This means callers cannot assume the {@link #getCurrentTagContext()
- * current context} is the same instance as the one {@link #withTagContext(TagContext) placed into
- * scope}
+ * <p>Implementations may have different constraints and are free to convert tag contexts to their
+ * own subtypes. This means callers cannot assume the {@link #getCurrentTagContext() current
+ * context} is the same instance as the one {@link #withTagContext(TagContext) placed into scope}.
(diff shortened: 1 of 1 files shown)


core/src/main/java/io/opencensus/tags/TagContexts.java
```diff
@@ -24,6 +24,11 @@
  *
  * <p>This class returns {@link TagContextBuilder builders} that can be used to create the
  * implementation-dependent {@link TagContext}s.
+ *
+ * <p>Implementations may have different constraints and are free to require their own subtypes of
+ * {@linkplain TagContext}. This means callers cannot assume the {@link #getCurrentTagContext()
```
sebright: > are free to require their own subtypes of {@linkplain TagContext}.

Something like "are free to convert TagContexts to their own subtypes" might be clearer, since the implementation shouldn't throw an exception when given a different subtype.

adriancole:
>
> Thanks for clarifying the behavior! I was thinking of #569
> <https://github.com/census-instrumentation/opencensus-java/pull/569> as
> handling an edge case, and I didn't consider that users might want to wrap
> a TagContext. Do you think that the new behavior would interfere with any
> important use cases?
>
I don't know if there's a compelling wrap use case, suffice to say people
intercept and wrap things often enough. So, this is mainly a defensive
comment.


> I forgot to mention that another reason I converted the TagContexts to
> TagContextImpls in #569
> <https://github.com/census-instrumentation/opencensus-java/pull/569> is
> that it is closer to the behavior of TagContext before we made it abstract.
> Users can assume that a TagContext that they obtain from this
> implementation is a simple value object.
>
great

> Something like "are free to convert TagContexts to their own subtypes"
> might be clearer, since the implementation shouldn't throw an exception
> when given a different subtype.
>
nice will update

sebright:
Thanks!

PR #582, opened 2017-09-05
Implement Duration.compareTo()

songy23:
Add compareTo() method for Duration. Duration will be ordered similar to Timestamp.

### DIFF
--- a/api/src/main/java/io/opencensus/common/Duration.java
+++ b/api/src/main/java/io/opencensus/common/Duration.java
@@ -20,6 +20,7 @@ import static io.opencensus.common.TimeUtil.MAX_NANOS;
 import static io.opencensus.common.TimeUtil.MAX_SECONDS;
 import static io.opencensus.common.TimeUtil.MILLIS_PER_SECOND;
 import static io.opencensus.common.TimeUtil.NANOS_PER_MILLI;
+import static io.opencensus.common.TimeUtil.compareLong;
 
 import com.google.auto.value.AutoValue;
 import javax.annotation.concurrent.Immutable;
@@ -31,7 +32,6 @@ import javax.annotation.concurrent.Immutable;
  */
 @Immutable
 @AutoValue
-// TODO(songya): implements Comparable<Duration>
 public abstract class Duration {
   private static final Duration ZERO = create(0, 0);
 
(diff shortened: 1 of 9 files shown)


sebright:
> I didn't implement Comparable or compareTo(), since there could be positive and negative Durations, and I think it's hard to determine the reasonable behavior when we compare a positive Duration with a negative one.

Shouldn't `compareTo` treat positive durations as greater than negative durations?

songy23:
That's my initial thought, but I'm not sure about the use case when users call compareTo(). For example, what if they just want to compare or sort durations by length?

api/src/main/java/io/opencensus/common/Duration.java
```diff
@@ -88,5 +88,21 @@ public static Duration fromMillis(long millis) {
...
+  /**
+   * Compares the length of this {@code Duration} to the specified {@code Duration}.
+   *
+   * @param otherDuration the other {@code Duration} to compare to, not {@code null}.
+   * @return the comparator value: zero if equal, negative if this duration is shorter
+   *     than otherDuration, positive if larger.
+   * @throws NullPointerException if otherDuration is {@code null}.
+   */
+  public int compareLength(Duration otherDuration) {
+    int cmp = compareLong(Math.abs(getSeconds()), Math.abs(otherDuration.getSeconds()));
```
sebright: Suggestion: You could call `com.google.common.primitives.Longs.compare`.

sebright: I think that a positive length should be greater than a negative length, similar to a unit-less comparison.

core_impl/src/main/java/io/opencensus/implcore/stats/IntervalBucket.java
```diff
@@ -75,7 +75,7 @@ void record(List<TagValue> tagValues, double value) {
    */
   double getFraction(Timestamp now) {
     Duration elapsedTime = now.subtractTimestamp(start);
-    checkArgument(toMillis(elapsedTime) >= 0 && toMillis(elapsedTime) < toMillis(duration),
+    checkArgument(toMillis(elapsedTime) >= 0 && elapsedTime.compareLength(duration) < 0,
```
sebright: `toMillis(elapsedTime) >= 0` could compare to a zero-length Duration.

api/src/main/java/io/opencensus/common/Duration.java
songy23: From my understanding lengths are all positive and should be the absolute value of Duration, while Duration could have negative values. I feel it's a little counterintuitive to have negative length.

songy23: Updated.

core_impl/src/main/java/io/opencensus/implcore/stats/IntervalBucket.java
songy23: Done.

songy23:
Updated to implement compareTo(), based on our discussion.

songy23:
Thanks!

PR #592, opened 2017-09-08
Add separate buckets for sampled and not sampled spans.

bogdandrutu:


impl_core/src/test/java/io/opencensus/implcore/trace/export/SampledSpanStoreImplTest.java
```diff
@@ -83,7 +88,7 @@ public void setUp() {
     sampleStore.registerSpanNamesForCollection(Arrays.asList(REGISTERED_SPAN_NAME));
   }
 
-  SpanImpl createSpan(String spanName) {
```
AlekseySanin: Should we keep the name? To ensure backward compatibility?

bogdandrutu: It is test (so no need) (nobody links this) + it is part of implementation package where we don't promise backwards compatibility.

impl_core/src/main/java/io/opencensus/implcore/trace/export/SampledSpanStoreImpl.java
```diff
@@ -51,25 +51,44 @@
...
-      // this may never sample again (at least for the next ~200 years). No real chance to
-      // overflow two times because that means the process runs for ~200 years.
-      if (spanEndNanoTime - lastSampleNanoTime > TIME_BETWEEN_SAMPLES) {
-        queue.add(span);
-        lastSampleNanoTime = spanEndNanoTime;
+      if (span.getContext().getTraceOptions().isSampled()) {
+        // Need to compare by doing the subtraction all the time because in case of an overflow,
+        // this may never sample again (at least for the next ~200 years). No real chance to
+        // overflow two times because that means the process runs for ~200 years.
+        if (spanEndNanoTime - lastSampledNanoTime > TIME_BETWEEN_SAMPLES) {
```
adriancole: since this is copy/paster, the comment also is. One way to declutter that is to make a method that is make a method duration(long endNano, long beginNano) use it everywhere instead of direct subtract, if only to centralize the comment. That or delete the comment and wait 200years for someone to ask.

PR #603, opened 2017-09-08
Change TagContext serialization methods to use byte[] (closes #213).

### LINKED ISSUE #213: Update StatsSerializer.serialze() and deserialze() to use byte[] instead of Input/OutputStream
From our conversation in PR https://github.com/google/instrumentation-java/pull/208, gRPC team suggested to use byte[] instead of Input[Output]Stream. Consider updating the input and output format of serialize() and deserialize().

sebright:
The signatures changed from

```java
    void serialize(TagContext tags, OutputStream output) throws IOException
    TagContext deserialize(InputStream input) throws IOException
```

to

```java
    byte[] toByteArray(TagContext tags)
    TagContext fromByteArray(byte[] bytes) throws IOException
```

fromByteArray can only fail if there is a parse error, so we should probably
change it to use a more appropriate exception type later.

This commit also refactors the toByteArray implementation to remove the
possibility of ~~an unchecked~~ a checked exception, since it is no longer doing IO.

/cc @bogdandrutu

### DIFF
--- a/core/src/main/java/io/opencensus/tags/TagContextBinarySerializer.java
+++ b/core/src/main/java/io/opencensus/tags/TagContextBinarySerializer.java
@@ -17,8 +17,6 @@
 package io.opencensus.tags;
 
 import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
 import javax.annotation.concurrent.Immutable;
 
 /** Object for serializing and deserializing {@link TagContext}s with the binary format. */
@@ -29,25 +27,24 @@ public abstract class TagContextBinarySerializer {
   /**
    * Serializes the {@code TagContext} into the on-the-wire representation.
    *
-   * <p>This method should be the inverse of {@link #deserialize}.
+   * <p>This method should be the inverse of {@link #fromByteArray}.
    *
    * @param tags the {@code TagContext} to serialize.
-   * @param output the {@link OutputStream} to write the serialized tags to.
-   * @throws IOException if there is an {@code IOException} while writing to {@code output}.
+   * @return the on-the-wire representation of a {@code TagContext}.
    */
-  public abstract void serialize(TagContext tags, OutputStream output) throws IOException;
+  public abstract byte[] toByteArray(TagContext tags);
 
   /**
    * Creates a {@code TagContext} from the given on-the-wire encoded representation.
    *
-   * <p>This method should be the inverse of {@link #serialize}.
+   * <p>This method should be the inverse of {@link #toByteArray}.
    *
-   * @param input on-the-wire representation of a {@code TagContext}.
-   * @return a {@code TagContext} deserialized from {@code input}.
-   * @throws IOException if there is a parse error or an {@code IOException} while reading from
-   *     {@code input}.
+   * @param bytes on-the-wire representation of a {@code TagContext}.
+   * @return a {@code TagContext} deserialized from {@code bytes}.
+   * @throws IOException if there is a parse error.
    */
-  public abstract TagContext deserialize(InputStream input) throws IOException;
+  // TODO(sebright): Use a more appropriate exception type, since this method doesn't do IO.
+  public abstract TagContext fromByteArray(byte[] bytes) throws IOException;
 
   /**
    * Returns a {@code TagContextBinarySerializer} that serializes all {@code TagContext}s to zero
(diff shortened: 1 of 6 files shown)


bogdandrutu:
So we throw only checked exceptions now?

sebright:
Sorry, there was a typo in my commit message.  I meant that it doesn't throw a checked exception.  I fixed the message.  The methods could still throw unchecked exceptions, like NPEs for null arguments.

adriancole:
I like byte array as less chance for io problems (as they happen prior) cool

adriancole:
Will look closer at this later today or tomorrow but if need to merge
before do it. I want to closer look at exceptions as for example in zipkin
we usually say encoding is an impl bug (as it should be impossible to not
be able to write a known type to a byte array) and decoding could possibly
throw IllegalArg or similar, but also there could be value in a wrapper

core/src/main/java/io/opencensus/tags/TagContextBinarySerializer.java
```diff
@@ -29,25 +27,24 @@
...
-   * <p>This method should be the inverse of {@link #serialize}.
+   * <p>This method should be the inverse of {@link #toByteArray}.
    *
-   * @param input on-the-wire representation of a {@code TagContext}.
-   * @return a {@code TagContext} deserialized from {@code input}.
-   * @throws IOException if there is a parse error or an {@code IOException} while reading from
-   *     {@code input}.
+   * @param bytes on-the-wire representation of a {@code TagContext}.
+   * @return a {@code TagContext} deserialized from {@code bytes}.
+   * @throws IOException if there is a parse error.
```
dinooliva: I agree, there is probably a better checked exception - possibly ParseException or DataFormatException but we can look into it more.

core/src/main/java/io/opencensus/tags/TagContextBinarySerializer.java
```diff
@@ -29,25 +27,24 @@
...
-   * <p>This method should be the inverse of {@link #deserialize}.
+   * <p>This method should be the inverse of {@link #fromByteArray}.
    *
    * @param tags the {@code TagContext} to serialize.
-   * @param output the {@link OutputStream} to write the serialized tags to.
-   * @throws IOException if there is an {@code IOException} while writing to {@code output}.
+   * @return the on-the-wire representation of a {@code TagContext}.
    */
-  public abstract void serialize(TagContext tags, OutputStream output) throws IOException;
+  public abstract byte[] toByteArray(TagContext tags);
```
dinooliva: Is this what the gRPC team was asking for? In the past, they had an issue with this approach because it requires always creating a new byte[] rather than appending to a given one e.g. appendToByteArray(tags, byte[], int offset) - that adds its own set of problems (e.g. if the input array isn't long enough) but is a concern for client code where performance is a major concern.

sebright: I tried to match the BinaryFormat API from the tracing package: https://github.com/census-instrumentation/opencensus-java/blob/bf42c2c695ba80eef4b94745b187b0d553309f0d/api/src/main/java/io/opencensus/trace/propagation/BinaryFormat.java

I'll cc @zhangkun83 .

zhangkun83: Returning a `byte[]` is not an issue for gRPC, because [Metadata.BinaryMarshaller's serialization method](https://github.com/grpc/grpc-java/blob/v1.6.1/core/src/main/java/io/grpc/Metadata.java#L76) returns a `byte[]` too.

core/src/main/java/io/opencensus/tags/TagContextBinarySerializer.java
bogdandrutu: Completely agreed. I think we can change the Exception later but it has to be checked.

core/src/main/java/io/opencensus/tags/TagContextBinarySerializer.java
adriancole: if writing to an offset is an issue, we could overload later, keeping simpler for now.

core/src/main/java/io/opencensus/tags/TagContextBinarySerializer.java
adriancole: why does it need to be a checked exception?

adriancole: inevitably some impl will throw a kind of unchecked exception, so we have to catch RuntimeException anyway. Languages like kotlin prefer unchecked. In the case of not doing I/O, I'd personally prefer a convention of using IllegalArgumentException (or Malformed subtype of RuntimeException), as it avoids bikeshedding about checked vs unchecked.

core_impl/src/main/java/io/opencensus/implcore/tags/SerializationUtils.java
```diff
@@ -91,57 +90,47 @@ private SerializationUtils() {}
...
+              return null;
+            }
+          },
+          new Function<TagLong, Void>() {
+            @Override
+            public Void apply(TagLong arg) {
+              throw new IllegalArgumentException("long tags are not supported.");
+            }
+          },
+          new Function<TagBoolean, Void>() {
```
adriancole: seems like we want Functions.booleanUnsupported() to avoid allocating one of these each time.

sebright: I opened #625 to address this.  I didn't add the constants to the API, though, since they won't be needed once we support boolean and long.

sebright:
I'm going to merge this now, and we can continue discussing the error handling in #604.  Thanks for the reviews!

Issue #584, opened 2017-09-05
Decide how to provide access to TagContext tags in the first release.


sebright:
Currently, `TagContext` doesn't implement `Iterable`, but it has a method `unsafeGetIterator()`.  I can think of three options for handling tag access in the first release:

#### Pros
- This option provides the most functionality and convenience for users.
- It is also convenient for accessing tags from unknown `TagContext` subclasses within the implementation.
#### Cons
- The implementation would need to have reasonable performance when accessing tags.
- It would be hard to remove access to tags in the future.

#### Pros
- This choice would give the library more flexibility to allow access to the tags in a later release.
#### Cons
- The implementation would not be able to access the tags contained in any `TagContext` subclass that isn't `TagContextImpl`, and all other `TagContexts` would be treated as empty.
- Adding an interator in a later release wouldn't work well with existing `TagContext` subclasses.  We would need to provide a default `iterator()` implementation, which would be empty.

#### Pros
- The iterator wouldn't need to have good performance.
#### Cons
- Users would need to implement an "unsafe" method to subclass `TagContext`.

sebright:
I implemented the first option in #605.

sebright:
Since we aren't going to support access to tags in the first release, I think we should go with (3).  It will allow us to provide a default implementation for `iterator()` if we make TagContext implement Iterable in the future, and it will simplify handling TagContexts of unknown subtype in the implementation.

sebright:
Closing for now, since we won't implement Iterable in the first release.

PR #605, opened 2017-09-08
Make TagContext implement Iterable<Tag>.

sebright:
This commit also removes TagContext.unsafeGetIterator().  It is one solution
to #584.

/cc @adriancole


dinooliva:
In the first version of the library, it was decided not to give general access to the tags in the TagContext - we can revisit the issue if there's a compelling argument for doing so.

PR #606, opened 2017-09-08
Add initial support for b3-propagation headers.

bogdandrutu:


api/src/main/java/io/opencensus/trace/propagation/B3Format.java
```diff
@@ -0,0 +1,143 @@
...
+ *   SpanContext spanContext = b3Format.fromHeaders(b3Headers);
+ *   try (Scope ss =
+ *            tracer.spanBuilderWithRemoteParent("Recv.MyRequest", spanContext).startScopedSpan()) {
+ *     // Handle request and send response back.
+ *   }
+ * }
+ * }</pre>
+ */
+public abstract class B3Format {
+  public enum HeaderName {
```
sebright: I think the name should include B3, such as `B3HeaderName`, since it can't represent all header names.

impl_core/src/main/java/io/opencensus/implcore/trace/propagation/B3FormatImpl.java
```diff
@@ -0,0 +1,80 @@
...
+    checkNotNull(headers, "headers");
+    try {
+      TraceId traceId = TraceId.INVALID;
+      String traceIdStr = headers.get(HeaderName.X_B3_TRACE_ID);
+      if (traceIdStr != null) {
+        if (traceIdStr.length() == TraceId.SIZE) {
+          // This is an 8-byte traceID.
+          traceId = TraceId.fromLowerBase16(UPPER_TRACE_ID + traceIdStr);
+        } else {
+          traceId = TraceId.fromLowerBase16(traceIdStr);
```
sebright: Should this case also check the length?

impl_core/src/test/java/io/opencensus/implcore/trace/propagation/B3FormatImplTest.java
```diff
@@ -0,0 +1,67 @@
...
+  private static final String TRACE_ID_BASE16 = "ff000000000000000000000000000041";
+  private static final TraceId TRACE_ID = TraceId.fromLowerBase16(TRACE_ID_BASE16);
+  private static final String SPAN_ID_BASE16 = "ff00000000000041";
+  private static final SpanId SPAN_ID = SpanId.fromLowerBase16(SPAN_ID_BASE16);
+  private static final byte[] TRACE_OPTIONS_BYTES = new byte[] {1};
+  private static final TraceOptions TRACE_OPTIONS = TraceOptions.fromBytes(TRACE_OPTIONS_BYTES);
+  private static final SpanContext SPAN_CONTEXT =
+      SpanContext.create(TRACE_ID, SPAN_ID, TRACE_OPTIONS);
+  private static final EnumMap<HeaderName, String> headers =
+      new EnumMap<HeaderName, String>(HeaderName.class);
```
sebright: Shouldn't `headers` be non-static, since it is updated in the setup method?  Alternatively, you could use an immutable map builder to initialize it statically.

impl_core/src/test/java/io/opencensus/implcore/trace/propagation/B3FormatImplTest.java
```diff
@@ -0,0 +1,67 @@
...
+  public void serializeValidContext() {
+    assertThat(b3Format.toHeaders(SPAN_CONTEXT)).isEqualTo(headers);
+  }
+
+  @Test
+  public void parseValidHeaders() throws ParseException {
+    assertThat(b3Format.fromHeaders(headers)).isEqualTo(SPAN_CONTEXT);
+  }
+
+  // TODO(bdrutu): Add tests for invalid inputs + 8-byte trace id.
```
sebright: I think this PR should include the tests for error handling.

api/src/main/java/io/opencensus/trace/propagation/B3Format.java
```diff
@@ -0,0 +1,143 @@
...
+    public String getValue() {
+      return value;
+    }
+
+    /**
+     * Returns the ASCII value of this header (lowercase).
+     *
+     * @return the ASCII value of this header (lowercase).
+     */
+    public String getLowerValue() {
```
sebright: How will this method be used?

impl_core/src/main/java/io/opencensus/implcore/trace/propagation/B3FormatImpl.java
```diff
@@ -0,0 +1,80 @@
...
+/** Implementation of the {@link B3Format}. */
+@VisibleForTesting
+public final class B3FormatImpl extends B3Format {
+  // Used as the upper 8-bytes of the traceId when a 8-bytes traceId is received.
+  private static final String UPPER_TRACE_ID = "0000000000000000";
+  private static final String SAMPLED_VALUE = "1";
+  private static final String FLAGS_VALUE = "1";
+
+  @Override
+  public Map<HeaderName, String> toHeaders(SpanContext spanContext) {
```
sebright: It looks like this implementation always ignore the parent span ID.  Since there isn't a one-to-one mapping between OpenCensus spans and B3 headers, I think there should be an explanation of the conversion in the API Javadocs.

adriancole:
In transit but will try to review between flights today

api/src/main/java/io/opencensus/trace/SpanId.java
```diff
@@ -84,6 +84,20 @@ public static SpanId fromBytes(byte[] src, int srcOffset) {
   }
 
   /**
+   * Returns a {@code SpanId} built from a lowercase base16 representation.
+   *
+   * @param src the lowercase base16 representation.
+   * @return a {@code SpanId} built from a lowercase base16 representation.
+   * @throws NullPointerException if {@code src} is null.
+   */
+  public static SpanId fromLowerBase16(String src) {
```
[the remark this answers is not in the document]
adriancole: possibly this is hinting at the trace context format which is also lower base 16? https://github.com/TraceContext/tracecontext-spec/blob/master/HTTP_HEADER_FORMAT.md#format-1

api/src/main/java/io/opencensus/trace/propagation/B3Format.java
adriancole: possibly the intended use is B3Format.HeaderName.X (I use this a lot in other libraries)

api/src/main/java/io/opencensus/trace/propagation/B3Format.java
adriancole: guess is in grpc? (there's existing grpc instrumentation that use B3, and grpc downcases header names)

api/src/main/java/io/opencensus/trace/propagation/B3Format.java
```diff
@@ -0,0 +1,143 @@
...
+  /**
+   * Returns the no-op implementation of the {@code B3Format}.
+   *
+   * @return the no-op implementation of the {@code B3Format}.
+   */
+  static B3Format getNoopB3Format() {
+    return NOOP_B3_FORMAT;
+  }
+
+  private static final class NoopB3Format extends B3Format {
```
adriancole: food for thought, but this is what brave uses. In practice `C` is usually an httprequest object. We have a supporting type Getter and Setter to replace elements in the carrier. This prevents us from having to play any games with making a copy of a map especially as almost no http api has request headers as a string-> string map.

```
 /**
   * Used to send the trace context downstream. For example, as http headers.
   */
  public interface Injector<C> {
    /**
     * Usually calls a setter for each propagation field to send downstream.
     *
     * @param traceContext possibly unsampled.
     * @param carrier holds propagation fields. For example, an outgoing message or http request.
     */
    void inject(TraceContext traceContext, C carrier);
  }

  /**
   * Used to join an incoming trace. For example, by reading http headers.
   *
   * @see brave.Tracer#joinSpan
   */
  public interface Extractor<C> {

    /**
     * Returns either a trace context or sampling flags parsed from the carrier. If nothing was
     * parsable, sampling flags will be set to {@link SamplingFlags#EMPTY}.
     *
     * @param carrier holds propagation fields. For example, an incoming message or http request.
     */
    TraceContextOrSamplingFlags extract(C carrier);
  }
```

sebright: Does Brave implement Injector and Extractor, and then the user implements a Getter and Setter specific to their request type?

adriancole: For simple string things like http headers, it is easy.

```java

or more simply for http:

```java
  static final Propagation.Setter<MultivaluedMap, String> SETTER = MultivaluedMap::putSingle;

  @Inject TracingClientFilter(HttpTracing httpTracing) {
--snip--
    injector = httpTracing.tracing().propagation().injector(SETTER);
  }

--snip--
    injector.inject(span.context(), request.getHeaders());
```

gRPC uses predefined keys, so requires a bit more coordination
```java
enum AsciiMetadataKeyFactory implements Propagation.KeyFactory<Key<String>> {
  INSTANCE;

  @Override public Key<String> create(String name) {
    return Key.of(name, Metadata.ASCII_STRING_MARSHALLER);
  }
}
...
  static final Setter<Metadata, Metadata.Key<String>> SETTER = (metadata, key, value) -> {
    metadata.removeAll(key);
    if (value != null) metadata.put(key, value);
  };
...
  final TraceContext.Injector<Metadata> injector;

  TracingClientInterceptor(Tracing tracing) {
    tracer = tracing.tracer();
    injector = tracing.propagationFactory().create(AsciiMetadataKeyFactory.INSTANCE).injector(SETTER);
  }
...
    @Override
    public void start(Listener<RespT> responseListener, Metadata headers) {
      injector.inject(span.context(), headers);
```

api/src/main/java/io/opencensus/trace/SpanId.java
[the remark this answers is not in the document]
bogdandrutu: we have to also add the X-Cloud-Trace-Context (the old SD defined format) support as well, which uses hex for traceId.

Also the new tracecontext format is using lower hex. So we need this between multiple implementation.

adriancole:
I started looking at this again, and not sure this matches completely what we need to achieve. For example, in grpc, you need to know header names up front in order to allocate keys for them. even if grpc always uses trace-context and we want to just export on other protocols B3, we'll still need this (as otherwise the impls have to know too much.. ex which keys to clear)

an integrated example might highlight this more clearly. a lot of the back/forth on brave's propagation design was on account of things like this. If you don't mind, I might spike a variation of that?

adriancole:
This is what I mean https://github.com/census-instrumentation/opencensus-java/pull/667

you can feel free to incorporate this into your PR if you find it helpful. I raised a PR mainly for discussion

bogdandrutu:
PTAL. Refactor based on https://github.com/census-instrumentation/opencensus-java/pull/667

PR #609, opened 2017-09-08
Implement equals and hashCode for TagContext.

sebright:
Both methods are non-final so that subclasses can provide more efficient
implementations.  For example, TagContextImpl can compare the tag maps directly
when it is compared with another TagContextImpl.

/cc @bogdandrutu

### DIFF
--- a/core/src/main/java/io/opencensus/tags/TagContext.java
+++ b/core/src/main/java/io/opencensus/tags/TagContext.java
@@ -62,12 +62,8 @@ public abstract class TagContext {
     return tags1.equals(tags2);
   }
 
-  /**
-   * Returns the sum of the hash codes of all tags in this {@code TagContext}, ignoring nulls.
-   * Implementations are free to override this method to provide better performance.
-   */
   @Override
-  public int hashCode() {
+  public final int hashCode() {
     int hashCode = 0;
     for (Iterator<Tag> i = unsafeGetIterator(); i.hasNext(); ) {
       Tag tag = i.next();
--- a/core/src/main/java/io/opencensus/tags/TagContext.java
+++ b/core/src/main/java/io/opencensus/tags/TagContext.java
@@ -16,6 +16,8 @@
 
 package io.opencensus.tags;
 
+import com.google.common.collect.HashMultiset;
+import com.google.common.collect.Lists;
 import io.opencensus.tags.TagValue.TagValueString;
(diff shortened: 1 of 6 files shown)


core/src/main/java/io/opencensus/tags/TagContext.java
```diff
@@ -41,6 +42,38 @@
...
+   * Returns true iff the other object is an instance of {@code TagContext} and contains the same
+   * key-value pairs. Implementations are free to override this method to provide better
+   * performance.
+   */
+  @Override
+  public boolean equals(Object other) {
+    if (!(other instanceof TagContext)) {
+      return false;
+    }
+    HashMultiset<Tag> tags1 = HashMultiset.create(Lists.newArrayList(unsafeGetIterator()));
```
songy23: I think unsafeGetIterator() has been replaced by iterator()?

sebright: I made a PR for that change (#605), but it hasn't been merged yet.  If it is approved, then I can just replace all uses of unsafeGetIterator() with iterator().

sebright:
I have a question about this implementation.  Would it be better to make `TagContext.hashCode()` final?  Then we wouldn't need to specify it's behavior, because no one would need to re-implement it.  We would be free to change it.  The downside is that classes that override `equals` wouldn't be able to override `hashCode` and might get style warnings for not implementing both.  Also, TagContextImpl wouldn't be to optimize `hashCode`.

songy23:
I would vote for not making hashCode() final, for the downsides you mentioned and to keep consistent with equals().

sebright:
Making TagContext a value class makes the class simpler to use and explain.  Users will know that it behaves just like a map from TagKeys to TagValues.  Making TagContext a value class also means that we can document that all subclasses of TagContext must behave like value classes, which makes it easier for the OpenCensus tagging and stats implementation to handle any other TagContext subclasses that it encounters.  This design is also more consistent with the design of TagContext before we made it abstract.

sebright:
I made `hashCode` final, since I realized that we can change it from final to non-final much more easily than we can change it from non-final to final.  If `hashCode` is final, we don't need to specify the hash function until we choose to let implementations override it.

I also saw that there was no Error Prone warning with this change, because the superclass already overrides both `equals` and `hashCode`.

PR #610, opened 2017-09-09
Add a section on enabling branch protection to RELEASING.md.

sebright:


RELEASING.md
```diff
@@ -69,7 +69,26 @@ token](https://help.github.com/articles/creating-a-personal-access-token-for-the
...
+
+    - Open the branch protection settings for the new branch, by following
+      [Github's instructions](https://help.github.com/articles/configuring-protected-branches/).
+    - Copy the settings from a previous branch, i.e., check
+      - `Protect this branch`
+      - `Require pull request reviews before merging`
+      - `Require status checks to pass before merging`
+      - `Include administrators`
+
+      Uncheck `Require branches to be up to date before merging`.
```
bogdandrutu: I would remove this and the next examples. This is a bit too much detail here and may get outdated very easily. I would prefer to have one separate doc about this things (maybe in the -spec repo) where we have this detailed config.

sebright: I thought that it was unlikely that Github would remove the options that we enable, though it might rename them.  I agree that the boxes that we uncheck could easily become outdated as Github adds features.  Why don't we leave the list of options here until we are ready to move it to the other doc?

sebright: I removed "Uncheck `Require branches to be up to date before merging`."

sebright:
Thanks!

PR #611, opened 2017-09-09
Override toString() in TagContext.

sebright:
TagContext should override Object.toString(), since it is a value class.  This
implementation doesn't print the tags because the tags are currently only
exposed by an "unsafe" method, unsafeGetIterator().

### DIFF
--- a/core/src/main/java/io/opencensus/tags/TagContext.java
+++ b/core/src/main/java/io/opencensus/tags/TagContext.java
@@ -40,6 +40,11 @@ public abstract class TagContext {
   // support fast access to tags.
   public abstract Iterator<Tag> unsafeGetIterator();
 
+  @Override
+  public String toString() {
+    return "TagContext";
+  }
+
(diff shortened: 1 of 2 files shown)


adriancole:
seems an interim change? Since this class controls the unsafe method, what would be the harm in using it in the toString?

sebright:
> seems an interim change?

Yes, I think the implementation depends on how we resolve #584.  If we expose the iterator (not as an unsafe method) then I think toString should include the tags.

> Since this class controls the unsafe method, what would be the harm in using it in the toString?

I was thinking of the advice in Effective Java item 10 that says that you should provide accessors for all fields used in toString, and unsafeGetIterator is only meant to called by implementations of the API.  I think we should just expose the iterator, though.

dinooliva:
Agreed, if toString() prints all of the tags, people will use it as a backdoor to access the tags of a TagContext.

PR #612, opened 2017-09-11
Rename TagContexts to Tagger (fixes #588).

### LINKED ISSUE #588: Consider renaming TagContexts class.
It is a little confusing to have both `TagContext` and `TagContexts`, especially since `TagContexts` isn't a utility class.  I can't think of a perfect name for the class, so I figured I would open an issue for discussion.  Here are some ideas:

`Tagger`
`TagContextFactory`
`TagContextManager`
`TagContexts` (current name)

The purpose of the class is to allow creating and manipulating `TagContext`s for implicit and explicit tag propagation.

See https://github.com/census-instrumentation/opencensus-java/pull/498#pullrequestreview-55777822.

sebright:


examples/src/main/java/io/opencensus/examples/stats/StatsRunner.java
```diff
@@ -22,9 +22,9 @@
 import io.opencensus.stats.Stats;
 import io.opencensus.stats.StatsRecorder;
 import io.opencensus.tags.TagContext;
-import io.opencensus.tags.TagContexts;
 import io.opencensus.tags.TagKey.TagKeyString;
 import io.opencensus.tags.TagValue.TagValueString;
+import io.opencensus.tags.Tagger;
 import io.opencensus.tags.Tags;
 
 /** Simple program that uses Stats contexts. */
```
songy23: Nit: this should be updated to TagContext now.

PR #618, opened 2017-09-12
Upgrade autovalue to 1.5

bogdandrutu:


bogdandrutu:
Cannot upgrade because autovalue 1.5 is not java7 compatible:
https://github.com/google/auto/releases/tag/auto-value-1.5

Issue #616, opened 2017-09-11
opencensus-contrib-agent not java6 compatible.


bogdandrutu:
The release 0.6.0 happened last week but unfortunately the code is not java6 compatible:

io.opencensus.contrib.agent.Resources:52  Undefined reference: void AutoCloseable.close()
io.opencensus.contrib.agent.Resources:52  Undefined reference: void AutoCloseable.close()

It uses try with resources. This may not be a problem in real environments.

PR #619, opened 2017-09-12
Fix opencensus agent to be java6 compatible.

bogdandrutu:
Fixes https://github.com/census-instrumentation/opencensus-java/issues/616


ubschmidt2:
Thanks!

Not yet sure, why how/why this issue surfaced - potentially it indicates an issue with Retrolambda (or maybe again the execution order of gradle tasks).
I haven't seen the issue before and cannot reproduce it (without this change, of change). If you happen to have additional input, PLKM.

PR #621, opened 2017-09-13
Add TagContextParseException to throw from deserialization method.

sebright:
This class will allow us to add fields to the exception if we need to add more information about the error in the future.

____________________________________________________________

The exception isn't used yet, because it depends on updating the deserialization method to not do IO (#213).

/cc @adriancole

### DIFF
--- a/core/src/main/java/io/opencensus/tags/TagContextParseException.java
+++ b/core/src/main/java/io/opencensus/tags/TagContextParseException.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2017, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


core/src/main/java/io/opencensus/tags/TagContextParseException.java
```diff
@@ -0,0 +1,41 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.tags;
+
+/** Exception thrown when a {@link TagContext} cannot be parsed. */
+public final class TagContextParseException extends Exception {
```
bogdandrutu: Do you think a more generic exception for the entire library is better?

ParseException/DeserializationException?

sebright: The reasons I see for sharing the exception type are: 1. to reuse an exception type that is already known to users 2. to allow users to handle a single exception type when one try block can throw exceptions with either cause.  I don't think either of those reasons apply here, though.  I would rather use separate exception types so that it is easier to add fields specific to each cause if we need to later.

sebright:
I updated `TagContextBinarySerializer.fromByteArray` to throw `TagContextParseException`.

Issue #622, opened 2017-09-13
Add listSpanNamesForCollection() to SampledSpanStore


zhangkun83:
This can be used in tests to verify that Span names are registered as expected.

PR #623, opened 2017-09-13
Add a method to list all registered span names.

bogdandrutu:
Fixes https://github.com/census-instrumentation/opencensus-java/issues/622

### DIFF
--- a/api/src/main/java/io/opencensus/trace/export/SampledSpanStore.java
+++ b/api/src/main/java/io/opencensus/trace/export/SampledSpanStore.java
@@ -27,6 +27,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
@@ -101,6 +102,14 @@ public abstract class SampledSpanStore {
    * @param spanNames list of span names for which the library will no longer collect samples.
    */
   public abstract void unregisterSpanNamesForCollection(Collection<String> spanNames);
+
+  /**
+   * Returns the list of unique span names registered to the library. For this set of span names the
+   * library will collect latency based sampled spans and error based sampled spans.
+   *
+   * @return the list of unique span names registered to the library.
+   */
+  public abstract Set<String> listRegisteredSpanNamesForCollection();
 
   /** The summary of all available data. */
   @AutoValue
--- a/api/src/main/java/io/opencensus/trace/export/SampledSpanStore.java
+++ b/api/src/main/java/io/opencensus/trace/export/SampledSpanStore.java
@@ -104,12 +104,12 @@ public abstract class SampledSpanStore {
   public abstract void unregisterSpanNamesForCollection(Collection<String> spanNames);
 
   /**
-   * Returns the list of unique span names registered to the library. For this set of span names the
+   * Returns the set of unique span names registered to the library. For this set of span names the
    * library will collect latency based sampled spans and error based sampled spans.
    *
-   * @return the list of unique span names registered to the library.
+   * @return the set of unique span names registered to the library.
    */
-  public abstract Set<String> listRegisteredSpanNamesForCollection();
+  public abstract Set<String> getRegisteredSpanNamesForCollection();
 
   /** The summary of all available data. */
   @AutoValue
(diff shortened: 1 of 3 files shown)


api/src/main/java/io/opencensus/trace/export/SampledSpanStore.java
```diff
@@ -102,6 +103,14 @@ protected SampledSpanStore() {}
    */
   public abstract void unregisterSpanNamesForCollection(Collection<String> spanNames);
 
+  /**
+   * Returns the list of unique span names registered to the library. For this set of span names the
+   * library will collect latency based sampled spans and error based sampled spans.
+   *
+   * @return the list of unique span names registered to the library.
```
sebright: s/list/set

api/src/main/java/io/opencensus/trace/export/SampledSpanStore.java
```diff
@@ -102,6 +103,14 @@ protected SampledSpanStore() {}
...
+   */
+  public abstract Set<String> listRegisteredSpanNamesForCollection();
```
sebright: I think getRegisteredSpanNamesForCollection would be more consistent with other method names.

impl_core/src/main/java/io/opencensus/implcore/trace/export/SampledSpanStoreImpl.java
```diff
@@ -276,6 +278,13 @@ public void unregisterSpanNamesForCollection(Collection<String> spanNames) {
   }
 
   @Override
+  public Set<String> listRegisteredSpanNamesForCollection() {
+    synchronized (samples) {
+      return new HashSet<String>(samples.keySet());
```
sebright: Would it be better to return an immutable set now, to avoid having client code depend on the mutability?

impl_core/src/test/java/io/opencensus/implcore/trace/export/SampledSpanStoreImplTest.java
```diff
@@ -171,19 +171,20 @@ public void addSpansWithoutRegisteredNamesInAllLatencyBuckets() {
...
   @Test
-  public void registerAndUnregisterSpanNames() {
-    addSpanNameToAllLatencyBuckets(NOT_REGISTERED_SPAN_NAME);
+  public void registerUnregisterAndListSpanNames() {
+    assertThat(
+        sampleStore.listRegisteredSpanNamesForCollection().contains(REGISTERED_SPAN_NAME))
+        .isTrue();
     assertThat(
-            sampleStore.getSummary().getPerSpanNameSummary().containsKey(NOT_REGISTERED_SPAN_NAME))
+        sampleStore.listRegisteredSpanNamesForCollection().contains(NOT_REGISTERED_SPAN_NAME))
```
sebright: I think it would be better to check the exact contents of the set with `containsExactly`, here and below.

api/src/main/java/io/opencensus/trace/export/SampledSpanStore.java
sebright: Should this method have a default implementation, to avoid breaking subclasses?

api/src/main/java/io/opencensus/trace/export/SampledSpanStore.java
bogdandrutu: done.

api/src/main/java/io/opencensus/trace/export/SampledSpanStore.java
bogdandrutu: done.

bogdandrutu: nobody else implements this for the moment and I would not worry that much about other implementation, I would feel very bad if we break compatibility on instrumentations.

@dinooliva any suggestion here?

impl_core/src/main/java/io/opencensus/implcore/trace/export/SampledSpanStoreImpl.java
bogdandrutu: Done.

impl_core/src/test/java/io/opencensus/implcore/trace/export/SampledSpanStoreImplTest.java
bogdandrutu: Done.

PR #625, opened 2017-09-13
Replace anonymous classes in SerializationUtils with nested classes.

sebright:
This commit also creates constants for two of the classes, since the instances
can be reused.


sebright:
Thanks!

PR #627, opened 2017-09-13
Rename TagContextBuilder methods to be more consistent with Map and Collection

sebright:
#### Rename TagContextBuilder.set methods to "put".

"put" is a more descriptive name, since these methods behave like Map.put.  The name is also more consistent with Span.putAttribute.

#### Rename TagContextBuilder.clear to "remove".
    
This method removes a specific key from the builder.  The corresponding methods
in Map and Collection are named "remove".  "clear" was slightly misleading,
because Map and Collection also have "clear" methods, but they remove all
values.


sebright:
Thanks!

PR #631, opened 2017-09-15
Refactor TagContexts to make the current context methods final.

sebright:
In order to make the methods final, this commit adds a non-final protected
method to TagContexts, transformTagContext, which handles any transformation of
the TagContext that must be done by the particular instance of TagContexts.  For
example, if the TagContexts allows tagging to be disabled, then
transformTagContext can return a no-op TagContext.


sebright:
This change might not be needed, depending on what we decide for #632.

dinooliva:
Not sure I understand the rationale behind this change - let's discuss it later.

sebright:
> Not sure I understand the rationale behind this change - let's discuss it later.

I thought this made the API too complex, so I took a different approach in #643.

Issue #624, opened 2017-09-13
Implementors unable to move away from Span.addAttributes().


zhangkun83:
Span.addAttributes() is deprecated, but remains abstract. Implementations of Span, e.g., mock implementations in tests can't stop implementing this method.

A solution for that is to provide a default implementation of the deprecated method that delegates to the new one, while the new one also delegates to the old one:
```java
public void putAttributes(Map<String, AttributeValue> attributes) {
  addAttributes(attributes);
}

@Deprecated
public void addAttributes(Map<String, AttributeValue> attributes) {
  putAttributes(attributes);
}
```

This allows implementors to completely switch away from the old one without breaking.

PR #633, opened 2017-09-15
Implement deprecated method addAttributes in the Span and remove it from anywhere else.

bogdandrutu:
Fixes https://github.com/census-instrumentation/opencensus-java/issues/624

### DIFF
--- a/api/src/main/java/io/opencensus/trace/Span.java
+++ b/api/src/main/java/io/opencensus/trace/Span.java
@@ -99,7 +99,7 @@ public abstract class Span {
    */
   public void putAttributes(Map<String, AttributeValue> attributes) {
     // Not final because we want to start overriding this method from the beginning, this will
-    // allow us to remove the addAttributes faster.
+    // allow us to remove the addAttributes faster. All implementations MUST override this method.
     addAttributes(attributes);
   }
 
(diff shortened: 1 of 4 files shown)


api/src/main/java/io/opencensus/trace/Span.java
```diff
@@ -109,7 +109,9 @@ public void putAttributes(Map<String, AttributeValue> attributes) {
    * @param attributes the attributes that will be added and associated with the {@code Span}.
    */
   @Deprecated
-  public abstract void addAttributes(Map<String, AttributeValue> attributes);
+  public void addAttributes(Map<String, AttributeValue> attributes) {
+    putAttributes(attributes);
```
sebright: Should there be some kind of warning about the circularity, telling users that they must implement one of addAttributes or putAttributes?

bogdandrutu: Done.

sebright: Shouldn't it go in the Javadocs?

bogdandrutu: I think this is not relevant for 99.9% of the users of our library so I would not put it in the javadoc. This is relevant only for other implementation which I don't really expect to exist.

PR #634, opened 2017-09-15
Clean up TagContext equals and hashCode

sebright:
#### Check for null in TagContext equals and hashCode.

This isn't a good place to throw a NullPointerException if the subclass
incorrectly returns a null Iterator.

#### Remove obsolete TagContext.hashCode() tests.

These tests aren't needed, because we don't specify the behavior of TagContext.hashCode().

### DIFF
--- a/core/src/main/java/io/opencensus/tags/TagContext.java
+++ b/core/src/main/java/io/opencensus/tags/TagContext.java
@@ -17,7 +17,9 @@
 package io.opencensus.tags;
 
 import com.google.common.collect.HashMultiset;
+import com.google.common.collect.ImmutableMultiset;
 import com.google.common.collect.Lists;
+import com.google.common.collect.Multiset;
 import io.opencensus.tags.TagValue.TagValueString;
 import java.util.Collections;
 import java.util.Iterator;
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/opencensus/tags/TagContext.java
```diff
@@ -56,16 +58,28 @@ public boolean equals(Object other) {
...
+        iter2 == null
+            ? ImmutableMultiset.<Tag>of()
+            : HashMultiset.create(Lists.newArrayList(iter2));
     return tags1.equals(tags2);
   }
 
   @Override
   public final int hashCode() {
     int hashCode = 0;
-    for (Iterator<Tag> i = unsafeGetIterator(); i.hasNext(); ) {
```
bogdandrutu: Should we use something like this: http://docs.oracle.com/javase/7/docs/api/java/util/Arrays.html?is-external=true#hashCode(java.lang.Object[]) or https://google.github.io/guava/releases/19.0/api/docs/com/google/common/base/Objects.html#hashCode(java.lang.Object...)

sebright: We need to ensure that the hash code is order independent, so I think it is better to explicitly sum the elements' hash codes.

sebright:
Thanks!

PR #636, opened 2017-09-15
Refactor MeasureMap.Builder

songy23:
1. Rename MeasureMap.Builder.set() to put(). Consistent with TagContextBuilder #627 
2. Overwrite existing Measure values, if there are subsequent updates to a Measure that is already put.

### DIFF
--- a/core/src/main/java/io/opencensus/stats/MeasureMap.java
+++ b/core/src/main/java/io/opencensus/stats/MeasureMap.java
@@ -54,7 +54,7 @@ public final class MeasureMap {
   public static class Builder {
     /**
      * Associates the {@link MeasureDouble} with the given value. Subsequent updates to the
-     * same {@link MeasureDouble} are ignored.
+     * same {@link MeasureDouble} will overwrite the previous value.
      *
      * @param measure the {@link MeasureDouble}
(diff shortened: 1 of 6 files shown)


sebright:
I agree with this change, though I think we should also make the builder keep the last value for a Measure so that the behavior is consistent with `Map.put`.

dinooliva:
It may make more sense to stay consistent with the behavior of: https://google.github.io/guava/releases/19.0/api/docs/com/google/common/collect/ImmutableMap.Builder.html#put(K, V)

sebright:
> It may make more sense to stay consistent with the behavior of: https://google.github.io/guava/releases/19.0/api/docs/com/google/common/collect/ImmutableMap.Builder.html#put(K, V)

I think we should avoid throwing exceptions during stats recording, if there is another option.  It might be hard for users to ensure that they aren't recording duplicate measures at compile time, or test all instrumented code.

songy23:
I updated MeasureMap.Builder.build() so that only the latest value for a Measure will be picked. This is equivalent to overwriting existing Measure values in put().

dinooliva:
Agree - using the last value is better than throwing an exception.

PR #638, opened 2017-09-17
Run the integration tests using different Java versions.

ubschmidt2:
This change let's Gradle execute the integration tests using different JRE versions. This makes it easy to test the agent on arbitrary JRE versions by just setting JAVA_HOMES appropriately. On Travis and AppVeyor, we test the Java 8-built agent on Java 6, 7, and 8 to increase our confidence in compatiblity with different Java versions. Also see #616, #619.

### DIFF
--- a/contrib/agent/build.gradle
+++ b/contrib/agent/build.gradle
@@ -126,8 +126,9 @@ jar.finalizedBy shadowJar
 
 // TODO(stschmidt): Proguard-shrink the agent JAR.
 
-// Integration tests. The setup is mostly based on
+// Integration tests. The setup was initially based on
 // https://www.petrikainulainen.net/programming/gradle/getting-started-with-gradle-integration-testing/.
+// The same suite of integration tests are run on different Java versions with the agent enabled.
 
 sourceSets {
   integrationTest {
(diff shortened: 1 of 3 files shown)


contrib/agent/build.gradle
```diff
@@ -155,21 +156,41 @@ checkstyleIntegrationTest.enabled = JavaVersion.current().isJava8Compatible()
...
+                   : "${it}/bin/java" }
+    .findAll { new File(it).canExecute() }
+    .unique()
+
+assert javaExecutables.size > 0 :
+       'No Java executables found for running integration tests'
+
+task integrationTest
+
+javaExecutables.eachWithIndex { javaExecutable, index ->
```
adriancole: Is there a way to print that a specific JRE is in use? For example, right now, there's a matrix build in travis (which I guess amplifies this test scenario as you would test all JREs for both JDK 7 and 8). We might want a flag to only run the nested multi-version integration test with one of the matrixes and if so, print output so we know it is happening

ubschmidt2: > Is there a way to print that a specific JRE is in use? For example, right now, there's a matrix build in travis (which I guess amplifies this test scenario as you would test all JREs for both JDK 7 and 8).

Yes, for debugging I had used something like 'doFirst { println "Testing with ${javaExecutable}..." }' and a similar 'doLast ...', but then removed it from this change. I'll add something like that (using the Gradle provided logger instead of plain printlns) - it might be just good enough to see what's going on.

> We might want a flag to only run the nested multi-version integration test with one of the matrixes and if so, print output so we know it is happening

Yes, this might be useful. I was thinking about using a single "java.homes" property (or environment variable?, TBD) for passing the list of (potential) java homes to Gradle instead of the separate environment variables JAVA6_HOME, JAVA7_HOME, and JAVA8_HOME. In the Travis build script, we could set this "java.homes" property differently for each build. I now like this idea better - let me know what you think before I make any changes in this direction.

adriancole:
I like the java, homes.

;)

ubschmidt2:
Changed it to use a single JAVA_HOMES environment variable. Also added minimal logging.

ubschmidt2:
I've done some minor cleanups. This is complete (ATM). PTAL.

adriancole:
Verified CI logs look good! Very good work (also having windows run all jres). Thanks.. I know this took some time, but infra like this is really helpful.

PR #642, opened 2017-09-17
Add a new SpanContextParseException and use it in BinaryFormat.

bogdandrutu:
Add new methods in BinaryFormat which throw the new exception and deprecate the old methods.

### DIFF
--- a/api/src/main/java/io/opencensus/trace/propagation/BinaryFormat.java
+++ b/api/src/main/java/io/opencensus/trace/propagation/BinaryFormat.java
@@ -32,7 +32,7 @@ import java.text.ParseException;
  *     Tracing.getPropagationComponent().getBinaryFormat();
  * void onSendRequest() {
  *   try (Scope ss = tracer.spanBuilder("Sent.MyRequest").startScopedSpan()) {
- *     byte[] binaryValue = binaryFormat.toBinaryValue(tracer.getCurrentContext().context());
+ *     byte[] binaryValue = binaryFormat.toByteArray(tracer.getCurrentContext().context());
  *     // Send the request including the binaryValue and wait for the response.
  *   }
  * }
@@ -49,9 +49,9 @@ import java.text.ParseException;
  *   SpanContext spanContext = SpanContext.INVALID;
  *   try {
  *     if (binaryValue != null) {
- *       spanContext = binaryFormat.fromBinaryValue(binaryValue);
+ *       spanContext = binaryFormat.fromByteArray(binaryValue);
  *     }
- *   } catch (ParseException e) {
+ *   } catch (SpanContextParseException e) {
  *     // Maybe log the exception.
  *   }
  *   try (Scope ss =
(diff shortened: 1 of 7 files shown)


api/src/main/java/io/opencensus/trace/propagation/BinaryFormat.java
```diff
@@ -65,23 +65,58 @@
...
+   *
+   * @param bytes a binary encoded buffer from which the {@code SpanContext} will be parsed.
+   * @return the parsed {@code SpanContext}.
+   * @throws NullPointerException if the {@code input} is {@code null}.
+   * @throws SpanContextParseException if the version is not supported or the input is invalid
+   */
+  public SpanContext fromByteArray(byte[] bytes) throws SpanContextParseException {
+    try {
+      return fromBinaryValue(bytes);
+    } catch (ParseException e) {
```
adriancole: Since the result type is checked exception should we also wrap runtime exceptions or let them propagate?

bogdandrutu: we do throw only one type of runtime exception which is NullPointerException if the buffer is null.

api/src/main/java/io/opencensus/trace/propagation/BinaryFormat.java
```diff
@@ -65,23 +65,58 @@
...
 
   /**
-   * Parses the {@link SpanContext} from the binary format.
+   * Serializes a {@link SpanContext} using the binary format into a byte array.
    *
+   * @param spanContext the {@code SpanContext} to serialize.
+   * @return the serialized binary value.
+   * @throws NullPointerException if the {@code spanContext} is {@code null}.
+   */
+  public byte[] toByteArray(SpanContext spanContext) {
```
songy23: Why not make this abstract? (and fromByteArray)

api/src/main/java/io/opencensus/trace/propagation/SpanContextParseException.java
```diff
@@ -0,0 +1,41 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.trace.propagation;
+
+/** Exception thrown when a {@link io.opencensus.trace.SpanContext} cannot be parsed. */
+public class SpanContextParseException extends Exception {
```
sebright: Should this class be final?  I made TagContextParseException final because I thought that subclasses would only be useful if we added them to the API, so that users could catch a subtype.  If we decided to add subclasses to the API, we could make TagContextParseException non-final at the same time.

impl_core/src/main/java/io/opencensus/implcore/trace/propagation/BinaryFormatImpl.java
```diff
@@ -77,7 +77,7 @@
       4 * ID_SIZE + TraceId.SIZE + SpanId.SIZE + TraceOptions.SIZE;
 
   @Override
-  public byte[] toBinaryValue(SpanContext spanContext) {
+  public byte[] toByteArray(SpanContext spanContext) {
```
sebright: Shouldn't BinaryFormatImpl continue to implement to/fromBinaryValue instead, for backwards compatibility with the API?

api/src/main/java/io/opencensus/trace/propagation/BinaryFormat.java
```diff
@@ -65,23 +65,58 @@
...
-   *
+   * @deprecated use {@link #toByteArray(SpanContext)}.
    * @param spanContext the {@code SpanContext} to serialize.
    * @return the serialized binary value.
    * @throws NullPointerException if the {@code spanContext} is {@code null}.
    */
-  public abstract byte[] toBinaryValue(SpanContext spanContext);
+  @Deprecated
+  public byte[] toBinaryValue(SpanContext spanContext) {
+    return toByteArray(spanContext);
```
sebright: There should probably be a note here, too, about subclasses needing to override at least one "to" and "from" method.

api/src/main/java/io/opencensus/trace/propagation/BinaryFormat.java
```diff
@@ -65,23 +65,58 @@
...
+  }
 
   /**
-   * Parses the {@link SpanContext} from the binary format.
+   * Serializes a {@link SpanContext} using the binary format into a byte array.
```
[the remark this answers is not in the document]
bogdandrutu: done.

api/src/main/java/io/opencensus/trace/propagation/BinaryFormat.java
bogdandrutu: done.

api/src/main/java/io/opencensus/trace/propagation/BinaryFormat.java
bogdandrutu: Because if in tests let's say somebody uses the previous API they override the toBinaryValue and if I make this abstract I will break their build.

api/src/main/java/io/opencensus/trace/propagation/BinaryFormat.java
```diff
@@ -94,13 +129,13 @@ static BinaryFormat getNoopBinaryFormat() {
...
     @Override
-    public byte[] toBinaryValue(SpanContext spanContext) {
+    public byte[] toByteArray(SpanContext spanContext) {
       checkNotNull(spanContext, "spanContext");
       return new byte[0];
     }
 
     @Override
-    public SpanContext fromBinaryValue(byte[] bytes) throws ParseException {
+    public SpanContext fromByteArray(byte[] bytes) throws SpanContextParseException {
```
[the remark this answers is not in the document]
bogdandrutu: I see.

api/src/main/java/io/opencensus/trace/propagation/SpanContextParseException.java
bogdandrutu: Done.

impl_core/src/main/java/io/opencensus/implcore/trace/propagation/BinaryFormatImpl.java
bogdandrutu: For me backwards compatibility means a code instrumented with API 0.6.0 can run with API 0.7.0 without a problem. 

Based on how dependency management work, (you can read more https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html) the final binary controls what API/implementation is used so we don't need to worry about Impl 0.7.0 working with API 0.6.0 (that's also forward compatibility I feel and we do not guarantee that).

sebright: I thought that the implementation also needed to be compatible with older versions of the API for some period of time, in case users don't exclude the API dependency.  I'll ask you about it later, though.

PR #643, opened 2017-09-18
Move current TagContext logic from the API to the implementation (closes #632).

### LINKED ISSUE #632: Should we move the current context logic from the API to the implementation library?
While implementing #608, I found that it was difficult to work with the split in the tagging logic between the API and implementation libraries.  The API library contains the gRPC context keys and all of the code for interacting with the current context.  However, the implementation needs to support dynamically enabling and disabling tagging, and we may want to optimize the case where tagging is disabled by not setting or getting the current context.  It would be nice if that logic could be contained completely within the implementation so that we could change it without any backwards compatibility issues.

Should we move all current context logic into the implementation, except for the gRPC context keys?  This would mean making methods like `TagContexts.getCurrentTagContext()` and `TagContexts.withTagContext(TagContext)` abstract.  The downside is that implementations would need to use `io.opencensus.tags.unsafe.ContextUtils` to manipulate the context themselves.  They would also be free to use a completely different implementation of context propagation, but that would be inconsistent with the gRPC Context key that is exposed in `ContextUtils`.

I think that the same issue applies to tracing, if there is no need to propagate the span when tracing is disabled.

/cc @bogdandrutu @dinooliva @songy23

sebright:
This commit makes all methods in Tagger and TagContextBuilder abstract.  It also
changes the no-op Tagger and TagContextBuilder so that they do not access the
current context.

This change gives the implementation more flexibility to control when it
accesses the current context.  For example, it could avoid setting the current
context when tagging is not enabled.  It also simplifies the code by not
splitting the current context logic between the API and implementation classes.

### DIFF
--- a/api/src/main/java/io/opencensus/internal/NoopScope.java
+++ b/api/src/main/java/io/opencensus/internal/NoopScope.java
@@ -19,7 +19,7 @@ package io.opencensus.internal;
 import io.opencensus.common.Scope;
 
 /**
- * A {@link Scope} that does nothing when it is closed.
+ * A {@link Scope} that does nothing when it is created or closed.
  */
 public final class NoopScope implements Scope {
   private static final Scope INSTANCE = new NoopScope();
--- a/api/src/main/java/io/opencensus/internal/NoopScope.java
+++ b/api/src/main/java/io/opencensus/internal/NoopScope.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2017, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.internal;
+
+import io.opencensus.common.Scope;
+
+public final class NoopScope implements Scope {
+  private static final Scope INSTANCE = new NoopScope();
+
+  private NoopScope() {}
+
+  /**
+   * Returns a {@code NoopScope}.
+   *
+   * @return a {@code NoopScope}.
+   */
+  public static Scope getInstance() {
+    return INSTANCE;
+  }
+
+  @Override
+  public void close() {}
+}
--- a/api/src/main/java/io/opencensus/internal/NoopScope.java
+++ b/api/src/main/java/io/opencensus/internal/NoopScope.java
@@ -18,6 +18,9 @@ package io.opencensus.internal;
 
 import io.opencensus.common.Scope;
 
+/**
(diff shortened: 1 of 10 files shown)


api/src/main/java/io/opencensus/internal/NoopScope.java
```diff
@@ -0,0 +1,37 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.internal;
+
+import io.opencensus.common.Scope;
+
+public final class NoopScope implements Scope {
```
songy23: Nit: javadoc?

core/src/main/java/io/opencensus/tags/TagContextBuilder.java
```diff
@@ -121,5 +120,10 @@ public TagContextBuilder remove(TagKey key) {
     public TagContext build() {
       return TagContext.getNoopTagContext();
     }
+
+    @Override
+    public Scope buildScoped() {
```
songy23: Tests for Noop?

core/src/main/java/io/opencensus/tags/Tagger.java
```diff
@@ -104,8 +97,28 @@ public TagContext empty() {
...
       return TagContextBuilder.getNoopTagContextBuilder();
     }
+
+    @Override
+    public TagContextBuilder currentBuilder() {
+      return TagContextBuilder.getNoopTagContextBuilder();
+    }
+
+    @Override
+    public Scope withTagContext(TagContext tags) {
```
songy23: Tests?

api/src/main/java/io/opencensus/internal/NoopScope.java
sebright: Done

core/src/main/java/io/opencensus/tags/Tagger.java
sebright: Done

core/src/main/java/io/opencensus/tags/TagContextBuilder.java
sebright: Done

api/src/main/java/io/opencensus/internal/NoopScope.java
```diff
@@ -0,0 +1,40 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.internal;
+
+import io.opencensus.common.Scope;
+
+/**
+ * A {@link Scope} that does nothing when it is closed.
```
bogdandrutu: s/when it is closed/neither when constructed or closed?

api/src/main/java/io/opencensus/internal/NoopScope.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+public final class NoopScope implements Scope {
+  private static final Scope INSTANCE = new NoopScope();
```
bogdandrutu: Let's decide on this and keep it consistent. Should we make INSTANCE public or always have function calls? Advantage when directly access variable is clear that is not creating a new instance, disadvantage is that is harder to implement any side effect (or deprecate).

sebright:
@bogdandrutu I opened #653.

sebright:
Thanks!

PR #647, opened 2017-09-19
Add SumDataDouble and SumDataLong

songy23:
API change to support long typed AggregationData.

### DIFF
--- a/core_impl/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
+++ b/core_impl/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
@@ -44,8 +44,6 @@ import io.opencensus.stats.AggregationData.MeanData;
 import io.opencensus.stats.AggregationData.SumDataDouble;
 import io.opencensus.stats.AggregationData.SumDataLong;
 import io.opencensus.stats.Measure;
-import io.opencensus.stats.Measure.MeasureDouble;
-import io.opencensus.stats.Measure.MeasureLong;
 import io.opencensus.stats.View;
 import io.opencensus.stats.View.AggregationWindow.Cumulative;
 import io.opencensus.stats.View.AggregationWindow.Interval;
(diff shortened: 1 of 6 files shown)


songy23:
Rebased, ready for review now.

core_impl/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
```diff
@@ -491,12 +496,50 @@ public MutableAggregation apply(Distribution arg) {
...
+      this.measure = measure;
+    }
+
     @Override
-    public AggregationData apply(MutableSum arg) {
-      return SumData.create(arg.getSum());
+    public AggregationData apply(final MutableSum arg) {
+      return measure.match(
+          new CreateSumDataDouble(arg),
+          new CreateSumDataLong(arg),
```
sebright: Suggestion: These function objects ignore their input, so I think you could write this as:

```java
      return measure.match(
          Functions.<AggregationData>returnConstant(SumDataDouble.create(arg.getSum())),
          Functions.<AggregationData>returnConstant(SumDataLong.create(Math.round(arg.getSum()))),
          Functions.<AggregationData>throwAssertionError());
```

core_impl/src/test/java/io/opencensus/implcore/stats/ViewManagerImplTest.java
```diff
@@ -199,17 +201,27 @@ public void disallowGettingNonexistentViewData() {
...
-    testRecordCumulative(MEASURE_LONG, 1000, 2000, 3000, 4000);
+  public void testRecordLong_distribution_cumulative() {
+    testRecordCumulative(MEASURE_LONG, DISTRIBUTION, 1000, 2000, 3000, 4000);
   }
 
-  private void testRecordCumulative(Measure measure, double... values) {
-    View view = createCumulativeView(VIEW_NAME, measure, DISTRIBUTION, Arrays.asList(KEY));
+  @Test
+  public void testRecordDouble_sum_cumulative() {
+    testRecordCumulative(MEASURE_DOUBLE, SUM, 10.0, 20.0, 30.0, 40.0);
```
sebright: I think it would help to include some non-integers in this test, to check that they aren't rounded.

core_impl/src/test/java/io/opencensus/implcore/stats/ViewManagerImplTest.java
```diff
@@ -288,7 +301,7 @@ private void testRecordInterval(
         viewManager.getView(VIEW_NAME).getAggregationMap(),
         ImmutableMap.of(
             Arrays.asList(VALUE),
-            StatsTestUtil.createAggregationData(MEAN, initialValues)),
+            StatsTestUtil.createAggregationData(MEAN, MEASURE_LONG, initialValues)),
```
sebright: Should this be `measure` instead of `MEASURE_LONG`, since this method is called with both subclasses?

core_impl/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
songy23: Nice suggestion, updated

core_impl/src/test/java/io/opencensus/implcore/stats/ViewManagerImplTest.java
songy23: Done.

core_impl/src/test/java/io/opencensus/implcore/stats/ViewManagerImplTest.java
songy23: Good catch, fixed. Also added tests for recording sum stats for interval view.

songy23:
Merging now, might want to revisit this API change later.

Issue #641, opened 2017-09-17
Move tag propagation related classes into a propagation package


bogdandrutu:
Consider moving all tag propagation related classes into a propagation package (similar with trace):
TagContextParseException.java
TagPropagationComponent.java
TagContextBinarySerializer.java

There will be more classes like TagContextHttpSerializer.

PR #650, opened 2017-09-20
Move tagging no-op classes into a new class, NoopTags.

sebright:
This change will allow us to move the public serialization-related classes into
a new package without moving their no-op implementations.  Then the no-op classes
can remain package-private.

_____________________________________________________________________

This is needed for #641.

### DIFF
--- a/core/src/main/java/io/opencensus/tags/NoopTags.java
+++ b/core/src/main/java/io/opencensus/tags/NoopTags.java
@@ -1,0 +1,179 @@
+/*
+ * Copyright 2017, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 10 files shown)


core/src/main/java/io/opencensus/tags/NoopTags.java
```diff
@@ -0,0 +1,179 @@
...
+    @Override
+    public TagContextBinarySerializer getBinarySerializer() {
+      return getNoopTagContextBinarySerializer();
+    }
+  }
+
+  @Immutable
+  private static final class NoopTagContextBinarySerializer extends TagContextBinarySerializer {
+    static final TagContextBinarySerializer INSTANCE = new NoopTagContextBinarySerializer();
+    static final byte[] EMPTY_BYTE_ARRAY = {};
```
bogdandrutu: private? Everything can be private because the class is private.

sebright: I don't think private has an effect in a private nested class.

sebright:
Thanks!

PR #654, opened 2017-09-20
Clean up tests for tagging implementation.

sebright:

### DIFF
--- a/core_impl/src/test/java/io/opencensus/implcore/tags/TagsTestUtil.java
+++ b/core_impl/src/test/java/io/opencensus/implcore/tags/TagsTestUtil.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2017, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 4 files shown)


core_impl/src/test/java/io/opencensus/implcore/tags/TagsTestUtil.java
```diff
@@ -0,0 +1,31 @@
...
+
+package io.opencensus.implcore.tags;
+
+import com.google.common.collect.Lists;
+import io.opencensus.tags.Tag;
+import io.opencensus.tags.TagContext;
+import java.util.Collection;
+
+/** Test utilities for tagging. */
+public class TagsTestUtil {
```
songy23: Does it need to a public class?

core_impl/src/test/java/io/opencensus/implcore/tags/TagsTestUtil.java
```diff
@@ -0,0 +1,31 @@
...
+
+  /** Returns a collection of all tags in a {@link TagContext}. */
+  public static Collection<Tag> tagContextToList(TagContext tags) {
```
songy23: Same

core_impl/src/test/java/io/opencensus/implcore/tags/TagsTestUtil.java
sebright: It will need to be public once I split the serialization classes into another package in #652.

PR #658, opened 2017-09-21
Make TagContext iterator method protected, and expose it through internal class.

sebright:
This commit makes TagContext.unsafeGetIterator() protected and renames it to
iterator().  The protected modifier means that subclasses can specify the tags
that they contain, but users cannot access the tags of subclasses that they
did not create.  Since the stats and tagging implementations still need to
access the TagContext iterator, I added a class, io.opencensus.tags.Internal,
that contains a static method for accessing the iterator.

This change helps to separate the two uses of an iterator in TagContext:
giving the implementation access to tags in a TagContext subclass, which should
be exposed, and accessing any TagContext's tags, which should be treated as
internal.

### DIFF
--- a/core/src/main/java/io/opencensus/tags/Internal.java
+++ b/core/src/main/java/io/opencensus/tags/Internal.java
@@ -26,6 +26,6 @@ public final class Internal {
   private Internal() {}
 
   public static Iterator<Tag> getTags(TagContext tags) {
-    return tags.iterator();
+    return tags.getIterator();
   }
 }
--- a/core/src/main/java/io/opencensus/tags/Internal.java
+++ b/core/src/main/java/io/opencensus/tags/Internal.java
@@ -14,19 +14,18 @@
  * limitations under the License.
  */
 
-package io.opencensus.implcore.tags;
+package io.opencensus.tags;
 
-import com.google.common.collect.Lists;
-import io.opencensus.tags.Internal;
-import io.opencensus.tags.Tag;
-import io.opencensus.tags.TagContext;
-import java.util.Collection;
+import java.util.Iterator;
 
-/** Test utilities for tagging. */
-public class TagsTestUtil {
+/**
+ * Internal tagging utilities.
+ */
+@io.opencensus.common.Internal
+public final class Internal {
+  private Internal() {}
 
-  /** Returns a collection of all tags in a {@link TagContext}. */
-  public static Collection<Tag> tagContextToList(TagContext tags) {
-    return Lists.newArrayList(Internal.getTags(tags));
+  public static Iterator<Tag> getTags(TagContext tags) {
+    return tags.iterator();
   }
 }
--- a/core/src/main/java/io/opencensus/tags/NoopTags.java
+++ b/core/src/main/java/io/opencensus/tags/NoopTags.java
@@ -186,7 +186,7 @@ final class NoopTags {
 
     // TODO(sebright): Is there any way to let the user know that their tags were ignored?
     @Override
-    public Iterator<Tag> unsafeGetIterator() {
+    public Iterator<Tag> iterator() {
       return Collections.<Tag>emptySet().iterator();
     }
   }
--- a/core/src/main/java/io/opencensus/tags/NoopTags.java
+++ b/core/src/main/java/io/opencensus/tags/NoopTags.java
@@ -186,7 +186,7 @@ final class NoopTags {
 
     // TODO(sebright): Is there any way to let the user know that their tags were ignored?
     @Override
-    public Iterator<Tag> iterator() {
+    protected Iterator<Tag> iterator() {
       return Collections.<Tag>emptySet().iterator();
     }
   }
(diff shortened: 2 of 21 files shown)


core/src/main/java/io/opencensus/tags/NoopTags.java
```diff
@@ -175,7 +175,7 @@ public Scope buildScoped() {
 
     // TODO(sebright): Is there any way to let the user know that their tags were ignored?
     @Override
-    public Iterator<Tag> unsafeGetIterator() {
+    public Iterator<Tag> iterator() {
       return Collections.<Tag>emptySet().iterator();
```
adriancole: Nit protected

core/src/main/java/io/opencensus/tags/TagContext.java
```diff
@@ -37,9 +37,12 @@
...
 
-  // TODO(sebright): Consider removing TagContext.unsafeGetIterator() so that we don't need to
-  // support fast access to tags.
-  public abstract Iterator<Tag> unsafeGetIterator();
+  /**
+   * Returns an iterator over the tags in this {@code TagContext}.
+   *
+   * @return an iterator over the tags in this {@code TagContext}.
+   */
+  protected abstract Iterator<Tag> iterator();
```
adriancole: Copy rationale note about protected here to reduce chance of misunderstanding later.

core/src/test/java/io/opencensus/tags/TagContextTest.java
```diff
@@ -47,7 +47,7 @@ public void equals_IgnoresTagOrderAndTagContextClass() {
             new SimpleTagContext(TAG2, TAG1),
             new TagContext() {
               @Override
-              public Iterator<Tag> unsafeGetIterator() {
+              public Iterator<Tag> iterator() {
```
bogdandrutu: Protected?

core_impl/src/test/java/io/opencensus/implcore/tags/CurrentTagContextUtilsTest.java
```diff
@@ -44,7 +43,7 @@
       new TagContext() {
 
         @Override
-        public Iterator<Tag> unsafeGetIterator() {
+        public Iterator<Tag> iterator() {
```
bogdandrutu: protected?

core/src/main/java/io/opencensus/tags/Internal.java
```diff
@@ -0,0 +1,31 @@
...
+
+package io.opencensus.tags;
+
+import java.util.Iterator;
+
+/**
+ * Internal tagging utilities.
+ */
+@io.opencensus.common.Internal
+public final class Internal {
```
bogdandrutu: Where do you use this class?

sebright: Internal.getTags() is used by the stats implementation in MutableViewData, and by the tagging implementation in TaggerImpl and SerializationUtils.

core/src/main/java/io/opencensus/tags/TagContext.java
sebright: Done

core/src/main/java/io/opencensus/tags/NoopTags.java
sebright: Done

core/src/test/java/io/opencensus/tags/TagContextTest.java
sebright: Done

core_impl/src/test/java/io/opencensus/implcore/tags/CurrentTagContextUtilsTest.java
sebright: Done

adriancole:
>
> I realized that if we want to expose iterator() later, we can't just make
> it public, because that would break subclasses that make it protected. I
> renamed it to getIterator(), so that we could deprecate the protected
> method and add a public method.
>
good thinking

core/src/main/java/io/opencensus/tags/Internal.java
bogdandrutu: If it is used by the stats/tags implementation only why do we need it in the public API? Stats and Tags can get this from TagContextImpl.

sebright: The implementation only accesses the internal state when the TagContext is a TagContextImpl.  It uses Internal.getTags() for any other subclass.  That way the TagContext's class only affects performance, not behavior.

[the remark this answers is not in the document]
sebright: Yes, InternalUtils is better, and it is consistent with other utility classes.

sebright: I don't think that the current design optimizes for that use case, it just ensures that the code behaves correctly on any input.  I think that having public methods that take arguments of type TagContext but only work on one subclass is confusing and not very user-friendly, even if it is unlikely that they will need to be called with different types.  This design choice also relates to the general argument against using `instanceof` in Java.

sebright:
@bogdandrutu Do you mind if I merge this PR?

Issue #608, opened 2017-09-08
Add support for multiple library states


bogdandrutu:
I think it will be useful to have a enabling level for each of the libraries:

Trace:
 * No-op (should be possible even if the implementation is linked)
 * Only trace ids propagation
 * Normal/Full

Tags:
 * No-op (should be possible even if the implementation is linked)
 * Enable

Stats:
 * No-op (should be possible even if the implementation is linked)
 * Enable

User should be able to configure for example: Trace propagation ids only + Tags enable without any recording.

sebright:
Some notes from discussing this with @bogdandrutu:

- The states will be set programmatically, probably with methods on the Tags, Stats, and Tracing classes.  There should also be methods for getting the state.
- The state for any component can be changed at any time.
- The default for each package is enabled.
- Disabling tagging also disables stats collection.
- Views can be registered at any time, even when stats is disabled.  Disabling stats doesn't clear the registered views, though it does clear any stats that were collected for those views.
- When a view is queried and stats collection is disabled, the result is a ViewData with no stats.
- It may be possible for stats to be recorded with no tags with this sequence of steps, though it is probably okay:
  - Tagging is disabled.
  - The user creates a TagContext, after adding some tags to the builder.
  - Stats and tagging are re-enabled.
  - The user records stats with the TagContext.

sebright:
@bogdandrutu What type of API do you think we should provide?  We could have something like a `setState` method on each component, which is also exposed in the corresponding class containing static methods to access the singleton.  The `setState` methods could each take a different enum type, since the different components have different sets of states.

bogdandrutu:
I think from user perspective they care only about the methods on the static classes (Tracing, Stats, Tags). I don't think we should have an setState on each different component like TagPropagationComponent because we complicate behaviors. Maybe we can have a setTags on the main components TagsComponent, TraceComponent, StatsComponent.

bogdandrutu:
Also I would not initially implement the bullet point:
```Disabling tagging also disables stats collection.```

sebright:
> I think from user perspective they care only about the methods on the static classes (Tracing, Stats, Tags). I don't think we should have an setState on each different component like TagPropagationComponent because we complicate behaviors. Maybe we can have a setTags on the main components TagsComponent, TraceComponent, StatsComponent.

Do you mean `setState`?  I agree.  I only meant the three top level components.

> Also I would not initially implement the bullet point:
Disabling tagging also disables stats collection.

[...]

Issue #585, opened 2017-09-05
Provide a way to determine whether tags, stats, or trace classes are no-op implementations.


sebright:
The no-op implementations can often be used to avoid unnecessary if statements, but it is important to be able to check whether they are no-op when necessary.

See https://github.com/census-instrumentation/opencensus-java/issues/215#issuecomment-296304442

bogdandrutu:
This probably should apply to all.

sebright:
> This probably should apply to all.

I'm not sure I understand.  Do you mean all of stats, tags, and tracing, or all classes?

sebright:
#659 adds a `getState` method to `Tags` and `TagsComponent`.

sebright:
Even if we have getState and setState methods, users may want to be able to check whether `setState(ENABLED)` will be ignored.  See the discussions in #659 and #673.

sebright:
@bogdandrutu Do you have any examples of use cases where users would want to distinguish between the no-op implementation and functionality turned off programmatically?

sebright:
One use case for checking whether the implementation is available is performance optimizations that rely on stats, tagging, and tracing never being turned on at runtime.  For example, @zhangkun83 mentioned that gRPC could skip initializing the [CensusStatsModule](https://github.com/grpc/grpc-java/blob/61d5baa5f219da280ccfa44afca77d2457cd8777/core/src/main/java/io/grpc/internal/CensusStatsModule.java) if there was no implementation.

sebright:
Another use case is allowing libraries that depend on OpenCensus functionality to report a more specific error when an implementation isn't available.

PR #659, opened 2017-09-21
Add API for enabling/disabling tagging (part of #608).

sebright:
This commit adds an enum to represent the tagging state (TaggingState) and adds
getters and setters for the state to TagsComponent and Tags.  The state has no
effect yet.

/cc @adriancole

### DIFF
--- a/core/src/main/java/io/opencensus/tags/TaggingState.java
+++ b/core/src/main/java/io/opencensus/tags/TaggingState.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2017, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.tags;
+
+/** State of the {@link TagsComponent}. */
+public enum TaggingState {
+  // TODO(sebright): Should we add a state that propagates the tags, but doesn't allow
+  // modifications?
+
+  /**
+   * State that fully enables tagging.
+   *
+   * <p>The {@link TagsComponent} can add tags to {@link TagContext}s, propagate {@code TagContext}s
+   * in the current context, and serialize {@code TagContext}s.
+   */
+  ENABLED,
+
+  /**
+   * State that disables tagging.
+   *
+   * <p>The {@link TagsComponent} may not add tags to {@link TagContext}s, propagate {@code
+   * TagContext}s in the current context, or serialize {@code TagContext}s.
+   */
+  DISABLED
+}
--- a/core/src/main/java/io/opencensus/tags/TaggingState.java
+++ b/core/src/main/java/io/opencensus/tags/TaggingState.java
@@ -35,5 +35,6 @@ public enum TaggingState {
    * <p>The {@link TagsComponent} may not add tags to {@link TagContext}s, propagate {@code
    * TagContext}s in the current context, or serialize {@code TagContext}s.
    */
+  // TODO(sebright): Document how this interacts with stats collection.
   DISABLED
 }
--- a/core/src/main/java/io/opencensus/tags/Tags.java
+++ b/core/src/main/java/io/opencensus/tags/Tags.java
@@ -50,21 +50,10 @@ public final class Tags {
   }
 
   /**
-   * Returns {@code true} if a tagging implementation is available.
-   *
-   * <p>When an implementation is not available, calling {@link #setState} has no effect.
-   *
-   * @return {@code true} if a tagging implementation is available.
-   */
-  public static boolean isImplementationAvailable() {
-    return tagsComponent.isImplementationAvailable();
-  }
-
-  /**
    * Returns the current {@code TaggingState}.
    *
-   * <p>When {@link #isImplementationAvailable} returns {@code false}, {@code getState} always
-   * returns {@link TaggingState#DISABLED}.
+   * <p>When no implementation is available, {@code getState} always returns {@link
+   * TaggingState#DISABLED}.
    *
    * @return the current {@code TaggingState}.
    */
@@ -75,8 +64,7 @@ public final class Tags {
   /**
    * Sets the current {@code TaggingState}.
    *
-   * <p>When {@link #isImplementationAvailable} returns {@code false}, {@code setState} has no
-   * effect.
+   * <p>When no implementation is available, {@code setState} has no effect.
    *
    * @param state the new {@code TaggingState}.
    */
--- a/core/src/main/java/io/opencensus/tags/Tags.java
+++ b/core/src/main/java/io/opencensus/tags/Tags.java
@@ -56,8 +56,8 @@ public final class Tags {
    *
    * @return {@code true} if a tagging implementation is available.
    */
-  public static boolean isTaggingAvailable() {
-    return tagsComponent.isTaggingAvailable();
+  public static boolean isImplementationAvailable() {
+    return tagsComponent.isImplementationAvailable();
   }
 
   /**
--- a/core/src/main/java/io/opencensus/tags/Tags.java
+++ b/core/src/main/java/io/opencensus/tags/Tags.java
@@ -49,6 +49,24 @@ public final class Tags {
     return tagsComponent.getTagPropagationComponent();
   }
 
+  /**
+   * Returns the current {@code TaggingState}.
+   *
+   * @return the current {@code TaggingState}.
+   */
+  public static TaggingState getState() {
+    return tagsComponent.getState();
+  }
+
+  /**
+   * Sets the current {@code TaggingState}.
+   *
+   * @param state the new {@code TaggingState}.
+   */
+  public static void setState(TaggingState state) {
+    tagsComponent.setState(state);
+  }
+
   // Any provider that may be used for TagsComponent can be added here.
   @VisibleForTesting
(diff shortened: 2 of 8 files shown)


core/src/main/java/io/opencensus/tags/TaggingState.java
```diff
@@ -0,0 +1,39 @@
...
+   */
+  ENABLED,
+
+  /**
+   * State that disables tagging.
+   *
+   * <p>The {@link TagsComponent} may not add tags to {@link TagContext}s, propagate {@code
+   * TagContext}s in the current context, or serialize {@code TagContext}s.
+   */
+  DISABLED
```
songy23: nit: Will this state disable stats collection? If we plan to do so maybe worth mentioning in the comment.

sebright: This is still under discussion in #608 (https://github.com/census-instrumentation/opencensus-java/issues/608#issuecomment-329828347), so I added a TODO.

core/src/main/java/io/opencensus/tags/TaggingState.java
```diff
@@ -0,0 +1,40 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.tags;
+
+/** State of the {@link TagsComponent}. */
+public enum TaggingState {
```
bogdandrutu: I think it is clear if you add a third state NO_OP and describe that as a state which cannot be modified and cannot get into that state.

What do you think?

core_impl/src/main/java/io/opencensus/implcore/tags/TagsComponentImplBase.java
```diff
@@ -35,4 +41,15 @@ public Tagger getTagger() {
...
   }
+
+  @Override
+  public TaggingState getState() {
+    return state.get();
+  }
+
+  @Override
+  public void setState(TaggingState newState) {
+    state.set(checkNotNull(newState));
```
bogdandrutu: it is better to use checkNotNull(newState, "newState") because the message will tell you what variable was null. gRPC people told me to do that because will help with debugging.

core/src/main/java/io/opencensus/tags/TaggingState.java
sebright: You're right, users would probably want to know whether calling `setState(TaggingState.ENABLED)` will have an effect.

sebright: You're right, users would probably want to know whether calling `setState(TaggingState.ENABLED)` would have an effect.  I'm not sure how we should represent it though.  The third state would be a valid output for getState, but not a valid input for setState.

core_impl/src/main/java/io/opencensus/implcore/tags/TagsComponentImplBase.java
sebright: Done

core/src/main/java/io/opencensus/tags/TaggingState.java
adriancole: I think we should just ignore on noop maybe linking to where noop is controlled. noop state shouldn't cause damage even to callers of this.

sebright: > I think we should just ignore on noop maybe linking to where noop is controlled. noop state shouldn't cause damage even to callers of this.

That makes sense, especially since we will also ignore setting ENABLED/DISABLED when the state is NO_OP.

We could also have a separate method for checking whether the tagging implementation is no-op.  Then TaggingState could just represent the different levels of tagging functionality.

sebright: I added a method `boolean isTaggingAvailable()` to Tags and TagsComponent.  I tried both options, but I thought that it was easier to document the new method than a new enum constant.  What do you think of this design?

bogdandrutu: I personally think that it is wrong to have conceptually one state split into actually 2 APIs "is available" and "set/get state".

I think it is easier to have 3 states as I explained in a previous comment and explain possible transitions that can happen (as a state machine).

Why do you think it was hard to explain? I would explain this way:
* NO_IMPLEMENTATION. This is a permanent state that can be change only by recompiling the library and include the implementation.
* ENABLE. This is the default state of the library and transition from ENABLE to DISABLE can be made via setState. (also include what enable means, propagate tags, etc.)
* DISABLE. same as ENABLE

What do you think? Also we may make SetState throw a checked exception that force people to catch it and take the correct action (it will not be used that often, I don't expect to have more than 1 call to the SetState per binary/library).

core/src/main/java/io/opencensus/tags/Tags.java
```diff
@@ -49,6 +49,35 @@ public static TagPropagationComponent getTagPropagationComponent() {
...
   }
 
+  /**
+   * Returns {@code true} if a tagging implementation is available.
+   *
+   * <p>When tagging is not available, calling {@link #setState} has no effect.
+   *
+   * @return {@code true} if a tagging implementation is available.
+   */
+  public static boolean isTaggingAvailable() {
```
bogdandrutu: Maybe isImplementationAvailable? Refer also to the readme how to do that with a @plainlink.

core/src/main/java/io/opencensus/tags/Tags.java
```diff
@@ -49,6 +49,35 @@ public static TagPropagationComponent getTagPropagationComponent() {
...
+  public static TaggingState getState() {
+    return tagsComponent.getState();
+  }
+
+  /**
+   * Sets the current {@code TaggingState}.
+   *
+   * @param state the new {@code TaggingState}.
+   */
+  public static void setState(TaggingState state) {
```
bogdandrutu: Please comment that this is no-op if {@link #isImplementationAvailable()} returns false.

core/src/main/java/io/opencensus/tags/Tags.java
```diff
@@ -49,6 +49,35 @@ public static TagPropagationComponent getTagPropagationComponent() {
...
+  public static boolean isTaggingAvailable() {
+    return tagsComponent.isTaggingAvailable();
+  }
+
+  /**
+   * Returns the current {@code TaggingState}.
+   *
+   * @return the current {@code TaggingState}.
+   */
+  public static TaggingState getState() {
```
bogdandrutu: Document that always returns DISABLE if {@link #isImplementationAvailable()} returns false.

core/src/main/java/io/opencensus/tags/TaggingState.java
bogdandrutu: Ok. Let's do this way. Can you add the same changes to tracing library later?

core/src/main/java/io/opencensus/tags/Tags.java
sebright: Done

core/src/main/java/io/opencensus/tags/Tags.java
sebright: Done

core/src/main/java/io/opencensus/tags/TaggingState.java
sebright: Sure, I opened an issue for the change: #673

adriancole:
Main thing is "when in doub leave it out" is a pretty good policy. Better to have users scream for something than have the code scream ;) In brave we kept it simple.. isNoop means dont bother.. doesnt matter why. We can do other things but I havent heard a request like this. I can check in spring if anyone else has encountered the need to split states.. if theres a workaround (like type check) we can at least defer the no impl state

All this said.. if strong feelings I can be on board with it.. so make a call!

bogdandrutu:
@adriancole not sure I understand what exactly do you propose here. Can you clarify?

Do you support disable at runtime? What is the default state?

adriancole:
Here's the background.

Request https://github.com/openzipkin/brave/issues/435
Impl https://github.com/openzipkin/brave/pull/445

Luckily we only had to make a simple switch to satisfy the customer in this
case, and we had no further requests. That doesnt mean more sophisticated
requests wont come in the future.

Still not sure if the feature matches what we are discussing but yeah hope
it helps.

adriancole:
On what is default state.. it is assumed tracing is enabled by default.
Because brave doesnt encourage static factory like Tracer.getDefault if you
have a tracer it is enabled (though if sampling is 0%  results are always
noop spans). We do have a noop span which is the result of a sampled false
decision. Brave 4 does up front sampling and doesnt retain data if
unsampled (noop span is same as unsampled span). Brave 5 probably will give
option to always record. Forgive if too much context.. not sure what you
want to know.

sebright:
@adriancole Thanks for the background.  That looks very similar to #608, including the comment about whether to disable propagation.  I didn't consider the need to stop stats/tracing functionality after startup, but I think this design will also support that use case.

sebright:
It sounds like we agree on at least providing get/setState methods, but not isImplementationAvailable().  How about I just remove the commits that added isImplementationAvailable(), and then we can decide whether we need to treat no-op differently in a separate issue.  @adriancole were you suggesting that we only use boolean, or do you think that the enum is okay?  I wanted to use an enum, because tracing will have three states (#608), and it will be easier to add states to tagging in the future.

bogdandrutu:
Sounds also good. We may not have 3 states for tracing at the beginning but I still think is better to have an extensible API.

You can remove that and mention in the comment that if no implementation linked the library will always return DISABLE and setState will be a no-op.

sebright:
I removed `isImplementationAvailable` and updated the Javadocs for `getState` and `setState`.  I also realized that we already have an issue for checking whether classes are no-op, #585, so we can continue the discussion there.

sebright:
@bogdandrutu @adriancole Do you mind if I merge this PR?  It only handles the enabled/disabled states, and we can discuss handling no-op in #585.

bogdandrutu:
@sebright go ahead.

sebright:
Thanks!

adriancole:
+1

PR #667, opened 2017-09-24
Adds TextFormat and usage example

adriancole:
This is a WIP to show how a text format injector adapted from Brave
might be used in census. This is a simplified variant, which can only
use String keys (as opposed to gRPC metadata keys, which are not
strings).

Ex.
```java
HttpURLConnection connection = (HttpURLConnection) new URL("http://myserver").openConnection();

// if using a method reference, you just do it directly
textFormat.putContext(spanContext, connection, URLConnection::setRequestProperty);

// if you can't use a method reference or lambda, you should save off
// a setter instead of allocating one for each request like below does
textFormat.putContext(spanContext, connection, new Setter<HttpURLConnection>() {
  @Override
  public void put(HttpURLConnection carrier, String field, String value) {
    carrier.setRequestProperty(field, value);
  }
});
```

### DIFF
--- a/api/src/main/java/io/opencensus/trace/propagation/PropagationComponent.java
+++ b/api/src/main/java/io/opencensus/trace/propagation/PropagationComponent.java
@@ -33,6 +33,14 @@ public abstract class PropagationComponent {
   public abstract BinaryFormat getBinaryFormat();
 
   /**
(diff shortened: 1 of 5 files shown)


adriancole:
closed in favor of #724

PR #668, opened 2017-09-25
Update ViewManager to support a push model

songy23:
Adding an exporter for `ViewData`.

EDIT: there's also a registerView() method for stats exporting handlers to support callback to handlers in ViewManager.regsiterView(), which represents a push model for getting ViewData.

EDIT 2: removed StatsExporter APIs and moved it to the implementation. Start with a minimal set of APIs to support push model.

### DIFF
--- a/core/src/main/java/io/opencensus/stats/NoopStats.java
+++ b/core/src/main/java/io/opencensus/stats/NoopStats.java
@@ -126,7 +126,7 @@ final class NoopStats {
     @Override
     public void setExportInterval(Duration duration) {
       checkNotNull(duration, "duration");
-      checkArgument(duration.compareTo(ZERO_DURATION) > 0, "Duration cannot be negative.");
+      checkArgument(duration.compareTo(ZERO_DURATION) > 0, "Duration must be positive.");
     }
 
     @Override
--- a/core/src/main/java/io/opencensus/stats/NoopStats.java
+++ b/core/src/main/java/io/opencensus/stats/NoopStats.java
@@ -20,6 +20,7 @@ import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.common.collect.Maps;
+import io.opencensus.common.Duration;
 import io.opencensus.common.Functions;
 import io.opencensus.common.Timestamp;
 import io.opencensus.stats.ViewData.AggregationWindowData;
@@ -97,6 +98,7 @@ final class NoopStats {
   @ThreadSafe
   private static final class NoopViewManager extends ViewManager {
     private static final Timestamp ZERO_TIMESTAMP = Timestamp.create(0, 0);
+    private static final Duration ZERO_DURATION = Duration.create(0, 0);
 
     @GuardedBy("views")
     private final Map<View.Name, View> views = Maps.newHashMap();
@@ -113,6 +115,18 @@ final class NoopStats {
           views.put(newView.getName(), newView);
         }
       }
+    }
+
+    @Override
+    public void registerView(View newView, List<? extends Handler> handlers) {
+      checkNotNull(handlers, "handlers");
+      registerView(newView);
+    }
+
+    @Override
+    public void setExportInterval(Duration duration) {
+      checkNotNull(duration, "duration");
+      checkArgument(duration.compareTo(ZERO_DURATION) > 0, "Duration cannot be negative.");
     }
 
     @Override
--- a/core/src/main/java/io/opencensus/stats/ViewManager.java
+++ b/core/src/main/java/io/opencensus/stats/ViewManager.java
@@ -49,7 +49,7 @@ public abstract class ViewManager {
 
   /**
    * Returns the current stats data, {@link ViewData}, associated with the given
-   * {@link View}.
+   * {@link View.Name}.
    */
   public abstract ViewData getView(View.Name view);
 
@@ -73,7 +73,7 @@ public abstract class ViewManager {
      * <p>Implementation SHOULD not block the calling thread. It should execute the export on a
      * different thread if possible.
      *
-     * @param viewDataList a list of {@code ViewData} objects to be exported.
+     * @param viewDataList a collection of {@code ViewData} objects to be exported.
      */
     public abstract void export(Collection<ViewData> viewDataList);
   }
--- a/core/src/main/java/io/opencensus/stats/ViewManager.java
+++ b/core/src/main/java/io/opencensus/stats/ViewManager.java
@@ -16,6 +16,10 @@
 
 package io.opencensus.stats;
 
+import io.opencensus.common.Duration;
+import java.util.Collection;
+import java.util.List;
+
 /**
  * Provides facilities to register {@link View}s for collecting stats and retrieving
  * stats data as a {@link ViewData}.
@@ -28,7 +32,49 @@ public abstract class ViewManager {
   public abstract void registerView(View view);
 
   /**
-   * Returns the current stats data, {@link ViewData}, associated with the given view name.
+   * Push model for stats. Registers a {@link View} that will collect data to be accessed
+   * via {@link Handler#export(Collection)} of the passed-in {@link Handler}s. Though it's
+   * also possible to manually get stats data via {@link #getView(View.Name)}.
+   */
+  public abstract void registerView(View view, List<? extends Handler> handlers);
+
+  /**
+   * Set the interval for exporting {@code ViewData} to the given {@code Duration}.
+   *
+   * <p>By default, {@code ViewData} will be pushed to {@code Handler}s every 10 seconds.
+   *
+   * @param duration the interval for exporting stats.
+   */
+  public abstract void setExportInterval(Duration duration);
+
+  /**
+   * Returns the current stats data, {@link ViewData}, associated with the given
+   * {@link View}.
    */
   public abstract ViewData getView(View.Name view);
+
+  /**
+   * An abstract class that registers {@code View} and exports recorded {@code ViewData} in their
+   * own format.
+   */
+  public abstract static class Handler {
+
+    /**
+     * Register a {@code View}. Each {@code View} must be registered first, so that its stats can be
+     * exported appropriately.
+     *
+     * @param view the {@code View} to be registered.
+     */
+    public abstract void registerView(View view);
+
+    /**
+     * Exports a list of {@code ViewData}.
+     *
+     * <p>Implementation SHOULD not block the calling thread. It should execute the export on a
+     * different thread if possible.
+     *
+     * @param viewDataList a list of {@code ViewData} objects to be exported.
+     */
+    public abstract void export(Collection<ViewData> viewDataList);
+  }
 }
--- a/core/src/main/java/io/opencensus/stats/ViewManager.java
+++ b/core/src/main/java/io/opencensus/stats/ViewManager.java
@@ -68,13 +68,13 @@ public abstract class ViewManager {
     public abstract void registerView(View view);
 
     /**
-     * Exports a list of {@code ViewData}.
+     * Exports a collection of {@code ViewData}.
      *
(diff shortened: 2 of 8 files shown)


core/src/main/java/io/opencensus/stats/export/StatsExporter.java
```diff
@@ -0,0 +1,80 @@
+/*
+ * Copyright 2016-17, OpenCensus Authors
```
bogdandrutu: don't copy-paste years ..

core/src/main/java/io/opencensus/stats/export/StatsExporter.java
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * Exports a list of {@code ViewData}.
+     *
+     * <p>Implementation SHOULD not block the calling thread. It should execute the export on a
+     * different thread if possible.
+     *
+     * @param viewDataList a list of {@code ViewData} objects to be exported.
+     */
+    public abstract void export(Collection<ViewData> viewDataList);
```
bogdandrutu: I don't know how this interacts with ViewManager and stuff. Can you give some details?

songy23: In `StatsManager` we have a `StatsEvent` class, which implements `EventQueue.Entry` and records stats: https://github.com/census-instrumentation/opencensus-java/blob/master/core_impl/src/main/java/io/opencensus/implcore/stats/StatsManager.java#L78. In the `process()` method, after stats are recorded, a `ViewData` will be added by `StatsExporter.addViewData()`. Similar to span end process: https://github.com/census-instrumentation/opencensus-java/blob/master/impl_core/src/main/java/io/opencensus/implcore/trace/StartEndHandlerImpl.java#L117

core/src/main/java/io/opencensus/stats/export/StatsExporter.java
songy23: Done. Didn't realize these are different..

core/src/main/java/io/opencensus/stats/export/StatsExporter.java
bogdandrutu: I am not asking about implementation details. I think the design that @dinooliva told me he wants to have a per-view registration. Does this register for all the views? how does this interact with the per-view registration?

songy23: We haven't talked about this yet. By now export is enabled for all views by default and doesn't interact with view registrations. Will discuss with @dinooliva later.

bogdandrutu: If that is the case the API looks good.

sebright: > By now export is enabled for all views by default and doesn't interact with view registrations.

Does that mean that the set of views used by StatsExporter is completely independent of the views registered with the ViewManager?  I thought that ViewManager would also handle pushing data.  Then we could have a separate library for exporting to StackDriver that registers StackDriver-specific handlers with the ViewManager.

songy23: > Does that mean that the set of views used by StatsExporter is completely independent of the views registered with the ViewManager?

No it's not independent, only stats of the views registered by Viewmanager will be exported.

> I thought that ViewManager would also handle pushing data.

My idea is pushing stats will be a implementation detail: whenever a record() [StatsEvent](https://github.com/census-instrumentation/opencensus-java/blob/master/core_impl/src/main/java/io/opencensus/implcore/stats/StatsManager.java#L64) happens, `StatsManager` will invoke some method of StatsExporter and push ViewData to it.

> we could have a separate library for exporting to StackDriver that registers StackDriver-specific handlers with the ViewManager.

StackDriver handlers will be in another library (under [exporters](https://github.com/census-instrumentation/opencensus-java/tree/master/exporters)). Are you suggesting we should also move StatsExporter to another library?

core/src/main/java/io/opencensus/stats/export/StatsExporter.java
```diff
@@ -0,0 +1,92 @@
...
+    return NOOP_STATS_EXPORTER;
+  }
+
+  /**
+   * Registers a new service handler that is used by the library to export {@code ViewData}.
+   *
+   * @param name the name of the service handler. Must be unique for each service.
+   * @param handler the service handler that is called for each ended sampled span.
+   */
+  public abstract void registerHandler(String name, Handler handler);
```
sebright: Why are there two methods for registering handlers?  It looks like a handler can be registered with `ViewManager.registerView(View view, List<? extends Handler> handlers)` or `StatsExporter.registerHandler(String name, Handler handler)`.

songy23: `StatsExporter.registerHandler(String name, Handler handler)` is used to register handlers to StatsExporter (so that stats can be pushed to them), `ViewManager.registerView(View view, List<? extends Handler> handlers)` is for registering views against handlers. A handler must be registered on StatsExporter first, and then views can be registered for the handler.

songy23: An alternative is to completely remove StatsExporter and ExportComponent from API. StatsExporter will just be a worker thread in the implementation, and once user calls ViewManager.registerView(View view, List<? extends Handler> handlers), stats for the given view will be automatically exported to the handlers.

This makes our API simpler to use, but might also make it hard to mock and test.

sebright: The alternative sounds simpler, and I like that it would move code out of the API.  How would it affect mocking or testing?  Would we lose any other functionality by moving it out of the API library?

songy23: Right, I just realized moving this to implementation won't affect the functionality or mocking. I'll go ahead with the alternative approach to minimize API changes.

core/src/main/java/io/opencensus/stats/NoopStats.java
```diff
@@ -116,6 +118,18 @@ public void registerView(View newView) {
...
     @Override
+    public void registerView(View newView, List<? extends Handler> handlers) {
+      checkNotNull(handlers, "handlers");
+      registerView(newView);
+    }
+
+    @Override
+    public void setExportInterval(Duration duration) {
+      checkNotNull(duration, "duration");
+      checkArgument(duration.compareTo(ZERO_DURATION) > 0, "Duration cannot be negative.");
```
sebright: "Duration must be positive."?

core/src/main/java/io/opencensus/stats/ViewManager.java
```diff
@@ -28,7 +32,49 @@
...
+   *
+   * <p>By default, {@code ViewData} will be pushed to {@code Handler}s every 10 seconds.
+   *
+   * @param duration the interval for exporting stats.
+   */
+  public abstract void setExportInterval(Duration duration);
+
+  /**
+   * Returns the current stats data, {@link ViewData}, associated with the given
+   * {@link View}.
```
sebright: Shouldn't this say "`{@link View.Name}`"?

core/src/main/java/io/opencensus/stats/ViewManager.java
```diff
@@ -28,7 +32,49 @@
...
+  public abstract void registerView(View view, List<? extends Handler> handlers);
+
+  /**
+   * Set the interval for exporting {@code ViewData} to the given {@code Duration}.
+   *
+   * <p>By default, {@code ViewData} will be pushed to {@code Handler}s every 10 seconds.
+   *
+   * @param duration the interval for exporting stats.
+   */
+  public abstract void setExportInterval(Duration duration);
```
sebright: Would it be better to set an export interval per handler, in case some exporters need more recent data?

core/src/main/java/io/opencensus/stats/ViewManager.java
```diff
@@ -28,7 +32,49 @@
...
+   */
+  public abstract static class Handler {
+
+    /**
+     * Register a {@code View}. Each {@code View} must be registered first, so that its stats can be
+     * exported appropriately.
+     *
+     * @param view the {@code View} to be registered.
+     */
+    public abstract void registerView(View view);
```
sebright: Could this method go in the StackDriver exporter?  It will already need to know which views to export in order to call ViewManager.registerView.

core/src/main/java/io/opencensus/stats/ViewManager.java
```diff
@@ -28,7 +32,49 @@
...
+
+    /**
+     * Exports a list of {@code ViewData}.
+     *
+     * <p>Implementation SHOULD not block the calling thread. It should execute the export on a
+     * different thread if possible.
+     *
+     * @param viewDataList a list of {@code ViewData} objects to be exported.
```
sebright: s/list/collection ?

core/src/main/java/io/opencensus/stats/ViewManager.java
```diff
@@ -28,7 +32,49 @@
...
+     */
+    public abstract void export(Collection<ViewData> viewDataList);
```
sebright: Why does this method take more than one ViewData?  Does the ViewManager try to only call this method once when the same handler is registered for more than one View?

core_impl/src/main/java/io/opencensus/implcore/stats/MeasureToViewMap.java
```diff
@@ -84,10 +86,13 @@ private synchronized MutableViewData getMutableViewData(View.Name viewName) {
   }
 
   /** Enable stats collection for the given {@link View}. */
-  synchronized void registerView(View view, Clock clock) {
+  synchronized void registerView(View view, List<? extends Handler> handlers, Clock clock) {
```
sebright: Would it be better to process the list of handlers in another class, and have MeasureToViewMap just hold the map of aggregated data?  It seems like the code that manages the handlers could use just the export intervals, handlers, and registered views to determine what views to query at different times.  Then it could call getView on the StatsManager or MeasureToViewMap and pass the results to the handlers.

core/src/main/java/io/opencensus/stats/ViewManager.java
songy23: It's like a batch call, if multiple views are registered against one handler, stats on them will be exported with one call. If a handler only has one view, ViewManager will just export ViewData for that view.

Also refer to Trace span export: https://github.com/census-instrumentation/opencensus-java/blob/master/impl_core/src/main/java/io/opencensus/implcore/trace/export/SpanExporterImpl.java#L125

core_impl/src/main/java/io/opencensus/implcore/stats/MeasureToViewMap.java
songy23: My idea is, if a view is registered using a push model (i.e with a list of handlers), we need to block the thread where the Stats library is running, until we make sure view is successfully registered against the given handlers.That's because if we don't block the thread, ViewData might be exported to handlers before view is registered for that handler. This is hard to achieve if we move the logic on registering view for handlers to other class.

sebright: I don't think that that race condition is possible if the new export class controls exporting to handlers.  `ViewManagerImpl.registerView(View view, List<? extends Handler> handlers)` could add the view to MeasureToViewMap and then call a method on the new export class to give it the view and handlers.  Then the new export class could determine when to export data to each handler and get that data from MeasureToViewMap when necessary.

core/src/main/java/io/opencensus/stats/ViewManager.java
sebright: There was one more occurrence at the top of the Javadoc.

songy23:
Per our discussion today, we're holding off supporting a push model for View in Stats library until next release. In the first release, exporting to Stackdriver will be an independent artifact.

PR #670, opened 2017-09-25
Check for null arguments in tagging no-op classes.

sebright:
This behavior is more consistent with the corresponding implementation classes.


sebright:
Thanks!

PR #671, opened 2017-09-26
Move stats no-op classes into a new class, NoopStats.

sebright:
This design is more consistent with the 'tags' package.


sebright:
Thanks!

PR #672, opened 2017-09-26
Throw TagContextParseException for invalid tags (fixes #660).

### LINKED ISSUE #660: Don't throw IllegalArgumentException when deserializing invalid tags.
Currently, `TagContextBinarySerializer.fromByteArray` calls `TagKeyString.create` and `TagValueString.create` to create the `TagContext`.  Those two methods can throw `IllegalArgumentException` if the Strings are not valid values for tag keys or values.  We should either throw a `TagContextParseException` or find a different way to handle invalid serialized tags.

sebright:


core_impl/src/main/java/io/opencensus/implcore/tags/propagation/SerializationUtils.java
```diff
@@ -152,6 +151,26 @@ static TagContextImpl deserializeBinary(byte[] bytes) throws TagContextParseExce
     }
   }
 
+  // TODO(sebright): Consider exposing a TagKey name validation method to avoid needing to catch an
+  // IllegalArgumentException here.
+  private static final TagKeyString createTagKey(String name) throws TagContextParseException {
+    try {
+      return TagKeyString.create(name);
+    } catch (IllegalArgumentException e) {
+      throw new TagContextParseException("Invalid tag key: " + name, e);
```
bogdandrutu: Is this just to have a better exception message? You can simply catch IllegalArgumentException in the deserializeBinary and avoid this 2 methods.

sebright: I was trying to give a better error message, so that we don't need to rely on the IllegalArgumentException giving the user all of the information that they need.  I also thought that we might want to add something like `TagKey.isValidName(String)` later, if throwing the IllegalArgumentException is inefficient, which would also require the TagContextParseException to be thrown where the TagKey is created.

bogdandrutu: I would not use TagContextParseException for TagKey creation because it is not a parsing exception in my opinion.

sebright: What do you think we should do?  Should we use a different custom checked exception (possibly sharing a superclass with TagContextParseException)?  I'd also like to document the error handling for invalid tags in https://github.com/census-instrumentation/opencensus-specs, since this should be consistent across languages.

bogdandrutu: For the moment we are ok we do not change the API so we are good. Please document this.

sebright: I created an issue for improving the handling of invalid tags: #674

sebright: I also updated fromByteArray's Javadoc: #675

core_impl/src/test/java/io/opencensus/implcore/tags/propagation/TagContextDeserializationTest.java
```diff
@@ -71,6 +75,38 @@ public void testDeserializeEmptyByteArrayThrowException() throws Exception {
...
+    os.write(SerializationUtils.VERSION_ID);
+    os.write(SerializationUtils.VALUE_TYPE_STRING);
+
+    // Encode a valid tag key and an invalid tag value:
+    encodeString("key", os);
+    encodeString("val\3", os);
+    final byte[] bytes = os.toByteArray();
+
+    thrown.expect(TagContextParseException.class);
+    thrown.expectMessage("Invalid tag value: val\3");
```
adriancole: Imho the most important part of the change is making sure we see the tag name that failed to parse. That helps the user figure out what they screwed up. Can we make the message include the tag name? We have to do things like this in zipkin to reduce support costs.

Issue #639, opened 2017-09-17
Consider making MeasureMap abstract.


sebright:
Making MeasureMap abstract would allow us to optimize the implementation more easily.  It would also allow the stats implementation to access state that isn't exposed through the API.  These are the same reasons that we made TagContext abstract.

sebright:
I made a PR that doesn't expose MeasureMap in the API: #676

sebright:
Implemented in #676.

PR #676, opened 2017-09-26
Move MeasureMap from the API to the implementation.

sebright:
I wrote a rough implementation of #639 and wanted to ask for feedback.

This commit adds an abstract builder-like class (`StatsBuilder`) for setting measures.  The new class has a record method to record all of the measures with the same timestamp.

Here is an example with implicit propagation.

master:

```java
        statsRecorder.record(
            MeasureMap.builder()
                .put(RPC_CLIENT_ROUNDTRIP_LATENCY, 100)
                .put(RPC_CLIENT_REQUEST_BYTES, 2000)
                .build());
```

this PR:

```java
        statsRecorder
            .newStatsBuilder()
            .put(RPC_CLIENT_ROUNDTRIP_LATENCY, 100)
            .put(RPC_CLIENT_REQUEST_BYTES, 2000)
            .record();
```

I'm not sure what names to use for the new class or methods.  I'm also not sure how to handle tags for explicit propagation, because they could be an argument to `newStatsBuilder` or `record`.  (They are currently an argument to `newStatsBuilder` in this PR.)

Here are some reasons for making this change.  The first two also applied to making TagContext abstract.
1. We don't need to put the concrete `MeasureMap` class in the API.  This gives us more flexibility to change it later.  For example, the stats implementation could access its list of measure values directly, rather than going through the iterator.  This API doesn't even require us to build the immutable map of measures.
2. We don't need to create a `MeasureMap` when stats collection isn't enabled.
3. This API requires slightly less code to record stats.

Reasons to keep `MeasureMap` in the API:
1. `MeasureMap` is simpler to understand and document than `StatsBuilder`.
2. `MeasureMap` is threadsafe and can be reused.
3. `StatsRecorder.record(MeasureMap stats)` is easier to mock.

/cc @adriancole


songy23:
> We don't need to put the concrete MeasureMap class in the API. This gives us more flexibility to change it later.

I'm not sure how likely it is for us to change `MeasureMap` later. To me I think `MeasureMap` is just a value class, and our current implementation is simple enough. As a contrast, `StatsBuilder` and its implementation seems more complicated.

> We don't need to create a MeasureMap when stats collection isn't enabled.

Could you give an example on how `StatsBuilder` will be used when stats collection isn't enabled?

core/src/main/java/io/opencensus/stats/StatsRecorder.java
```diff
@@ -21,22 +21,23 @@
...
 
   /**
-   * Records a set of measurements with the tags in the current context.
+   * Returns an object for recording multiple measures with the tags in the current context.
    *
-   * @param measureValues the measurements to record.
+   * @return an object for recording multiple measures with the tags in the current context.
    */
-  public final void record(MeasureMap measureValues) {
+  public final StatsBuilder builder() {
```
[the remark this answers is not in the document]
sebright: Thanks for the suggestion.  How about `statsBuilder()` or `newStatsBuilder()`, to emphasize that it is empty and independent of any stats recorded previously?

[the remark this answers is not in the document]
songy23: `newStatsBuilder()` sounds better to me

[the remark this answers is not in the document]
sebright: I updated the code and example.

sebright:
> > We don't need to put the concrete MeasureMap class in the API. This gives us more flexibility to change it later.
>
> I'm not sure how likely it is for us to change MeasureMap later. To me I think MeasureMap is just a value class, and our current implementation is simple enough. As a contrast, StatsBuilder and its implementation seems more complicated.

I thought that we might want to make the same types of changes to `MeasureMap` that we could make to `TagContext`, such as changing the data structure that holds the values or changing how the implementation accesses those values.  For example, we could store the measures in a `HashMap<Measure, Number>` as the user specifies them, and `MeasureToViewMap.record` could use the `Measure`s and values from the map without creating intermediate `Iterator` or `Measurement` objects.  I think it is important to be able to make these changes if we find any performance problems later.

> > We don't need to create a MeasureMap when stats collection isn't enabled.
>
> Could you give an example on how StatsBuilder will be used when stats collection isn't enabled?

In the example in the PR description, `statsRecorder.newStatsBuilder()` would return a [NoopStatsBuilder](https://github.com/sebright/opencensus-java/blob/d3cdfc554e9bd5c6a7ce56a9c0dff6fa9f1bde82/core/src/main/java/io/opencensus/stats/NoopStats.java#L109).  Then each call to `put` would return the same builder, and `record` would do nothing.

core/src/main/java/io/opencensus/stats/NoopStats.java
```diff
@@ -88,10 +99,28 @@ public StatsRecorder getStatsRecorder() {
...
+      return this;
+    }
+
+    @Override
+    public StatsBuilder put(MeasureLong measure, long value) {
+      return this;
     }
+
+    @Override
+    public void record() {}
```
bogdandrutu: because Java I may call this "buildAndRecord". You may keep a build as well and the record function:
1. If we want to count active/finished requests we will always record a long metric with value 1 so I can allocate that object once.
2. I think there are other examples like this when you record the same metrics every time.

sebright: I didn't consider that users might need to record the same values twice.  Do you think that would be likely for a map with size greater than 1?  I think we could handle the 1 measure case by adding a convenience method like `record(MeasureDouble, double)`.

If we also added a `build` method, it might be better to separate it from the `record` method, something like

```java
MeasureMap measures =
     statsRecorder.measureMapBuilder().
            .put(RPC_CLIENT_ROUNDTRIP_LATENCY, 100)
            .put(RPC_CLIENT_REQUEST_BYTES, 2000)
            .build();
statsRecorder.record(measures);
```

/cc @dinooliva

core/src/main/java/io/opencensus/stats/NoopStats.java
```diff
@@ -88,10 +99,28 @@ public StatsRecorder getStatsRecorder() {
     static final StatsRecorder INSTANCE = new NoopStatsRecorder();
 
     @Override
-    public void record(TagContext tags, MeasureMap measureValues) {
+    public StatsBuilder newStatsBuilder(TagContext tags) {
```
songy23: nit: might be better to have a method like `StatsBuilder.withExplicitTagContext(TagContext tags)` instead of passing TagContext to the builder method? I think if there's a parameter to newBuilder(), usually that parameter should be required, which is not the case here.

sebright: As we discussed offline, the API should make the distinction between the two use cases more clear.  I renamed `StatsRecorder.newStatsBuilder(TagContext)` to `newStatsBuilderWithExplicitTagContext`.

sebright:
I think I found better names.  I made the following renames in the last commit:

`StatsBuilder` -> `StatsRecord`
`StatsRecorder.newStatsBuilder` -> `StatsRecorder.newRecord`
`StatsRecorder.newStatsBuilderWithExplicitTagContext` -> `StatsRecorder.newRecordWithExplicitTagContext`

sebright:
I modified the API so that StatsRecord.record can take a TagContext.  Here is an example of recording with an explicit TagContext now:

```java
        statsRecorder
            .newRecord()
            .put(RPC_CLIENT_ROUNDTRIP_LATENCY, 100)
            .put(RPC_CLIENT_REQUEST_BYTES, 2000)
            .recordWithExplicitTagContext(tags);
```

I think it is slightly simpler, since the StatsRecord doesn't need to store the tags.

core/src/main/java/io/opencensus/stats/NoopStats.java
```diff
@@ -99,9 +110,28 @@ public void setState(StatsCollectionState state) {
...
+      return this;
+    }
+
+    @Override
+    public StatsRecord put(MeasureLong measure, long value) {
+      return this;
+    }
+
+    @Override
+    public void recordWithExplicitTagContext(TagContext tags) {
```
bogdandrutu: Probably at one point we should add another API that is:

public void recordWithExplicitTagContextAndTags(TagContext tags, Map<String, String> extraTags) {
}

To allow users to record stats with extra tags without creating a new TagContext only for this call.

Not necessary now but something to keep in mind.

core/src/main/java/io/opencensus/stats/StatsRecord.java
```diff
@@ -0,0 +1,59 @@
...
+   * MeasureLong} will overwrite the previous value.
+   *
+   * @param measure the {@link MeasureLong}
+   * @param value the value to be associated with {@code measure}
+   * @return this
+   */
+  public abstract StatsRecord put(MeasureLong measure, long value);
+
+  /** Records all of the measures at the same time. */
+  public final void record() {
```
bogdandrutu: Should we also mention here and to the other method that this can be called multiple times?

core/src/main/java/io/opencensus/stats/StatsRecord.java
```diff
@@ -0,0 +1,59 @@
...
+    // Use the context key directly, to avoid depending on the tags implementation.
+    recordWithExplicitTagContext(ContextUtils.TAG_CONTEXT_KEY.get());
```
bogdandrutu: I would keep this abstract to be consistent with the fact that we want to allow ourselves to be backwards compatible, and move this to impl.

api/src/main/java/io/opencensus/stats/NoopStats.java
```diff
@@ -99,9 +110,31 @@ public void setState(StatsCollectionState state) {
     static final StatsRecorder INSTANCE = new NoopStatsRecorder();
 
     @Override
-    public void record(TagContext tags, MeasureMap measureValues) {
+    public StatsRecord newRecord() {
```
adriancole: Maybe recorder as opposed to record

api/src/main/java/io/opencensus/stats/StatsRecord.java
```diff
@@ -0,0 +1,63 @@
...
+   */
+  public abstract void record();
+
+  /**
+   * Records all of the measures at the same time, with an explicit {@link TagContext}.
+   *
+   * <p>This method records all of the stats in the {@code StatsRecord} every time it is called.
+   *
+   * @param tags the tags associated with the measurements.
+   */
```
adriancole: Too wordy imho just say record(context) or recordInContext(context)

api/src/main/java/io/opencensus/stats/NoopStats.java
sebright: I like that name, except that the class containing this method is already called StatsRecorder.  I opened an issue for finding better names for StatsRecord and the related methods: #726

api/src/main/java/io/opencensus/stats/StatsRecord.java
sebright: I added "WithExplicitTagContext" to the method name because @songy23 thought that having an overloaded method to create the StatsRecord was unclear (https://github.com/census-instrumentation/opencensus-java/pull/676#discussion_r143073529), but I wonder if the distinction between the use cases is clearer now that `StatsRecord.record` takes the optional TagContext.  If the TagContext can only be provided at the same time that the stats are recorded, it should be clear which tags will be used.

/cc @songy23

songy23: Agreed. In this situation, overloading record() with an optional TagContext is less ambiguous (than having it at newRecord()). I'm fine with renaming this and just use record().

sebright: Thanks for the feedback.  I made a PR: #728

PR #677, opened 2017-09-27
Add a no-op ViewManager (closes #614).

### LINKED ISSUE #614: Decide whether to add  a no-op ViewManager.
Currently, `Stats.getViewManager()` returns null if no implementation is available.  We could instead return a no-op ViewManager, for consistency with `Stats.getStatsRecorder()` and most of the tracing classes.  However, a no-op ViewManager is much less useful than a no-op StatsRecorder, because the ViewManager isn't used by instrumented code.  The ViewManager is used by code that may already need to check whether stats collection is enabled.  That code would need to call an `isNoop()` method instead of checking for null.

I looked at the other packages quickly, and the only other classes that use null instead of no-op implementations are `RunningSpanStore` and `SampledSpanStore`.

EDIT: [ViewManager](https://github.com/census-instrumentation/opencensus-java/blob/master/core/src/main/java/io/opencensus/stats/ViewManager.java) contains the methods for registering views and querying the stats collected for those views.

sebright:
NoopViewManager maintains a map of views so that it can return empty ViewDatas
from getView.  It also prevents different views with the same name from being
registered.

### DIFF
--- a/core/src/main/java/io/opencensus/stats/NoopStats.java
+++ b/core/src/main/java/io/opencensus/stats/NoopStats.java
@@ -129,7 +129,6 @@ final class NoopStats {
               existing
                   .getWindow()
                   .match(
-                      // TODO(sebright): Should we use the zero timestamp or the current time?
                       Functions.<AggregationWindowData>returnConstant(
                           CumulativeData.create(ZERO_TIMESTAMP, ZERO_TIMESTAMP)),
                       Functions.<AggregationWindowData>returnConstant(
--- a/core/src/main/java/io/opencensus/stats/NoopStats.java
+++ b/core/src/main/java/io/opencensus/stats/NoopStats.java
@@ -16,11 +16,23 @@
 
 package io.opencensus.stats;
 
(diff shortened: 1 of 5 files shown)


core/src/main/java/io/opencensus/stats/NoopStats.java
```diff
@@ -72,4 +93,50 @@ public void record(TagContext tags, MeasureMap measureValues) {
...
+        if (existing == null) {
+          throw new IllegalArgumentException("View is not registered.");
+        } else {
+          return ViewData.create(
+              existing,
+              Collections.<List<TagValue>, AggregationData>emptyMap(),
+              existing
+                  .getWindow()
+                  .match(
+                      // TODO(sebright): Should we use the zero timestamp or the current time?
```
songy23: I think zero timestamp is better.

sebright: Thanks.  I removed the TODO.

sebright:
@bogdandrutu I created an issue: #679

sebright:
Thanks for the review!

PR #678, opened 2017-09-27
Add tag key to message about deserializing invalid tag value.

sebright:
This PR addresses https://github.com/census-instrumentation/opencensus-java/pull/672#discussion_r141222897.

### DIFF
--- a/core_impl/src/test/java/io/opencensus/implcore/tags/propagation/TagContextDeserializationTest.java
+++ b/core_impl/src/test/java/io/opencensus/implcore/tags/propagation/TagContextDeserializationTest.java
@@ -28,13 +28,11 @@ import io.opencensus.tags.propagation.TagContextBinarySerializer;
 import io.opencensus.tags.propagation.TagContextParseException;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
-import org.hamcrest.Matcher;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
-import org.mockito.internal.matchers.Matches;
 
 /**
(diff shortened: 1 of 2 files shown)


core_impl/src/test/java/io/opencensus/implcore/tags/propagation/TagContextDeserializationTest.java
```diff
@@ -97,12 +99,17 @@ public void testDeserializeInvalidTagValue() throws Exception {
...
     // Encode a valid tag key and an invalid tag value:
-    encodeString("key", os);
+    encodeString("my key", os);
     encodeString("val\3", os);
     final byte[] bytes = os.toByteArray();
 
+    // TODO(sebright): Is there a better way to match a regular expression? 'Matches' extends
+    // Matcher<Object>, but 'expectMessage' takes a Matcher<String>.
+    @SuppressWarnings({"rawtypes", "unchecked"})
+    Matcher<String> matcher = (Matcher) new Matches("Invalid tag value for key .*my key.*: val\3");
```
bogdandrutu: Maybe something from this package can help.
http://hamcrest.org/JavaHamcrest/javadoc/1.3/org/hamcrest/core/package-summary.html

adriancole: Why not jusr equals or prefix match? Do we need regex for this test?

sebright: @bogdandrutu I found prefix, substring, and suffix matchers in the Hamcrest library, but not regex.

@adriancole I thought it would be better to not depend on the toString format used by Autovalue, but avoiding the exact match might not be worth the complexity.  I changed the test to use an exact match.

PR #682, opened 2017-09-28
Clean up io.opencensus.internal.StringUtil.

sebright:
We don't need StringUtil.sanitize, since we no longer sanitize tags.
Additionally, we don't need to throw an AssertionError from the constructor to
discourage instantiation of the class, because the class is now in the internal
package.

_________________________________________________________________

@songy23 Do you think we can remove these constants and methods now?


sebright:
Thanks!  I'll merge when the build passes.

PR #685, opened 2017-09-29
Implement Tags.getState and Tags.setState (part of #608).

sebright:
Disabling tagging does not disable stats collection, so disabling tagging could
cause stats to be recorded with empty or out-of-date tags.

/cc @adriancole

### DIFF
--- a/core_impl/src/main/java/io/opencensus/implcore/tags/TaggerImpl.java
+++ b/core_impl/src/main/java/io/opencensus/implcore/tags/TaggerImpl.java
@@ -22,7 +22,9 @@ import io.opencensus.tags.Tag;
 import io.opencensus.tags.TagContext;
 import io.opencensus.tags.TagContextBuilder;
 import io.opencensus.tags.Tagger;
+import io.opencensus.tags.TaggingState;
 import java.util.Iterator;
+import java.util.concurrent.atomic.AtomicReference;
 
 public final class TaggerImpl extends Tagger {
   // All methods in this class use TagContextImpl and TagContextBuilderImpl. For example,
@@ -32,6 +34,12 @@ public final class TaggerImpl extends Tagger {
   // TODO(sebright): Consider treating an unknown TagContext as empty.  That would allow us to
   // remove TagContext.unsafeGetIterator().
 
+  private final AtomicReference<TaggingState> state;
+
+  TaggerImpl(AtomicReference<TaggingState> state) {
+    this.state = state;
+  }
+
   @Override
   public TagContextImpl empty() {
     return TagContextImpl.EMPTY;
@@ -39,12 +47,14 @@ public final class TaggerImpl extends Tagger {
 
   @Override
   public TagContextImpl getCurrentTagContext() {
-    return toTagContextImpl(CurrentTagContextUtils.getCurrentTagContext());
+    return toTagContextImpl(CurrentTagContextUtils.getCurrentTagContext(), state.get());
   }
 
   @Override
-  public TagContextBuilderImpl emptyBuilder() {
-    return new TagContextBuilderImpl();
+  public TagContextBuilder emptyBuilder() {
+    return state.get() == TaggingState.DISABLED
+        ? NoopTagContextBuilder.INSTANCE
+        : new TagContextBuilderImpl();
   }
 
   @Override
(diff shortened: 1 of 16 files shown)


core_impl/src/main/java/io/opencensus/implcore/tags/TaggerImpl.java
```diff
@@ -31,19 +33,27 @@
   // TODO(sebright): Consider treating an unknown TagContext as empty.  That would allow us to
   // remove TagContext.unsafeGetIterator().
 
+  private final AtomicReference<TaggingState> state;
```
bogdandrutu: private volatile TaggingState state.

You don't need getAndSet or lazySet.

core_impl/src/main/java/io/opencensus/implcore/tags/TaggerImpl.java
```diff
@@ -52,16 +62,19 @@ public TagContextBuilder currentBuilder() {
...
   }
 
   @Override
   public Scope withTagContext(TagContext tags) {
-    return CurrentTagContextUtils.withTagContext(toTagContextImpl(tags));
+    return CurrentTagContextUtils.withTagContext(toTagContextImpl(tags, state.get()));
   }
 
-  private static TagContextImpl toTagContextImpl(TagContext tags) {
+  private static TagContextImpl toTagContextImpl(TagContext tags, TaggingState state) {
```
bogdandrutu: I would rather make this non-static, up to you.

core_impl/src/main/java/io/opencensus/implcore/tags/TaggerImpl.java
```diff
@@ -80,7 +93,10 @@ private static TagContextImpl toTagContextImpl(TagContext tags) {
     }
   }
 
-  private static TagContextBuilderImpl toTagContextBuilderImpl(TagContext tags) {
+  private static TagContextBuilder toTagContextBuilderImpl(TagContext tags, TaggingState state) {
```
bogdandrutu: same comment.

core_impl/src/main/java/io/opencensus/implcore/tags/TaggerImpl.java
bogdandrutu: Also I think we discussed to not change this here and do it only when we need to do it (record usage maybe when add tags).

core_impl/src/main/java/io/opencensus/implcore/tags/TaggerImpl.java
sebright: I also updated this method.

core_impl/src/main/java/io/opencensus/implcore/tags/TaggerImpl.java
bogdandrutu: I want you just to replace your current line with my proposed line:
s/private final AtomicReference<TaggingState> state;/private volatile TaggingState state;

core_impl/src/main/java/io/opencensus/implcore/tags/TaggerImpl.java
bogdandrutu: I meant because you passed an internal state (the state) as argument to keep this static instead of making it instance method.

core_impl/src/main/java/io/opencensus/implcore/tags/TaggerImpl.java
sebright: I don't understand how that change would work.  How would the TagsComponent update the state?

core_impl/src/main/java/io/opencensus/implcore/tags/TaggerImpl.java
sebright: Now I see what you mean.  I passed in the state enum (instead of giving access to the AtomicReference) to make it clear that the method only needs a snapshot of the state and doesn't need to update it.

core_impl/src/main/java/io/opencensus/implcore/tags/TaggerImpl.java
bogdandrutu: Oh what you do is you share an atomic reference object between multiple classes. I would like to propose changes:
1) Create a class for the CurrentState (or Eanbled).
2) Share that object if you don't want to have it as a global state (using static functions).

What do you think? That is much cleared that you share that state that is owned by the TraceComponent.

sebright: Done

sebright:
Thanks!

PR #686, opened 2017-09-29
Add link to the cross-language binary serialization format for TagConext.

sebright:
/cc @adriancole

### DIFF
--- a/core/src/main/java/io/opencensus/tags/propagation/TagContextBinarySerializer.java
+++ b/core/src/main/java/io/opencensus/tags/propagation/TagContextBinarySerializer.java
@@ -18,7 +18,13 @@ package io.opencensus.tags.propagation;
 
 import io.opencensus.tags.TagContext;
 
-/** Object for serializing and deserializing {@link TagContext}s with the binary format. */
+/**
+ * Object for serializing and deserializing {@link TagContext}s with the binary format.
+ *
+ * <p>See <a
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/opencensus/tags/propagation/TagContextBinarySerializer.java
```diff
@@ -18,7 +18,13 @@
...
 
-/** Object for serializing and deserializing {@link TagContext}s with the binary format. */
+/**
+ * Object for serializing and deserializing {@link TagContext}s with the binary format.
+ *
+ * <p>See <a
+ * href="https://github.com/census-instrumentation/opencensus-specs/blob/master/encodings/BinaryEncoding.md#tag-context">opencensus-specs</a>
+ * for the specification of the cross-language binary serialization format.
+ */
 public abstract class TagContextBinarySerializer {
```
bogdandrutu: I would name this BinarySerializer or BinaryFormat. Already in tags/propagation so kind of TagContext is redundant.

sebright: I wanted to avoid adding a class with the same name as a class in the trace package, especially since code that serializes TagContexts may also need to serialize SpanContexts.

bogdandrutu: Up to you. Please merge the PR because it is ready.

sebright: Any class renaming should at least be done in a separate pull request, so I'll merge this one now.

PR #697, opened 2017-10-05
 Remove dependency between i.o.implcore.tags and i.o.internal.

sebright:
The tagging implementation was using NoopScope from the API library, even though
it was in an "internal" package.  This commit adds a NoopScope class to
io.opencensus.implcore.internal.

______________________________________

It would be nice if we could check for access to internal packages in CI, but I'm not sure how.


sebright:
Thanks!

Issue #655, opened 2017-09-21
Consider implementing serialization for long and boolean tag values in the first release.


sebright:
If the first stats library release doesn't support deserializing long and boolean tags, then it will have to ignore those tags when receiving tags from a client that is using a later version of the library.  I'm not sure how to implement deserialization, though, since the factory methods for `TagKeyLong` and `TagKeyBoolean` are currently package-private.

bogdandrutu:
So if we do not implement right now the deserialization, we will have a bit to work because we will need to always put all the new fields at the end (you don't know how to skip a field with an unknown field id so we will drop everything up to the end).

Probably just skip these fields is good, don't necessary need to propagate them, though is better to do it if we can. Let's keep the work minimal and maybe skip them initially.

bogdandrutu:
By implementing deserialization I mean skip the tags. Correct, if you add the full support later you will have problems with propagation because some services will drop them, but there is no other option (the same problem will apply if we start supporting doubles for example later). I think it is fine, we keep everything backwards compatible so we should be fine.

sebright:
@bogdandrutu Thanks for the feedback.  I think skipping the tags makes sense.  We should at least add tests for deserializing TagContexts that contain other tag types, where the other types are between string tags or at the end.

sebright:
We are no longer supporting non-string tags, but this issue still applies to any unknown TagContext fields.  `fromByteArray` should skip the rest of the input when it encounters an unknown field ID.

sebright:
#715 implemented skipping unknown fields.

PR #702, opened 2017-10-07
Refactor deserialization unit tests, and test error messages.

sebright:
The refactoring will simplify testing more cases for #655.  Checking error
messages ensures that we are testing the right error cases.


sebright:
Thanks!

PR #703, opened 2017-10-08
Add an option in EndSpanOptions to register the span name for sampled store.

bogdandrutu:


bogdandrutu:
@adriancole I got a strong opinion from the gRPC team that they need to modify the generated code from the proto otherwise and they feel uncomfortable to do that.

adriancole:
>
> @adriancole <https://github.com/adriancole> I got a strong opinion from
> the gRPC team that they need to modify the generated code from the proto
> otherwise and they feel uncomfortable to do that.
>

mind just for disclosure pasting what the problem is code-wise? Ex what the
code looks like now to use the flag and where the thing that hits codegen
wall is? just so we know in case that changes a year from now.

bogdandrutu:
Here is the initial commit:
https://github.com/grpc/grpc-java/commit/a6653bb135ccf8352815f41036dd6cdf76f38fb8

Problem is that they register the name maybe multiple times (every time when the MethodDescriptor is built), because they try to avoid a dependency on opencensus directly in the generated stub.

As a bonus the current register implementation is a global lock so it may cause performance problems for them as well.

adriancole:
thanks

api/src/main/java/io/opencensus/trace/SpanBuilder.java
```diff
@@ -135,6 +137,23 @@
...
+   * the library will save samples for each name.
+   *
+   * <p>It is strongly recommended to use the {@link
+   * io.opencensus.trace.export.SampledSpanStore#registerSpanNamesForCollection(Collection)} API
+   * instead.
+   *
+   * @return this.
+   */
+  @ExperimentalApi
+  public abstract SpanBuilder setRegisterNameToLocalSpanStore(boolean registerNameToLocalSpanStore);
```
zhangkun83: This method has nothing to do with "registration" per se. Something like `setSampleToLocalSpanStore()` is more accurate.

bogdandrutu: done.

zhangkun83:
> Here is the initial commit:
> grpc/grpc-java@a6653bb
> Problem is that they register the name maybe multiple times (every time when the MethodDescriptor is built), because they try to avoid a dependency on opencensus directly in the generated stub.
> As a bonus the current register implementation is a global lock so it may cause performance problems for them as well.

Besides the global lock, it's hard to remove a name from the global registry when it's no longer needed, e.g., when the generated stub and its MethodDescriptors are garbage-collected routinely, e.g., if they are loaded by a separate ClassLoader.

zhangkun83:
@bogdandrutu, I just realized that having the option on `SpanBuilder` works for the client-side, but not for the server-side, because on server-side the Span is created before `MethodDescriptor`, which propagates this option from the generated code, is available. It could work if the option were on `Span` instead.

zhangkun83:
Or, in `EndSpanOptions`.

bogdandrutu:
@zhangkun83 can you be more specific here. I think this was the requirement from our meeting.

adriancole:
Besides the global lock, it's hard to remove a name from the global
registry when it's no longer needed, e.g., when the generated stub and its
MethodDescriptors are garbage-collected routinely, e.g., if they are loaded
by a separate ClassLoader


So suppose this is an impact of the census tracer usually living at a
parent classloader (due to static initialization and access) right?

zhangkun83:
@bogdandrutu Can you put the option in the `EndSpanOptions` rather than the builder? After our meeting I realize on the server side the Span is started before gRPC knows what value this option should be, thus we can't set it on the Span builder. IIUC this option will have impact only when a Span is closed, it doesn't need to be set at the beginning.

zhangkun83:
> So suppose this is an impact of the census tracer usually living at a
parent classloader (due to static initialization and access) right?

To my best understanding, yes.

bogdandrutu:
Done. PTAL

impl_core/src/main/java/io/opencensus/implcore/trace/export/SampledSpanStoreImpl.java
```diff
@@ -252,7 +256,11 @@ public Summary getSummary() {
    */
   public void considerForSampling(SpanImpl span) {
     synchronized (samples) {
-      PerSpanNameSamples perSpanNameSamples = samples.get(span.getName());
+      String spanName = span.getName();
+      if (span.getSampleToLocalSpanStore() && !samples.containsKey(spanName)) {
```
zhangkun83: You can get rid of the synchronization in this class by changing `samples` to a `ConcurrentHashMap`.
Here and in `registerSpanNamesForCollection()` you can use `putIfAbsent()`:
```java
PerSpanNameSamples perSpanNameSamples = samples.get(spanName);
if (perSpanNameSamples == null && span.getSampleToLocalSpanStore()) {
  perSpanNameSamples = new PerSpanNameSamples();
  PerSpanNameSamples existingSamples = samples.putIfAbsent(spanName, perSpanNameSamples);
  if (existingSamples != null) {
    perSpanNameSamples = existingSamples;
  }
}
```

impl_core/src/main/java/io/opencensus/implcore/trace/SpanImpl.java
```diff
@@ -191,6 +194,17 @@ public long getLatencyNs() {
   }
 
   /**
+   * Returns if the name of this {@code Span} must be register to the {@code SampledSpanStore}.
```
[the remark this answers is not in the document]
bogdandrutu: Done.

impl_core/src/main/java/io/opencensus/implcore/trace/export/SampledSpanStoreImpl.java
bogdandrutu: https://github.com/census-instrumentation/opencensus-java/issues/711

bogdandrutu:
PTAL.

api/src/main/java/io/opencensus/common/ExperimentalApi.java
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * <p>Usage guidelines:
+ * <ol>
+ * <li>This annotation is used only on public API. Internal interfaces should not use it.</li>
+ * <li>After OpenCensus has gained API stability, this annotation can only be added to new API.
+ * Adding it to an existing API is considered API-breaking.</li>
+ * <li>Removing this annotation from an API gives it stable status.</li>
+ * </ol>
+ */
+@Internal
```
sebright: Shouldn't ExperimentalApi not be annotated with Internal, since it is meant to be used in the API?

bogdandrutu: Is marked as Internal for others to not use it.

sebright: I thought that `@Internal` meant that the element should be treated as private.  However, we do want users to see ExperimentalApi's definition, documentation, and uses.

PR #704, opened 2017-10-09
Silently skip tags with unknown type during deserialization (issue #655).

sebright:
Now TagContextBinarySerializerImpl.fromByteArray skips unknown tag types instead
of throwing a TagContextParseException.  When the tag has type long or boolean,
it only skips the individual tag.  When the tag has an unknown type, it skips
the rest of the input, since it doesn't know the tag's length.

EDIT:
This change will increase compatibility between the first version of the tagging
implementation and later versions, because the first version will be able to
accept tags that it supports, even if they are mixed with tags that it doesn't
support.

/cc @adriancole

### DIFF
--- a/core_impl/src/main/java/io/opencensus/implcore/tags/propagation/SerializationUtils.java
+++ b/core_impl/src/main/java/io/opencensus/implcore/tags/propagation/SerializationUtils.java
@@ -39,6 +39,7 @@ import java.nio.BufferUnderflowException;
 import java.nio.ByteBuffer;
 import java.util.HashMap;
 import java.util.Iterator;
+import java.util.Map;
 
 /**
  * Methods for serializing and deserializing {@link TagContext}s.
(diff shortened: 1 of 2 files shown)


songy23:
I would prefer throwing exception for unsupported tags instead. I think it might be better to hard crash users' code when there's an incompatibility between tagging versions, so that they know it's time to update their serialization code.

sebright:
@songy23 This PR is based on the Serialization Rules and Deserialization Rules sections of [BinaryEncoding.md](https://github.com/census-instrumentation/opencensus-specs/blob/master/encodings/BinaryEncoding.md) in opencensus-specs.  I think that if we throw exceptions for unrecognized fields, we should also update opencensus-specs.  I opened an issue for adding more details on how to handle invalid input, which relates to this issue: https://github.com/census-instrumentation/opencensus-specs/issues/12

bogdandrutu:
I think we are a stats/trace/debugging library and we should not crash user's binary for these kind of reasons.

songy23:
I think we're going to remove long and boolean typed tags, may be better to hold off this change after discussion on that.

sebright:
Yes, I'll also need to rebase this onto #710.

sebright:
I opened a new PR (#715) instead of rebasing this onto #710, since I needed to make significant changes.

PR #710, opened 2017-10-17
Remove support for multiple tag types.

sebright:
This commit removes the class hierarchies for TagKey, TagValue, and Tag.  The
three classes are now concrete AutoValue classes that use Strings for keys and
values.  The serialization format isn't affected, because it already only
supported string tags.

________________________________________

/cc @bogdandrutu @adriancole


core/src/main/java/io/opencensus/tags/TagContextBuilder.java
```diff
@@ -17,12 +17,6 @@
...
 import io.opencensus.common.Scope;
-import io.opencensus.tags.TagKey.TagKeyBoolean;
-import io.opencensus.tags.TagKey.TagKeyLong;
-import io.opencensus.tags.TagKey.TagKeyString;
-import io.opencensus.tags.TagValue.TagValueBoolean;
-import io.opencensus.tags.TagValue.TagValueLong;
-import io.opencensus.tags.TagValue.TagValueString;
 
 /** Builder for the {@link TagContext} class. */
 // TODO(sebright): Decide what to do when 'put' is called with a key that has the same name as an
```
songy23: This TODO can be removed now.

core/src/main/java/io/opencensus/tags/TagContextBuilder.java
```diff
@@ -36,25 +30,7 @@
    * @param value the value to set for the given key.
```
songy23: nit: s/value/{@code TagValue}

sebright: Done

core/src/main/java/io/opencensus/tags/TagContextBuilder.java
sebright: :+1:

bogdandrutu:
@dinooliva can you approve this if you think it is fine?

sebright:
I'll hold off merging this until we're sure that we don't need the other types of tags.

adriancole:
Fyi from a serialization pov. I know jonah from appD mentioned they have a
separate header for numeric tags

sebright:
Thanks!

PR #712, opened 2017-10-19
Add API for enabling/disabling stats collection (part of #608).

sebright:
This commit adds an enum that represents the state of the StatsComponent,
StatsCollectionState, and getters and setters for that state.  The state has no
effect yet.

____________________________________________________________

/cc @adriancole


sebright:
Thanks!

PR #713, opened 2017-10-20
Add a StackDriver exporter for stats using a Pull model

songy23:
Per our discussion yesterday, we're not going to support a push model for exporting ViewData in Stats library in the first release. Instead, the StackDriver exporter will be more like an independent agent: users need to register views against exporter first, and then the exporter will poll stats from the Stats library for all registered views and upload it to StackDriver periodically.

### DIFF
--- a/exporters/stats/stackdriver/README.md
+++ b/exporters/stats/stackdriver/README.md
@@ -35,6 +35,7 @@ This uses the default configuration for authentication and a given project ID.
 ```java
 public class MyMainClass {
   public static void main(String[] args) {
+    // Exporter will export to Stackdriver every 10 seconds.
     StackdriverStatsExporter.createWithProjectId("MyStackdriverProjectId", Duration.create(10, 0));
     StackdriverStatsExporter.registerView(myView);
   }
--- a/exporters/stats/stackdriver/README.md
+++ b/exporters/stats/stackdriver/README.md
@@ -1,0 +1,73 @@
+# OpenCensus Stackdriver Stats Exporter
+
+The *OpenCensus Stackdriver Stats Exporter* is a stats exporter that exports data to 
+Stackdriver Monitoring. [Stackdriver Monitoring][stackdriver-monitoring] provides visibility into 
+the performance, uptime, and overall health of cloud-powered applications. Stackdriver ingests that 
+data and generates insights via dashboards, charts, and alerts.
+
+## Quickstart
+
+### Prerequisites
+
+To use this exporter, you must have an application that you'd like to monitor. The app can be on 
+Google Cloud Platform, on-premise, or another cloud platform.
+
+In order to be able to push your stats to [Stackdriver Monitoring][stackdriver-monitoring], you must:
+
+1. [Create a Cloud project](https://support.google.com/cloud/answer/6251787?hl=en).
+2. [Enable billing](https://support.google.com/cloud/answer/6288653#new-billing).
+3. [Enable the Stackdriver Monitoring API](https://app.google.stackdriver.com/).
+
+These steps enable the API but don't require that your app is hosted on Google Cloud Platform.
+
+### Hello "Stackdriver Stats"
+
+#### Add the dependencies to your project
+
+For Maven add to your `pom.xml`: TODO
+
+For Gradle add to your dependencies: TODO
+
+#### Register the exporter
+
+This uses the default configuration for authentication and a given project ID.
+
+```java
+public class MyMainClass {
+  public static void main(String[] args) {
+    StackdriverStatsExporter.createWithProjectId("MyStackdriverProjectId", Duration.create(10, 0));
+    StackdriverStatsExporter.registerView(myView);
+  }
+}
+```
+
+#### Authentication
+
+This exporter uses [google-cloud-java](https://github.com/GoogleCloudPlatform/google-cloud-java),
+for details about how to configure the authentication see [here](https://github.com/GoogleCloudPlatform/google-cloud-java#authentication).
+
+If you prefer to manually set the credentials use:
+```
+StackdriverStatsExporter.createWithCredentialsAndProjectId(
+    new GoogleCredentials(new AccessToken(accessToken, expirationTime)), 
+    "MyStackdriverProjectId",
+    Duration.create(10, 0));
+```
+
+#### Specifying a Project ID
+
+This exporter uses [google-cloud-java](https://github.com/GoogleCloudPlatform/google-cloud-java),
+for details about how to configure the project ID see [here](https://github.com/GoogleCloudPlatform/google-cloud-java#specifying-a-project-id).
+
+If you prefer to manually set the project ID use:
+```
+StackdriverStatsExporter.createWithProjectId("MyStackdriverProjectId", Duration.create(10, 0));
+```
+
+#### Java Versions
+
+Java 7 or above is required for using this exporter.
+
+## FAQ
+
+[stackdriver-monitoring]: https://cloud.google.com/monitoring/
--- a/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExportUtils.java
+++ b/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExportUtils.java
@@ -107,9 +107,9 @@ final class StackdriverExportUtils {
   @VisibleForTesting
   static MetricKind createMetricKind(AggregationWindow window) {
     return window.match(
-        Functions.returnConstant(MetricKind.CUMULATIVE),               // Cumulative
+        Functions.returnConstant(MetricKind.CUMULATIVE), // Cumulative
         // TODO(songya): We don't support exporting Interval stats to StackDriver in this version.
-        Functions.returnConstant(MetricKind.UNRECOGNIZED),             // Interval
+        Functions.returnConstant(MetricKind.UNRECOGNIZED), // Interval
         Functions.returnConstant(MetricKind.UNRECOGNIZED));
   }
 
@@ -120,12 +120,12 @@ final class StackdriverExportUtils {
     return aggregation.match(
         Functions.returnConstant(
             measure.match(
-                Functions.returnConstant(MetricDescriptor.ValueType.DOUBLE),  // Sum Double
-                Functions.returnConstant(MetricDescriptor.ValueType.INT64),   // Sum Long
+                Functions.returnConstant(MetricDescriptor.ValueType.DOUBLE), // Sum Double
+                Functions.returnConstant(MetricDescriptor.ValueType.INT64), // Sum Long
                 Functions.returnConstant(MetricDescriptor.ValueType.UNRECOGNIZED))),
-        Functions.returnConstant(MetricDescriptor.ValueType.INT64),           // Count
-        Functions.returnConstant(MetricDescriptor.ValueType.DOUBLE),          // Mean
-        Functions.returnConstant(MetricDescriptor.ValueType.DISTRIBUTION),    // Distribution
+        Functions.returnConstant(MetricDescriptor.ValueType.INT64), // Count
+        Functions.returnConstant(MetricDescriptor.ValueType.DOUBLE), // Mean
+        Functions.returnConstant(MetricDescriptor.ValueType.DISTRIBUTION), // Distribution
         Functions.returnConstant(MetricDescriptor.ValueType.UNRECOGNIZED));
   }
 
@@ -166,8 +166,8 @@ final class StackdriverExportUtils {
         String.format("custom.googleapis.com/opencensus/%s", view.getName().asString()));
     Map<String, String> stringTagMap = Maps.newHashMap();
     List<TagKey> columns = view.getColumns();
-    checkArgument(tagValues.size() == columns.size(),
-        "TagKeys and TagValues don't have same size.");
+    checkArgument(
+        tagValues.size() == columns.size(), "TagKeys and TagValues don't have same size.");
     for (int i = 0; i < tagValues.size(); i++) {
       TagKey key = columns.get(i);
       TagValue value = tagValues.get(i);
@@ -249,7 +249,8 @@ final class StackdriverExportUtils {
         new Function<DistributionData, Void>() {
           @Override
           public Void apply(DistributionData arg) {
-            checkArgument(aggregation instanceof Aggregation.Distribution,
+            checkArgument(
+                aggregation instanceof Aggregation.Distribution,
                 "Aggregation and AggregationData mismatch.");
             builder.setDistributionValue(
                 createDistribution(
@@ -265,8 +266,7 @@ final class StackdriverExportUtils {
   @VisibleForTesting
   static Distribution createDistribution(
       DistributionData distributionData, BucketBoundaries bucketBoundaries) {
-    return Distribution
-        .newBuilder()
+    return Distribution.newBuilder()
         .setBucketOptions(createBucketOptions(bucketBoundaries))
         .addAllBucketCounts(distributionData.getBucketCounts())
         .setCount(distributionData.getCount())
@@ -283,23 +283,19 @@ final class StackdriverExportUtils {
   // Create BucketOptions from BucketBoundaries
   @VisibleForTesting
   static BucketOptions createBucketOptions(BucketBoundaries bucketBoundaries) {
-    return BucketOptions
-        .newBuilder()
-        .setExplicitBuckets(
-            Explicit.newBuilder().addAllBounds(bucketBoundaries.getBoundaries()))
+    return BucketOptions.newBuilder()
+        .setExplicitBuckets(Explicit.newBuilder().addAllBounds(bucketBoundaries.getBoundaries()))
         .build();
   }
 
   // Convert a Census Timestamp to a StackDriver Timestamp
   @VisibleForTesting
   static Timestamp convertTimestamp(io.opencensus.common.Timestamp censusTimestamp) {
-    return Timestamp
-        .newBuilder()
+    return Timestamp.newBuilder()
         .setSeconds(censusTimestamp.getSeconds())
         .setNanos(censusTimestamp.getNanos())
         .build();
   }
 
-  private StackdriverExportUtils() {
-  }
+  private StackdriverExportUtils() {}
 }
--- a/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExportUtils.java
+++ b/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExportUtils.java
@@ -55,9 +55,7 @@ import io.opencensus.stats.ViewData.AggregationWindowData;
 import io.opencensus.stats.ViewData.AggregationWindowData.CumulativeData;
 import io.opencensus.stats.ViewData.AggregationWindowData.IntervalData;
 import io.opencensus.tags.TagKey;
-import io.opencensus.tags.TagKey.TagKeyString;
 import io.opencensus.tags.TagValue;
-import io.opencensus.tags.TagValue.TagValueString;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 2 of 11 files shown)


exporters/stats_stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporter.java
```diff
@@ -0,0 +1,269 @@
...
+
+  /**
+   * Register a {@link View} against this exporter, and upload it as a {@link MetricDescriptor} to
+   * StackDriver.
+   *
+   * @param view the {@code View} to be registered.
+   */
+  public void registerView(View view) {
+    workerThread.registerView(view);
+    // This is required only for custom view definitions. Canonical views should be pre-registered.
```
bogdandrutu: How do we enforce this?

exporters/stats_stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExportUtils.java
```diff
@@ -0,0 +1,331 @@
...
+    builder.setValueType(createValueType(view.getAggregation(), view.getMeasure()));
+    return builder.build();
+  }
+
+  // Construct a name for a MetricDescriptor
+  // Name format refers to cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.metricDescriptors
+  private static String createMetricDescriptorName(
+      String projectId, String viewName) {
+    return String.format(
+        "projects/%s/metricDescriptors/custom.googleapis.com%%2Fcensus%%2F%s%%2F%s",
```
bogdandrutu: See here for name:
https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.metricDescriptors/create

return String.format(projects/%s, project_id) -> also you can cache this because it is constant

exporters/stats_stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExportUtils.java
```diff
@@ -0,0 +1,331 @@
...
+        projectId, NAMESPACE, viewName);
+  }
+
+  // Construct a MetricDescriptor.Type. From the documentation:
+  // cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.metricDescriptors
+  // For now only support custom metrics.
+  private static String createMetricType(String viewName) {
+    return String.format("custom.googleapis.com/census/%s/%s", NAMESPACE, viewName);
```
bogdandrutu: "custom.googleapis.com/opencensus/%s", viewName?

exporters/stats_stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExportUtils.java
```diff
@@ -0,0 +1,331 @@
...
+    }
+
+    MetricDescriptor.Builder builder = MetricDescriptor.newBuilder();
+    String name = view.getName().asString();
+    builder.setName(createMetricDescriptorName(projectId, name));
+    builder.setType(createMetricType(name));
+    builder.setDescription(view.getDescription());
+    builder.setUnit(view.getMeasure().getUnit());
+    // TODO(songya): do we want a different display name?
+    builder.setDisplayName(NAMESPACE + "/" + name);
```
bogdandrutu: Better to use:
"Opencensus " + name

songy23: Sounds good, updated.

exporters/stats_stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExportUtils.java
songy23: Done.

exporters/stats_stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExportUtils.java
songy23: Updated

exporters/stats_stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporter.java
songy23: We can check if the view is one of [RpcViewConstants](https://github.com/census-instrumentation/opencensus-java/blob/master/core/src/main/java/io/opencensus/stats/RpcViewConstants.java), if so then it doesn't need to be posted to StackDriver. This is not needed for now, since we haven't defined canonical metrics on StackDriver. I've added a TODO here.

adriancole:
Curious.. are you thinking the same for trace data?

songy23:
@adriancole probably not, since views need to be registered as metrics first in order to export stats to Stackdriver, while span data do not. In addition, we're considering modifying Stats APIs to support a push model for stats and having an export component like trace. That's probably for next release though.

exporters/stats_stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExportUtils.java
```diff
@@ -0,0 +1,322 @@
...
+    builder.setValueType(valueType);
+    return builder.build();
+  }
+
+  // Construct a MetricKind from an AggregationWindow
+  @VisibleForTesting
+  static MetricKind createMetricKind(AggregationWindow window) {
+    return window.match(
+        Functions.returnConstant(MetricKind.CUMULATIVE),               // Cumulative
+        // We don't support exporting Interval stats to StackDriver in v0.1
```
bogdandrutu: I would rephrase it as:
// We don't support exporting Interval stats to StackDriver in this version.

Same everywhere. The current version of the library is 0.8.0 I know that this is the first stats release but it is confusing.

exporters/stats_stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporter.java
```diff
@@ -0,0 +1,271 @@
...
+      }
+    }
+
+    // Polls ViewData from Stats library for all registered views, and upload them as TimeSeries to
+    // StackDriver.
+    private void export() {
+      CreateTimeSeriesRequest.Builder builder = CreateTimeSeriesRequest.newBuilder();
+      synchronized (monitor) {
+        for (View.Name view : registeredViews.keySet()) {
+          builder.addAllTimeSeries(createTimeSeriesList(viewManager.getView(view), projectId));
```
bogdandrutu: To avoid holding the lock for too long better to just extract the ViewData, add to a list then outside the lock create the proto.

exporters/stats_stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExportUtils.java
```diff
@@ -0,0 +1,322 @@
...
+    builder.setMetricKind(createMetricKind(view.getWindow()));
+    builder.setValueType(createValueType(view.getAggregation(), view.getMeasure()));
+    return builder.build();
+  }
+
+  // Construct a MetricDescriptor.Type. From the documentation:
+  // cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.metricDescriptors
+  // For now only support custom metrics.
+  private static String createMetricType(String viewName) {
+    return String.format("custom.googleapis.com/opencensus/%s", viewName);
```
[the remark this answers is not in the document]
songy23: Done.

exporters/stats_stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExportUtils.java
songy23: Done.

exporters/stats_stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporter.java
songy23: Updated

exporters/stats_stackdriver/src/test/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporterTest.java
```diff
@@ -0,0 +1,208 @@
...
+        CumulativeData.create(Timestamp.fromMillis(100), Timestamp.fromMillis(200)));
+    doNothing().when(mockViewManager).registerView(view);
+    doReturn(viewData).when(mockViewManager).getView(VIEW_NAME);
+
+    StackdriverExporter exporter = new StackdriverExporter(
+        PROJECT_ID, new FakeMetricServiceClient(mockStub), ONE_SECOND, mockViewManager);
+    exporter.registerView(view);
+
+    verify(mockStub, times(1)).createMetricDescriptorCallable();
+    verify(mockStub, timeout(1000).times(1)).createTimeSeriesCallable();
```
sebright: Does the timeout need to match the export interval for this test to pass?  It would help to have a comment explaining the timing in this test.

exporters/stats_stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporter.java
```diff
@@ -0,0 +1,277 @@
...
+      MetricServiceClient metricServiceClient,
+      Duration exportInterval,
+      ViewManager viewManager) {
+    checkNotNull(exportInterval, "exportInterval");
+    checkArgument(exportInterval.compareTo(ZERO) > 0, "Duration must be positive");
+    this.projectId = checkNotNull(projectId, "projectId");
+    this.metricServiceClient = metricServiceClient;
+    this.workerThread = new WorkerThread(
+        projectId, metricServiceClient, exportInterval, viewManager);
+    workerThread.start();
```
sebright: It would be better to not start the thread in the constructor, because the "this" reference could escape from the constructor if WorkerThread were ever made non-static.  You could start it in `createInternal` instead.

exporters/stats_stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporter.java
```diff
@@ -0,0 +1,277 @@
...
+   * Creates a StackdriverExporter for an explicit project ID and using explicit credentials.
+   *
+   * <p>Only one Stackdriver exporter can be created at any point.
+   *
+   * @param credentials a credentials used to authenticate API calls.
+   * @param projectId the cloud project id.
+   * @param exportInterval the interval to export stats.
+   * @throws IllegalStateException if a Stackdriver exporter already exists.
+   */
+  public static StackdriverExporter createWithCredentialsAndProjectId(
```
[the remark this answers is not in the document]
sebright: I thought about this a bit more, and I think I prefer making `registerView` static, if it is not possible to export to StackDriver with two different values for credentials, project ID, and time interval.  An instance method could cause users to think that they can register different sets of views on different instances of StackdriverExporter.  A static method makes it clear that registered views are global state.

exporters/stats_stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporter.java
```diff
@@ -0,0 +1,206 @@
...
+
+  /**
+   * Register a {@link View} against this exporter, and upload it as a {@link MetricDescriptor} to
+   * StackDriver.
+   *
+   * @param view the {@code View} to be registered.
+   * @throws IllegalStateException if a Stackdriver stats exporter has not been created yet.
+   */
+  public static synchronized void registerView(View view) {
+    synchronized (monitor) {
```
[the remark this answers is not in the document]
songy23: Done.

bogdandrutu:
Very small request. Probably better to move the exporter into:
exporters/stats/stackdriver instead of exporters/stats_stackdriver

See this PR https://github.com/census-instrumentation/opencensus-java/pull/719 for trace.

songy23:
Sure, updated.

exporters/stats/stackdriver/README.md
```diff
@@ -0,0 +1,73 @@
...
+For Gradle add to your dependencies: TODO
+
+#### Register the exporter
+
+This uses the default configuration for authentication and a given project ID.
+
+```java
+public class MyMainClass {
+  public static void main(String[] args) {
+    StackdriverStatsExporter.createWithProjectId("MyStackdriverProjectId", Duration.create(10, 0));
```
adriancole: Add comment or otherwise what the duration is.. ex if interval do

// install the exporter with a 10 second window for aggregation (or whatever it is)

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExportUtils.java
```diff
@@ -0,0 +1,301 @@
...
+        // TODO(songya): We don't support exporting Interval stats to StackDriver in this version.
+        Functions.returnConstant(MetricKind.UNRECOGNIZED), // Interval
+        Functions.returnConstant(MetricKind.UNRECOGNIZED));
+  }
+
+  // Construct a MetricDescriptor.ValueType from an Aggregation and a Measure
+  @VisibleForTesting
+  static MetricDescriptor.ValueType createValueType(
+      Aggregation aggregation, final Measure measure) {
+    return aggregation.match(
```
adriancole: Maybe revisit this with a builder or some fluent api as this looks cryptic in practice

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExportUtils.java
```diff
@@ -0,0 +1,301 @@
...
+  static Point createPoint(
+      AggregationData aggregationData, AggregationWindowData windowData, Aggregation aggregation) {
+    Point.Builder builder = Point.newBuilder();
+    builder.setInterval(createTimeInterval(windowData));
+    builder.setValue(createTypedValue(aggregation, aggregationData));
+    return builder.build();
+  }
+
+  // Convert AggregationWindowData to TimeInterval, currently only support CumulativeData.
+  @VisibleForTesting
```
adriancole: Visible for test or only used in test? If not only used in tests avoid anonymous types (extract or constant and/or declare class with nice tostring for debug)

adriancole: If only used in test move to test tree

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporter.java
```diff
@@ -0,0 +1,205 @@
...
+   *
+   * @param exportInterval the interval between pushing stats to StackDriver.
+   * @throws IllegalStateException if a Stackdriver exporter is already created.
+   */
+  public static void create(Duration exportInterval) throws IOException {
+    checkNotNull(exportInterval, "exportInterval");
+    createInternal(null, ServiceOptions.getDefaultProjectId(), exportInterval);
+  }
+
+  private static void createInternal(
```
adriancole: Comment create internal instead of just hidden ctor to enforce singleton

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporter.java
```diff
@@ -0,0 +1,205 @@
...
+      @Nullable Credentials credentials, String projectId, Duration exportInterval)
+      throws IOException {
+    synchronized (monitor) {
+      checkState(exporter == null, "Stackdriver stats exporter is already created.");
+      MetricServiceClient metricServiceClient;
```
adriancole: Unless this has side effect or we expect tons of calls i would initialize this outside the lock. If there is a good reason to init this inside lock comment why

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporter.java
```diff
@@ -0,0 +1,205 @@
...
+                    .build());
+      }
+      exporter =
+          new StackdriverStatsExporter(
+              projectId, metricServiceClient, exportInterval, Stats.getViewManager());
+      exporter.workerThread.start();
+    }
+  }
+
+  // Method for setting exporter to a fake exporter or null (reset) for unit tests.
```
adriancole: One crap thing about visible for tests is hard to tell only for tests or visible also for tests. I will just mention anyting that doesnt need to go in src/main should not be in src/main. Usually i put any heavy logic in test tree and only poke holes or allow minimum hooks in main. That may be the case here

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/WorkerThread.java
```diff
@@ -0,0 +1,114 @@
...
+import io.opencensus.stats.View;
+import io.opencensus.stats.ViewData;
+import io.opencensus.stats.ViewManager;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import javax.annotation.concurrent.GuardedBy;
+
+/** Worker {@code Thread} that polls ViewData from Stats library and batch export to StackDriver. */
+final class WorkerThread extends Thread {
```
adriancole: If this is to be run as daemon mention? Otherwise say something like interrupt to stop loop (as there is no other way to cancel implemented here).

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporter.java
songy23: Sorry for the confusion. This method is only for tests, however it cannot go into src/tests since it needs to access a private field, which is also guarded by a private lock of this class.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/WorkerThread.java
songy23: It's a daemon thread. Added comment here.

songy23:
Per our discussion today, since we'll probably update to use a push model for `ViewData`, it's better not to have `registerView()` at the exporter level for backwards compatibility. Instead we'll need an additional field in `View` that determine whether it should be local or exported to exporters.

For this version, since Stackdriver exporter is using a pull model, we'll need another API `getExportedViews` in our library. Exporter needs to call `getExportedViews` to get those `ViewData` that should be exported, instead of registering them against the exporter.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporter.java
dinooliva: Does it make sense to call it unsafeSetExporter() then?

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/WorkerThread.java
```diff
@@ -0,0 +1,118 @@
...
+      String projectId,
+      MetricServiceClient metricServiceClient,
+      Duration exportInterval,
+      ViewManager viewManager) {
+    this.scheduleDelayMillis = toMillis(exportInterval);
+    this.projectId = projectId;
+    this.metricServiceClient = metricServiceClient;
+    this.viewManager = viewManager;
+    setDaemon(true);
+    setName("ExportWorkerThread");
```
dinooliva: Consider "StackdriverExporterWorkerThread" in case there are ever other exporters.

songy23: Done.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporter.java
songy23: Sounds better, updated.

songy23:
Thanks for review!

PR #714, opened 2017-10-20
Start adding a new library containing RPC constants for gRPC (fixes #705).

### LINKED ISSUE #705: Move RPC measure and view constants out of the API library.
The RPC constants aren't needed for the core stats collection functionality, so it would be better to put them in a separate library.

sebright:
This commit moves RpcMeasureConstants.java and RpcViewConstants.java into the
new package so that they are no longer part of the core API. The new package is
called opencensus-grpc-rpc-constants.  It isn't enabled for release yet.

/cc @zhangkun83


bogdandrutu:
High level comment. I would put this under contrib/rpc_constants directory.

sebright:
@bogdandrutu Do you think that the artifact name is okay, and I should only change the directory?

bogdandrutu:
@sebright I would change that name as well to something like opencensus-contrib-rpc-constants. One of the reason is that the constants can be used for other RPC libraries like thrift.

sebright:
The measure and view names contain "grpc", so I don't think that they could be used by other RPC frameworks.  @dinooliva mentioned that we'll probably move the constants into gRPC eventually.

bogdandrutu:
Then maybe name it contrib/grpc_metrics?

sebright:
I moved the library into contrib/grpc_metrics and renamed it to `opencensus-contrib-grpc-metrics`.

bogdandrutu:
Just minor maybe add a note in the README that this may be moved in the grpc repo.

sebright:
@bogdandrutu Thanks for the suggestion.  I added a note.

Issue #709, opened 2017-10-17
Record "g.co/agent" label for Stackdriver Trace emitter


draffensperger:
Could we record a "g.co/agent" label for traces written to Stackdriver Trace with a value of "opencensus-java [library version]"?

That allows users (and Stackdriver Trace tech support / application developers) to associate spans back to what instrumentation library generated them.

PR #717, opened 2017-10-22
Record the agent label for stackdriver trace exporter.

bogdandrutu:
Fixes https://github.com/census-instrumentation/opencensus-java/issues/709


exporters/trace_stackdriver/src/main/java/io/opencensus/exporter/trace/stackdriver/StackdriverV1ExporterHandler.java
```diff
@@ -58,6 +59,8 @@
   private static final String STATUS_DESCRIPTION = "g.co/status/description";
   private static final String ANNOTATION_LABEL = "ANNOTATION-";
   private static final String NETWORK_EVENT_LABEL = "NETWORK-";
+  private static final String agentLabelKey = "g.co/agent";
```
songy23: Please fix the constant names

api/src/main/java/io/opencensus/common/Constants.java
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.common;
+
+/**
+ * Class holder for all common constants for the OpenCensus Java library.
+ */
+public final class Constants {
```
sebright: What other constants will go in this class?  Is Constants.VERSION meant to be used by all users?

bogdandrutu: Moved to internal and add @Internal annotation.

exporters/trace_stackdriver/src/main/java/io/opencensus/exporter/trace/stackdriver/StackdriverV1ExporterHandler.java
bogdandrutu: Done.

bogdandrutu:
PTAL

api/src/main/java/io/opencensus/common/Constants.java
sebright: I thought that we previously only used `Internal` for elements that were internal to the API library, to indicate that they weren't stable.  Since this field needs to be backwards compatible for the exporter, I would rather make it part of the API, but maybe change the class name.  I've seen other libraries expose a version constant.

bogdandrutu:
@sebright I picked Constants because we may have different other constants to add in the future. Also I marked the class as Experimental for the moment.

RELEASING.md
```diff
@@ -64,6 +64,7 @@ token](https://help.github.com/articles/creating-a-personal-access-token-for-the
     $ MAJOR=0 MINOR=4 PATCH=0 # Set appropriately for new release
     $ VERSION_FILES=(
       build.gradle
+      api/src/main/java/io/opencensus/common/internal/Constants.java
```
sebright: I think this is out of date.

bogdandrutu: Done.

PR #721, opened 2017-10-23
Make DistributionData.create take a List instead of a long[] for bucket counts.

sebright:
`List<Long>` is consistent with the return type of
`DistributionData.getBucketCounts()`.

### DIFF
--- a/core_impl/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
+++ b/core_impl/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
@@ -531,8 +531,12 @@ abstract class MutableViewData {
       implements Function<MutableDistribution, AggregationData> {
     @Override
     public AggregationData apply(MutableDistribution arg) {
+      List<Long> boxedBucketCounts = new ArrayList<Long>();
+      for (long bucketCount : arg.getBucketCounts()) {
+        boxedBucketCounts.add(bucketCount);
+      }
       return DistributionData.create(arg.getMean(), arg.getCount(), arg.getMin(), arg.getMax(),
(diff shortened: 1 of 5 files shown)


bogdandrutu:
Please rebase this PR.

sebright:
I rebased.

core_impl/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
```diff
@@ -531,8 +531,12 @@ public AggregationData apply(MutableMean arg) {
       implements Function<MutableDistribution, AggregationData> {
     @Override
     public AggregationData apply(MutableDistribution arg) {
+      List<Long> boxedBucketCounts = new ArrayList<Long>();
+      for (long bucketCount : arg.getBucketCounts()) {
+        boxedBucketCounts.add(bucketCount);
+      }
       return DistributionData.create(arg.getMean(), arg.getCount(), arg.getMin(), arg.getMax(),
-          arg.getSumOfSquaredDeviations(), arg.getBucketCounts());
+          arg.getSumOfSquaredDeviations(), boxedBucketCounts);
```
dinooliva: Doesn't MutableDistribution.getBucketCounts() return a List<Long>? Seems like we're going to do a lot of unnecessary copying (and boxing/unboxing).

sebright: MutableDistribution.getBucketCounts() returns a long[], but we could change it since it isn't part of the API.  I thought it was more important to avoid the boxing/unboxing when stats are recorded and aggregated in MutableDistribution than when we return the aggregated data in a DistributionData, since stats should be queried much less frequently.

PR #724, opened 2017-10-24
Adds TextFormat and usage example.

bogdandrutu:
Based on @adriancole work from here https://github.com/census-instrumentation/opencensus-java/pull/667

### DIFF
--- a/api/src/main/java/io/opencensus/trace/propagation/TextFormat.java
+++ b/api/src/main/java/io/opencensus/trace/propagation/TextFormat.java
@@ -45,7 +45,7 @@ import javax.annotation.Nullable;
  *   }
  * }
  *
- * void makeHttpRequest() {
+ * void makeHttpRequest(HttpURLConnection connection) {
  *   Span span = tracer.spanBuilder("Sent.MyRequest").startSpan();
  *   try (Scope s = tracer.withSpan(span)) {
  *     HttpURLConnection connection =
@@ -62,10 +62,10 @@ import javax.annotation.Nullable;
  * <pre>{@code
  * private static final Tracer tracer = Tracing.getTracer();
  * private static final TextFormat textFormat = Tracing.getPropagationComponent().getTextFormat();
- * private static final TextFormat.Getter getter = ...;
+ * private static final TextFormat.Getter<HttpRequest> getter = ...;
  *
- * void onRequestReceived() {
- *   SpanContext spanContext = textFormat.extract(carrier, getter);
+ * void onRequestReceived(HttpRequest request) {
+ *   SpanContext spanContext = textFormat.extract(request, getter);
  *   Span span = tracer.spanBuilderWithRemoteParent("Recv.MyRequest", spanContext).startSpan();
  *   try (Scope s = tracer.withSpan(span)) {
  *     // Handle request and send response back.
@@ -101,7 +101,7 @@ public abstract class TextFormat {
   public abstract <C> void inject(SpanContext spanContext, C carrier, Setter<C> setter);
 
   /**
-   * Class that allows to be saved as a constant to avoid runtime allocations.
+   * {@code Setter} is stateless and allows to be saved as a constant to avoid runtime allocations.
    *
    * @param <C> carrier of propagation fields, such as an http request
    */
@@ -123,9 +123,6 @@ public abstract class TextFormat {
   /**
    * Extracts the span context from upstream. For example, as http headers.
    *
-   * <p>If no particular format is known to be always received on from the wire it is recommended to
-   * use {@code Encoding.ALL}.
-   *
    * @param carrier holds propagation fields. For example, an outgoing message or http request.
    * @param getter invoked for each propagation key to get.
    * @throws SpanContextParseException if the input is invalid
@@ -134,7 +131,7 @@ public abstract class TextFormat {
       throws SpanContextParseException;
 
   /**
-   * Class that allows to be saved as a constant to avoid runtime allocations.
+   * {@code Getter} is stateless and allows to be saved as a constant to avoid runtime allocations.
    *
    * @param <C> carrier of propagation fields, such as an http request
    */
@@ -172,11 +169,13 @@ public abstract class TextFormat {
     @Override
     public <C> void inject(SpanContext spanContext, C carrier, Setter<C> setter) {
       checkNotNull(spanContext, "spanContext");
+      checkNotNull(carrier, "carrier");
       checkNotNull(setter, "setter");
     }
 
     @Override
     public <C> SpanContext extract(C carrier, Getter<C> getter) {
+      checkNotNull(carrier, "carrier");
       checkNotNull(getter, "getter");
       return SpanContext.INVALID;
     }
--- a/api/src/main/java/io/opencensus/trace/propagation/TextFormat.java
+++ b/api/src/main/java/io/opencensus/trace/propagation/TextFormat.java
@@ -1,0 +1,202 @@
+/*
+ * Copyright 2017, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.trace.propagation;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import io.opencensus.common.ExperimentalApi;
+import io.opencensus.trace.SpanContext;
+import java.util.Collections;
+import java.util.List;
+import javax.annotation.Nullable;
+
+/**
+ * Injects and extracts {@link SpanContext trace identifiers} as text into carriers that travel
+ * in-band across process boundaries. Identifiers are often encoded as messaging or RPC request
+ * headers.
+ *
+ * <p>When using http, the carrier of propagated data on both the client (injector) and server
+ * (extractor) side is usually an http request. Propagation is usually implemented via library-
+ * specific request interceptors, where the client-side injects span identifiers and the server-side
+ * extracts them.
+ *
+ * <p>Example of usage on the client:
+ *
+ * <pre>{@code
+ * private static final Tracer tracer = Tracing.getTracer();
+ * private static final TextFormat textFormat = Tracing.getPropagationComponent().getTextFormat();
+ * private static final TextFormat.Setter setter = new TextFormat.Setter<HttpURLConnection>() {
+ *   public void put(HttpURLConnection carrier, String key, String value, Encodings.B3_FORMAT) {
+ *     carrier.setRequestProperty(field, value);
+ *   }
+ * }
+ *
+ * void makeHttpRequest() {
+ *   Span span = tracer.spanBuilder("Sent.MyRequest").startSpan();
+ *   try (Scope s = tracer.withSpan(span)) {
+ *     HttpURLConnection connection =
+ *         (HttpURLConnection) new URL("http://myserver").openConnection();
+ *     textFormat.inject(span.getContext(), connection, httpURLConnectionSetter);
+ *     // Send the request, wait for response and maybe set the status if not ok.
+ *   }
+ *   span.end();  // Can set a status.
+ * }
+ * }</pre>
+ *
+ * <p>Example of usage on the server:
+ *
+ * <pre>{@code
+ * private static final Tracer tracer = Tracing.getTracer();
+ * private static final TextFormat textFormat = Tracing.getPropagationComponent().getTextFormat();
+ * private static final TextFormat.Getter getter = ...;
+ *
+ * void onRequestReceived() {
+ *   SpanContext spanContext = textFormat.extract(carrier, getter, Encodings.ALL);
+ *   Span span = tracer.spanBuilderWithRemoteParent("Recv.MyRequest", spanContext).startSpan();
+ *   try (Scope s = tracer.withSpan(span)) {
+ *     // Handle request and send response back.
+ *   }
+ *   span.end()
+ * }
+ * }</pre>
+ */
+@ExperimentalApi
+public abstract class TextFormat {
+  private static final NoopTextFormat NOOP_TEXT_FORMAT = new NoopTextFormat();
+
+  public enum Encoding {
+    /**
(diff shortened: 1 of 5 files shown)


api/src/main/java/io/opencensus/trace/propagation/TextFormat.java
```diff
@@ -0,0 +1,202 @@
...
+ *   }
+ *   span.end()
+ * }
+ * }</pre>
+ */
+@ExperimentalApi
+public abstract class TextFormat {
+  private static final NoopTextFormat NOOP_TEXT_FORMAT = new NoopTextFormat();
+
+  public enum Encoding {
```
adriancole: bikeshed: encoding vs format? I think encoding like binary and format like B3

api/src/test/java/io/opencensus/trace/propagation/TextFormatTest.java
```diff
@@ -0,0 +1,105 @@
...
+@RunWith(JUnit4.class)
+public class TextFormatTest {
+  private static final TextFormat textFormat = TextFormat.getNoopTextFormat();
+
+  @Test(expected = NullPointerException.class)
+  public void inject_NullSpanContext() {
+    textFormat.inject(
+        null,
+        new Object(),
+        new Setter<Object>() {
```
adriancole: we need to change test source to java 8 so it can use lambda :)

api/src/main/java/io/opencensus/trace/propagation/TextFormat.java
```diff
@@ -0,0 +1,202 @@
...
+    /**
+     * If used in the {@link #extract(Object, Getter, Encoding)} API the implementation will try to
+     * extract the headers from the carrier in the following order: TBD. If used in the
+     * {@link #inject(SpanContext, Object, Setter, Encoding)} then all available
+     * encodings will be used.
+     */
+    ALL,
```
adriancole: upside and downside of using enum is that it is a registry you cannot change. also if there are 3 encodings, ALL is possibly good for extract might not make sense for inject.

I read this as the required encodings to support.. if we expect users to change encodings on the fly (without a change to the registry), the supporting an override with a required encoding can make sense. OTOH, I think there should be an implicit encoding setup at the time the tracer is..

api/src/main/java/io/opencensus/trace/propagation/TextFormat.java
```diff
@@ -0,0 +1,202 @@
...
+
+  /**
+   * Injects the span context downstream. For example, as http headers.
+   *
+   * @param spanContext possibly not sampled.
+   * @param carrier holds propagation fields. For example, an outgoing message or http request.
+   * @param setter invoked for each propagation key to add or remove.
+   * @param encoding the encoding to be used to inject the {@code SpanContext} to the carrier.
+   */
+  public abstract <C> void inject(
```
adriancole: curious if we can't get by with tracer-scoped configuration, so that instrumentation don't have to care about which  encoding is used. for example, we don't need "ALL" as you can configure a composite inject/extract thing which can accomplish the same.

bogdandrutu:
PTAL. Changed based on an offline discussion.

api/src/test/java/io/opencensus/trace/propagation/TextFormatTest.java
bogdandrutu: Agreed.

api/src/main/java/io/opencensus/trace/propagation/TextFormat.java
```diff
@@ -0,0 +1,184 @@
...
+ *
+ * <p>Example of usage on the server:
+ *
+ * <pre>{@code
+ * private static final Tracer tracer = Tracing.getTracer();
+ * private static final TextFormat textFormat = Tracing.getPropagationComponent().getTextFormat();
+ * private static final TextFormat.Getter getter = ...;
+ *
+ * void onRequestReceived() {
+ *   SpanContext spanContext = textFormat.extract(carrier, getter);
```
sebright: Should `carrier` be passed in as an argument?

api/src/main/java/io/opencensus/trace/propagation/TextFormat.java
```diff
@@ -0,0 +1,184 @@
...
+     * @param value the value of the field.
+     */
+    public abstract void put(C carrier, String key, String value);
+  }
+
+  /**
+   * Extracts the span context from upstream. For example, as http headers.
+   *
+   * <p>If no particular format is known to be always received on from the wire it is recommended to
+   * use {@code Encoding.ALL}.
```
sebright: > `If no particular format is known to be always received on from the wire it is recommended to
    use {@code Encoding.ALL}.`

Is this left over from the previous design?

api/src/main/java/io/opencensus/trace/propagation/TextFormat.java
```diff
@@ -0,0 +1,184 @@
...
+   * before calling {@link #inject(SpanContext, Object, Setter)}.
+   *
+   * <p>For example, if the carrier is a single-use or immutable request object, you don't need to
+   * clear fields as they couldn't have been set before. If it is a mutable, retryable object,
+   * successive calls should clear these fields first.
+   */
+  // The use cases of this are:
+  // * allow pre-allocation of fields, especially in systems like gRPC Metadata
+  // * allow a single-pass over an iterator (ex OpenTracing has no getter in TextMap)
+  public abstract List<String> fields();
```
sebright: Does order matter? Should this return a Set?

api/src/main/java/io/opencensus/trace/propagation/TextFormat.java
```diff
@@ -0,0 +1,184 @@
...
+   * Injects the span context downstream. For example, as http headers.
+   *
+   * @param spanContext possibly not sampled.
+   * @param carrier holds propagation fields. For example, an outgoing message or http request.
+   * @param setter invoked for each propagation key to add or remove.
+   */
+  public abstract <C> void inject(SpanContext spanContext, C carrier, Setter<C> setter);
+
+  /**
+   * Class that allows to be saved as a constant to avoid runtime allocations.
```
sebright: I don't really understand this phrase.  Is this accurate?  "`Setter for arbitrary fields in {@code C}. An instance can be saved as a constant to avoid runtime allocations.`"  The same applies to `Getter`.

api/src/main/java/io/opencensus/trace/propagation/TextFormat.java
```diff
@@ -0,0 +1,184 @@
...
+
+    @Override
+    public <C> void inject(SpanContext spanContext, C carrier, Setter<C> setter) {
+      checkNotNull(spanContext, "spanContext");
+      checkNotNull(setter, "setter");
+    }
+
+    @Override
+    public <C> SpanContext extract(C carrier, Getter<C> getter) {
+      checkNotNull(getter, "getter");
```
sebright: Should this method also check the carrier for null?  If it can be null, then I think it would help to mention what can be null in the Javadocs above.

api/src/test/java/io/opencensus/trace/propagation/TextFormatTest.java
```diff
@@ -0,0 +1,75 @@
...
+import io.opencensus.trace.propagation.TextFormat.Getter;
+import io.opencensus.trace.propagation.TextFormat.Setter;
+import javax.annotation.Nullable;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/** Unit tests for {@link TextFormat}. */
+@RunWith(JUnit4.class)
+public class TextFormatTest {
```
sebright: Should this be NoopTextFormatTest?

api/src/main/java/io/opencensus/trace/propagation/TextFormat.java
```diff
@@ -0,0 +1,184 @@
...
+ *   Span span = tracer.spanBuilderWithRemoteParent("Recv.MyRequest", spanContext).startSpan();
+ *   try (Scope s = tracer.withSpan(span)) {
+ *     // Handle request and send response back.
+ *   }
+ *   span.end()
+ * }
+ * }</pre>
+ */
+@ExperimentalApi
+public abstract class TextFormat {
```
sebright: Are you planning to add this to the PropagationComponent?

sebright: Nevermind, I saw your previous comment.

api/src/main/java/io/opencensus/trace/propagation/TextFormat.java
adriancole: yeah carrier shouldn't be allowed to be null

api/src/main/java/io/opencensus/trace/propagation/TextFormat.java
adriancole: personally not too bothered, but list is more efficient (you can iterate w/o creating an iterator). I don't tend to use set unless set semantics are important

api/src/main/java/io/opencensus/trace/propagation/TextFormat.java
adriancole: carrier in this case is usually a request object (or headers object). what alternative to passing as an argument did you have in mind?

api/src/main/java/io/opencensus/trace/propagation/TextFormat.java
adriancole: setters and getters are stateless, so can be shared across invocations. This makes them a good choice to be constants (avoiding allocating an extra object per-request)

Ex. here's a similar usage that defines getters/setters for kafka
```java
final class KafkaPropagation {

  static final Charset UTF_8 = Charset.forName("UTF-8");

  static final Setter<Headers, String> HEADER_SETTER = (carrier, key, value) -> {
    carrier.remove(key);
    carrier.add(key, value.getBytes(UTF_8));
  };

  static final Getter<Headers, String> HEADER_GETTER = (carrier, key) -> {
    Header header = carrier.lastHeader(key);
    if (header == null) return null;
    return new String(header.value(), UTF_8);
  };

  KafkaPropagation() {
  }
}
```

api/src/main/java/io/opencensus/trace/propagation/TextFormat.java
adriancole: ack leftover

api/src/main/java/io/opencensus/trace/propagation/TextFormat.java
sebright: I expected `carrier` to be passed in as an argument to onRequestReceived, but onRequestReceived doesn't take any arguments in this example.

api/src/main/java/io/opencensus/trace/propagation/TextFormat.java
sebright: Thanks.  That makes sense.  I think that the Javadoc is just a little unclear.

api/src/main/java/io/opencensus/trace/propagation/TextFormat.java
sebright: I suggested Set because it's a public method, so the return type helps document it.  I think that List usually implies that order matters.  I didn't consider that users would try to avoid allocating an iterator.  Do you think that would have a significant performance impact here?

api/src/main/java/io/opencensus/trace/propagation/TextFormat.java
adriancole: agreed javadoc should be corrected to pass request as a parameter and use the name request instead of the word carried in textFormat.extract( arg

api/src/main/java/io/opencensus/trace/propagation/TextFormat.java
adriancole: we are bordering on bikeshed, but I summarily use list unless there's a good reason not to, as they are cheap. The common case of this is clearing keys on a mutable input. at least in android, folks tend to avoid allocating objects for iteration

api/src/main/java/io/opencensus/trace/propagation/TextFormat.java
bogdandrutu: Done.

api/src/main/java/io/opencensus/trace/propagation/TextFormat.java
bogdandrutu: ACK

api/src/main/java/io/opencensus/trace/propagation/TextFormat.java
bogdandrutu: Updated.

api/src/main/java/io/opencensus/trace/propagation/TextFormat.java
bogdandrutu: Done. removed.

api/src/main/java/io/opencensus/trace/propagation/TextFormat.java
bogdandrutu: Done.

api/src/test/java/io/opencensus/trace/propagation/TextFormatTest.java
bogdandrutu: We don't name the tests NoopXXX in other cases. Happy to change if you think it is important.

api/src/main/java/io/opencensus/trace/propagation/TextFormat.java
```diff
@@ -45,7 +45,7 @@
  *   }
  * }
  *
- * void makeHttpRequest() {
+ * void makeHttpRequest(HttpURLConnection connection) {
  *   Span span = tracer.spanBuilder("Sent.MyRequest").startSpan();
  *   try (Scope s = tracer.withSpan(span)) {
  *     HttpURLConnection connection =
```
sebright: Should this declaration of `connection` be removed now?

api/src/main/java/io/opencensus/trace/propagation/TextFormat.java
sebright: I still think that the Javadoc could be clearer, since it doesn't say what the class does.

api/src/test/java/io/opencensus/trace/propagation/TextFormatTest.java
sebright: It's fine the way it is.

api/src/main/java/io/opencensus/trace/propagation/TextFormat.java
sebright: Okay.  I didn't consider the cost of the iterator on Android.

api/src/main/java/io/opencensus/trace/propagation/TextFormat.java
bogdandrutu: done. my bad of adding as param.

bogdandrutu:
PTAL. Updated javadoc.

Issue #649, opened 2017-09-19
Interval stats currently don't support sum of squared deviations


songy23:
For interval stats implementation, we're using sub-intervals: within each sub-interval, we store mean, count, min, max, bucket counts and sum of squared deviations, and when we need to get stats for a whole interval, we combine stats in all sub-intervals and generate the result. For example:

sub-interval 1 has values 5, 10, 15, its sum of squared deviations is 50;
sub-interval 2 has value -10, its sum of squared deviations is 0;
when we combine these 2 (5, 10, 15, -10), the correct sum of squared deviations should be 350

The problem is we cannot store all the input values in each sub-interval. Instead, all we have are the stats mentioned above (mean, count, etc). By so far I haven't found a way to combine sum of squared deviations using just these stats. As a result, currently for interval stats, we don't support sum of squared deviations - it will always be 0.

PR #727, opened 2017-10-24
Calculate sum of squared deviations for IntervalView

songy23:
Fix https://github.com/census-instrumentation/opencensus-java/issues/649

Calculate sum of squared deviations by using the Parallel algorithm. Thanks @isturdy for providing a solution to this issue.

### DIFF
--- a/core_impl/src/main/java/io/opencensus/implcore/stats/MutableAggregation.java
+++ b/core_impl/src/main/java/io/opencensus/implcore/stats/MutableAggregation.java
@@ -305,7 +305,7 @@ abstract class MutableAggregation {
         double delta = mutableDistribution.mean - this.mean;
         this.sumOfSquaredDeviations = this.sumOfSquaredDeviations
             + mutableDistribution.sumOfSquaredDeviations
-            + delta * delta * this.count * mutableDistribution.count
+            + Math.pow(delta, 2) * this.count * mutableDistribution.count
             / (this.count + mutableDistribution.count);
       }
 
--- a/core_impl/src/main/java/io/opencensus/implcore/stats/MutableAggregation.java
+++ b/core_impl/src/main/java/io/opencensus/implcore/stats/MutableAggregation.java
@@ -301,12 +301,10 @@ abstract class MutableAggregation {
 
       // Algorithm for calculating the combination of sum of squared deviations:
       // https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Parallel_algorithm.
-      double delta = mutableDistribution.mean - this.mean;
-      double ma = this.sumOfSquaredDeviations * (this.count - 1);
-      double mb = mutableDistribution.sumOfSquaredDeviations * (mutableDistribution.count - 1);
-      this.sumOfSquaredDeviations = 0;
       if (this.count + mutableDistribution.count > 0) {
-        this.sumOfSquaredDeviations = ma + mb
+        double delta = mutableDistribution.mean - this.mean;
+        this.sumOfSquaredDeviations = this.sumOfSquaredDeviations
+            + mutableDistribution.sumOfSquaredDeviations
             + delta * delta * this.count * mutableDistribution.count
             / (this.count + mutableDistribution.count);
       }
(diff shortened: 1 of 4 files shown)


core_impl/src/main/java/io/opencensus/implcore/stats/MutableAggregation.java
```diff
@@ -299,6 +299,18 @@ void combine(MutableAggregation other, double fraction) {
       checkArgument(this.bucketBoundaries.equals(mutableDistribution.bucketBoundaries),
           "Bucket boundaries should match.");
 
+      // Algorithm for calculating the combination of sum of squared deviations:
+      // https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Parallel_algorithm.
+      double delta = mutableDistribution.mean - this.mean;
+      double ma = this.sumOfSquaredDeviations * (this.count - 1);
+      double mb = mutableDistribution.sumOfSquaredDeviations * (mutableDistribution.count - 1);
```
isturdy: I think these should just be sumOfSquaredDeviations--the pseudocode from Wikipedia takes variances as arguments (using the unbiased estimator formula) and thus needs to multiply by N-1 to get back the sum of the squared deviations, but you already have the sum rather than the variance.

The test looks correct, but both the distributions have count == 2, so multiplying by count - 1 does nothing--I think it will fail if you give it a 3+ element distribution.

songy23: You're right, fixed. Also added a test for the case when there are >2 values.

core_impl/src/main/java/io/opencensus/implcore/stats/MutableAggregation.java
```diff
@@ -299,6 +299,16 @@ void combine(MutableAggregation other, double fraction) {
       checkArgument(this.bucketBoundaries.equals(mutableDistribution.bucketBoundaries),
           "Bucket boundaries should match.");
 
+      // Algorithm for calculating the combination of sum of squared deviations:
+      // https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Parallel_algorithm.
+      if (this.count + mutableDistribution.count > 0) {
+        double delta = mutableDistribution.mean - this.mean;
+        this.sumOfSquaredDeviations = this.sumOfSquaredDeviations
```
isturdy: Could be a bit more concise (and parallel with the other updates) with +=, but I think it obscures the correspondence to the linked algorithm--your call.

core_impl/src/main/java/io/opencensus/implcore/stats/MutableAggregation.java
```diff
@@ -299,6 +299,16 @@ void combine(MutableAggregation other, double fraction) {
...
+            + mutableDistribution.sumOfSquaredDeviations
+            + delta * delta * this.count * mutableDistribution.count
```
isturdy: I would prefer Math.pow(delta, 2).

core_impl/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
```diff
@@ -170,7 +170,7 @@ void record(TagContext tags, long value, Timestamp timestamp) {
     for (int i = 0; i < columns.size(); ++i) {
       TagKey tagKey = columns.get(i);
       if (!tags.containsKey(tagKey)) {
-        // replace not found key values by “unknown/not set”.
+        // replace not found tag values with null.
```
isturdy: Is this related to the other changes? If not I would prefer a separate pull. Regardless, this seems misleading--null suggests an empty string to me, not "UNKNOWN_TAG_VALUE".

core_impl/src/main/java/io/opencensus/implcore/stats/MutableAggregation.java
songy23: Updated.

core_impl/src/main/java/io/opencensus/implcore/stats/MutableAggregation.java
songy23: I'll prefer to leave this as it is - to be clear and more consistent with the formula.

core_impl/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
songy23: Sure, I'll make another PR to fix this.

songy23:
Thanks!

Issue #711, opened 2017-10-18
Simplify SampledSpanStore synchronization


bogdandrutu:
@zhangkun83 proposed this:

You can get rid of the synchronization in this class by changing `samples` to a `ConcurrentHashMap`.
Here and in `registerSpanNamesForCollection()` you can use `putIfAbsent()`:
```java
PerSpanNameSamples perSpanNameSamples = samples.get(spanName);
if (perSpanNameSamples == null && span.getSampleToLocalSpanStore()) {
  perSpanNameSamples = new PerSpanNameSamples();
  PerSpanNameSamples existingSamples = samples.putIfAbsent(spanName, perSpanNameSamples);
  if (existingSamples != null) {
    perSpanNameSamples = existingSamples;
  }
}
```

zhangkun83:
It is not only for simplicity, it also removes a contention on a global lock.

bogdandrutu:
I think ConcurrentHashMap is still a global lock, isn't it?

zhangkun83:
No, it's thread-safe without synchronization.

bogdandrutu:
Read more about ConcurrentHashMap and decided to fix the contention in a different way by using our queue and enqueue the event and let the worker thread to process the event. Main reason is that ConcurrentHashMap is optimized for intensive writes but our behavior is read intense.

zhangkun83:
ConcurrentHashMap may be optimized for intensive writes, but it doesn't necessarily mean read performance is bad? Less writes is actually preferable because it means lower chance of falling back to the slow path.

Using a queue sounds more complex and there is always a question about its size management.

bogdandrutu:
We need not only to protect the Map we also need to protect each Bucket and fields in each bucket. Will get more complicated to do that on the critical path.

PR #729, opened 2017-10-25
Remove contention between threads and worker thread when register span names.

bogdandrutu:
Change the logic of register/unregister the span name to local span store form directly accessing the lock to log an event into our event queue and consume it by the worker thread.

This way we remove a global lock (contention) between main threads and the worker thread.

Fixes https://github.com/census-instrumentation/opencensus-java/issues/711

Reason to not use https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html is that the map is read intense not write and ConcurrentHashMap is optimized for writes.

### DIFF
--- a/impl_core/src/main/java/io/opencensus/implcore/trace/export/SampledSpanStoreImpl.java
+++ b/impl_core/src/main/java/io/opencensus/implcore/trace/export/SampledSpanStoreImpl.java
@@ -17,6 +17,7 @@
 package io.opencensus.implcore.trace.export;
 
 import com.google.common.collect.EvictingQueue;
+import io.opencensus.implcore.internal.EventQueue;
 import io.opencensus.implcore.trace.SpanImpl;
 import io.opencensus.trace.Status;
 import io.opencensus.trace.Status.CanonicalCode;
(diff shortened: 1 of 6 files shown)


impl_core/src/main/java/io/opencensus/implcore/trace/export/SampledSpanStoreImpl.java
```diff
@@ -48,6 +49,10 @@
       NUM_SAMPLES_PER_LATENCY_BUCKET * NUM_LATENCY_BUCKETS
           + NUM_SAMPLES_PER_ERROR_BUCKET * NUM_ERROR_BUCKETS;
 
+  // Used to stream the register/unregister events to the implementation to avoid lock contention
+  // between the main threads and the worker thread.
+  private final EventQueue eventQueue;
```
zhangkun83: It doesn't prevent the contention on `samples`. There is always a contention if `considerForSampling` ever has a `synchronized` in the main path.

bogdandrutu: The way it works is the following:

Critical path Thread -> EventQueues(Disruptor) -> Worker Thread -> SampledSpanStoreImpl.lock

As you can see the critical path Thread does not touch the lock in the SampledSpanStoreImpl only the Worker thread interacts with that mutex.

For the end event is the same because the considerForSampling is also recorded as part of the SpanEndEvent not directly applied to the SampledSpanStoreImpl.

zhangkun83: I see. You are already streaming `considerForSampling()` using the EventQueue. I misunderstood.

bogdandrutu: For the moment we have one thread. We will work later to improve and support multiple threads then your suggestion of using thread safe data structures will make a lot of sense.

Also currently getRegisteredSpanNamesForCollection is a place where the critical path can interact with the lock, but we marked that as test only (because indeed it is used only in tests).

PR #736, opened 2017-10-27
Change TagValue.MAX_LENGTH to 255, to match TagKey.MAX_LENGTH.

sebright:


songy23:
Do we want the same restriction for Measure name as well? (currently it's 256)

sebright:
> Do we want the same restriction for Measure name as well? (currently it's 256)

I think that makes sense, if we need to restrict it.  We can always loosen the restrictions on measure names later, though.

Issue #718, opened 2017-10-22
Specify the behavior when ViewManager.getView() is called for unregistered views


songy23:
Currently in our implementation, if someone called `ViewManager.getView()` for an unregistered view, an `IllegalArgumentException` will be thrown: https://github.com/census-instrumentation/opencensus-java/blob/master/core_impl/src/main/java/io/opencensus/implcore/stats/StatsManager.java#L52. 

However, this API might be simpler to use if we return null for unregistered views instead. Users could simply check for null instead of trying to deal with unchecked exceptions.

Also, as @sebright pointed out, 

> Indicating that a view isn't registered would become even more important if we ever added ViewManager.unregisterView.

See more discussion at: https://github.com/census-instrumentation/opencensus-java/pull/713/commits/37ffb47fc4c7cdc0f57d2f71ed6e202218d3abd0#diff-453a0897e016f2c5fe149602aba1e353R198

sebright:
We will probably need to support unregistering views, so we should return null, or at least throw a checked exception.

PR #737, opened 2017-10-27
Returns null when getting unregistered views

songy23:
Fix https://github.com/census-instrumentation/opencensus-java/issues/718

### DIFF
--- a/api/src/main/java/io/opencensus/stats/ViewManager.java
+++ b/api/src/main/java/io/opencensus/stats/ViewManager.java
@@ -16,6 +16,8 @@
 
 package io.opencensus.stats;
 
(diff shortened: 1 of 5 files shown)


api/src/main/java/io/opencensus/stats/ViewManager.java
```diff
@@ -29,6 +31,9 @@
 
   /**
    * Returns the current stats data, {@link ViewData}, associated with the given view name.
+   *
+   * <p>Returns {@code null} if the {@code View} is not registered.
```
bogdandrutu: @return statement?

songy23: Done.

PR #739, opened 2017-10-27
Update max length of Measure name to 255

songy23:
To be consistent with TagKey and TagValue.

### DIFF
--- a/api/src/main/java/io/opencensus/stats/Measure.java
+++ b/api/src/main/java/io/opencensus/stats/Measure.java
@@ -28,7 +28,7 @@ import javax.annotation.concurrent.Immutable;
 @Immutable
 public abstract class Measure {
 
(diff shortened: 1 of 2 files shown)


api/src/main/java/io/opencensus/stats/Measure.java
```diff
@@ -86,7 +86,7 @@ private Measure() {
      */
     public static MeasureDouble create(String name, String description, String unit) {
       checkArgument(StringUtil.isPrintableString(name) && name.length() <= NAME_MAX_LENGTH,
-          "Name should be a ASCII string with a length no greater than 256 characters.");
+          "Name should be a ASCII string with a length no greater than 255 characters.");
```
bogdandrutu: Nit: Use the constant value instead of literal value to make things easier to change in the future.

songy23: Sure, updated.

PR #742, opened 2017-10-30
Update the jmhreport plugin to its latest version.

ubschmidt2:


ubschmidt2:
Triggered a new build because cla/google was failing.

PR #743, opened 2017-10-30
Revert "Fix opencensus agent to be java6 compatible. (#619)"

ubschmidt2:
This reverts commit d2263490e5353b005657bed1fbd60585559a85f2.

I'm reverting this because it seems unrelated to the issue in https://github.com/census-instrumentation/opencensus-java/issues/616#issue-256872886, which is actually about a different code location (line 52).

These "Undefined reference" warnings come from Animal Sniffer. In theory, Animal Sniffer may have seen the class files before they were processed by Retrolambda. However, the gradle task dependency tree (from dorongold/gradle-task-tree) doesn't support this theory.

In any case, we have test coverage for running on Java 6 since https://github.com/census-instrumentation/opencensus-java/pull/638.

### DIFF
--- a/contrib/agent/src/main/java/io/opencensus/contrib/agent/Resources.java
+++ b/contrib/agent/src/main/java/io/opencensus/contrib/agent/Resources.java
@@ -61,11 +61,8 @@ final class Resources {
 
     file.deleteOnExit();
 
-    InputStream is = getResourceAsStream(resourceName);
-    try {
+    try (InputStream is = getResourceAsStream(resourceName)) {
       ByteStreams.copy(is, outputStream);
-    } finally {
-      is.close();
     }
   }
 


adriancole:
Sounds reasonable

bogdandrutu:
Try with resources does not work on java6, am I missing something?

adriancole:
Retrolambda rewrites bytecode of try/resources to normal try finally

ubschmidt2:
Yes, though source/target is 1.7, retrolambda rewrites the classes to 1.6. Try-with-resources works fine (https://github.com/orfjackal/retrolambda#backported-language-features).

ubschmidt2:
PTAL.

bogdandrutu:
I think we should keep the code as 1.6 (without fancy retrolambda). My 2 cents.

ubschmidt2:
Bogdan that's actually a good point. I'd assumed to get more benefit from not restricting the code to 1.6, but only using try-with-resource in this class is not convincing. Thus, happy to just stick with 1.6. Will send a separate PR for that.

ubschmidt2:
Obsoleted by https://github.com/census-instrumentation/opencensus-java/pull/799.

PR #747, opened 2017-10-30
Move the Travis "script" from .travis.yml to a separate file.

### LINKED ISSUE #693: Create a separate file for the Travis "script" section.
The ["script"](https://github.com/census-instrumentation/opencensus-java/blob/a9558ba6f680beecaef5a77eda690dda83bc215e/.travis.yml#L47) section of .travis.yml is becoming large enough that we should probably split it into a separate .sh file.  A .sh file would let us avoid the restrictions of the .travis.yml file format.  For example, we could use colons or add comments within the script.

ubschmidt2:
Fixes #693.


scripts/travis_script
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+#
+# Travis build script, cf.
+# https://docs.travis-ci.com/user/customizing-the-build/#Implementing-Complex-Build-Steps.
+
```
sebright: I think this script should also use "set -ev" from the link.

ubschmidt2: I've added this to this change for now, but will likely have to revert again because -v seems fairly useless (-x would seem better), and, more importantly, -e causes the script to fail (jdk_switcher is a function and somewhere in there it encounters an exit code 1).

ubschmidt2: Oh, interesting, "jdk_switcher use ..." calls "sudo update-java-alternatives ...", but "sudo" would only work if .travis.yml contains "sudo: true".

ubschmidt2: Now using -o xtrace (instead of -o verbose) and JAVA_HOME="$(jdk_switcher home openjdk8)" ... (instead of jdk_switcher use ...). This looks better. PTAL.

PR #748, opened 2017-10-30
Rename ContextManager to ContextTrampoline.

ubschmidt2:
This is a follow-up on https://github.com/census-instrumentation/opencensus-java/pull/548 (https://github.com/census-instrumentation/opencensus-java/pull/548#issuecomment-324511205, https://github.com/census-instrumentation/opencensus-java/pull/548#issuecomment-324781496), just renaming ContextManager to ContextTrampoline, which is better name. Going to add the TraceTrampoline (etc.) mentioned in https://github.com/census-instrumentation/opencensus-java/pull/548 in a separate PR.


adriancole:
ContextHulaHoop would be better if you can backport an engineering pattern
to the term.

/me ducks

PR #753, opened 2017-10-31
Add initial support for configuring the agent.

ubschmidt2:
This is just an initial version to get basic infrastructure in place together with preliminary end-to-end examples. To keep this PR small, I want to leave most refinements to future PRs, such as the introduction of a Settings class with proper accessors. Same for the config-driven instrumentation for tracing.

### DIFF
--- a/contrib/agent/build.gradle
+++ b/contrib/agent/build.gradle
@@ -24,9 +24,10 @@ dependencies {
   compileOnly libraries.auto_service
   compileOnly libraries.grpc_context
   compileOnly project(':opencensus-api')
+  compile libraries.byte_buddy
+  compile libraries.config
   compile libraries.findbugs_annotations
   compile libraries.guava
-  compile libraries.byte_buddy
 
   signature 'org.codehaus.mojo.signature:java16:+@signature'
 }
@@ -61,6 +62,7 @@ shadowJar {
   // Include only the following dependencies (excluding transitive dependencies).
   dependencies {
     include(dependency(libraries.byte_buddy))
+    include(dependency(libraries.config))
     include(dependency(libraries.guava))
   }
 
@@ -72,6 +74,8 @@ shadowJar {
   // classes, which are loaded by the same classloader (the system classloader).
   // Byte Buddy:
   relocate 'net.bytebuddy', agentRepackaged + '.bytebuddy'
+  // Config:
+  relocate 'com.typesafe.config', agentRepackaged + '.config'
   // Guava:
   relocate 'com.google.common', agentRepackaged + '.guava'
   relocate 'com.google.thirdparty.publicsuffix', agentRepackaged + '.publicsuffix'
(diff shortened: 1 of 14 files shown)


ubschmidt2:
PTAL.

contrib/agent/build.gradle
```diff
@@ -93,6 +95,8 @@ shadowJar {
   // classes, which are loaded by the same classloader (the system classloader).
   // Byte Buddy:
   relocate 'net.bytebuddy', agentRepackaged + '.bytebuddy'
+  // Config:
+  relocate 'com.typesafe.config', agentRepackaged + '.config'
```
adriancole: not quite sure the impact of source 1.6, but target 1.8.. suppose the JRE 6 tests will ensure this works with JRE 6? Or will we have to watch execution paths? https://github.com/lightbend/config/blob/master/config/build.sbt#L75

contrib/agent/src/main/resources/reference.conf
```diff
@@ -0,0 +1,21 @@
+# Reference configuration for the OpenCensus Agent for Java.
+
+opencensus.contrib.agent {
+
+  # Configuration settings related to automatic context propagation.
+  context-propagation {
+
+    # Enable/disable automatic context propagation for Executors.
+    executor = on
```
adriancole: is on syntax normal? just used to .enabled which maybe isn't idomatic in this library..

contrib/agent/build.gradle
ubschmidt2: Nice catch! I haven't looked into that project's build scripts before.

According to https://github.com/lightbend/config/blob/master/README.md#binary-releases, the last version supporting Java 6 is 1.2.1, which I'm using in this PR. It's interesting to see that typesafe config is using source 1.6 and target 1.8 (https://github.com/lightbend/config/commit/e72decad9251ea2c31e7e53dbfc9902f0f922194) - I haven't looked that close. Ideally, I'd like to use a newer version of typesafe config, but not bothering about that for now.

In any case the integration tests running with JRE 6 will catch any obvious issue (class file version, usage of java 7+ APIs at runtime).

contrib/agent/src/main/resources/reference.conf
ubschmidt2: Yes, `on` is one of the six strings (`true`/`false`, `on`/`off`, `yes`/`no`) that can be parsed to a boolean by typesafe config (https://github.com/lightbend/config/blob/master/HOCON.md#automatic-type-conversions).

Re `.enabled`: I haven't yet seen enough configs to make up my mind about this. It would definitely make sense if there are more details to configure, something like:
```
feature1.enabled = yes
feature1.host = www.example.com
feature1.port = 80
```
BTW, in HOCON, this is the same as:
```
feature1 {
  enabled = yes
  host = www.example.com
  port = 80
```
Now, considering that I later want to use this feature/detail pattern for configuring which (and how) methods will be instrumented, it makes sense to use the same pattern everywhere, if even there are no feature details to configure. Will update the PR later. (Ignoring nested features for now)

ubschmidt2:
PTAL.

contrib/agent/src/main/resources/reference.conf
```diff
@@ -0,0 +1,23 @@
...
+
+    # Enable/disable automatic context propagation for Threads.
+    thread.enabled = true
+  }
+
+  # The "trace" section configures which Java methods the agent instruments for
+  # tracing.
+  trace {
+
+    java.net.URL.getContent {
```
adriancole: clever

ubschmidt2:
Thanks, Adrian! I'll wait a bit in case of additional comments, otherwise will merge by EOW. (busy this week)

Issue #726, opened 2017-10-24
Consider renaming the methods for recording stats.


sebright:
#676 replaced `MeasureMap` with an abstract class for recording multiple values at once, `StatsRecord`.  The names of the methods and class could be improved.  Here is the code that is required to record stats with the current context:

```java
        statsRecorder
            .newRecord()
            .put(RPC_CLIENT_ROUNDTRIP_LATENCY, 100)
            .put(RPC_CLIENT_REQUEST_BYTES, 2000)
            .record();
```

and an explicit context:

```java
        statsRecorder
            .newRecord()
            .put(RPC_CLIENT_ROUNDTRIP_LATENCY, 100)
            .put(RPC_CLIENT_REQUEST_BYTES, 2000)
            .recordWithExplicitTagContext(tags);
```

The word "record" appears too many times.

/cc @dinooliva @songy23 @bogdandrutu @adriancole 

EDIT:

After #728, the code required to record with an explicit TagContext is:

```java
        statsRecorder
            .newRecord()
            .put(RPC_CLIENT_ROUNDTRIP_LATENCY, 100)
            .put(RPC_CLIENT_REQUEST_BYTES, 2000)
            .record(tags);
```

PR #757, opened 2017-11-01
Rename StatsRecord to MeasureMap

songy23:
Fix https://github.com/census-instrumentation/opencensus-java/issues/726

Per our discussion today, we're renaming `StatsRecord` to `MeasureMap`. There are also some temporary changes to the implementation (renamed the old internal `MeasureMap` to `MeasureMapInternal`), which will be updated after the API release.


songy23:
Thanks!

PR #758, opened 2017-11-01
Make TagContextBinarySerializer.toByteArray throw a checked exception.

sebright:
The first version of the implementation will throw
TagContextSerializationException when the serialized TagContext would be larger
than 8192 bytes.

### DIFF
--- a/api/src/main/java/io/opencensus/tags/propagation/TagContextBinarySerializer.java
+++ b/api/src/main/java/io/opencensus/tags/propagation/TagContextBinarySerializer.java
@@ -34,8 +34,9 @@ public abstract class TagContextBinarySerializer {
    *
    * @param tags the {@code TagContext} to serialize.
    * @return the on-the-wire representation of a {@code TagContext}.
+   * @throws TagContextSerializationException if the {@code TagContext} is too large to serialize.
    */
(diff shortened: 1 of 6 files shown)


api/src/main/java/io/opencensus/tags/propagation/TagContextBinarySerializer.java
```diff
@@ -34,8 +34,9 @@
    *
    * @param tags the {@code TagContext} to serialize.
    * @return the on-the-wire representation of a {@code TagContext}.
+   * @throws TagContextSerializationException if the {@code TagContext} is too large to serialize.
```
songy23: Shall we mention the 8192 bytes limit here?

sebright:
I was also wondering if we should rename TagContextParseException to TagContextDeserializationException, to make it more generic.  The name might be better for an exception that can be used for other errors, like invalid tags.

api/src/main/java/io/opencensus/tags/propagation/TagContextBinarySerializer.java
sebright: I realized that the maximum could change, so we probably shouldn't put it in the API.  We could include it in the documentation for the implementation instead.  I made a PR to remove it: #760

PR #759, opened 2017-11-01
Rename TagContextParseException to TagContextDeserializationException. 

sebright:
The new name is more consistent with TagContextSerializationException.


sebright:
@bogdandrutu also agreed with the renaming, so I'll merge this.

PR #760, opened 2017-11-01
Avoid specifying maximum size of serialized TagContext.

sebright:
The value will probably be configurable in a future version, and the default
could change.


sebright:
> This could also be specified in opencensus-specs.

Yes, I commented on https://github.com/census-instrumentation/opencensus-specs/issues/12.

PR #766, opened 2017-11-01
Add a trampoline for tracing-related methods.

ubschmidt2:
This change adds a separate trampoline for tracing-related methods (cf. https://github.com/census-instrumentation/opencensus-java/pull/548#issuecomment-324511205). Besides the different naming and purpose, it's just a copy of the existing ContextTrampoline.

### DIFF
--- a/contrib/agent/src/main/java/io/opencensus/contrib/agent/bootstrap/TraceTrampoline.java
+++ b/contrib/agent/src/main/java/io/opencensus/contrib/agent/bootstrap/TraceTrampoline.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2017, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


contrib/agent/src/main/java/io/opencensus/contrib/agent/bootstrap/TraceTrampoline.java
```diff
@@ -0,0 +1,82 @@
...
+   * Starts a new span and sets it as the current span.
+   *
+   * @param spanName the name of the returned {@link Span}
+   * @return an object that defines a scope where the newly created {@code Span} will be set to the
+   *     current Context
+   * @see Tracer#spanBuilder(java.lang.String)
+   * @see SpanBuilder#startScopedSpan()
+   */
+  @MustBeClosed
+  public static Closeable startScopedSpan(String spanName) {
```
adriancole: Iirc this finishes the span on close? Maybe note that (as it is good for synch tracing less so for async)

adriancole: Also comment why must be closed (to ensure context doesnt leak)

ubschmidt2: Thanks for the review! I've improved the Javadoc (copied quite a few bits...).

Issue #596, opened 2017-09-08
Add support to start a span wrapped on top of a Runnable/Callable


bogdandrutu:
Add equivalent of startScoped in the SpanBuilder that wraps Runnable and Callable. Something like:

// Starts a new span which will be made active while execute the returned Runnable and ends it.
public Runnable startWrapped(Runnable);

Caveats:
* Decide if we propagate the grpc::Context as well.
* Probably needed for Tags as well.

bogdandrutu:
This is currently possible with:

```java
Span span = spanBuilder.startSpan();
Runnable runnableWithSpan = tracer.withSpan(span, runnable);
```

Caveat: The "withSpan" methods do not wrap the whole context.

Closing until someone will ask for this.

PR #778, opened 2017-11-04
Add methods to wrap Runnable and Callbacks and to run them.

bogdandrutu:
This is the fix for https://github.com/census-instrumentation/opencensus-java/issues/596 for tracing.


api/src/main/java/io/opencensus/trace/SpanBuilder.java
```diff
@@ -223,6 +224,60 @@ public final Scope startScopedSpan() {
...
+   *   tracer.wrap(span, myRunnable).run();
+   * } finally {
+   *   span.end();
+   * }
+   * }</pre>
+   *
+   * @param runnable the {@code Runnable} to run in the {@code Span}.
+   */
+  public final void startSpanAndRun(Runnable runnable) {
+    Span span = startSpan();
```
adriancole: One choice here is whether to keep adding methods or do something like Spans.run(spanBuilder, Runnable)

bogdandrutu: I think both are valid and nice solutions. I think if we have a builder it makes sense to have them here. In languages where the Builder pattern is not common your suggestion makes a lot of sense.

bogdandrutu:
PTAL.

Added TODO to record an error status in case of an exception in Runnable or Callable. Also added startSpanAndWrap functions.

api/src/main/java/io/opencensus/trace/Tracer.java
```diff
@@ -151,6 +152,78 @@ public final Scope withSpan(Span span) {
   }
 
   /**
+   * Wraps a {@link Runnable} so that it executes with the {@code span} as the current {@code Span}
+   * and saves the rest of the Context.
```
sebright: I wonder if it would be better for this method to only set the current span.  Then it would compose better with a corresponding method for tags. The same applies to startSpanAndWrap.

api/src/test/java/io/opencensus/trace/SpanBuilderTest.java
```diff
@@ -0,0 +1,121 @@
...
+import io.opencensus.common.Scope;
+import java.util.concurrent.Callable;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+
+/** Unit tests for {@link Tracer}. */
```
sebright: s/Tracer/SpanBuilder

api/src/test/java/io/opencensus/trace/TracerTest.java
```diff
@@ -94,6 +94,31 @@ public void run() {
     assertThat(noopTracer.getCurrentSpan()).isSameAs(BlankSpan.INSTANCE);
```
sebright: Should this test also check that `end` wasn't called in the wrapped Runnable?

api/src/test/java/io/opencensus/trace/TracerTest.java
```diff
@@ -71,20 +71,20 @@ public void getCurrentSpan_WithSpan() {
   }
 
   @Test
-  public void propagationViaRunnable() {
+  public void wrapRunnable() {
```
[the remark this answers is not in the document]
bogdandrutu: I fixed it by removing the previous test. No need to test in our unit-tests that gRPC Context does the correct thing.

api/src/test/java/io/opencensus/trace/TracerTest.java
bogdandrutu: Done.

api/src/test/java/io/opencensus/trace/TracerTest.java
```diff
@@ -94,6 +94,31 @@ public void run() {
...
+                  assertThat(noopTracer.getCurrentSpan()).isSameAs(span);
+                  return null;
+                }
+              });
+    } finally {
+      ws.close();
+    }
+    assertThat(noopTracer.getCurrentSpan()).isSameAs(BlankSpan.INSTANCE);
+    // When we call the callable we will have the span in the current Context.
+    callable.call();
```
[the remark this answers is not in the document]
bogdandrutu: Done.

api/src/test/java/io/opencensus/trace/SpanBuilderTest.java
bogdandrutu: Done.

api/src/test/java/io/opencensus/trace/CurrentSpanUtilsTest.java
```diff
@@ -69,25 +70,51 @@ public void withSpan() {
   }
 
   @Test
-  public void propagationViaRunnable() {
-    Runnable runnable = null;
+  public void wrapRunnable() {
```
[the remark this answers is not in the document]
bogdandrutu: Done.

api/src/main/java/io/opencensus/trace/Tracer.java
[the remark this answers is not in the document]
bogdandrutu: Done.

api/src/main/java/io/opencensus/trace/Tracer.java
```diff
@@ -151,6 +152,78 @@ public final Scope withSpan(Span span) {
...
+   *
+   * <p>This method saves the entire {@link io.grpc.Context} when wraps the {@code Runnable}.
+   *
+   * <p>IMPORTANT: this method does not end the {@code Span}, caller must do that after the {@code
+   * Runnable} is executed.
```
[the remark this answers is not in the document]
bogdandrutu: Done.

api/src/main/java/io/opencensus/trace/Tracer.java
bogdandrutu: @sebright let me think more about this.
@adriancole any suggestion?

api/src/main/java/io/opencensus/trace/Tracer.java
```diff
@@ -151,6 +152,78 @@ public final Scope withSpan(Span span) {
...
+   * Callable}, or after the {@code Callable} is executed.
+   *
+   * <pre><code>
+   * class MyClass {
+   *   private static Tracer tracer = Tracing.getTracer();
+   *   void handleRequest(Executor executor) {
+   *     Span span = new Tracer.spanBuilder("MyRunnableSpan");
+   *     executor.execute(tracer.wrap(span, new Callable() {
+   *      {MessageEventl @}Override
+   *       public void call() throws Exception {
```
sebright: call() should return an object.

api/src/main/java/io/opencensus/trace/Tracer.java
```diff
@@ -151,6 +152,78 @@ public final Scope withSpan(Span span) {
...
+   * <p>IMPORTANT: this method does not end the {@code Span}, caller must do that within the {@code
+   * Runnable}, or after the {@code Runnable} is executed.
+   *
+   * <pre><code>
+   * class MyClass {
+   *   private static Tracer tracer = Tracing.getTracer();
+   *   void handleRequest(Executor executor) {
+   *     Span span = new Tracer.spanBuilder("MyRunnableSpan");
+   *     executor.execute(tracer.wrap(span, new Runnable() {
+   *      {MessageEventl @}Override
```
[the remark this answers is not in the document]
bogdandrutu: you are correct. done.

api/src/main/java/io/opencensus/trace/Tracer.java
bogdandrutu: done.

api/src/main/java/io/opencensus/trace/Tracer.java
```diff
@@ -151,6 +152,78 @@ public final Scope withSpan(Span span) {
...
+   *     }));
+   *   }
+   * }
+   * }</pre>
+   *
+   * @param span the {@code Span} to be set as current.
+   * @param runnable the {@code Runnable} to wrap in the {@code Span}.
+   * @return the wrapped {@code Runnable}.
+   */
+  public final Runnable wrap(Span span, Runnable runnable) {
```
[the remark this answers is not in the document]
bogdandrutu: The name has to be consistent with io.grpc.Context#wrap()

api/src/main/java/io/opencensus/trace/SpanBuilder.java
```diff
@@ -223,6 +224,96 @@ public final Scope startScopedSpan() {
...
+   *       } finally {
+   *         span.end();
+   *       }
+   *     }
+   * });
+   * }</pre>
+   *
+   * @param runnable the {@code Runnable} to run in the {@code Span}.
+   */
+  public final Runnable startSpanAndWrap(final Runnable runnable) {
```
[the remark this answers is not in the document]
bogdandrutu: wrap is a terminology added by the io.grpc.Context. I think it is better to keep it consistent.

api/src/main/java/io/opencensus/trace/CallableSpanHandle.java
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  @Override
+  public V call() throws Exception {
+    try {
+      return callable.call();
+    } catch (Exception e) {
+      span.setStatus(Status.UNKNOWN.withDescription(e.getMessage()));
+      throw e;
+    } catch (Throwable t) {
```
adriancole: heh guarding against the terrible programming idea that you can make a subclass of throwable that's neither exception nor error, huh?

api/src/main/java/io/opencensus/trace/CurrentSpanUtils.java
```diff
@@ -48,6 +49,28 @@ static Scope withSpan(Span span) {
     return new WithSpan(span, ContextUtils.CONTEXT_SPAN_KEY);
   }
 
+  /**
+   * Wraps a {@link Runnable} so that it executes with the {@code span} as the current {@code Span}.
```
adriancole: mention any error will end up as a status

api/src/main/java/io/opencensus/trace/CallableSpanHandle.java
```diff
@@ -0,0 +1,46 @@
...
+      span.setStatus(Status.UNKNOWN.withDescription(t.getMessage()));
```
adriancole: guard against getMessage() == null. Ex I tend to use getClass().getSimpleName when message is null.

api/src/main/java/io/opencensus/trace/SpanBuilder.java
```diff
@@ -223,6 +224,98 @@ public final Scope startScopedSpan() {
...
+   *       } finally {
+   *         span.end();
+   *       }
+   *     }
+   * });
+   * </code></pre>
+   *
+   * @param runnable the {@code Runnable} to run in the {@code Span}.
+   */
+  public final Runnable startSpanAndWrap(final Runnable runnable) {
```
adriancole: maybe recordInSpan? or decorate or wrap?

ex. spanBuilder.decorate(runnable), spanBuilder.wrap(runnable) makes sense to me. wrap makes most sense sense, if we start the span deferred.

api/src/main/java/io/opencensus/trace/Tracer.java
```diff
@@ -151,6 +152,78 @@ public final Scope withSpan(Span span) {
...
+   * <p>IMPORTANT: this method does not end the {@code Span}, caller must do that within the {@code
+   * Callable}, or after the {@code Callable} is executed.
+   *
+   * <pre><code>
+   * class MyClass {
+   *   private static Tracer tracer = Tracing.getTracer();
+   *   void handleRequest(Executor executor) {
+   *     Span span = new Tracer.spanBuilder("MyRunnableSpan");
+   *     executor.execute(tracer.wrap(span, new Callable() {
+   *      {MessageEventl @}Override
```
adriancole: fix

api/src/main/java/io/opencensus/trace/SpanBuilder.java
```diff
@@ -223,6 +224,96 @@ public final Scope startScopedSpan() {
     return new ScopedSpanHandle(startSpan());
   }
 
+  /**
+   * Wraps a {@link Runnable} so that it executes with the {@code span} as the current {@code Span}
+   * and saves the rest of the Context, and ends the {@code Span} after the {@code runnable} is run.
+   *
+   * <p>This method saves the entire {@link io.grpc.Context} when wraps the {@code Runnable}.
+   *
+   * <p>This is equivalent with following code:
```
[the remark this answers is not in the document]
bogdandrutu: Can you tell me a bit more details. I think I miss something.

api/src/main/java/io/opencensus/trace/Tracer.java
```diff
@@ -151,6 +152,78 @@ public final Scope withSpan(Span span) {
...
+    return CurrentSpanUtils.wrap(span, runnable);
+  }
+
+  /**
+   * Wraps a {@link Callable} so that it executes with the {@code Span} as the current {@code Span}
+   * and saves the rest of the current Context.
+   *
+   * <p>This method saves the entire {@link io.grpc.Context} when wraps the {@code Callable}.
+   *
+   * <p>IMPORTANT: this method does not end the {@code Span}, caller must do that within the {@code
```
adriancole: before saying what it doesn't say what it does. Ex this is used to ensure the current span is readable when invoking callbacks.

api/src/test/java/io/opencensus/trace/SpanBuilderTest.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+    verify(span).end(EndSpanOptions.DEFAULT);
+    assertThat(tracer.getCurrentSpan()).isSameAs(BlankSpan.INSTANCE);
+  }
+
+  @Test
+  public void startSpanAndWrap_Runnable() {
+    assertThat(tracer.getCurrentSpan()).isSameAs(BlankSpan.INSTANCE);
+    spanBuilder
+        .startSpanAndWrap(
```
adriancole: switch test source to java8 as it will make some of this stuff easier to see (also you can play with something that throws.)

api/src/main/java/io/opencensus/trace/SpanBuilder.java
```diff
@@ -223,6 +224,96 @@ public final Scope startScopedSpan() {
...
+   *         span.end();
+   *       }
+   *     }
+   * });
+   * }</pre>
+   *
+   * @param runnable the {@code Runnable} to run in the {@code Span}.
+   */
+  public final Runnable startSpanAndWrap(final Runnable runnable) {
+    final Span span = startSpan();
```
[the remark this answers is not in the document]
adriancole: it would be nice if you don't have to start a span in the builder. in brave we don't because there are a number of cases where you need a span visible before it is started (due to awkward callbacks ex http asyncclient)

api/src/main/java/io/opencensus/trace/SpanBuilder.java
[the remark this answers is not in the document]
adriancole: ps usually I create one javadoc with a full example, then make others refer to it (vs a lot of copy/paste) ex.

```
Like {@link #foo}, except this accepts a callable. Notably, callables can throw checked exceptions, so blah blah
```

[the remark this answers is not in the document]
sebright: @bogdandrutu I meant something like

```
   * This code starts a span and returns a Runnable with the span in scope:
   *
   * <pre><code>
   * Runnable newRunnable = tracer.spanBuilder("MyRunnableSpan").startSpanAndWrap(myRunnable);
   * </code></pre>
   *
   * <p>It is equivalent with the following code:
   *
   * <pre><code>
   * Span span = tracer.spanBuilder("MyRunnableSpan").startSpan();
   * Runnable newRunnable = tracer.wrap(span, new Runnable() {
   *    {@literal @}Override
   *     public void run() {
   *       try {
   *         myRunnable.run();
   *       } finally {
   *         span.end();
   *       }
   *     }
   * });
   * </code></pre>
```

api/src/main/java/io/opencensus/trace/Tracer.java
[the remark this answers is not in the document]
sebright: Why does the name need to be consistent with Context.wrap?  I think it is much more important to have consistency within OpenCensus than with gRPC Context.  Some users will be able to use these APIs without needing to manipulate the context directly.

api/src/main/java/io/opencensus/trace/CallableSpanHandle.java
bogdandrutu: I think this is safe to do.

api/src/main/java/io/opencensus/trace/Tracer.java
bogdandrutu: done.

api/src/test/java/io/opencensus/trace/SpanBuilderTest.java
bogdandrutu: Not in this PR. I will consider this.

api/src/main/java/io/opencensus/trace/Tracer.java
bogdandrutu: Done.

api/src/main/java/io/opencensus/trace/CurrentSpanUtils.java
bogdandrutu: Done.

bogdandrutu:
PTAL. I tried to address all the comments but I may missed some, feel free to ping me if I missed any.

api/src/main/java/io/opencensus/trace/SpanBuilder.java
```diff
@@ -223,6 +224,121 @@ public final Scope startScopedSpan() {
...
+   * and ends the {@code Span} after the {@code Callable} is run. See more details in {@link
+   * Tracer#wrap(Span, Callable)}.
+   *
+   * <p>Any error will end up as a {@link Status#UNKNOWN}.
+   *
+   * <p>This is a shorter version of the following code:
+   *
+   * <pre><code>
+   * Span span = tracer.spanBuilder("MyCallableSpan").startSpan();
+   * tracer.wrap(span, {@code new Callable<V>()} {
```
sebright: Suggestion: The Callable examples would be simpler if they used concrete types instead of `V`.

api/src/main/java/io/opencensus/trace/SpanBuilder.java
```diff
@@ -223,6 +224,121 @@ public final Scope startScopedSpan() {
...
+   *
+   * <p>Any error will end up as a {@link Status#UNKNOWN}.
+   *
+   * <pre><code>
+   * class MyClass {
+   *   private static Tracer tracer = Tracing.getTracer();
+   *   void handleRequest() {
+   *     new Tracer.spanBuilder("MyCallableSpan").startSpanAndCall({@code new Callable<V>()} {
+   *      {@literal @}Override
+   *       public V void call() throws Exception {
```
sebright: You forgot to remove the "void", here and in Tracer.java.

api/src/main/java/io/opencensus/trace/SpanBuilder.java
bogdandrutu: @sebright said

"I think "wrap" is a little unclear. One suggestion is "startSpanInRunnable"."

Currently we have options:
startSpanAndWrap
startSpanInRunnable/startSpanInCallable
wrap
decorate

I personally like the first one, but I would like to hear other opinions.

bogdandrutu:
From a previous comment:
@sebright said:

> I think withSpan would be more descriptive, and also more consistent with the existing API.

I like the "withSpan". The only one cons that I see is that the current "withSpan" applies right now but for runnable/callable will be for the future. That's why I think "wrap" makes more sense because it signals that the action is delayed to when the runnable/callable are executed.

api/src/test/java/io/opencensus/trace/CurrentSpanUtilsTest.java
```diff
@@ -86,25 +88,210 @@ public void withSpan_CloseDetachesAndEndsSpan() {
...
+          @Override
+          public Object call() throws Exception {
+            // When we run the runnable we will have the span in the current Context.
+            assertThat(CurrentSpanUtils.getCurrentSpan()).isSameAs(span);
+            throw exception;
+          }
+        };
+    try {
+      CurrentSpanUtils.wrap(span, callable, true).call();
+      Assert.fail();
```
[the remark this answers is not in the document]
bogdandrutu: I remembered now why:

  Did you mean 'Throwable e = expectThrows(Throwable.class, () -> CurrentSpanUtils.wrap(span, callable, true).call());'?
/home/travis/build/census-instrumentation/opencensus-java/api/src/test/java/io/opencensus/trace/CurrentSpanUtilsTest.java:293: error: [AssertionFailureIgnored] This assertion throws an AssertionError if it fails, which will be caught by an enclosing try block.
      Assert.fail();
                 ^
    (see http://errorprone.info/bugpattern/AssertionFailureIgnored)

Pushed a different fix.

api/src/main/java/io/opencensus/trace/SpanBuilder.java
sebright: I suggested startSpanInRunnable when I had assumed that startSpan was called inside the runnable, so I don't think it's the best choice now.  I think that startSpanAndWrap is probably the most descriptive, though the word "wrap" sounds too general.  I'll comment again if I think of anything else.

api/src/main/java/io/opencensus/trace/SpanBuilder.java
```diff
@@ -319,21 +325,17 @@ public final void startSpanAndRun(final Runnable runnable) {
...
-   *     new Tracer.spanBuilder("MyCallableSpan").startSpanAndCall({@code new Callable<V>()} {
-   *      {@literal @}Override
-   *       public V void call() throws Exception {
-   *         // The newly created Span is current in the Context.
-   *         return sendResult();
-   *       }
-   *     }));
-   *   }
-   * }
+   * MyResult myResult = tracer.spanBuilder("MyRunnableSpan").startSpanAndCall(myCallable);
```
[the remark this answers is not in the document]
bogdandrutu: Done.

bogdandrutu:
All comments should be addressed. PTAL.

api/src/main/java/io/opencensus/trace/CurrentSpanUtils.java
```diff
@@ -49,6 +51,30 @@ static Scope withSpan(Span span, boolean endSpan) {
...
+   *
+   * @param span the {@code Span} to be set as current.
+   * @param runnable the {@code Runnable} to run in the {@code Span}.
+   * @param endSpan if {@code true} the returned {@code Runnable} will close the {@code Span}.
+   * @return the wrapped {@code Runnable}.
+   */
+  static Runnable wrap(Span span, final Runnable runnable, boolean endSpan) {
```
adriancole: nit you have an extra final here and in the below method

api/src/main/java/io/opencensus/trace/CurrentSpanUtils.java
```diff
@@ -74,4 +100,73 @@ public void close() {
       }
     }
   }
+
+  private static final class RunnableInSpan implements Runnable {
+    private final Span span;
```
adriancole: nit since the type is private, everything else isn't accessible. Removing redundant visibility modifiers results in smaller bytecode iirc

api/src/main/java/io/opencensus/trace/CurrentSpanUtils.java
```diff
@@ -74,4 +100,73 @@ public void close() {
...
+      this.endSpan = endSpan;
+    }
+
+    @Override
+    public void run() {
+      Context origContext =
+          Context.current().withValue(ContextUtils.CONTEXT_SPAN_KEY, span).attach();
+      try {
+        runnable.run();
+      } catch (Throwable t) {
```
adriancole: if you catch (RuntimeException|Error e), then you don't need to use guava to throw

api/src/main/java/io/opencensus/trace/CurrentSpanUtils.java
```diff
@@ -74,4 +100,73 @@ public void close() {
...
+      this.endSpan = endSpan;
+    }
+
+    @Override
+    public V call() throws Exception {
+      Context origContext =
+          Context.current().withValue(ContextUtils.CONTEXT_SPAN_KEY, span).attach();
+      try {
+        return callable.call();
+      } catch (Exception e) {
```
adriancole: if you catch (RuntimeException|Exception|Error e), then you don't need to use guava to throw (also reduces a duplicate catch)

api/src/main/java/io/opencensus/trace/SpanBuilder.java
```diff
@@ -223,6 +224,129 @@ public final Scope startScopedSpan() {
...
+   *         span.end();
+   *       }
+   *     }
+   * });
+   * </code></pre>
+   *
+   * @param callable the {@code Callable} to run in the {@code Span}.
+   * @since 0.11.0
+   */
+  public final <V> Callable<V> startSpanAndWrap(final Callable<V> callable) throws Exception {
```
adriancole: this startSpanAndWrap stuff feels a bit odd to me, unless you are trying to do a lambda trace or something. Particularly wrapping a callable (ex for use in an executorService) usually would defer the start. Hard to see why we optimize for this as there isn't a lot of use cases to look at.

api/src/main/java/io/opencensus/trace/Tracer.java
```diff
@@ -150,6 +151,135 @@ public final Scope withSpan(Span span) {
...
+   *   }
+   * }
+   * </code></pre>
+   *
+   * @param span the {@code Span} to be set as current.
+   * @param runnable the {@code Runnable} to wrap in the {@code Span}.
+   * @return the wrapped {@code Runnable}.
+   * @since 0.11.0
+   */
+  public final Runnable wrap(Span span, Runnable runnable) {
```
adriancole: this api feels weird as I see a lot of wrap utilities which start and stop the span during the work method (ex callable.call). This wraps, but doesn't start it. I wouldn't use this api

adriancole:
minimum change I'd like to see is removal of the `startSpanAndWrap` methods as they feel awkward, particularly married to the executor types Runnable and Callable. Other things look good to me.

api/src/main/java/io/opencensus/trace/CurrentSpanUtils.java
bogdandrutu: You still want to catch any error to set the span status. If I don't use guava (because java6) the code will be like:

```
      try {
        runnable.run();
      } catch (Error e) {
        setErrorStatus(span, e);
        throw e;
      } catch (RuntimeException e) {
        setErrorStatus(span, e);
        throw e;
      } catch (Throwable t) {
        setErrorStatus(span, e);
        throw new RuntimeException("unexpected", t);
      } finally {
        Context.current().detach(origContext);
        if (endSpan) {
          span.end();
        }
      }
```

I think the current code is a bit nicer.

api/src/main/java/io/opencensus/trace/CurrentSpanUtils.java
bogdandrutu: See the previous comment.

api/src/main/java/io/opencensus/trace/SpanBuilder.java
bogdandrutu: The other 2 APIs are very useful, to move this PR forward I will remove this API and we can consider to add it back later if people ask for.

api/src/main/java/io/opencensus/trace/Tracer.java
bogdandrutu: I think this comment is based on the misunderstanding of the CurrentSpanUtil privacy. This is the only public "wrap" API.

bogdandrutu:
For wrap is easy to say "Waps X ...", for withSpan I will still have to say something "Wraps X ..." which makes me think that withSpan is not probably the best name for this. Do you have a proposal for the comment as well that makes more sense?

sebright:
I would write something like, "Returns a Runnable that runs the given task with the given span in the current context".  I don't think that the method name needs to match the comment, especially with the difference in length.

adriancole:
I would write something like, "Returns a Runnable that runs the given task
with the given span in the current context". I don't think that the method
name needs to match the comment, especially with the difference in length.

Javadoc sounds good

adriancole:
Ps I would consider retrolambda as it rewrites multicatch to jre6.

bogdandrutu:
PTAL. Changed the name.

api/src/main/java/io/opencensus/trace/SpanBuilder.java
```diff
@@ -238,25 +238,17 @@ public final Scope startScopedSpan() {
...
-   *       try {
-   *         myRunnable.run();
-   *       } finally {
-   *         span.end();
-   *       }
-   *     }
-   * });
+   * Runnable newRunnable = tracer.withSpan(span, myRunnable);
    * newRunnable.run();
+   * span.end();
```
sebright: Shouldn't span.end() be called in a finally block?

api/src/main/java/io/opencensus/trace/SpanBuilder.java
```diff
@@ -273,16 +265,9 @@ public final void startSpanAndRun(final Runnable runnable) {
...
-   *     public MyResult call() throws Exception {
-   *       try {
-   *         return myCallable.call();
-   *       } finally {
-   *         span.end();
-   *       }
-   *     }
+   * {@code Callable<MyResult>} newCallable = tracer.withSpan(span, myCallable);
    * MyResult myResult = newCallable.call();
+   * span.end();
```
sebright: Same here

bogdandrutu: done.

api/src/main/java/io/opencensus/trace/SpanBuilder.java
bogdandrutu: done.

Issue #597, opened 2017-09-08
Add SetStatus in the Span API.


bogdandrutu:
There are clear examples when setting the status must happen at different moments especially with the ScopedSpan API and the newly probable API for startWrapped.

PR #779, opened 2017-11-04
Add a setStatus method in the Span.

bogdandrutu:
Fixes https://github.com/census-instrumentation/opencensus-java/issues/597

### DIFF
--- a/api/src/main/java/io/opencensus/trace/EndSpanOptions.java
+++ b/api/src/main/java/io/opencensus/trace/EndSpanOptions.java
@@ -59,7 +59,10 @@ public abstract class EndSpanOptions {
   /**
    * Returns the status.
    *
-   * @return the status or {@code null} if not set.
+   * <p>If {@code null} then the {@link Span} will record the {@link Status} set via {@link
+   * Span#setStatus(Status)} or the default {@link Status#OK} if no status was set.
+   *
+   * @return the status.
    */
   @Nullable
   public abstract Status getStatus();
--- a/api/src/main/java/io/opencensus/trace/EndSpanOptions.java
+++ b/api/src/main/java/io/opencensus/trace/EndSpanOptions.java
@@ -38,7 +38,7 @@ public abstract class EndSpanOptions {
    * @return a new {@code Builder} with default options.
    */
   public static Builder builder() {
-    return new AutoValue_EndSpanOptions.Builder().setStatus(null).setSampleToLocalSpanStore(false);
+    return new AutoValue_EndSpanOptions.Builder().setSampleToLocalSpanStore(false);
   }
 
   /**
@@ -75,7 +75,7 @@ public abstract class EndSpanOptions {
      * @param status the status.
      * @return this.
      */
-    public abstract Builder setStatus(@Nullable Status status);
+    public abstract Builder setStatus(Status status);
 
     /**
      * If set to {@code true} this is equivalent with calling the {@link
--- a/api/src/main/java/io/opencensus/trace/EndSpanOptions.java
+++ b/api/src/main/java/io/opencensus/trace/EndSpanOptions.java
@@ -16,11 +16,10 @@
 
 package io.opencensus.trace;
 
-import static com.google.common.base.Preconditions.checkNotNull;
-
 import com.google.auto.value.AutoValue;
 import io.opencensus.common.ExperimentalApi;
 import java.util.Collection;
+import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 
 /**
@@ -39,9 +38,7 @@ public abstract class EndSpanOptions {
    * @return a new {@code Builder} with default options.
    */
   public static Builder builder() {
-    return new AutoValue_EndSpanOptions.Builder()
-        .setStatus(Status.OK)
-        .setSampleToLocalSpanStore(false);
+    return new AutoValue_EndSpanOptions.Builder().setStatus(null).setSampleToLocalSpanStore(false);
   }
 
   /**
@@ -53,7 +50,8 @@ public abstract class EndSpanOptions {
    * io.opencensus.trace.export.SampledSpanStore#registerSpanNamesForCollection(Collection)} API
    * instead.
    *
-   * @return this.
+   * @return {@code true} if the name of the {@code Span} should be registered to the {@code
+   *     io.opencensus.trace.export.SampledSpanStore}.
    */
   @ExperimentalApi
   public abstract boolean getSampleToLocalSpanStore();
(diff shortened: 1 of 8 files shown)


bogdandrutu:
This PR is not about (for the moment) being able to handle Throwable with Status. Is about allowing users to set the Status before the end of the Span.

Status can be modified to support the Throwable case later.

api/src/main/java/io/opencensus/trace/EndSpanOptions.java
```diff
@@ -53,16 +50,18 @@ public static Builder builder() {
...
+   *     io.opencensus.trace.export.SampledSpanStore}.
    */
   @ExperimentalApi
   public abstract boolean getSampleToLocalSpanStore();
 
   /**
    * Returns the status.
    *
-   * @return the status.
+   * @return the status or {@code null} if not set.
```
songy23: IIUC this will also return null if the status is explicitly set to null.

api/src/main/java/io/opencensus/trace/Span.java
```diff
@@ -158,6 +158,20 @@ public final void addAnnotation(String description) {
   public abstract void addLink(Link link);
 
   /**
+   * Sets the {@link Status} to the {@code Span}.
+   *
+   * <p>If used, this will override the default {@code Span} status. Default is {@link Status#OK}.
+   *
+   * <p>If called multiple times, or status set via
```
songy23: Seems an incomplete sentence here.

api/src/main/java/io/opencensus/trace/EndSpanOptions.java
```diff
@@ -71,12 +70,12 @@ public static Builder builder() {
...
      * Sets the status for the {@link Span}.
      *
-     * <p>If set, this will override the default {@code Span} status. Default is {@link Status#OK}.
+     * <p>If set, this will override the status set via {@link Span#setStatus(Status)}.
      *
      * @param status the status.
      * @return this.
      */
-    public abstract Builder setStatus(Status status);
+    public abstract Builder setStatus(@Nullable Status status);
```
songy23: Probably document why we want to set status to null here.

api/src/main/java/io/opencensus/trace/EndSpanOptions.java
```diff
@@ -39,9 +38,7 @@
    * @return a new {@code Builder} with default options.
    */
   public static Builder builder() {
-    return new AutoValue_EndSpanOptions.Builder()
-        .setStatus(Status.OK)
-        .setSampleToLocalSpanStore(false);
+    return new AutoValue_EndSpanOptions.Builder().setStatus(null).setSampleToLocalSpanStore(false);
```
sebright: Why does the status need to be set to null here?

api/src/main/java/io/opencensus/trace/Span.java
sebright: Is this a comment about the EndSpanOptions? I think it is important to explain how the two status methods interact.

api/src/test/java/io/opencensus/trace/EndSpanOptionsTest.java
```diff
@@ -26,19 +26,20 @@
 /** Unit tests for {@link EndSpanOptions}. */
 @RunWith(JUnit4.class)
 public class EndSpanOptionsTest {
-  @Test(expected = NullPointerException.class)
-  public void setNullStatus() {
-    EndSpanOptions.builder().setStatus(null).build();
```
sebright: It might be useful to test clearing the status by setting the status and then setting it back to null.

impl_core/src/main/java/io/opencensus/implcore/trace/export/SampledSpanStoreImpl.java
```diff
@@ -173,17 +173,13 @@ private Bucket getErrorBucket(CanonicalCode code) {
 
     private void considerForSampling(SpanImpl span) {
       Status status = span.getStatus();
-      // Null status means running Span, this should not happen in production, but the library
-      // should not crash because of this.
-      if (status != null) {
```
sebright: Why does this method no longer need to check whether the span in running?

api/src/main/java/io/opencensus/trace/EndSpanOptions.java
bogdandrutu: Done.

api/src/main/java/io/opencensus/trace/EndSpanOptions.java
bogdandrutu: Better to not allow it to be set to null. So comment is fine.

api/src/main/java/io/opencensus/trace/EndSpanOptions.java
bogdandrutu: Better to not allow it to be set to null.

api/src/main/java/io/opencensus/trace/Span.java
bogdandrutu: Done.

bogdandrutu: Done.

api/src/test/java/io/opencensus/trace/EndSpanOptionsTest.java
bogdandrutu: EndSpanOptions.Builder#setStatus does not accept null Status anymore. No reason for the builder to support set/unset/set in my opinion.

impl_core/src/main/java/io/opencensus/implcore/trace/export/SampledSpanStoreImpl.java
bogdandrutu: My bad. Tried to do something else. Reverted

api/src/main/java/io/opencensus/trace/Span.java
```diff
@@ -158,6 +158,22 @@ public final void addAnnotation(String description) {
...
+   *
+   * <p>If used, this will override the default {@code Span} status. Default is {@link Status#OK}.
+   *
+   * <p>Only the value of the last call will be recorded, and implementations are free to ignore
+   * previous calls. If the status is set via {@link EndSpanOptions.Builder#setStatus(Status)} that
+   * will always be the last call.
+   *
+   * @param status the {@link Status} to set.
+   */
+  public void setStatus(Status status) {
```
sebright: Should this method also disallow null?

api/src/main/java/io/opencensus/trace/EndSpanOptions.java
sebright: It looks like setStatus never checks for null.  I don't know if it's possible with the AutoValue builder, though.

bogdandrutu: Don't think this is needed.

api/src/main/java/io/opencensus/trace/Span.java
bogdandrutu: We don't do this for any method in Span. I created an issues to add this https://github.com/census-instrumentation/opencensus-java/issues/785

PR #780, opened 2017-11-05
Wrap the execution of URL#getContent in a trace span.

ubschmidt2:


contrib/agent/src/main/java/io/opencensus/contrib/agent/instrumentation/UrlInstrumentation.java
```diff
@@ -0,0 +1,96 @@
...
+     *
+     * <p>NB: This method is never called as is. Instead, Byte Buddy copies the method's bytecode
+     * into Executor#execute.
+     *
+     * @see Advice
+     */
+    @Advice.OnMethodEnter
+    @SuppressFBWarnings("UPM_UNCALLED_PRIVATE_METHOD")
+    @MustBeClosed
+    private static Closeable enter(@Advice.Origin("#t\\##m") String classAndMethodName) {
```
adriancole: "#t\\##m" might be obvious, but maybe comment what this is doing

adriancole: nevermind.. guessing it is escaping the hash

contrib/agent/src/main/java/io/opencensus/contrib/agent/instrumentation/UrlInstrumentation.java
```diff
@@ -0,0 +1,96 @@
...
+      return TraceTrampoline.startScopedSpan(classAndMethodName);
+    }
+
+    /**
+     * Closes the current span when exiting the method.
+     *
+     * <p>NB: This method is never called as is. Instead, Byte Buddy copies the method's bytecode
+     * into Executor#execute.
+     *
+     * <p>NB: By default, any {@link Throwable} thrown during the advice's execution is silently
```
adriancole: needs a test case (eventhough I trust the span indeed closes on throw)

ubschmidt2: Added a simple test that the span is closed in case of an exception.

PR #781, opened 2017-11-05
Close the span with a status derived from the (optional) Throwable.

ubschmidt2:


contrib/agent/src/main/java/io/opencensus/contrib/agent/bootstrap/TraceStrategy.java
```diff
@@ -36,4 +37,13 @@
    */
   @MustBeClosed
   Closeable startScopedSpan(String spanName);
```
bogdandrutu: I would suggest to use Scope instead of Closeable to avoid having to always catch exception.

bogdandrutu: Good comments. I think we should actually have a NonThrowingCloseable to simplify the code, what do you think?

ubschmidt2: Added Javadoc for that. Classes in the .bootstrap package are loaded by the bootstrap classloader and thus cannot directly see the classes loaded by the system classloader.

contrib/agent/src/main/java/io/opencensus/contrib/agent/instrumentation/TraceStrategyImpl.java
```diff
@@ -38,4 +42,21 @@ public Closeable startScopedSpan(String spanName) {
...
   }
+
+  @Override
+  public void endScope(Closeable scope, @Nullable Throwable throwable) {
+    checkNotNull(scope, "scope");
+
+    if (throwable != null) {
+      Tracing.getTracer()
+          .getCurrentSpan()
+          .end(EndSpanOptions.builder().setStatus(Status.UNKNOWN).build());
```
bogdandrutu: I would suggest:

.end(EndSpanOptions.builder().setStatus(Status.UNKNOWN.withDescription(throwable.getMessage()).build());

ubschmidt2: Sounds good! Done.

ubschmidt2:
Re non-throwing: future PR sounds good.

PR #789, opened 2017-11-07
Update Stats implementation to use Stats state

songy23:
Implement stats collection state. Stats library has 3 major operations that are affected by state: `registerView()`, `record()` and `getView()`. I think there are 2 options for implementing state for them:

1. When state is `DISABLED`, implement all the 3 operations above as no-op. `registerView()` won't really add the view to `MeasureToViewMap`, and `getView()` will always return null. (It's not possible to return empty `ViewData` in this case, since no `View` will be preserved.)

2. When state is `DISABLED`, we still add `View`s to `MeasureToViewMap` when `registerView()` is called. `record()` will be ignored, and `getView()` will return an empty `ViewData` for registered `View`s. When users re-enabled stats collection, they don't need to register views again.

I think either of these two makes sense. ~~I went with the first option, since I think we expect users only set state at the very beginning and should not restore stats collection in the middle of other operations.~~

EDIT: update to go with option 2, to be more consistent with the No-op impl. Also, setting state to DISABLED will clear all previous stats.

### DIFF
--- a/core_impl/src/main/java/io/opencensus/implcore/stats/CurrentStatsState.java
+++ b/core_impl/src/main/java/io/opencensus/implcore/stats/CurrentStatsState.java
@@ -47,8 +47,15 @@ public final class CurrentStatsState {
     return currentState;
   }
 
-  synchronized void set(StatsCollectionState state) {
+  // Sets current state to the given state. Returns true if the current state is changed, false
+  // otherwise.
+  synchronized boolean set(StatsCollectionState state) {
     checkState(!isRead, "State was already read, cannot set state.");
-    currentState = checkNotNull(state, "state");
+    if (state == currentState) {
+      return false;
+    } else {
+      currentState = checkNotNull(state, "state");
+      return true;
+    }
   }
 }
--- a/core_impl/src/main/java/io/opencensus/implcore/stats/CurrentStatsState.java
+++ b/core_impl/src/main/java/io/opencensus/implcore/stats/CurrentStatsState.java
@@ -17,6 +17,7 @@
 package io.opencensus.implcore.stats;
 
 import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
 
 import io.opencensus.stats.StatsCollectionState;
 import io.opencensus.stats.StatsComponent;
@@ -30,12 +31,18 @@ import javax.annotation.concurrent.ThreadSafe;
 @ThreadSafe
 public final class CurrentStatsState {
   private volatile StatsCollectionState currentState = StatsCollectionState.ENABLED;
+  private volatile boolean isRead;
 
   public StatsCollectionState get() {
     return currentState;
   }
 
+  void setRead() {
+    isRead = true;
+  }
+
   void set(StatsCollectionState state) {
+    checkState(!isRead, "State was already read, cannot set state.");
     currentState = checkNotNull(state, "state");
   }
 }
--- a/core_impl/src/main/java/io/opencensus/implcore/stats/CurrentStatsState.java
+++ b/core_impl/src/main/java/io/opencensus/implcore/stats/CurrentStatsState.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2017, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.implcore.stats;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import io.opencensus.stats.StatsCollectionState;
+import io.opencensus.stats.StatsComponent;
+import javax.annotation.concurrent.ThreadSafe;
+
+/**
+ * The current {@link StatsCollectionState} for a {@link StatsComponent}.
+ *
+ * <p>This class allows different stats classes to share the state in a thread-safe way.
+ */
+@ThreadSafe
+public final class CurrentStatsState {
+  private volatile StatsCollectionState currentState = StatsCollectionState.ENABLED;
+
+  public StatsCollectionState get() {
+    return currentState;
+  }
+
+  void set(StatsCollectionState state) {
+    currentState = checkNotNull(state, "state");
+  }
+}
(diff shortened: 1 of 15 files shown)


core_impl/src/main/java/io/opencensus/implcore/stats/CurrentStatsState.java
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * The current {@link StatsCollectionState} for a {@link StatsComponent}.
+ *
+ * <p>This class allows different stats classes to share the state in a thread-safe way.
+ */
+@ThreadSafe
+public final class CurrentStatsState {
+  private volatile StatsCollectionState currentState = StatsCollectionState.ENABLED;
+  private volatile boolean isRead;
```
sebright: Suggestion: You could encapsulate the `isRead` state in this class by exposing the methods `setState`, `getState`, and `getInternal`.  Then `getState` could update `isRead`.

core_impl/src/main/java/io/opencensus/implcore/stats/CurrentStatsState.java
```diff
@@ -0,0 +1,48 @@
...
+    return currentState;
+  }
+
+  void setRead() {
+    isRead = true;
+  }
+
+  void set(StatsCollectionState state) {
+    checkState(!isRead, "State was already read, cannot set state.");
+    currentState = checkNotNull(state, "state");
```
sebright: If the goal is to ensure that getState always returns the same value, then I think this class should synchronize access to the two fields.  Otherwise, the state could be read between these two lines.

core_impl/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
```diff
@@ -208,11 +213,24 @@ void record(TagContext context, double value, Timestamp timestamp) {
...
+      if (state == StatsCollectionState.ENABLED) {
+        return ViewData.create(
+            super.view,
+            createAggregationMap(tagValueAggregationMap, super.view.getMeasure()),
+            CumulativeData.create(start, now));
+      } else {
+        // If Stats state is DISABLED, return an empty ViewData.
+        return ViewData.create(
+            super.view,
+            Maps.<List<TagValue>, AggregationData>newHashMap(),
```
sebright: This line could use Collections.emptyMap().

core_impl/src/main/java/io/opencensus/implcore/stats/StatsComponentImplBase.java
```diff
@@ -55,11 +55,15 @@ public StatsRecorderImpl getStatsRecorder() {
...
+    return state.get();
   }
 
   @Override
   public void setState(StatsCollectionState newState) {
-    state = Preconditions.checkNotNull(newState, "newState");
+    state.set(Preconditions.checkNotNull(newState, "newState"));
+    if (newState == StatsCollectionState.DISABLED) {
+      viewManager.clearStats();
+    }
```
sebright: I think this method also needs to reset the cumulative view start times when stats collection is changed from disabled to enabled.

core_impl/src/main/java/io/opencensus/implcore/stats/StatsManager.java
```diff
@@ -32,25 +33,34 @@
...
   }
 
   ViewData getView(View.Name viewName) {
-    return measureToViewMap.getView(viewName, clock);
+    return measureToViewMap.getView(viewName, clock, state.get());
   }
 
   void record(TagContext tags, MeasureMapInternal measurementValues) {
-    queue.enqueue(new StatsEvent(this, tags, measurementValues));
+    if (state.get() == StatsCollectionState.ENABLED) {
```
sebright: I think we could improve performance when stats is disabled by checking the state in StatsRecorderImpl.newMeasureMap and returning a no-op MeasureMap in that case.  It doesn't need to be in this PR, though.

core_impl/src/test/java/io/opencensus/implcore/stats/ViewManagerImplTest.java
```diff
@@ -781,6 +783,105 @@ public void testGetCumulativeViewDataWithoutBucketBoundaries() {
...
+            Arrays.asList(KEY),
+            CUMULATIVE);
+    testFailedToRegisterView(
+        view1, view2, "A different view with the same name is already registered");
+  }
+
+  @Test
+  public void settingStateToDisabledWillClearStats() {
+    View view1 = createCumulativeView(VIEW_NAME, MEASURE_DOUBLE, MEAN, Arrays.asList(KEY));
+    View view2 =
```
sebright: Does this test create two views in order to test interval and cumulative views?  I think it would be better to split this into two separate tests with names describing the difference.

core_impl/src/test/java/io/opencensus/implcore/stats/ViewManagerImplTest.java
```diff
@@ -854,32 +855,61 @@ public void registerDifferentViewWithSameNameWithStatsDisabled() {
...
-    assertThat(viewManager.getView(VIEW_NAME_2).getAggregationMap()).isEmpty();
-    assertThat(viewManager.getView(VIEW_NAME_2)).isNotEqualTo(createEmptyViewData(view2));
+
+    Timestamp timestamp4 = Timestamp.create(4, 0);
+    clock.setTime(timestamp4);
+    // This ViewData does not have any stats, but it should not be an empty ViewData, since it has
+    // non-zero TimeStamps.
+    ViewData viewData = viewManager.getView(view.getName());
+    assertThat(viewData.getAggregationMap()).isEmpty();
+    assertThat(viewData).isNotEqualTo(createEmptyViewData(view));
```
sebright: Should this assertion be removed?  It looks like the assertions below check the same thing but are more specific.

songy23: Right, removed.

core_impl/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
```diff
@@ -317,6 +327,11 @@ void clearStats() {
       }
     }
 
+    @Override
+    void resumeStatsCollection(Timestamp now) {
+      refreshBucketList(now);
```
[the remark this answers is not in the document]
songy23: I expect users will begin recording stats as soon as they re-enable stats collection, and it's better to refresh buckets as early as possible to be ready for `record()`. `refreshBucketList()` is a somewhat heavy operation, and we have already indicated that turning stats state on/off is also heavy-weight, I think it's reasonable to do the refreshing here. I've left a comment here.

core_impl/src/main/java/io/opencensus/implcore/stats/StatsComponentImplBase.java
```diff
@@ -55,15 +55,20 @@ public StatsRecorderImpl getStatsRecorder() {
...
     return state.get();
   }
 
   @Override
   public void setState(StatsCollectionState newState) {
+    StatsCollectionState oldState = state.getInternal();
+    if (newState == oldState) {
+      return;
+    }
     state.set(Preconditions.checkNotNull(newState, "newState"));
```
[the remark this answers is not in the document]
songy23: Make sense, updated.

songy23:
Thanks!

PR #790, opened 2017-11-08
Deprecate Tags.setState, and throw an exception when it is called after getState.

sebright:
This commit is related to #608.  It deprecates Tags.setState and
TagsComponent.setState and makes NoopTagsComponent.setState and
TagsComponentImplBase.setState throw IllegalStateException when they are called
after getState.

_________________________________________________________________

This PR is similar to #789, but applied to tagging.

### DIFF
--- a/api/src/main/java/io/opencensus/tags/TagsComponent.java
+++ b/api/src/main/java/io/opencensus/tags/TagsComponent.java
@@ -37,6 +37,9 @@ public abstract class TagsComponent {
    * <p>When no implementation is available, {@code getState} always returns {@link
    * TaggingState#DISABLED}.
    *
+   * <p>Once {@link #getState()} is called, subsequent calls to {@link #setState(TaggingState)} will
+   * throw an {@code IllegalStateException}.
+   *
    * @return the current {@code TaggingState}.
    */
   public abstract TaggingState getState();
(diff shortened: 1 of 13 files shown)


core_impl/src/main/java/io/opencensus/implcore/tags/CurrentTaggingState.java
```diff
@@ -29,13 +31,25 @@
...
+
+  @GuardedBy("this")
+  private boolean isRead;
+
+  public synchronized TaggingState get() {
+    isRead = true;
+    return getInternal();
+  }
+
+  public synchronized TaggingState getInternal() {
```
songy23: This can be package-private.

sebright: This method is called from TagContextBinarySerializerImpl, which is in a different package.

api/src/main/java/io/opencensus/tags/TagsComponent.java
```diff
@@ -37,16 +37,25 @@
    * <p>When no implementation is available, {@code getState} always returns {@link
    * TaggingState#DISABLED}.
    *
+   * <p>Once {@link #getState()} is called, subsequent calls to {@link #setState(TaggingState)} will
```
bogdandrutu: You can say that user of this API can cache the value.

core_impl/src/main/java/io/opencensus/implcore/tags/CurrentTaggingState.java
```diff
@@ -29,13 +31,25 @@
...
-  private volatile TaggingState currentState = TaggingState.ENABLED;
 
-  public TaggingState get() {
+  @GuardedBy("this")
+  private TaggingState currentState = TaggingState.ENABLED;
+
+  @GuardedBy("this")
+  private boolean isRead;
+
+  public synchronized TaggingState get() {
```
bogdandrutu: I think it is fine for the moment, but I would like to create an issue to remove this global lock.

impl_core/src/main/java/io/opencensus/implcore/tags/TagsComponentImplBase.java
```diff
@@ -50,6 +50,7 @@ public TaggingState getState() {
   }
 
   @Override
+  @Deprecated
```
bogdandrutu: Same comment from stats. This is not visible so no need to mark as deprecated.

sebright:
> I don't believe in deprecated internal methods, but up to you, I don't mind either way.

I'll leave it as it is, since it also means that I don't need to suppress a deprecation warning.

sebright:
Thanks for the reviews!

Issue #762, opened 2017-11-01
Implement API for enabling/disabling stats collection.


sebright:
#712 added the API to get and set the state, but the state currently has no effect.  This is part of #608.

PR #792, opened 2017-11-08
Deprecate Stats.setState, and throw an exception when it is called after getState.

songy23:
Changes on Stats state APIs, similar to https://github.com/census-instrumentation/opencensus-java/pull/790.

Closes https://github.com/census-instrumentation/opencensus-java/issues/762.

### DIFF
--- a/api/src/test/java/io/opencensus/stats/StatsTest.java
+++ b/api/src/test/java/io/opencensus/stats/StatsTest.java
@@ -82,13 +82,4 @@ public final class StatsTest {
     thrown.expectMessage("state");
     Stats.setState(null);
   }
-
-  @Test
-  @SuppressWarnings("deprecation")
-  public void disallowSetStateAfterGetState() {
-    Stats.getState();
-    thrown.expect(IllegalStateException.class);
-    thrown.expectMessage("State was already read, cannot set state.");
-    Stats.setState(StatsCollectionState.DISABLED);
-  }
 }
--- a/api/src/test/java/io/opencensus/stats/StatsTest.java
+++ b/api/src/test/java/io/opencensus/stats/StatsTest.java
@@ -74,8 +74,18 @@ public final class StatsTest {
     assertThat(Stats.getState()).isEqualTo(StatsCollectionState.DISABLED);
   }
 
-  @Test(expected = NullPointerException.class)
+  @Test
   public void setState_DisallowsNull() {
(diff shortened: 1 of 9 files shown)


api/src/test/java/io/opencensus/stats/StatsTest.java
```diff
@@ -74,8 +74,18 @@ public void setState_IgnoresInput() {
...
-  @Test(expected = NullPointerException.class)
+  @Test
   public void setState_DisallowsNull() {
+    thrown.expect(NullPointerException.class);
+    thrown.expectMessage("state");
     Stats.setState(null);
   }
+
+  @Test
+  public void disallowSetStateAfterGetState() {
```
sebright: It looks like this test could cause other tests to fail if the tests ran in the right order, since Stats has global state.  It would probably be better to only test this functionality in NoopStatsTest.  Another option is to combine all of the state testing into one method and add a comment.

songy23: Oh that's true, I've removed this test since this should have been covered in NoopStatsTest.

core_impl/src/main/java/io/opencensus/implcore/stats/StatsComponentImplBase.java
```diff
@@ -59,6 +59,7 @@ public StatsCollectionState getState() {
   }
 
   @Override
+  @Deprecated
```
bogdandrutu: Don't need to deprecated the internal function.

songy23: I think this method still needs to be marked as Deprecated, since it's a public method (though it's in impl) and users can call it if they have compile-time dependency on opencensus-impl.

bogdandrutu: Yes, but we do not promise/guarantee any backwards compatibility in this package. So we can remove this at any moment.

songy23: Make sense to me. Removed.

songy23:
I reverted the javadoc that mentioned `State` could be cached, to be consistent with https://github.com/census-instrumentation/opencensus-java/pull/790.

PR #794, opened 2017-11-09
Add ViewManager.getAllExportedViews().

sebright:
This method currently returns all registered cumulative views.  We could add a
way to specify which views should be exported later.  The API is currently
needed for the StackDriver exporter.

The NoopViewManager also returns all cumulative views.  However, the exporter
could choose to not export data when stats collection isn't enabled.

### DIFF
--- a/api/src/main/java/io/opencensus/stats/NoopStats.java
+++ b/api/src/main/java/io/opencensus/stats/NoopStats.java
@@ -22,10 +22,12 @@ import static com.google.common.base.Preconditions.checkState;
 
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Maps;
+import com.google.common.collect.Sets;
 import io.opencensus.common.Functions;
 import io.opencensus.common.Timestamp;
 import io.opencensus.stats.Measure.MeasureDouble;
 import io.opencensus.stats.Measure.MeasureLong;
+import io.opencensus.stats.View.AggregationWindow;
 import io.opencensus.stats.ViewData.AggregationWindowData;
 import io.opencensus.stats.ViewData.AggregationWindowData.CumulativeData;
 import io.opencensus.stats.ViewData.AggregationWindowData.IntervalData;
@@ -34,6 +36,7 @@ import io.opencensus.tags.TagValue;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.GuardedBy;
 import javax.annotation.concurrent.Immutable;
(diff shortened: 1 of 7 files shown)


core_impl/src/main/java/io/opencensus/implcore/stats/ViewManagerImpl.java
```diff
@@ -38,6 +42,17 @@ public ViewData getView(View.Name viewName) {
     return statsManager.getView(viewName);
   }
 
+  @Override
+  public Set<View> getAllExportedViews() {
```
bogdandrutu: The way how I think this will be used is by querying this every X seconds before pulling the data (in the exporter) correct? Probably worth having a cached version of this result in the StatsManager:

private volatile List<View> exportedViews;  // gets updated all the time when register/unregister a view and you have the sync there so this will be consistent.

api/src/main/java/io/opencensus/stats/ViewManager.java
```diff
@@ -42,4 +43,15 @@
...
   public abstract ViewData getView(View.Name view);
+
+  /**
+   * Returns all registered views that should be exported.
+   *
+   * <p>This method should be used by any stats exporter that automatically exports data for views
+   * registered with the {@link ViewManager}.
+   *
+   * @return all registered views that should be exported.
+   * @since 0.9
```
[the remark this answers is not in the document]
sebright: I think it would be useful.  It is also easier to add now that we are releasing all of the APIs.

core_impl/src/main/java/io/opencensus/implcore/stats/ViewManagerImpl.java
sebright: I cached the set of views and recalculated them when they are needed.  It does add some complexity, though.

api/src/main/java/io/opencensus/stats/NoopStats.java
```diff
@@ -180,5 +183,18 @@ public ViewData getView(View.Name name) {
         }
       }
     }
+
+    @Override
+    public Set<View> getAllExportedViews() {
```
bogdandrutu: Why does not return empty set for no-op?  I think this is more like "testing" implementation. If you need a testing implementation you should add one in the testing package, and keep the no-op as no-op implementation.

impl_core/src/main/java/io/opencensus/implcore/stats/MeasureToViewMap.java
```diff
@@ -57,35 +61,41 @@
...
-  private synchronized MutableViewData getMutableViewData(View.Name viewName) {
-    View view = registeredViews.get(viewName);
-    if (view == null) {
-      return null;
+  Set<View> getExportedViews() {
+    Set<View> views = exportedViews;
+    if (views == null) {
+      synchronized (this) {
+        exportedViews =
+            views = Collections.unmodifiableSet(filterExportedViews(registeredViews.values()));
```
bogdandrutu: Double wrap with unmodifiableSet. Please remove one of them (filterExportedViews also wraps with unmodifiableSet).

sebright: Oops!  I removed one call.

sebright:
Thanks!  I rebased and merged.

PR #797, opened 2017-11-09
Increase timeout for an asynchronous verification

songy23:
`StackdriverStatsExporterTest.registerViewAndExport()` is slower on some machine. Previously it has an 1 second timeout for the asynchronous verification, however this limit seems too strict. Relax the timeout to 2 seconds to make it more stable.

Test failures:
https://travis-ci.org/census-instrumentation/opencensus-java/jobs/299758118

### DIFF
--- a/exporters/stats/stackdriver/src/test/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporterTest.java
+++ b/exporters/stats/stackdriver/src/test/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporterTest.java
@@ -165,8 +165,9 @@ public class StackdriverStatsExporterTest {
       StackdriverStatsExporter.registerView(view);
 
       verify(mockStub, times(1)).createMetricDescriptorCallable();
-      // The timeout for verifying createTimeSeries needs to match the export interval of exporter.
-      verify(mockStub, timeout(1000).times(1)).createTimeSeriesCallable();
+      // This verification is slower on some machine. Relax the timeout be greater than the export
+      // interval to make this more steady.
+      verify(mockStub, timeout(2000).times(1)).createTimeSeriesCallable();
 
       MetricDescriptor descriptor = StackdriverExportUtils.createMetricDescriptor(view, PROJECT_ID);
       List<TimeSeries> timeSeries =
--- a/exporters/stats/stackdriver/src/test/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporterTest.java
+++ b/exporters/stats/stackdriver/src/test/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporterTest.java
@@ -18,6 +18,7 @@ package io.opencensus.exporter.stats.stackdriver;
 
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.atLeast;
 import static org.mockito.Mockito.doNothing;
 import static org.mockito.Mockito.doReturn;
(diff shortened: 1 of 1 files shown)


exporters/stats/stackdriver/src/test/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporterTest.java
```diff
@@ -165,8 +165,9 @@ public void registerViewAndExport() throws IOException {
       StackdriverStatsExporter.registerView(view);
 
       verify(mockStub, times(1)).createMetricDescriptorCallable();
-      // The timeout for verifying createTimeSeries needs to match the export interval of exporter.
-      verify(mockStub, timeout(1000).times(1)).createTimeSeriesCallable();
+      // This verification is slower on some machine. Relax the timeout be greater than the export
+      // interval to make this more steady.
+      verify(mockStub, timeout(2000).times(1)).createTimeSeriesCallable();
```
sebright: Shouldn't this test use `atLeast(1)` instead of `times(1)` when verifying `createTimeSeriesCallable()`, since the method could be called twice in the new time interval?

sebright: It would help to give more detail about what this line is testing, such as "createTimeSeriesCallable() should be called once per second.  This test uses a timeout that is longer than the export interval in order to decrease the chance of failures due to timing."

songy23: Done. Actually `timeout()` only waits for the first invocation, so `atLeast(1)` and `times(1)` will be the same here.

songy23: Sure, documented.

exporters/stats/stackdriver/src/test/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporterTest.java
```diff
@@ -165,8 +166,10 @@ public void registerViewAndExport() throws IOException {
       StackdriverStatsExporter.registerView(view);
 
       verify(mockStub, times(1)).createMetricDescriptorCallable();
-      // The timeout for verifying createTimeSeries needs to match the export interval of exporter.
-      verify(mockStub, timeout(1000).times(1)).createTimeSeriesCallable();
+      // createTimeSeriesCallable() should be called once per second. This test uses a timeout that
+      // is longer than the export interval in order to decrease the chance of failures due to
+      // timing.
+      verify(mockStub, timeout(2000).atLeast(1)).createTimeSeriesCallable();
```
bogdandrutu: Can we make this test deterministic and avoid using sleep?

songy23: Another option would be to use CountDownLatch to make the tests more parallel, but that would make the test more complicated and I'm not sure if it's worthwhile.

songy23: I've left a TODO here to see if there's a better way to fix this.

songy23:
Going to merge this for now to unblock other pending PRs. These tests will be updated once https://github.com/census-instrumentation/opencensus-java/pull/794 is merged.

PR #802, opened 2017-11-11
Upgrade to Error Prone 2.1.2.

sebright:
#### Force gradle-errorprone-plugin to use newer Error Prone, version 2.1.2.

Other changes:

 - Remove references to warnings that were removed from Error Prone.

 - Suppress occurrences of a new warning, JdkObsolete. All of the warnings are
   about uses of LinkedList. We need to decide whether to use a different class
   or continue suppressing the warnings. I left TODOs, since each occurrence may
   need to be handled differently.

________________________________________________________________

@bogdandrutu I wasn't sure how to update the build file, so let me know if there is a better way to specify the Error Prone version.

### DIFF
--- a/api/src/test/java/io/opencensus/trace/export/SpanDataTest.java
+++ b/api/src/test/java/io/opencensus/trace/export/SpanDataTest.java
@@ -34,9 +34,9 @@ import io.opencensus.trace.export.SpanData.Attributes;
 import io.opencensus.trace.export.SpanData.Links;
 import io.opencensus.trace.export.SpanData.TimedEvent;
 import io.opencensus.trace.export.SpanData.TimedEvents;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Random;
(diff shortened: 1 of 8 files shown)


build.gradle
```diff
@@ -34,6 +34,12 @@ subprojects {
     if (JavaVersion.current().isJava8Compatible()) {
         apply plugin: "net.ltgt.errorprone"
         apply plugin: 'com.github.sherter.google-java-format'
+
+        configurations.errorprone {
+            dependencies {
+                errorprone 'com.google.errorprone:error_prone_core:2.1.2'
```
bogdandrutu: you have some versions of errorprone at lines: 126, 158. Please use a version variable and be consistent.

Also I think you can remove this and just update the one at line 158.

api/src/test/java/io/opencensus/trace/export/SpanDataTest.java
```diff
@@ -68,11 +68,21 @@
           TraceId.generateRandomId(random), SpanId.generateRandomId(random), TraceOptions.DEFAULT);
   private final SpanId parentSpanId = SpanId.generateRandomId(random);
   private final Map<String, AttributeValue> attributesMap = new HashMap<String, AttributeValue>();
+
+  // TODO: Decide whether to use a different class instead of LinkedList.
```
bogdandrutu: In tests is not a problem. Please use what is suggested.

build.gradle
sebright: I missed the version on line 158!  I was surprised that the Error Prone plugin wasn't using the latest version.  I updated that version number instead.

api/src/test/java/io/opencensus/trace/export/SpanDataTest.java
sebright: Okay, this test was using List for each of the field types, so I just replaced LinkedList with ArrayList.

api/src/test/java/io/opencensus/trace/export/SpanDataTest.java
```diff
@@ -34,9 +34,9 @@
 import io.opencensus.trace.export.SpanData.Links;
 import io.opencensus.trace.export.SpanData.TimedEvent;
 import io.opencensus.trace.export.SpanData.TimedEvents;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.LinkedList;
```
adriancole: haha yeah.. yanked a bunch of these as well

sebright:
Thanks for the reviews!  I rebased, and I'll merge when it's green.

PR #803, opened 2017-11-13
 Remove StackdriverStatsExporter.registerView() and rearrange unit tests.

songy23:
~~This PR depends on #794.~~

Remove `StackdriverStatsExporter.registerView()`. Instead, export views that are returned by `ViewManager.getAllExportedViews()`. Now `StackdriverStatsExporter` only has factory methods, and uploading `View` to Stackdriver will be handled by the worker thread.

In addition, due to the changes above, I've also moved tests on `register()` and `export()` to `StackdriverExporterWorkerThreadTest`. This also resolves a TODO from https://github.com/census-instrumentation/opencensus-java/pull/797, since there are no asynchronous verification in the unit tests now.

### DIFF
--- a/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporterWorkerThread.java
+++ b/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporterWorkerThread.java
@@ -29,12 +29,17 @@ import io.opencensus.stats.ViewManager;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import javax.annotation.concurrent.NotThreadSafe;
 
 /**
  * Worker {@code Thread} that polls ViewData from Stats library and batch export to StackDriver.
  *
  * <p>{@code StackdriverExporterWorkerThread} is a daemon {@code Thread}.
+ *
+ * <p>The state of this class should only be accessed from the {@link
+ * StackdriverExporterWorkerThread} thread.
  */
+@NotThreadSafe
 final class StackdriverExporterWorkerThread extends Thread {
 
   private final long scheduleDelayMillis;
@@ -67,8 +72,8 @@ final class StackdriverExporterWorkerThread extends Thread {
         // Ignore views that are already registered.
         return;
       } else {
-        // TODO(songya): Do we still need to check this? Or can we assume that
-        // ViewManager.getAllExportedViews() will always return correct results?
+        // If we upload a view that has the same name with a registered view but with different
+        // attributes, Stackdriver client will throw an exception.
         throw new IllegalArgumentException(
             "A different view with the same name is already registered: " + existing);
       }
--- a/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporterWorkerThread.java
+++ b/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporterWorkerThread.java
@@ -16,8 +16,11 @@
 
 package io.opencensus.exporter.stats.stackdriver;
 
+import com.google.api.MetricDescriptor;
 import com.google.cloud.monitoring.v3.MetricServiceClient;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Lists;
+import com.google.monitoring.v3.CreateMetricDescriptorRequest;
 import com.google.monitoring.v3.CreateTimeSeriesRequest;
 import io.opencensus.common.Duration;
 import io.opencensus.stats.View;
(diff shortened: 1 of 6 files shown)


exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporterWorkerThread.java
```diff
@@ -35,16 +37,10 @@
  */
 final class StackdriverExporterWorkerThread extends Thread {
 
-  private static final Object monitor = new Object();
```
sebright: It would help to add a comment or a NotThreadSafe annotation if this class' state should now only be accessed from the StackdriverExporterWorkerThread thread.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporterWorkerThread.java
```diff
@@ -60,31 +56,44 @@
...
+    View existing = registeredViews.get(view.getName());
+    if (existing != null) {
+      if (existing.equals(view)) {
+        // Ignore views that are already registered.
+        return;
+      } else {
+        // TODO(songya): Do we still need to check this? Or can we assume that
+        // ViewManager.getAllExportedViews() will always return correct results?
+        throw new IllegalArgumentException(
+            "A different view with the same name is already registered: " + existing);
```
sebright: I think it is better to handle this case, if possible, since there could be a version mismatch between the exporter and the API, and the API could support unregistering a view.  Does Stack Driver allow a view to be changed or unregistered?  If not, I think it would at least be better to log an error and ignore the view.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporterWorkerThread.java
songy23: That's the case, done.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporterWorkerThread.java
songy23: Stackdriver doesn't allow changing a `View`, if we upload a `MetricDescriptor` (equivalent to `View`) which has the same name but different attributes to Stackdriver, it will throw an exception. Stackdriver does support unregistering (delete) a `MetricDescriptor`, though.

songy23: I've removed the TODO and left a comment explaining this.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporter.java
```diff
@@ -46,7 +43,6 @@
  *   public static void main(String[] args) {
  *     StackdriverStatsExporter.createWithProjectId(
```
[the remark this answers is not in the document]
songy23: Make sense, I would prefer to rename this to createAndRegister to be consistent with tracing exporter, and keep backwards compatibility when we switch to push model.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporter.java
```diff
@@ -180,32 +172,4 @@ static void unsafeSetExporter(StackdriverStatsExporter exporter) {
       }
```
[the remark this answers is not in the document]
songy23: That's right, updated.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporterWorkerThread.java
sebright: I still think it would be better to just log an error and skip the view, if you don't think we should unregister and re-register the MetricDescriptor.  Currently, a view mismatch would cause export thread to stop running.

songy23: I see, I've updated to use log instead. This also matches the behavior of trace exporter: https://github.com/census-instrumentation/opencensus-java/blob/master/impl_core/src/main/java/io/opencensus/implcore/trace/export/SpanExporterImpl.java#L136.

PR #804, opened 2017-11-14
Update lists of enabled Error Prone warnings with the latest bug patterns.

sebright:
I also had to fix some warnings.  The biggest change was hiding the constructor of `io.opencensus.common.OpenCensusLibraryInformation`, but that shouldn't be a problem because the class is marked experimental.


sebright:
@ubschmidt2 Thanks for finding that!  I was considering opening an issue.

PR #824, opened 2017-11-21
Add no-op RunningSpanStore and SampledSpanStore

HailongWen:
This change is a follow up of a comment by @bogdandrutu in #677 .

NoopRunningSpanStore will always return empty list when querying running [SpanData](https://github.com/HailongWen/opencensus-java/blob/ff66e5a8ed7f1d2d0959232435d41a4ece87fc90/api/src/main/java/io/opencensus/trace/export/SpanData.java).
NoopSampledSpanStore has a similar no-op implementation. It maintains a set of span names for data collections and return empty [SampledSpanStore.Summary](https://github.com/HailongWen/opencensus-java/blob/ff66e5a8ed7f1d2d0959232435d41a4ece87fc90/api/src/main/java/io/opencensus/trace/export/SampledSpanStore.java) for each span name.

Similar to [NoopViewManager](https://github.com/HailongWen/opencensus-java/blob/ff66e5a8ed7f1d2d0959232435d41a4ece87fc90/api/src/main/java/io/opencensus/stats/NoopStats.java), since NoopSampledSpanStore is not stateless, several apis of "getXXXX()" has been refactored to "newXXXX()".

### DIFF
--- a/api/src/main/java/io/opencensus/trace/TraceComponent.java
+++ b/api/src/main/java/io/opencensus/trace/TraceComponent.java
@@ -29,7 +29,6 @@ import io.opencensus.trace.propagation.PropagationComponent;
  * <p>Unless otherwise noted all methods (on component) results are cacheable.
  */
 public abstract class TraceComponent {
-  private static final NoopTraceComponent noopTraceComponent = new NoopTraceComponent();
 
   /**
    * Returns the {@link Tracer} with the provided implementations. If no implementation is provided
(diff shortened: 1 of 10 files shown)


HailongWen:
Abandon this PR since it is polluted by merges and cannot pass travis history check. 
I'll create a new one. Sorry for the troubles.

sebright:
No problem.  You can usually update the history of an existing PR by doing a forced push (git push -f).

PR #826, opened 2017-11-21
Add NoopRunningSpanStore and NoopSampledSpanStore.

HailongWen:
This change is a follow up of a comment by @bogdandrutu in #677 .

[NoopRunningSpanStore](https://github.com/census-instrumentation/opencensus-java/compare/master...HailongWen:noop-span-store?expand=1#diff-23ac90b4be2dc7cd150cb6ee4d849721) will always return empty list when querying running SpanData.
[NoopSampledSpanStore](https://github.com/HailongWen/opencensus-java/blob/d89b938b29497ad234a466c97dedd3002c7aa414/api/src/main/java/io/opencensus/trace/export/SampledSpanStore.java) has a similar no-op implementation. It maintains a set of span names for data collections and return empty Summary for each span name.

Similar to [NoopViewManager](https://github.com/HailongWen/opencensus-java/blob/d89b938b29497ad234a466c97dedd3002c7aa414/api/src/main/java/io/opencensus/stats/NoopStats.java), since [NoopSampledSpanStore](https://github.com/HailongWen/opencensus-java/blob/d89b938b29497ad234a466c97dedd3002c7aa414/api/src/main/java/io/opencensus/trace/export/SampledSpanStore.java) is not stateless, several apis of "getXXXX()" has been refactored to "newXXXX()".

### DIFF
--- a/api/src/main/java/io/opencensus/trace/TraceComponent.java
+++ b/api/src/main/java/io/opencensus/trace/TraceComponent.java
@@ -79,6 +79,8 @@ public abstract class TraceComponent {
   }
 
   private static final class NoopTraceComponent extends TraceComponent {
+    private final ExportComponent noopExportComponent = ExportComponent.newNoopExportComponent();
+
     @Override
     public Tracer getTracer() {
       return Tracer.getNoopTracer();
@@ -96,7 +98,7 @@ public abstract class TraceComponent {
 
     @Override
     public ExportComponent getExportComponent() {
-      return ExportComponent.newNoopExportComponent();
+      return noopExportComponent;
     }
 
     @Override
--- a/api/src/main/java/io/opencensus/trace/TraceComponent.java
+++ b/api/src/main/java/io/opencensus/trace/TraceComponent.java
@@ -29,7 +29,6 @@ import io.opencensus.trace.propagation.PropagationComponent;
  * <p>Unless otherwise noted all methods (on component) results are cacheable.
  */
 public abstract class TraceComponent {
-  private static final NoopTraceComponent noopTraceComponent = new NoopTraceComponent();
 
   /**
    * Returns the {@link Tracer} with the provided implementations. If no implementation is provided
@@ -75,8 +74,8 @@ public abstract class TraceComponent {
    *
    * @return an instance that contains no-op implementations for all the instances.
    */
-  static TraceComponent getNoopTraceComponent() {
-    return noopTraceComponent;
+  static TraceComponent newNoopTraceComponent() {
+    return new NoopTraceComponent();
   }
 
   private static final class NoopTraceComponent extends TraceComponent {
@@ -97,7 +96,7 @@ public abstract class TraceComponent {
 
     @Override
     public ExportComponent getExportComponent() {
-      return ExportComponent.getNoopExportComponent();
+      return ExportComponent.newNoopExportComponent();
     }
 
     @Override
--- a/api/src/main/java/io/opencensus/trace/export/ExportComponent.java
+++ b/api/src/main/java/io/opencensus/trace/export/ExportComponent.java
@@ -48,7 +48,7 @@ public abstract class ExportComponent {
    * Returns the {@link RunningSpanStore} that can be used to get useful debugging information about
    * all the current active spans.
    *
-   * @return the {@code RunningSpanStore} or {@code null} if not supported.
+   * @return the {@code RunningSpanStore}.
    */
   public abstract RunningSpanStore getRunningSpanStore();
 
@@ -56,11 +56,14 @@ public abstract class ExportComponent {
    * Returns the {@link SampledSpanStore} that can be used to get useful debugging information, such
    * as latency based sampled spans, error based sampled spans.
    *
-   * @return the {@code SampledSpanStore} or {@code null} if not supported.
+   * @return the {@code SampledSpanStore}.
    */
   public abstract SampledSpanStore getSampledSpanStore();
 
   private static final class NoopExportComponent extends ExportComponent {
+    private final SampledSpanStore noopSampledSpanStore =
+        SampledSpanStore.newNoopSampledSpanStore();
+
     @Override
     public SpanExporter getSpanExporter() {
       return SpanExporter.getNoopSpanExporter();
@@ -73,7 +76,7 @@ public abstract class ExportComponent {
 
     @Override
     public SampledSpanStore getSampledSpanStore() {
-      return SampledSpanStore.newNoopSampledSpanStore();
+      return noopSampledSpanStore;
     }
   }
 }
--- a/api/src/main/java/io/opencensus/trace/export/ExportComponent.java
+++ b/api/src/main/java/io/opencensus/trace/export/ExportComponent.java
@@ -17,7 +17,6 @@
 package io.opencensus.trace.export;
 
 import io.opencensus.trace.TraceOptions;
-import javax.annotation.Nullable;
 
 /**
  * Class that holds the implementation instances for {@link SpanExporter}, {@link RunningSpanStore}
@@ -27,15 +26,13 @@ import javax.annotation.Nullable;
  */
 public abstract class ExportComponent {
 
-  private static final NoopExportComponent NOOP_EXPORT_COMPONENT = new NoopExportComponent();
-
   /**
    * Returns the no-op implementation of the {@code ExportComponent}.
    *
    * @return the no-op implementation of the {@code ExportComponent}.
    */
-  public static ExportComponent getNoopExportComponent() {
-    return NOOP_EXPORT_COMPONENT;
+  public static ExportComponent newNoopExportComponent() {
+    return new NoopExportComponent();
   }
 
   /**
@@ -53,7 +50,6 @@ public abstract class ExportComponent {
    *
    * @return the {@code RunningSpanStore} or {@code null} if not supported.
    */
-  @Nullable
   public abstract RunningSpanStore getRunningSpanStore();
 
   /**
@@ -62,7 +58,6 @@ public abstract class ExportComponent {
    *
    * @return the {@code SampledSpanStore} or {@code null} if not supported.
    */
-  @Nullable
   public abstract SampledSpanStore getSampledSpanStore();
 
   private static final class NoopExportComponent extends ExportComponent {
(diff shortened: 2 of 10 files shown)


api/src/main/java/io/opencensus/trace/export/SampledSpanStore.java
```diff
@@ -48,6 +51,17 @@
 
   protected SampledSpanStore() {}
 
+  /**
+   * Returns a {@code SampledSpanStore} that maintains a map of span names, but always returns an
+   * empty list of {@link SpanData}.
+   *
+   * @return a {@code SampledSpanStore} that maintains a map of span names, but always returns an
```
sebright: s/map/set ?

api/src/test/java/io/opencensus/trace/export/NoopSampledSpanStoreTest.java
```diff
@@ -0,0 +1,97 @@
...
+import org.junit.rules.ExpectedException;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/** Unit tests for {@link NoopSampledSpanStore}. */
+@RunWith(JUnit4.class)
+public final class NoopSampledSpanStoreTest {
+
+  @Rule public final ExpectedException thrown = ExpectedException.none();
+  private final SampledSpanStore.PerSpanNameSummary emptyPerSpanNameSummary =
```
sebright: This field could be static.

api/src/main/java/io/opencensus/trace/export/SampledSpanStore.java
```diff
@@ -370,4 +384,59 @@ public static ErrorFilter create(
...
+    @Override
+    public Summary getSummary() {
+      Map<String, PerSpanNameSummary> result = Maps.newHashMap();
+      synchronized (registeredSpanNames) {
+        for (String registeredSpanName : registeredSpanNames) {
+          result.put(
+              registeredSpanName,
+              PerSpanNameSummary.create(
+                  Collections.<SampledSpanStore.LatencyBucketBoundaries, Integer>emptyMap(),
+                  Collections.<CanonicalCode, Integer>emptyMap()));
```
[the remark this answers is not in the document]
HailongWen: Hi sebright, I can make the PerSpanNameSummary constant, but I think the the Summary itself should vary across different calls.

api/src/main/java/io/opencensus/trace/TraceComponent.java
```diff
@@ -97,7 +96,7 @@ public Clock getClock() {
 
     @Override
     public ExportComponent getExportComponent() {
-      return ExportComponent.getNoopExportComponent();
+      return ExportComponent.newNoopExportComponent();
```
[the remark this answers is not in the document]
HailongWen: Addressed in new commit.

api/src/main/java/io/opencensus/trace/export/ExportComponent.java
```diff
@@ -53,7 +50,6 @@ public static ExportComponent getNoopExportComponent() {
    *
    * @return the {@code RunningSpanStore} or {@code null} if not supported.
```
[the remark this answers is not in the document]
HailongWen: Addressed in new commit.

api/src/main/java/io/opencensus/trace/export/ExportComponent.java
```diff
@@ -71,16 +66,14 @@ public SpanExporter getSpanExporter() {
...
     public RunningSpanStore getRunningSpanStore() {
-      return null;
+      return RunningSpanStore.getNoopRunningSpanStore();
     }
 
-    @Nullable
     @Override
     public SampledSpanStore getSampledSpanStore() {
-      return null;
+      return SampledSpanStore.newNoopSampledSpanStore();
```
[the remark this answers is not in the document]
HailongWen: Addressed in new commit.

api/src/main/java/io/opencensus/trace/export/RunningSpanStore.java
```diff
@@ -37,8 +37,19 @@
...
+  private static final RunningSpanStore NOOP_RUNNING_SPAN_STORE = new NoopRunningSpanStore();
+
   protected RunningSpanStore() {}
 
+  /**
+   * Returns the no-op implementation of the {@code RunningSpanStore}.
+   *
+   * @return the no-op implementation of the {@code RunningSpanStore}.
+   */
+  public static RunningSpanStore getNoopRunningSpanStore() {
```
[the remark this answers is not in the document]
HailongWen: Addressed in new commit.

api/src/main/java/io/opencensus/trace/export/RunningSpanStore.java
```diff
@@ -151,4 +162,18 @@ public static Filter create(String spanName, int maxSpansToReturn) {
      */
     public abstract int getMaxSpansToReturn();
   }
+
+  private static final class NoopRunningSpanStore extends RunningSpanStore {
+
+    @Override
+    public Summary getSummary() {
+      return Summary.create(Collections.<String, PerSpanNameSummary>emptyMap());
```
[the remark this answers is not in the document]
HailongWen: Addressed in new commit.

api/src/main/java/io/opencensus/trace/export/SampledSpanStore.java
HailongWen: Hi sebright, what does this comment mean? Should I use other libraries?

api/src/main/java/io/opencensus/trace/export/SampledSpanStore.java
```diff
@@ -370,4 +384,59 @@ public static ErrorFilter create(
...
+        }
+      }
+      return Summary.create(result);
+    }
+
+    @Override
+    public Collection<SpanData> getLatencySampledSpans(LatencyFilter filter) {
```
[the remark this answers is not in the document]
HailongWen: Addressed in new commit.

api/src/main/java/io/opencensus/trace/export/SampledSpanStore.java
```diff
@@ -370,4 +384,59 @@ public static ErrorFilter create(
...
+      checkNotNull(spanNames);
+      synchronized (registeredSpanNames) {
+        registeredSpanNames.removeAll(spanNames);
+      }
+    }
+
+    @Override
+    public Set<String> getRegisteredSpanNamesForCollection() {
+      synchronized (registeredSpanNames) {
+        return registeredSpanNames;
```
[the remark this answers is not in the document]
HailongWen: Addressed in new commit.

api/src/test/java/io/opencensus/trace/export/NoopSampledSpanStoreTest.java
HailongWen: Addressed in new commit.

api/src/main/java/io/opencensus/trace/export/SampledSpanStore.java
sebright: I was wondering if the Javadoc should say "set" instead of "map" since the no-op class doesn't pair the span names with anything.

api/src/main/java/io/opencensus/trace/export/SampledSpanStore.java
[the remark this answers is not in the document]
sebright: Sorry, I only meant the PerSpanNameSummary.  I forgot that there were two "Summary" classes.

api/src/main/java/io/opencensus/trace/export/SampledSpanStore.java
HailongWen: I see. I agree that it should be updated.

api/src/main/java/io/opencensus/trace/TraceComponent.java
```diff
@@ -75,11 +74,13 @@
    *
    * @return an instance that contains no-op implementations for all the instances.
    */
-  static TraceComponent getNoopTraceComponent() {
```
bogdandrutu: Why do we change this?

api/src/main/java/io/opencensus/trace/export/SampledSpanStore.java
```diff
@@ -370,4 +384,61 @@ public static ErrorFilter create(
      */
     public abstract int getMaxSpansToReturn();
   }
+
+  @ThreadSafe
+  private static final class NoopSampledSpanStore extends SampledSpanStore {
```
bogdandrutu: This is not a no-op implementation. @sebright you did the same for ViewManager but I think it is wrong to say that this is a no-op implementation. Based on my knowledge no-op means "do nothing".

sebright: @bogdandrutu Do you object to just the name, or also the behavior?  I think it makes sense to rename the "Noop" classes that have state.

HailongWen: @bogdandrutu I do agree the name is misleading, but I kind of understand why @sebright introduce such behavior at first. 

Unlike the get() apis, whose results rely on actual metrics we collected, the behaviors of these register/unregister apis are deterministic no matter we collect metrics or not. I think some users might have tests or condition checks on these apis?

api/src/main/java/io/opencensus/trace/export/RunningSpanStore.java
```diff
@@ -37,8 +37,19 @@
...
+  private static final RunningSpanStore NOOP_RUNNING_SPAN_STORE = new NoopRunningSpanStore();
+
   protected RunningSpanStore() {}
 
+  /**
+   * Returns the no-op implementation of the {@code RunningSpanStore}.
+   *
+   * @return the no-op implementation of the {@code RunningSpanStore}.
+   */
+  static RunningSpanStore getNoopRunningSpanStore() {
```
bogdandrutu: Why do we sometimes use getNoop and sometimes newNoop we should be consistent.

api/src/main/java/io/opencensus/trace/export/SampledSpanStore.java
bogdandrutu: @sebright @HailongWen I don't like that we have a state (at all) in the no-op implementation in our API. But we should move forward with this PR and probably decide later what to do.

api/src/main/java/io/opencensus/trace/export/RunningSpanStore.java
HailongWen: The logic here is that stateless no-ops are static; for stateful no-ops, **getNoop** will make user think it is static and that's why we rename it to newNoop.

HailongWen:
Offline had a discussion with @bogdandrutu and I agree that implementations should not be present in opencensus-api, and users should depend on opencensus-impl when writing tests.

api/src/main/java/io/opencensus/trace/export/RunningSpanStore.java
bogdandrutu: I think we can do better (probably in a new PR) and always do newXXX and cache them in the component.

HailongWen:
Thanks for the review!

PR #827, opened 2017-11-21
Fix the api name sent to SD stats exporter.

bogdandrutu:


exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporterWorkerThread.java
```diff
@@ -118,15 +127,13 @@ void export() {
...
         export();
-        Thread.sleep(scheduleDelayMillis);
-      } catch (InterruptedException ie) {
-        // Preserve the interruption status as per guidance and stop doing any work.
-        Thread.currentThread().interrupt();
-        return;
+      } catch (Throwable e) {
+        logger.log(Level.WARNING, "Exception thrown by the Stackdriver stats exporter.", e);
       }
+      Uninterruptibles.sleepUninterruptibly(scheduleDelayMillis, TimeUnit.MILLISECONDS);
```
sebright: Why did you switch to sleepUninterruptibly?  Couldn't you just add another catch block for Throwable?

sebright:
LGTM, though I'm not very familiar with the StackDriver API, and I didn't know the answer to your question.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporterWorkerThread.java
bogdandrutu: I could but I think this is nicer. Also I think is better to end the Thread nicely.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExportUtils.java
```diff
@@ -79,7 +79,11 @@ static MetricDescriptor createMetricDescriptor(View view, String projectId) {
     String viewName = view.getName().asString();
     // Name format refers to
     // cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.metricDescriptors/create
-    builder.setName(String.format("projects/%s", projectId));
+    builder.setName(
+        // DO_NOT_SUBMIT: Should this also be "custom.googleapis.com/opencensus/%s"
```
songy23: You're right, all custom metric names need to be under "custom.googleapis.com/" domain:
(https://cloud.google.com/monitoring/custom-metrics/creating-metrics)

So the whole name should follow the format: projects/my-project-id/metricDescriptors/custom.googleapis.com/opencensus/viewName

bogdandrutu: Done.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporterWorkerThread.java
sebright: I'm not sure I understand.  The only difference I see is that the new implementation always sleeps for the full time interval, but the old implementation could return early on interruption.  That behavior doesn't matter now, though, since we don't have a shutdown method.   If we added a shutdown method, I think it would be better for the thread to return as soon as possible.

bogdandrutu: Reverted to limit the changes in this PR because we need to cherry pick it.

rakyll:
This naming looks very weird on the UI with the escaped metric name:
metric

PR #836, opened 2017-11-22
Batch export 3 Timeseries at one RPC call

songy23:
If we batch export too many `Timeseries` with one `CreateTimeSeriesRequest`, Stackdriver client might throw an `Http2Exception: Header size exceeded max allowed size (10240)`

Fix https://github.com/census-instrumentation/opencensus-java/issues/840

### DIFF
--- a/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporterWorkerThread.java
+++ b/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporterWorkerThread.java
@@ -121,17 +121,18 @@ final class StackdriverExporterWorkerThread extends Thread {
       timeSeriesList.addAll(StackdriverExportUtils.createTimeSeriesList(viewData, projectId));
     }
 
-    for (int i = 0; i < timeSeriesList.size(); i += MAX_BATCH_EXPORT_SIZE) {
+    for (List<TimeSeries> batchedTimeSeries :
+        Lists.partition(timeSeriesList, MAX_BATCH_EXPORT_SIZE)) {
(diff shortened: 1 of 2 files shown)


exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporterWorkerThread.java
```diff
@@ -113,13 +116,24 @@ void export() {
...
     for (ViewData viewData : viewDataList) {
-      builder.addAllTimeSeries(StackdriverExportUtils.createTimeSeriesList(viewData, projectId));
+      timeSeriesList.addAll(StackdriverExportUtils.createTimeSeriesList(viewData, projectId));
     }
-    if (!builder.getTimeSeriesList().isEmpty()) {
-      metricServiceClient.createTimeSeries(builder.build());
+
+    for (List<TimeSeries> batchedTimeSeries :
+        Lists.partition(timeSeriesList, MAX_BATCH_EXPORT_SIZE)) {
+      // Batch export 3 TimeSeries at one call, to avoid exceeding RPC header size limit.
```
ubschmidt2: Is there a relevant GitHub issue, discussion on SO, etc.? If so, it would be great to add a reference here.
Otherwise, LGTM.

songy23: @bogdandrutu caught this bug in an E2E test (https://github.com/bogdandrutu/java-demo-apps). I've just created an issue for this: https://github.com/census-instrumentation/opencensus-java/issues/840.

PR #837, opened 2017-11-23
Introduce a Settings class.

ubschmidt2:


contrib/agent/src/main/java/io/opencensus/contrib/agent/Settings.java
```diff
@@ -0,0 +1,66 @@
...
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Strings;
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigFactory;
+
+/** The {@code Settings} class provides access to user-configurable settings. */
+public class Settings {
```
adriancole: so this hides the typesafe stuff internally

Issue #838, opened 2017-11-27
AggregationData and Aggregation should match for ViewData


songy23:
For example, if a `View` has `Sum` aggregation, and a `ViewData` is created from this `View`, then all `AggregationData` associated with this `ViewData` should be `SumData`.

PR #839, opened 2017-11-27
Validate types of Aggregation and AggregationData when creating ViewData

songy23:
Fix https://github.com/census-instrumentation/opencensus-java/issues/838

### DIFF
--- a/api/src/main/java/io/opencensus/stats/ViewData.java
+++ b/api/src/main/java/io/opencensus/stats/ViewData.java
@@ -132,25 +132,25 @@ public abstract class ViewData {
           @Override
           public Void apply(Sum arg) {
             measure.match(
-                new Function<MeasureDouble, Object>() {
+                new Function<MeasureDouble, Void>() {
                   @Override
-                  public Object apply(MeasureDouble arg) {
+                  public Void apply(MeasureDouble arg) {
                     checkArgument(
                         aggregationData instanceof SumDataDouble,
                         createErrorMessageForAggregation(aggregation, aggregationData));
                     return null;
                   }
                 },
-                new Function<MeasureLong, Object>() {
+                new Function<MeasureLong, Void>() {
                   @Override
-                  public Object apply(MeasureLong arg) {
+                  public Void apply(MeasureLong arg) {
                     checkArgument(
                         aggregationData instanceof SumDataLong,
                         createErrorMessageForAggregation(aggregation, aggregationData));
                     return null;
                   }
                 },
-                Functions.throwIllegalArgumentException());
+                Functions.<Void>throwIllegalArgumentException());
             return null;
           }
         },
--- a/api/src/main/java/io/opencensus/stats/ViewData.java
+++ b/api/src/main/java/io/opencensus/stats/ViewData.java
@@ -16,11 +16,25 @@
 
 package io.opencensus.stats;
 
+import static com.google.common.base.Preconditions.checkArgument;
+
 import com.google.auto.value.AutoValue;
 import com.google.common.collect.Maps;
 import io.opencensus.common.Function;
 import io.opencensus.common.Functions;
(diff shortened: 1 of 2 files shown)


api/src/main/java/io/opencensus/stats/ViewData.java
```diff
@@ -53,52 +67,132 @@
...
+                new Function<MeasureLong, Void>() {
+                  @Override
+                  public Void apply(MeasureLong arg) {
+                    checkArgument(
+                        aggregationData instanceof SumDataLong,
+                        createErrorMessageForAggregation(aggregation, aggregationData));
+                    return null;
+                  }
+                },
+                Functions.<Void>throwIllegalArgumentException());
```
sebright: You could throw an AssertionError, because this code is in the API, so it never needs to handle unknown subclasses.

api/src/main/java/io/opencensus/stats/ViewData.java
```diff
@@ -53,52 +67,132 @@
...
+        new Function<Distribution, Void>() {
+          @Override
+          public Void apply(Distribution arg) {
+            checkArgument(
+                aggregationData instanceof DistributionData,
+                createErrorMessageForAggregation(aggregation, aggregationData));
+            return null;
+          }
+        },
+        Functions.<Void>throwIllegalArgumentException());
```
sebright: Same here

api/src/test/java/io/opencensus/stats/ViewDataTest.java
```diff
@@ -171,10 +181,48 @@ public void preventStartTimeLaterThanEndTime() {
...
+    thrown.expect(IllegalArgumentException.class);
+    thrown.expectMessage("Aggregation and AggregationData types mismatch. ");
+    ViewData.create(view, entries, cumulativeData);
+  }
+
   // tag keys
   private static final TagKey K1 = TagKey.create("k1");
   private static final TagKey K2 = TagKey.create("k2");
-  private final List<TagKey> tagKeys = Arrays.asList(K1, K2);
+  private static final List<TagKey> tagKeys = Arrays.asList(K1, K2);
```
sebright: This variable could use constant case.

api/src/test/java/io/opencensus/stats/ViewDataTest.java
```diff
@@ -171,10 +181,48 @@ public void preventStartTimeLaterThanEndTime() {
     CumulativeData.create(Timestamp.fromMillis(3000), Timestamp.fromMillis(2000));
   }
 
+  @Test
+  public void preventAggregationAndAggregationDataMismatch() {
+    aggregationAndAggregationDataMismatch(
+        createView(Sum.create(), MEASURE_DOUBLE),
+        ImmutableMap.<List<TagValue>, AggregationData>of(
+            Arrays.asList(V1, V2), SumDataLong.create(100)));
```
sebright: I think each call to `aggregationAndAggregationDataMismatch` should be a separate test, since it throws an exception.

api/src/main/java/io/opencensus/stats/ViewData.java
songy23: Make sense, done.

api/src/main/java/io/opencensus/stats/ViewData.java
songy23: Done.

api/src/test/java/io/opencensus/stats/ViewDataTest.java
songy23: Done.

api/src/test/java/io/opencensus/stats/ViewDataTest.java
songy23: Done.

Issue #833, opened 2017-11-21
Support custom MonitoredResource labels for Stackdriver Stats Exporter


songy23:

songy23:
PR https://github.com/census-instrumentation/opencensus-java/pull/832 removes the project_id label. For now the Stackdriver Stats Exporter will always use a global MonitoredResource with no associated labels: https://cloud.google.com/monitoring/api/resources#tag_global.

In the next version we may consider allowing users to set custom labels (e.g GCE instance-id) for their MonitoredResource in the exporter.

PR #870, opened 2017-12-04
Support custom MonitoredResource for Stackdriver Stats Exporter

songy23:
Fix https://github.com/census-instrumentation/opencensus-java/issues/833 and https://github.com/census-instrumentation/opencensus-java/issues/865. This PR allows users to use custom MonitoredResource to further break down their stats in Stackdriver.

/cc @rakyll @jcd2 

**Background:**
Due to some backend constraints, Stackdriver monitoring doesn't allow uploading out of order `Timeseries` stats for any `MetricDescriptor`. This means if we have multiple clients writing to the same metric, we can't upload stale `Timeseries` between them, otherwise Stackdriver will return an error. We have to make sure `Timeseries` in all the distributed binaries are uploaded in monotonically increased time order. This will be difficult in distributed systems. 

To address this, Stackdriver monitoring breaks down the `Timeseries` stats by one more dimension called [MonitoredResource](https://cloud.google.com/monitoring/api/resources). MonitoredResource can have a type and a set of resource labels, and each label combination for a metric is stored as a separate `Timeseries`. However, for Opencensus:
- we don't have an equivalent concept as Stackdriver `MonitoredResource` in our data model;
- we won't be able to create unique `MonitoredResource` for users.

Therefore I think it's better to have users create and set `MonitoredResource` for exporter. If they have multiple exporters exporting to a same `MetricDescriptor` concurrently, they need to associate each exporter with a unique `MonitoredResource`, and this could prevent the race condition of uploading out-of-order stats.


exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsMonitoredResource.java
```diff
@@ -0,0 +1,52 @@
...
+ * {@code ThreadSafe} class that holds Stackdriver {@link MonitoredResource} for this {@link
+ * StackdriverStatsExporter}.
+ */
+@ThreadSafe
+final class StackdriverStatsMonitoredResource {
+
+  private static final Object monitor = new Object();
+
+  @GuardedBy("monitor")
+  private static MonitoredResource monitoredResource =
```
bogdandrutu: This can be a volatile reference.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsMonitoredResource.java
```diff
@@ -0,0 +1,52 @@
...
+      MonitoredResource.newBuilder().setType("global").build();
+
+  static void setMonitoredResource(MonitoredResource monitoredResource) {
+    checkNotNull(monitoredResource);
+    synchronized (monitor) {
+      StackdriverStatsMonitoredResource.monitoredResource = monitoredResource;
```
bogdandrutu: MonitoredResource is not immutable so you need to make a copy.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsMonitoredResource.java
```diff
@@ -0,0 +1,52 @@
...
+import javax.annotation.concurrent.ThreadSafe;
+
+/**
+ * {@code ThreadSafe} class that holds Stackdriver {@link MonitoredResource} for this {@link
+ * StackdriverStatsExporter}.
+ */
+@ThreadSafe
+final class StackdriverStatsMonitoredResource {
+
+  private static final Object monitor = new Object();
```
bogdandrutu: You can remove this by using atomic/volatile.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporter.java
```diff
@@ -169,4 +171,25 @@ static void unsafeResetExporter() {
...
+   * sure each {@code StackdriverStatsExporter} has a unique {@code MonitoredResource}.
+   *
+   * <p>If {@link #setMonitoredResource(MonitoredResource)} is called after {@code
+   * StackdriverStatsExporter} is created, all the previous stats that were uploaded won't be
+   * affected, all the future stats will be associated with the given {@code MonitoredResource}.
+   *
+   * @param monitoredResource the {@code MonitoredResource} used by this exporter.
+   * @throws NullPointerException if the given {@code MonitoredResource} is null.
+   */
+  public static void setMonitoredResource(MonitoredResource monitoredResource) {
```
bogdandrutu: Can this API be part of the creating the exporter?

sebright: Maybe we should switch to a builder-like API for creating the StackdriverStatsExporter.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsMonitoredResource.java
songy23: Done.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsMonitoredResource.java
songy23: Done.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsMonitoredResource.java
songy23: Done.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExportUtils.java
```diff
@@ -151,8 +150,7 @@ static MetricKind createMetricKind(AggregationWindow window) {
     // Shared fields for all TimeSeries generated from the same ViewData
     TimeSeries.Builder shared = TimeSeries.newBuilder();
     shared.setMetricKind(createMetricKind(view.getWindow()));
-    // TODO(songya): add support for custom resource labels.
-    shared.setResource(MonitoredResource.newBuilder().setType("global"));
+    shared.setResource(StackdriverStatsMonitoredResource.getMonitoredResource());
```
sebright: Suggestion: This method would be clearer and easier to test if it took the monitored resource as an argument instead of accessing a static variable.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporter.java
songy23: That'a another alternative. My concern is we already have a lot of parameters in the constructor and several factory methods, adding one more parameter and factory method seems too much. Also, Stackdriver client allows users to change MonitoredResource at run time, I think it's better to support that as well.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsMonitoredResource.java
```diff
@@ -0,0 +1,45 @@
...
+      MonitoredResource.newBuilder().setType("global").build();
+
+  static void setMonitoredResource(MonitoredResource monitoredResource) {
+    checkNotNull(monitoredResource);
+    // Make a deep copy of monitoredResource
+    StackdriverStatsMonitoredResource.monitoredResource = monitoredResource.toBuilder().build();
+  }
+
+  static MonitoredResource getMonitoredResource() {
+    return monitoredResource;
```
sebright: It would be safer to also copy the MonitoredResource here.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsMonitoredResource.java
```diff
@@ -0,0 +1,45 @@
...
+ * StackdriverStatsExporter}.
+ */
+@ThreadSafe
+final class StackdriverStatsMonitoredResource {
+
+  private static volatile MonitoredResource monitoredResource =
+      MonitoredResource.newBuilder().setType("global").build();
+
+  static void setMonitoredResource(MonitoredResource monitoredResource) {
+    checkNotNull(monitoredResource);
```
bogdandrutu: checkNotNull(monitoredResource, "monitoredResource")

exporters/stats/stackdriver/README.md
```diff
@@ -67,6 +67,26 @@ public class MyMainClass {
...
+
+By default, the Stackdriver Stats Exporter uses [a global Stackdriver monitored resource with no 
+labels](https://cloud.google.com/monitoring/api/resources#tag_global). If you want to use a custom Monitored Resource, do:
+
+```java
+public class MyMainClass {
+  public static void main(String[] args) {
+    // Set a custom MonitoredResource. Please make sure each Stackdriver Stats Exporter has a 
+    // unique MonitoredResource.
+    StackdriverStatsExporter.setMonitoredResource(myResource);
```
bogdandrutu: setMonitoredResource is package protected so you cannot do this.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsMonitoredResource.java
bogdandrutu: This should be a internal API so making a copy is useless extra overhead.

sebright: Yes, it's probably better to add a comment to warn against modifying the object.

exporters/stats/stackdriver/README.md
songy23: Updated since I've removed this API.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExportUtils.java
songy23: Done.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporter.java
songy23: Updated to have MonitoredResource in the factory methods instead. I'll create an issue about changing to builder-like API for exporter, since there are too many factory methods now.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsMonitoredResource.java
songy23: Done.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsMonitoredResource.java
sebright: I just realized that MonitoredResource is a protocol buffer, so it should be immutable.  See the "Builders" section in https://developers.google.com/protocol-buffers/docs/reference/java-generated.

songy23: Oh you're right. Once it's built, its fields cannot be changed so should be immutable. I've removed the deep copy code.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporter.java
```diff
@@ -57,21 +58,25 @@
   private static StackdriverStatsExporter exporter = null;
 
   private static final Duration ZERO = Duration.create(0, 0);
+  private static final MonitoredResource defaultResource =
```
bogdandrutu: DEFAULT_RESOURCE?

exporters/stats/stackdriver/README.md
```diff
@@ -67,6 +67,24 @@ public class MyMainClass {
 }
 ```
 
+#### Set Monitored Resource for exporter
```
bogdandrutu: You need to mention that if user does not do this then will not be able to have multiple binaries sending the same metric.

songy23: Done.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporter.java
songy23: Correct - done.

exporters/stats/stackdriver/README.md
```diff
@@ -67,6 +67,27 @@ public class MyMainClass {
...
+have only one exporter running. However, if you want to have multiple binaries exporting stats for 
+the same metric concurrently, using the default monitored resource for all exporters will not work. 
+In this case you need to associate a unique monitored resource with each exporter:
+
+```java
+public class MyMainClass {
+  public static void main(String[] args) {
+    // Set a custom MonitoredResource. Please make sure each Stackdriver Stats Exporter has a 
+    // unique MonitoredResource.      
+    StackdriverStatsExporter.createAndRegisterWithProjectIdAndMonitoredResource("MyStackdriverProjectId", Duration.create(10, 0), myResource);
```
bogdandrutu: We should give an example for myResource definition as well.

songy23: Sure, updated.

exporters/stats/stackdriver/README.md
```diff
@@ -67,6 +67,38 @@ public class MyMainClass {
...
+By default, the Stackdriver Stats Exporter uses [a global Stackdriver monitored resource with no 
+labels](https://cloud.google.com/monitoring/api/resources#tag_global), and this works fine when you 
+have only one exporter running. However, if you want to have multiple binaries exporting stats for 
+the same metric concurrently, using the default monitored resource for all exporters will not work. 
+In this case you need to associate a unique monitored resource with each exporter:
+
+```java
+public class MyMainClass {
+  public static void main(String[] args) {
+    // A sample AWS EC2 monitored resource.
```
bogdandrutu: Mention in a comment that this will work only if in each instance is only one process that records data. If multiple process an extra label like pid should be used.

exporters/stats/stackdriver/README.md
```diff
@@ -67,6 +67,38 @@ public class MyMainClass {
 }
 ```
 
+#### Set Monitored Resource for exporter
+
+By default, the Stackdriver Stats Exporter uses [a global Stackdriver monitored resource with no 
+labels](https://cloud.google.com/monitoring/api/resources#tag_global), and this works fine when you 
+have only one exporter running. However, if you want to have multiple binaries exporting stats for
```
sebright: Suggestion: s/binaries/processes

songy23: Done.

exporters/stats/stackdriver/README.md
songy23: Added.

songy23:
Thanks for the review!

PR #873, opened 2017-12-04
Cherry-pick commits for V0.9.2

songy23:
Please let me know if there are other commits that we want to include in v0.9.2.


songy23:
Closing since we decide to do a minor release instead of patch release.

PR #879, opened 2017-12-05
Use continue when skip a value in a while-loop.

bogdandrutu:


impl_core/src/test/java/io/opencensus/implcore/stats/StatsRecorderImplTest.java
```diff
@@ -110,6 +114,37 @@ public void record_CurrentContextSet() {
...
+            Cumulative.create());
+    viewManager.registerView(view);
+    Context orig =
+        Context.current()
+            .withValue(ContextUtils.TAG_CONTEXT_KEY, new SimpleTagContext(Tag.create(KEY, VALUE)))
+            .attach();
+    try {
+      statsRecorder
+          .newMeasureMap()
+          .put(MEASURE_DOUBLE_NO_VIEW_1, 1.0)
```
songy23: nit: might be useful to have 3 different values recorded, and verify that the value associated with MEASURE_DOUBLE is in `viewData`.

impl_core/src/test/java/io/opencensus/implcore/stats/StatsRecorderImplTest.java
```diff
@@ -110,6 +114,37 @@ public void record_CurrentContextSet() {
...
+          .put(MEASURE_DOUBLE, 1.0)
+          .put(MEASURE_DOUBLE_NO_VIEW_2, 1.0)
+          .record();
```
sebright: You could call record with the TagContext to avoid modifying the current context.

impl_core/src/test/java/io/opencensus/implcore/stats/StatsRecorderImplTest.java
bogdandrutu: Done.

impl_core/src/test/java/io/opencensus/implcore/stats/StatsRecorderImplTest.java
bogdandrutu: Done.

PR #881, opened 2017-12-06
Make stackdriver stats exporter GAE Java7 compatible. (fixes #878)

### LINKED ISSUE #878: Stackdriver stats exporter not GAE Java7 compatible
The worker thread in SD stats exporter is not GAE java7 compatible:
https://github.com/census-instrumentation/opencensus-java/blob/master/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporterWorkerThread.java

HailongWen:
This commit makes the following changes:
* Renamed `io.opencensus.implcore.internal.DaemonThreadFactory` to `io.opencensus.interl.OpenCensusThreadFactory` so that other components can reuse the logic to detect the runtime and create threads accordingly.
* Refactored the `StackdriverExporterWorkerThread` to a `Runnable` called `StackdriverExporterWorker`.
* Adapt all occurrences and invocation of the changes above.

### DIFF
--- a/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporterWorker.java
+++ b/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporterWorker.java
@@ -37,18 +37,17 @@ import java.util.logging.Logger;
 import javax.annotation.concurrent.NotThreadSafe;
 
 /**
- * Worker {@code Thread} that polls ViewData from Stats library and batch export to StackDriver.
+ * Worker {@code Runnable} that polls ViewData from Stats library and batch export to StackDriver.
  *
- * <p>{@code StackdriverExporterWorkerThread} is a daemon {@code Thread}.
+ * <p>{@code StackdriverExporterWorker} will be started in a daemon {@code Thread}.
  *
- * <p>The state of this class should only be accessed from the {@link
- * StackdriverExporterWorkerThread} thread.
+ * <p>The state of this class should only be accessed from the thread which {@link
+ * StackdriverExporterWorker} resides in. thread.
  */
 @NotThreadSafe
-final class StackdriverExporterWorkerThread extends Thread {
+final class StackdriverExporterWorker implements Runnable {
 
-  private static final Logger logger =
-      Logger.getLogger(StackdriverExporterWorkerThread.class.getName());
+  private static final Logger logger = Logger.getLogger(StackdriverExporterWorker.class.getName());
 
   @VisibleForTesting static final int MAX_BATCH_EXPORT_SIZE = 3;
 
@@ -60,7 +59,7 @@ final class StackdriverExporterWorkerThread extends Thread {
   private final MonitoredResource monitoredResource;
   private final Map<View.Name, View> registeredViews = new HashMap<View.Name, View>();
 
-  StackdriverExporterWorkerThread(
+  StackdriverExporterWorker(
       String projectId,
       MetricServiceClient metricServiceClient,
       Duration exportInterval,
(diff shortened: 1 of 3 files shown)


exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporter.java
```diff
@@ -52,7 +53,8 @@
 
   private static final Object monitor = new Object();
 
-  private final StackdriverExporterWorkerThread workerThread;
+  private final StackdriverExporterWorker worker;
```
songy23: nit: this can be removed, having a reference to the daemon thread should be enough.

HailongWen:
Sure. Will do.

api/src/main/java/io/opencensus/internal/OpenCensusThreadFactory.java
```diff
@@ -14,14 +14,16 @@
  * limitations under the License.
  */
 
-package io.opencensus.implcore.internal;
+package io.opencensus.internal;
```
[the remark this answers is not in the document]
HailongWen: Thanks for the comment!

I agree. opencensus-api is not a good place for this utility class since technically users may decide to keep an older version of opencensus-api (e.g. for compatibility concern) and in such case the exporter will fail.

I can live with duplicating the logic in stack exporter. Actually I think to some extent we've already done such duplication on the worker thread logic.

api/src/main/java/io/opencensus/internal/OpenCensusThreadFactory.java
```diff
@@ -31,22 +33,27 @@
...
 
   /**
-   * Constructs a new {@code DaemonThreadFactory}.
+   * Constructs a new {@code OpenCensusThreadFactory}.
    *
    * @param threadPrefix used to prefix all thread names. (E.g. "CensusDisruptor").
+   * @param startAsDaemon whether to start new threads as daemon.
    */
-  public DaemonThreadFactory(String threadPrefix) {
+  public OpenCensusThreadFactory(String threadPrefix, Boolean startAsDaemon) {
```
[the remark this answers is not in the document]
HailongWen: I am trying to make it a general configurable ThreadFactory so that it can create non-daemon threads.

But I think this is no longer valid now since we are trying to constrain the change within exporter side. I will leave the file as it was and create a new ThreadFactory in stats exporter.

bogdandrutu:
FYI the 0.10.1 patch release is done. You can merge this anytime now.

HailongWen:
@bogdandrutu @songy23 @sebright PTAL. I've moved the DaemonThreadFactory back to its original path. The SD stats exporter now duplicate the thread spawn logic. We can remove that in the future once we done refactoring the stats exporter to use a callback.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporterWorker.java
```diff
@@ -37,18 +37,17 @@
...
- * Worker {@code Thread} that polls ViewData from Stats library and batch export to StackDriver.
+ * Worker {@code Runnable} that polls ViewData from Stats library and batch export to StackDriver.
  *
- * <p>{@code StackdriverExporterWorkerThread} is a daemon {@code Thread}.
+ * <p>{@code StackdriverExporterWorker} will be started in a daemon {@code Thread}.
  *
- * <p>The state of this class should only be accessed from the {@link
- * StackdriverExporterWorkerThread} thread.
+ * <p>The state of this class should only be accessed from the thread which {@link
+ * StackdriverExporterWorker} resides in. thread.
```
sebright: Is "thread." a typo?

HailongWen: Done.

PR #882, opened 2017-12-06
Update google-auth-library-credentials.

bogdandrutu:


bogdandrutu:
Looks like truth 0.36 is not java6 compatible, should we postpone the update? grpc already switched to this version.

sebright:
It might be easier to postpone the upgrade.  Are there any other reasons that we should upgrade soon, such as deprecation of APIs?

songy23:
Postpone sounds good to me, if this update is not in urgent need.

bogdandrutu:
Reverted the truth update for the moment.

Issue #855, opened 2017-11-30
Make examples independent gradle project.


bogdandrutu:
Currently our example gradle project is a child of the opencensus-java project. We should do what gRPC did and extract the examples as a standalone gradle project:
https://github.com/grpc/grpc-java/tree/master/examples

Also we should add  build support for:
* maven
* bazel

HailongWen:
All PR merged. Close this issue.

PR #890, opened 2017-12-07
Make `examples` an independent gradle project.

HailongWen:
This PR makes the `examples` subproject standalone (#855). 
Two separate PR will be created to add bazel and maven support.


examples/README.md
```diff
@@ -5,6 +5,8 @@
 ```
 $ ./gradlew installDist
```
sebright: IIUC, this command now needs to be run inside the `examples` directory, so the next two commands need to be updated.

examples/build.gradle
```diff
@@ -1,16 +1,193 @@
...
+repositories {
+    mavenCentral()
+    mavenLocal()
+}
+
+group = "io.opencensus"
+version = "0.11.0-SNAPSHOT"
+
+// change to the version you want to use.
+def opencensusVersion = "0.11.0-SNAPSHOT" // CURRENT_OPENCENSUS_VERSION
```
sebright: I think we should just use the latest released version here, like we do in the example dependency declarations in the readme.

examples/README.md
```diff
@@ -5,6 +5,8 @@
...
 ```
+
+* Remember to change `opencensusVersion` in `build.gradle` to a released version you want to use.
```
bogdandrutu: Suggest people to checkout to a release tag.

examples/buildscripts/checkstyle.xml
```diff
@@ -0,0 +1,233 @@
+<?xml version="1.0"?>
```
bogdandrutu: Can we share this file between the root project and examples? Same for license.

examples/gradle/errorprone/experimental_errors
```diff
@@ -0,0 +1,27 @@
+errorProneExperimentalErrors = \
```
bogdandrutu: Can we share this with the errorprone files from the root project?

examples/build.gradle
```diff
@@ -1,16 +1,193 @@
...
+buildscript {
+    repositories {
+        mavenCentral()
+        mavenLocal()
+        maven {
+            url "https://plugins.gradle.org/m2/"
+        }
+    }
+    dependencies {
+        classpath 'ru.vyarus:gradle-animalsniffer-plugin:1.4.2'
```
sebright: I'm not sure we need as much functionality in the examples directory as we have for the main project.  I think we should try to keep the build file as simple as possible to make it a good example for users who just want to depend on OpenCensus.

Checking the license is important, though.  I wonder if it is possible to check the example classes from the main project instead.

examples/buildscripts/checkstyle.xml
HailongWen: Sure.

The reason I copied that file here is to make examples "standalone" without referencing files outside the folder. The examples is still a sub-folder, not in a separate project so I think it should be fine to reuse those configurations.

examples/README.md
HailongWen: Thanks for pointing out. I'll update that later.

examples/build.gradle
bogdandrutu: We can do that but then you cannot add example for an API that is not released yet. Maybe not a big deal

HailongWen: Are you suggesting using `version` only?

My logic here is that the example project itself has a release version (defined by `version`), while it could be built against other version of opencensus api/impl (defined by `opencensusVersion`, and these dependencies will be downloaded from maven instead of referencing the project itself).

examples/build.gradle
HailongWen: @bogdandrutu @sebright what plugins do you suggest we keep in this standalone project? grpc only uses java/protobuf/idea/application (no signing, jacoco, findbugs and checkstyle). If we want to keep the build file minimal, we should remove them, too.

sebright: I think that `java` and `application` would be enough for now.  `idea` would be fine, too, especially if it is only one line.  We can add testing back if we need to add tests later.

examples/gradle/errorprone/experimental_errors
HailongWen: Yes. I'll remove these files.

examples/build.gradle
bogdandrutu: you can ask users to do a "./gradle install" for the main project, which builds and install locally the main project. I am fine with using the latest release.

examples/buildscripts/checkstyle.xml
bogdandrutu: +1 to reuse.

HailongWen:
@bogdandrutu @sebright @songy23 PTAL.

examples/build.gradle
```diff
@@ -1,16 +1,68 @@
...
+buildscript {
+    repositories {
+        mavenCentral()
+        mavenLocal()
+        maven {
+            url "https://plugins.gradle.org/m2/"
+        }
+    }
+    dependencies {
+        classpath "gradle.plugin.com.github.sherter.google-java-format:google-java-format-gradle-plugin:0.6"
```
sebright: Can this plugin also be removed?

examples/build.gradle
```diff
@@ -1,16 +1,68 @@
...
+    ignoreFailures = false
+    if (rootProject.hasProperty("checkstyle.ignoreFailures")) {
+        ignoreFailures = rootProject.properties["checkstyle.ignoreFailures"].toBoolean()
+    }
+    configProperties["rootDir"] = file("$rootDir/../")
+}
+
+// Disable checkstyle if no java8.
+checkstyleMain.enabled = JavaVersion.current().isJava8Compatible()
+checkstyleTest.enabled = JavaVersion.current().isJava8Compatible()
```
sebright: Should we assume that the Java version is at least 8 here, since the source also requires Java 8?

scripts/travis_script
```diff
@@ -41,6 +42,10 @@ case "$TASK" in
         # https://docs.travis-ci.com/user/reference/osx/#JDK-and-OS-X
         ./gradlew clean assemble --stacktrace
         ./gradlew check
+        java_version=`java -version 2>&1 | awk -F '"' '/version/ {print $2}'`
+        if [[ ! "$java_version" < "1.8" ]]; then
```
sebright: I don't think this check is necessary since we are currently only testing with Java 8 on OS X.

HailongWen:
@sebright Thanks for the review. I'll address them tomorrow.
For the example command I think I mis-added them in another PR. I'll add them back here.

scripts/travis_script
HailongWen: Done.

examples/build.gradle
HailongWen: Yes, I leave it here just because it might add productivity. But given the fact that we don't have that many codes in examples, I suppose we can just do it manually.

Another approach is to apply the plugin in root project so that we can run formatter from parent folder. But this somehow contradict purpose of making example standalone.

So I think I should just remove this plugin.

sebright: I think it would be okay to enforce style checks from the main project, since `examples` could still be built without those plugins.  We could add that later, though.

examples/build.gradle
HailongWen: Done.

PR #891, opened 2017-12-08
Add maven build support for `examples`.

HailongWen:
This PR adds a `pom.xml` in examples to enable support for maven build.

Currently I only add `appassembler` plugin, which is an equivalent for `application` plugin in gradle and it has [MIT license](http://www.mojohaus.org/appassembler/appassembler-maven-plugin/license.html).


examples/pom.xml
```diff
@@ -0,0 +1,115 @@
...
+            <source>1.8</source>
+            <target>1.8</target>
+          </configuration>
+        </plugin>
+      </plugins>
+    </pluginManagement>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-checkstyle-plugin</artifactId>
```
sebright: I don't think we need to add Checkstyle to more than one build file, since we only need to check the licenses once in CI.

HailongWen: Done.

HailongWen:
Subitem of #855, depending on #890 to merge.

sebright:
@HailongWen Could you please rebase this branch onto master?  It looks like the diff still contains #890.

bogdandrutu:
You need to use rebase when sync your branch with the master, not merge. Please also consider to use "squash and merge" when merge to the master.

examples/pom.xml
```diff
@@ -0,0 +1,86 @@
...
+  <artifactId>opencensus-examples</artifactId>
+  <packaging>jar</packaging>
+  <version>0.11.0-SNAPSHOT</version><!-- CURRENT_OPENCENSUS_VERSION -->
+  <name>opencensus-examples</name>
+  <url>http://maven.apache.org</url>
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <!-- change to the version you want to use. -->
+    <opencensus.version>0.10.1</opencensus.version><!-- LATEST_OPENCENSUS_VERSION -->
+    <checkstyle.failsOnError>false</checkstyle.failsOnError>
```
sebright: This property can also be removed.

HailongWen: Done.

PR #893, opened 2017-12-09
Add bazel build support for `examples`.

HailongWen:
To resolve the transitive dependency, I used [migration-tooling](https://github.com/bazelbuild/migration-tooling) to auto-generate libraries definitions. Command can be found in the generated file, which might help when updating the build file in the future.

This PR is a sub-item of #855 and depends on #890 to be merged. I'll update the `README.md` later.

### DIFF
--- a/examples/BUILD.bazel
+++ b/examples/BUILD.bazel
@@ -2,7 +2,7 @@ load("//:opencensus_workspace.bzl", "opencensus_java_libraries")
 opencensus_java_libraries()
 
 java_library(
-    name = "examples",
+    name = "opencensus_examples",
     srcs = glob(
         ["src/main/java/**/*.java"]
     ),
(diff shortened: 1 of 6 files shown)


bogdandrutu:
How do we make sure that we don't break this?

HailongWen:
As @sebright suggested downloading bazel while building might take a while. Maybe we should add another build task to verify it?

sebright:
I would rather create a new build job, since it allows more parallelism, and it might simplify the Travis script.  Do you think we should also build the examples on Windows and OS X with each build tool?

bogdandrutu:
@sebright I am fine only on linux.

HailongWen:
@bogdandrutu Of course. Done.

examples/BUILD.bazel
```diff
@@ -0,0 +1,61 @@
+load("//:opencensus_workspace.bzl", "opencensus_java_libraries")
+opencensus_java_libraries()
+
+java_library(
+    name = "examples",
```
sebright: opencensus-examples?

PR #894, opened 2017-12-09
Add MessageEvent and deprecate NetworkEvent. (closes #858)

### LINKED ISSUE #858: Rename NetworkEvent to MessageEvent
Data-model has changed and our API should use the new terminology (see https://github.com/census-instrumentation/opencensus-proto/blob/master/trace/trace.proto).

The NE class has to stay around but be marked as deprecated and ideally implemented using a ME.

HailongWen:
This PR responds to the renaming of `NetworkEvent` to `MessageEvent` in [proto](https://github.com/census-instrumentation/opencensus-proto/pull/40#partial-pull-merging).

The PR itself is complete while I am still working on testing its compatibility in grpc to ensure that it won't break grpc build.

@bogdandrutu @sebright @songy23 please kindly help review and leave comments, thanks!


impl_core/src/main/java/io/opencensus/implcore/trace/SpanImpl.java
```diff
@@ -28,7 +28,7 @@
 import io.opencensus.trace.AttributeValue;
 import io.opencensus.trace.EndSpanOptions;
 import io.opencensus.trace.Link;
-import io.opencensus.trace.NetworkEvent;
+import io.opencensus.trace.MessageEvent;
```
sebright: If the implementation uses MessageEvent, then it won't work with previous versions of the API.  I don't know if it's worth continuing to support that use case, though.  /cc @bogdandrutu

api/src/main/java/io/opencensus/trace/Span.java
```diff
@@ -144,8 +144,20 @@ public final void addAnnotation(String description) {
...
+    addMessageEvent(networkEvent.getMessageEvent());
+  }
+
+  /**
+   * Adds a MessageEvent to the {@code Span}.
+   *
+   * @param messageEvent the message to add.
+   * @since 0.10.0
+   */
+  public abstract void addMessageEvent(MessageEvent messageEvent);
```
sebright: This method should have a default implementation that calls addNetworkEvent, so that it doesn't break any existing subclasses, such as fake Span implementations for tests.

api/src/main/java/io/opencensus/trace/NetworkEvent.java
```diff
@@ -28,6 +29,8 @@
  * serves to uniquely identify each network message. It can optionally can have information about
  * the kernel time and message size.
  */
+// TODO(@Hailong): add `Deprecated` and `deprecated` annotation once we've removed most of the
+// imports.
```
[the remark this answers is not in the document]
HailongWen: To maintain backward compatibility, we have lots of classes which still use `NetworkEvent`. If we deprecate this class now, the compiler will generate warning messages on the deprecated import, which cannot be suppressed.

But of course, we could deprecate it as long as we remove all the imports, substitute all occurrences of `NetworkEvent` to `io.opencensus.trace.NetworkEvent` and add `@SuppressWarnings` annotations to those methods who use it.

api/src/main/java/io/opencensus/trace/Span.java
```diff
@@ -144,8 +144,20 @@ public final void addAnnotation(String description) {
    * higher level applications.
    *
    * @param networkEvent the network event to add.
+   * @deprecated Use {@link #addMessageEvent} for higher level applications.
```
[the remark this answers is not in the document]
HailongWen: Yes. I intend to explain why we deprecate the method (because another API for higher level application is provided). Maybe I should just omit it to avoid misunderstanding.

api/src/main/java/io/opencensus/trace/export/SpanData.java
```diff
@@ -70,7 +71,7 @@ public static SpanData create(
       Timestamp startTimestamp,
       Attributes attributes,
       TimedEvents<Annotation> annotations,
-      TimedEvents<NetworkEvent> networkEvents,
+      TimedEvents<MessageEvent> messageEvents,
```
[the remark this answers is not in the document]
HailongWen: Thanks @sebright. I initially thought this is only used by our impl_core but maybe I am wrong.

This is the most tricky part in this issue. My thoughts here are:

* Add another create() with `TimeEvents<MessageEvent>` as input in parallel: CANNOT be done due to Java compiler's type erasure. The added method will be considered to have the same signature.
* Make a superclass of `MessageEvent` and `NetworkEvent`: I initially thought about that but got trapped by the enum of `NetworkEvent.Type` (enum can not be inherited). So if we do so, we still have to break the compatibility of using NetworkEvent.Type (especially that MessageEvent.Type changed the name of "RECV" to "RECEIVED".
* Make one message extend another: same issue with above - users will end up using MessageEvent with NetworkEvent.Type (or the other way around).

The enum issue can somehow be worked around by changing it to a constant. Though it is not a recommended way.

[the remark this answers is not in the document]
sebright: I thought that the common superclass could have a package-private constructor and no methods.  Then SpanData.create could assume that the subclass is either MessageEvent or NetworkEvent (because no one else would be able to extend Event) and use instanceof/casting to handle the two cases.  Then the two enums could remain independent.

api/src/main/java/io/opencensus/trace/Span.java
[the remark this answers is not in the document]
HailongWen: I mean `addMessageEvent`, which is a more generalized API for higher level applications to use (including network application). The addNetworkEvent only applies to network application.

Anyway, I'll just use `Use {@link #addMessageEvent}.` to avoid misunderstanding.

api/src/main/java/io/opencensus/trace/Span.java
HailongWen: Actually it is the opposite way - the `addNetworkEvent` is implemented by `addMessageEvent`.

As to subclasses, I've added implementation for all of them in current version so tests run with same version of api and impl won't break.
If users have older version of opencensus-api library, they won't be able to access this API so everything is fine.

impl_core/src/main/java/io/opencensus/implcore/trace/SpanImpl.java
HailongWen: Yes, that will be a problem. I'll reverse the implementation.

api/src/main/java/io/opencensus/trace/export/SpanData.java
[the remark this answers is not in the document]
HailongWen: I can keep `SpanData.create` signature unchanged if I address your next comment (use `NetworkEvent` in implementation). So this change should not be a problem.

api/src/main/java/io/opencensus/trace/MessageEvent.java
```diff
@@ -0,0 +1,144 @@
...
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.google.auto.value.AutoValue;
+import io.opencensus.common.Timestamp;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.Immutable;
+
+/**
+ * A class that represents a message event. It requires a {@link Type type} and a message id that
```
adriancole: this javadoc needs help, especially as it replaces network event. Instead of mentioning the field names, how about what this is used for? For example, messaging is like pubsub? What use cases does this help with? If the enum values are to be added later, what might they be? Is this application layer, network layer, transport layer? Just some things to help jog what this is for.

HailongWen:
Thanks for all the comments!

I probably need to redo this PR in another way - taking into account that we want to support "old API lib + new IMPL lib" case, it is not possible to implement `NetworkEvent` using newly added `MessageEvent` since the implementation of `NetworkEvent` (`AutoValue_NetworkEvent`) is along with the API lib.

HailongWen:
>I'm not sure I understand. AutoValue shouldn't affect the NetworkEvent API. 

Yes. The autovalue does not affect the API, while it provides an implementation of `NetworkEvent`, which makes it difficult to fulfill this:
> The NE class has to stay around but be marked as deprecated and **ideally implemented using a ME**.

This PR initially change `NetworkEvent` to be implemented via wrapping a `MessageEvent`, so as to reduce code duplication. But it won't work in old API lib `NetworkEvent` will always come with an auto-value implementation. Unlike `Span` whose interface is defined in opencensus-api and implementation in opencensus-impl-core, the interface and implementation of `NetworkEvent` are all in opencensus-api.

>Also, I thought that you could continue to use AutoValue for NetworkEvent and MessageEvent if you handled SpanData.create by adding a common superclass for the two event classes.

We can not use superclass due to the same reason. If user relies on the`SpanData#create()` method in old version of API, a superclass and wildcard inheritance does not resolve this issue because in old API lib `NetworkEvent` is not a subclass of the newly added superclass. I think I might have to add a new method, something like `createWithMessageEvent()`, to get around the type erase check.

api/src/main/java/io/opencensus/trace/MessageEvent.java
HailongWen: Thanks! I will update that later.

api/src/main/java/io/opencensus/trace/Span.java
HailongWen: Thanks! I understand and fully agree. I initially thought gRPC relies on no-op (but explicitly not depending on opencensus-impl) but turns out I was wrong.

api/src/main/java/io/opencensus/trace/export/SpanData.java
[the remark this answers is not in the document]
HailongWen: I will not change it. The old API should be preserved. Instead adding another method with different name and signature (e.g. `createWithMessageEvents`) to get around the type erase check during compilation.

HailongWen:
The basic rule of maintaining compatibility of a published API is to add new things while not change/delete old things. That is why I want to redo this PR to remove current implementation change of `NetworkEvent`.

>If we don't make any backwards-incompatible changes to the API, then the implementation should work without any changes. I don't know if it's worth delaying the update to the implementation, though.

IIUC, if we don't implement the new API in opencensus-impl-core, we need to implement in opencensus-api. This somehow violates the rule that api and implementation should be separated.

bogdandrutu:
Ping me when this is ready for review.

HailongWen:
@bogdandrutu @songy23 @sebright PTAL.
I've made some changes to the PR. The most significant one is adding a `BaseMessageEvent` as superclass of `MessageEvent` and `NetworkEvent` to solve backward compatibility of `SpanData.create(...)`. I've also modified the javadoc to tell why we use `MessageEvent` to deprecate `NetworkEvent`.

api/src/main/java/io/opencensus/internal/BaseMessageEvent.java
```diff
@@ -0,0 +1,64 @@
...
+import io.opencensus.trace.MessageEvent;
+import io.opencensus.trace.MessageEvent.Type;
+
+/**
+ * Superclass for {@link MessageEvent} and {@link io.opencensus.trace.NetworkEvent} to resolve API
+ * backward compatibility issue.
+ *
+ * @since 0.12.0
+ */
+public abstract class BaseMessageEvent {
```
bogdandrutu: Why public?

api/src/main/java/io/opencensus/trace/MessageEvent.java
```diff
@@ -0,0 +1,152 @@
...
+
+  /**
+   * Returns the kernel timestamp associated with the {@code MessageEvent} or {@code null} if not
+   * set.
+   *
+   * @return the kernel timestamp associated with the {@code MessageEvent} or {@code null} if not
+   *     set.
+   */
+  @Nullable
+  public abstract Timestamp getKernelTimestamp();
```
bogdandrutu: We do not have this in the MessageEvent.

api/src/main/java/io/opencensus/internal/BaseMessageEvent.java
sebright: BaseMessageEvent should have a package-private constructor to ensure that there are only two subclasses.

api/src/main/java/io/opencensus/trace/Span.java
```diff
@@ -146,9 +146,23 @@ public final void addAnnotation(String description) {
    * higher level applications.
    *
    * @param networkEvent the network event to add.
+   * @deprecated Use {@link #addMessageEvent}.
    */
+  @Deprecated
   public abstract void addNetworkEvent(NetworkEvent networkEvent);
```
sebright: addNetworkEvent should be implemented in terms of addMessageEvent, so that users can migrate Span subclasses to the new API, as I described in my previous comment.

api/src/main/java/io/opencensus/trace/export/SpanData.java
```diff
@@ -151,12 +171,32 @@ public static SpanData create(
...
+              timedEvent.getTimestamp(), timedEvent.getEvent().asNetworkEvent()));
+    }
+    return TimedEvents.<io.opencensus.trace.NetworkEvent>create(
+        networkEventsList, timedEvents.getDroppedEventsCount());
+  }
+
   /**
    * Returns network events recorded for this {@code Span}.
    *
    * @return network events recorded for this {@code Span}.
```
sebright: s/network/message

api/src/main/java/io/opencensus/trace/export/SpanData.java
```diff
@@ -69,6 +72,7 @@
    *     Span} is still active.
    * @return a new immutable {@code SpanData}.
    */
+  @SuppressWarnings({"deprecation", "unchecked"})
```
sebright: It would be safer to limit `@SuppressWarnings("unchecked")` to one line by assigning the cast value to a local variable and annotating only the variable.

api/src/main/java/io/opencensus/trace/export/SpanData.java
```diff
@@ -151,12 +171,32 @@ public static SpanData create(
...
   public abstract TimedEvents<Annotation> getAnnotations();
 
+  /**
+   * Returns network events recorded for this {@code Span}.
+   *
+   * @return network events recorded for this {@code Span}.
+   * @deprecated Use {@link #getMessageEvents}.
+   */
+  @Deprecated
+  @SuppressWarnings({"deprecation", "unchecked"})
```
sebright: Same here

contrib/zpages/src/main/java/io/opencensus/contrib/zpages/TraceConfigzZPageHandler.java
```diff
@@ -117,7 +117,7 @@ public void emitHtml(Map<String, String> queryMap, OutputStream outputStream) {
           QUERY_COMPONENT_MAX_NUMBER_OF_ANNOTATIONS,
           TraceParams.DEFAULT.getMaxNumberOfAnnotations(),
           QUERY_COMPONENT_MAX_NUMBER_OF_NETWORK_EVENTS,
-          TraceParams.DEFAULT.getMaxNumberOfNetworkEvents(),
+          TraceParams.DEFAULT.getMaxNumberOfMessageEvents(),
```
sebright: Does this library need to be compatible with older versions of the API?  It might be safer to call getMaxNumberOfNetworkEvents().  The same applies to the other "contrib" libraries.

api/src/main/java/io/opencensus/trace/MessageEvent.java
HailongWen: @bogdandrutu if users have already been using `NetworkEvent.getKernelTimestamp()` to record kernelTimeStamp, what should they use to do so when switching to `MessageEvent`? Or they should just drop that field?

api/src/main/java/io/opencensus/internal/BaseMessageEvent.java
HailongWen: I can remove it and make it package protected in `io.opencensus.trace`.

api/src/main/java/io/opencensus/trace/MessageEvent.java
bogdandrutu: Nobody uses it. Dropping is fine.

api/src/main/java/io/opencensus/trace/Span.java
HailongWen: Are you suggesting implementing  `addNetworkEvent` and leave `addMessageEvent` abstract?
I am not sure if I am understanding your previous comment correctly :) 
>This method (addMessageEvent) should have a default implementation that calls addNetworkEvent, so that it doesn't break any existing subclasses, such as fake Span implementations for tests.

I myself do agree with you that it is better to implement `addNetworkEvent` using `addMessageEvent`.

sebright: I meant that addNetworkEvent should call addMessageEvent, and addMessageEvent should call addNetworkEvent, so that either one can be overridden.  Users can start using the new version of OpenCensus while they override addNetworkEvent and can then switch to only overriding addMessageEvent.  That is how we have handled other deprecated overridable methods.

HailongWen: Found a similar case:
https://github.com/census-instrumentation/opencensus-java/blob/a19d288a49a6635bfc3e99740fcf697234cd476e/api/src/main/java/io/opencensus/trace/Span.java#L112

Here `putAttributes` deprecated old API `addAttributes`. In API definition, the new API is by default implemented using old one. However, in real implementation, both of the APIs are overridden and old API is implemented using the new one.

contrib/zpages/src/main/java/io/opencensus/contrib/zpages/TraceConfigzZPageHandler.java
HailongWen: As to the text in html, I think we will have cross-language template to take care of that so now I will just revert the change as well.

api/src/main/java/io/opencensus/trace/Span.java
```diff
@@ -158,10 +158,31 @@ public final void addAnnotation(String description) {
...
+   * <p>For testing purposes, this method should always be overridden.
+   *
    * @param networkEvent the network event to add.
+   * @deprecated Use {@link #addMessageEvent}.
    * @since 0.5
    */
-  public abstract void addNetworkEvent(NetworkEvent networkEvent);
+  @Deprecated
+  public void addNetworkEvent(NetworkEvent networkEvent) {
+    throw new UnsupportedOperationException(
```
HailongWen: @sebright suggested that we can implement this as calling `addMessageEvent()` and save effort when deprecating this method. But since `addMessageEvent()` is also implemented as `addNetworkEvent()` below, I am worrying that new adopters who do not mock any of these methods may hit a recursive calling issue. 

Since there is not enforcement of override in the API contract, I made `addNetworkEvent` non-abstract here but throws an exception when directly called. This is to enforce this method to be overridden in testing environment, such that it is still be treated as abstract.

Users with opencensus-impl will not be affected since it is overridden in the impl.
Users who mock this class (especially grpc) should not be affected since they have previously overridden this method.

HailongWen:
Thanks again for all the comments and sorry about the back-and-forth.
Remove WIP for another round of review. @sebright @bogdandrutu PTAL, especially the deprecation of `Span.addNetworkEvent()` part.

HailongWen:
Compatibility report can be found [here](https://rawgit.com/HailongWen/19f404e5e1f983eaba1502ef19bdc7a4/raw/compatibility_report_opencensus_api_0.11.1_0.12.0-SNAPSHOT.html).
Some high-severity problems are false positive about `AutoValue` abstract methods.

api/src/main/java/io/opencensus/trace/BaseMessageEvent.java
```diff
@@ -0,0 +1,82 @@
...
+   * @param event the {@code BaseMessageEvent} that is being cast or converted.
+   * @return a {@code MessageEvent} representation of the input.
+   * @since 0.12
+   */
+  public static MessageEvent asMessageEvent(BaseMessageEvent event) {
+    checkNotNull(event);
+    if (event instanceof MessageEvent) {
+      return (MessageEvent) event;
+    }
+    assert event instanceof NetworkEvent;
```
sebright: I don't think the assertion is necessary, because `checkNotNull(event)` or the cast to NetworkEvent would already catch this problem without assertions enabled.

api/src/main/java/io/opencensus/trace/BaseMessageEvent.java
```diff
@@ -0,0 +1,82 @@
...
+   * @param event the {@code BaseMessageEvent} that is being cast or converted.
+   * @return a {@code NetworkEvent} representation of the input.
+   * @since 0.12
+   */
+  public static NetworkEvent asNetworkEvent(BaseMessageEvent event) {
+    checkNotNull(event);
+    if (event instanceof NetworkEvent) {
+      return (NetworkEvent) event;
+    }
+    assert event instanceof MessageEvent;
```
sebright: Same here

api/src/main/java/io/opencensus/trace/BaseMessageEvent.java
```diff
@@ -0,0 +1,82 @@
...
+   * Cast or convert a {@link BaseMessageEvent} to {@link MessageEvent}.
+   *
+   * <p>Warning: if the input is a {@code NetworkEvent} and contains {@code kernelTimestamp}
+   * information, this information will be dropped.
+   *
+   * @param event the {@code BaseMessageEvent} that is being cast or converted.
+   * @return a {@code MessageEvent} representation of the input.
+   * @since 0.12
+   */
+  public static MessageEvent asMessageEvent(BaseMessageEvent event) {
```
sebright: I think that the two static methods in this class could go in a utility class in the `internal` package, since they are only used by opencensus-api.

api/src/main/java/io/opencensus/trace/Span.java
HailongWen: Just wondering that StackOverflowError might seems a defect to users who doesn't read the javadoc carefully, while checked UnsupportedOperationException give more information on what users should do. I am fine with recursive calling since we've done that before.

api/src/main/java/io/opencensus/trace/BaseMessageEvent.java
HailongWen: I am trying to get rid the `BC_UNCONFIRMED_CAST` from findbugs. It cannot be suppressed by add annotation. The framework is not smart enough to tell that there are only two subclasses. Assertion is one way to work around, another one is to use judgement `if (event instanceof NetworkEvent)` to wrap the body. I'll go for the second option.

api/src/main/java/io/opencensus/trace/BaseMessageEvent.java
bogdandrutu: Please move it to the internal package if they are not really public.

api/src/main/java/io/opencensus/trace/BaseMessageEvent.java
sebright: You could suppress the warning in https://github.com/census-instrumentation/opencensus-java/blob/master/findbugs-exclude.xml.

api/src/main/java/io/opencensus/trace/BaseMessageEvent.java
bogdandrutu: What about moving the whole class to the internal package?

api/src/main/java/io/opencensus/trace/NetworkEvent.java
```diff
@@ -28,14 +28,16 @@
...
+ * @deprecated Use {@link MessageEvent}.
  * @since 0.5
  */
 @Immutable
 @AutoValue
 // Suppress Checker Framework warning about missing @Nullable in generated equals method.
 @AutoValue.CopyAnnotations
 @SuppressWarnings("nullness")
-public abstract class NetworkEvent {
+@Deprecated
```
bogdandrutu: Probably a comment to use MessageEvent.

api/src/main/java/io/opencensus/trace/BaseMessageEvent.java
HailongWen: I just move the class out from the package :)
Now I am moving the covert logic back in internal, and leaves the class itself in `io.opencensus.trace` as a part of the API.

api/src/main/java/io/opencensus/trace/NetworkEvent.java
HailongWen: It should be at line 31 :)

api/src/main/java/io/opencensus/trace/BaseMessageEvent.java
```diff
@@ -0,0 +1,27 @@
...
+
+package io.opencensus.trace;
+
+/**
+ * Superclass for {@link MessageEvent} and {@link NetworkEvent} to resolve API backward
+ * compatibility issue.
+ *
+ * @since 0.12
+ */
+public abstract class BaseMessageEvent {
```
bogdandrutu: Do we need this?

api/src/main/java/io/opencensus/trace/export/SpanData.java
```diff
@@ -82,11 +87,30 @@ public static SpanData create(
       Timestamp startTimestamp,
       Attributes attributes,
       TimedEvents<Annotation> annotations,
-      TimedEvents<NetworkEvent> networkEvents,
+      TimedEvents<? extends BaseMessageEvent> messageOrNetworkEvents,
```
bogdandrutu: Why do we not deprecate this create and add a new one which accepts MessageEvent and completely remove the BaseMessageEvent which becomes a completely internal thing?

sebright: This method can't be overloaded with parameter types that differ only in the type of the TimedEvent, because the signatures are the same after generic type erasure.  BaseMessageEvent allows the same method to accept both `TimedEvents<NetworkEvent>` and `TimedEvents<MessageEvent>`.

bogdandrutu: Sure, but then can BaseMessageEvent be package protected?

sebright: BaseMessageEvent is part of the SpanData API, so I think it should be public.

HailongWen: Plus, `SpanData` is not in the same package so it has to be public for `SpanData` to access it.

HailongWen: Sure. Will do.

api/src/main/java/io/opencensus/internal/BaseMessageEventUtil.java
```diff
@@ -0,0 +1,81 @@
...
+public class BaseMessageEventUtil {
+  /**
+   * Cast or convert a {@link BaseMessageEvent} to {@link MessageEvent}.
+   *
+   * <p>Warning: if the input is a {@code io.opencensus.trace.NetworkEvent} and contains {@code
+   * kernelTimestamp} information, this information will be dropped.
+   *
+   * @param event the {@code BaseMessageEvent} that is being cast or converted.
+   * @return a {@code MessageEvent} representation of the input.
+   * @since 0.12
```
sebright: The `@since` tag is unnecessary here, since classes in the internal package should only be used within opencensus-api.

api/src/main/java/io/opencensus/internal/BaseMessageEventUtil.java
```diff
@@ -0,0 +1,81 @@
...
+import io.opencensus.trace.BaseMessageEvent;
+import io.opencensus.trace.MessageEvent;
+
+/**
+ * Helper class to convert/cast between for {@link MessageEvent} and {@link NetworkEvent}.
+ *
+ * @since 0.12
+ */
+@SuppressWarnings("deprecation")
+public class BaseMessageEventUtil {
```
sebright: This class could be final.

HailongWen: Done. PTAL.

api/src/main/java/io/opencensus/trace/BaseMessageEvent.java
```diff
@@ -0,0 +1,35 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.trace;
+
+/**
+ * Superclass for {@link MessageEvent} and {@link NetworkEvent} to resolve API backward
```
adriancole: I understand why we are doing this, because eventhough the version is 0.12, we are treating it as a major release. That said, `BaseMessageEvent` seems a bit crap to keep forever. Why was `Event` or similar not ok?

api/src/main/java/io/opencensus/trace/BlankSpan.java
```diff
@@ -58,8 +58,13 @@ public void addAnnotation(Annotation annotation) {}
 
   /** No-op implementation of the {@link Span#addNetworkEvent(NetworkEvent)} method. */
   @Override
+  @SuppressWarnings("deprecation")
```
adriancole: why surpress vs re-deprecate?

api/src/main/java/io/opencensus/trace/MessageEvent.java
```diff
@@ -0,0 +1,154 @@
...
+package io.opencensus.trace;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.google.auto.value.AutoValue;
+import javax.annotation.concurrent.Immutable;
+
+/**
+ * A class that represents a generic messaging event. This class can represent messaging happened in
+ * any layer, especailly higher application layer. Thus, it can be used when recording events in
```
adriancole: typo especailly

HailongWen: Thanks. Done.

api/src/main/java/io/opencensus/trace/BaseMessageEvent.java
HailongWen: The naming itself did seem crap. I didn't use `Event` since it might be misleading to refer to anything happened in the system.
We don't want this class forever, we plan to remove `NetworkEvent` and `BaseMessageEvent` in the future.

api/src/main/java/io/opencensus/trace/BlankSpan.java
HailongWen: Done. Re-deprecated it.

api/src/main/java/io/opencensus/trace/BaseMessageEvent.java
adriancole: since we are exposing BaseMessageEvent public, even if hiding ctor, if it is temporary, mark it deprecated. that way no one in their right mind will start casting to it.

HailongWen:
Thanks for the review!

Issue #872, opened 2017-12-04
Consider using a Builder pattern for creating Stackdriver exporter


songy23:
Since there are a lot of parameters to the constructor of Stackdriver exporter, factory pattern seems less desirable for creating exporter. Considering changing to a builder pattern to make the API easier to use.

sebright:
Maybe we could add a StackDriver configuration class, with an AutoValue builder, and have a single factory method that takes an instance of that class.

songy23:
Sounds great. I post PR https://github.com/census-instrumentation/opencensus-java/pull/896 using the AutoValue.Builder approach.

PR #896, opened 2017-12-11
Add a StackdriverConfiguration class and builder to create StackdriverStatsExporter

songy23:
Fix https://github.com/census-instrumentation/opencensus-java/issues/872. Use a builder pattern to replace the old factory methods, since we now have many parameters when creating StackdriverStatsExporter. 

The stats exporter README file needs to be updated after v0.11.0 release. Tracked in https://github.com/census-instrumentation/opencensus-java/issues/895.

### DIFF
--- a/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverConfiguration.java
+++ b/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverConfiguration.java
@@ -1,0 +1,119 @@
+/*
+ * Copyright 2017, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.exporter.stats.stackdriver;
+
+import com.google.api.MonitoredResource;
+import com.google.auth.Credentials;
+import com.google.auto.value.AutoValue;
+import io.opencensus.common.Duration;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.Immutable;
+
+/** Configurations for {@link StackdriverStatsExporter}. */
+@AutoValue
+@Immutable
+public abstract class StackdriverConfiguration {
+
+  StackdriverConfiguration() {}
+
+  /**
+   * Returns the {@link Credentials}.
+   *
+   * @return the {@code Credentials}.
+   */
+  @Nullable
+  public abstract Credentials getCredentials();
+
+  /**
+   * Returns the project id.
+   *
+   * @return the project id.
+   */
+  @Nullable
+  public abstract String getProjectId();
+
+  /**
+   * Returns the export interval between pushes to StackDriver.
+   *
+   * @return the export interval.
+   */
+  @Nullable
+  public abstract Duration getExportInterval();
+
+  /**
+   * Returns the Stackdriver {@link MonitoredResource}.
+   *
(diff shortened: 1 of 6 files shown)


exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporter.java
```diff
@@ -189,21 +205,41 @@ public static void createAndRegisterWithProjectIdAndMonitoredResource(
...
-    createInternal(null, ServiceOptions.getDefaultProjectId(), exportInterval, monitoredResource);
+    createInternal(null, null, exportInterval, monitoredResource);
+  }
+
+  /**
+   * Creates a Stackdriver Stats exporter with a {@link StackdriverConfiguration}.
+   *
+   * @param configuration the {@code StackdriverConfiguration}.
+   */
+  public static void createAndRegisterWithConfiguration(StackdriverConfiguration configuration)
```
sebright: Should we just call this method "createAndRegister", since it will be the only way to initialize the exporter?

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporter.java
```diff
@@ -60,7 +64,11 @@
   private static StackdriverStatsExporter exporter = null;
 
   private static final Duration ZERO = Duration.create(0, 0);
-  private static final MonitoredResource DEFAULT_RESOURCE =
+
+  @VisibleForTesting static final Duration DEFAUL_INTERVAL = Duration.create(60, 0);
```
sebright: DEFAULT_INTERVAL

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporter.java
sebright: I think this Javadoc should include the same information as the Javadocs of the deprecated methods, including the default configuration values.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverConfiguration.java
```diff
@@ -0,0 +1,119 @@
...
+import com.google.auth.Credentials;
+import com.google.auto.value.AutoValue;
+import io.opencensus.common.Duration;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.Immutable;
+
+/** Configurations for {@link StackdriverStatsExporter}. */
+@AutoValue
+@Immutable
+public abstract class StackdriverConfiguration {
```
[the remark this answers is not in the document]
songy23: Make sense, and in case if we want to add StackdriverTraceConfiguration later. Updated.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporter.java
songy23: Done.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporter.java
songy23: Unfortunately that name is used by a previous factory method, so currently we can only use this name.

songy23: Done.

sebright: Couldn't we have two overloads, `createAndRegister(Duration)` and `createAndRegister(StackdriverStatsConfiguration)`?  I don't think anyone would confuse the two methods, because one would be deprecated.

songy23: You're right, just forgot this is an overload method. Updated.

songy23:
Thanks!

PR #900, opened 2017-12-13
Add initial support fox X-Cloud-Trace-Context.

HailongWen:
Implementation based on Bogdan's [initial commit](https://github.com/bogdandrutu/opencensus-java/commit/e933b283ba3a5c23d66d9126c3408c79ff243a5a).

### DIFF
--- a/impl_core/src/main/java/io/opencensus/implcore/trace/propagation/CloudTraceFormat.java
+++ b/impl_core/src/main/java/io/opencensus/implcore/trace/propagation/CloudTraceFormat.java
@@ -1,0 +1,139 @@
+/*
+ * Copyright 2017, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.implcore.trace.propagation;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.google.common.primitives.UnsignedLongs;
+import io.opencensus.trace.SpanContext;
+import io.opencensus.trace.SpanId;
+import io.opencensus.trace.TraceId;
+import io.opencensus.trace.TraceOptions;
+import io.opencensus.trace.propagation.SpanContextParseException;
+import io.opencensus.trace.propagation.TextFormat;
+import java.nio.ByteBuffer;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Implementation of the "X-Cloud-Trace-Context" format, defined by the Google Cloud Trace.
+ *
+ * <p>The supported format is the following:
+ *
+ * <pre>
+ * &lt;TRACE_ID&gt;/&lt;SPAN_ID&gt;[;o=&lt;TRACE_OPTIONS&gt;]
+ * </pre>
+ *
+ * <ul>
+ *   <li>TRACE_ID is a 32-character hex value;
+ *   <li>SPAN_ID is a decimal representation of a unsigned long;
+ *   <li>TRACE_OPTIONS is a hex representation of a 32-bit unsigned integer. The least significant
+ *       bit defines whether a request is traced (1 - enabled, 0 - disabled). Behaviors of other
+ *       bits are currently undefined. This value is optional. Although upstream service may leave
+ *       this value unset to leave the sampling decision up to downstream client, this utility will
+ *       always default it to 0 if absent.
+ * </ul>
+ *
+ * <p>Valid values:
+ *
+ * <ul>
+ *   <li>"105445aa7843bc8bf206b120001000/123;o=1"
+ *   <li>"105445aa7843bc8bf206b120001000/123"
+ *   <li>"105445aa7843bc8bf206b120001000/123;o=0"
+ * </ul>
+ */
+final class CloudTraceFormat extends TextFormat {
+  static final String HEADER_NAME = "X-Cloud-Trace-Context";
+  static final List<String> FIELDS = Collections.singletonList(HEADER_NAME);
+  static final char SPAN_ID_DELIMITER = '/';
+  static final String TRACE_OPTION_DELIMITER = ";o=";
+  static final String SAMPLED = "1";
+  static final String NOT_SAMPLED = "0";
+  static final int TRACE_ID_SIZE = 2 * TraceId.SIZE;
+  static final int TRACE_OPTION_DELIMITER_SIZE = TRACE_OPTION_DELIMITER.length();
+
+  @Override
+  public List<String> fields() {
+    return FIELDS;
+  }
+
+  @Override
+  public <C> void inject(SpanContext spanContext, C carrier, Setter<C> setter) {
+    checkNotNull(spanContext, "spanContext");
+    checkNotNull(setter, "setter");
+    checkNotNull(carrier, "carrier");
+    StringBuilder builder =
+        new StringBuilder()
+            .append(spanContext.getTraceId().toLowerBase16())
+            .append(SPAN_ID_DELIMITER)
+            .append(UnsignedLongs.toString(spanIdToLong(spanContext.getSpanId())))
+            .append(TRACE_OPTION_DELIMITER)
+            .append(spanContext.getTraceOptions().isSampled() ? SAMPLED : NOT_SAMPLED);
+
+    setter.put(carrier, HEADER_NAME, builder.toString());
+  }
+
+  @Override
+  public <C> SpanContext extract(C carrier, Getter<C> getter) throws SpanContextParseException {
+    checkNotNull(carrier, "carrier");
+    checkNotNull(getter, "getter");
+    try {
+      String headerStr = getter.get(carrier, HEADER_NAME);
+      if (headerStr == null) {
+        throw new SpanContextParseException("Missing " + HEADER_NAME);
+      }
+
+      int delimiterPos = headerStr.indexOf(SPAN_ID_DELIMITER);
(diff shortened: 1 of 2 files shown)


impl_core/src/main/java/io/opencensus/implcore/trace/propagation/CloudTraceFormat.java
```diff
@@ -0,0 +1,139 @@
...
+    checkNotNull(getter, "getter");
+    try {
+      String headerStr = getter.get(carrier, HEADER_NAME);
+      if (headerStr == null) {
+        throw new SpanContextParseException("Missing " + HEADER_NAME);
+      }
+
+      int delimiterPos = headerStr.indexOf(SPAN_ID_DELIMITER);
+      int traceOptionsPos = headerStr.indexOf(TRACE_OPTION_DELIMITER);
+      checkArgument(delimiterPos == TRACE_ID_SIZE, "Invalid TRACE_ID size");
```
bogdandrutu: This can be re-written as:
checkArgument(headerStr[TRACE_ID_SIZE] != SPAN_ID_DELIMITER, "Invalid TRACE_ID size");

impl_core/src/main/java/io/opencensus/implcore/trace/propagation/CloudTraceFormat.java
```diff
@@ -0,0 +1,139 @@
...
+    setter.put(carrier, HEADER_NAME, builder.toString());
+  }
+
+  @Override
+  public <C> SpanContext extract(C carrier, Getter<C> getter) throws SpanContextParseException {
+    checkNotNull(carrier, "carrier");
+    checkNotNull(getter, "getter");
+    try {
+      String headerStr = getter.get(carrier, HEADER_NAME);
+      if (headerStr == null) {
```
bogdandrutu: To avoid any out of range:

if (headerStr == null  || headerStr.length() < TRACE_ID_SIZE + 2 /* span_id at least 1 char */) {
  throw new SpanContextParseException("Missing or to short header: " + HEADER_NAME);
}

impl_core/src/main/java/io/opencensus/implcore/trace/propagation/CloudTraceFormat.java
```diff
@@ -0,0 +1,139 @@
...
+        throw new SpanContextParseException("Missing " + HEADER_NAME);
+      }
+
+      int delimiterPos = headerStr.indexOf(SPAN_ID_DELIMITER);
+      int traceOptionsPos = headerStr.indexOf(TRACE_OPTION_DELIMITER);
+      checkArgument(delimiterPos == TRACE_ID_SIZE, "Invalid TRACE_ID size");
+
+      TraceId traceId = TraceId.fromLowerBase16(headerStr.substring(0, TRACE_ID_SIZE));
+      String spanIdStr =
+          headerStr.substring(
```
bogdandrutu: Use:
CharSequence spanIdStr = headerStr.subSequence(...)

impl_core/src/main/java/io/opencensus/implcore/trace/propagation/CloudTraceFormat.java
```diff
@@ -0,0 +1,139 @@
...
+              TRACE_ID_SIZE + 1, traceOptionsPos < 0 ? headerStr.length() : traceOptionsPos);
+      SpanId spanId = longToSpanId(UnsignedLongs.parseUnsignedLong(spanIdStr, 10));
+      TraceOptions traceOptions = TraceOptions.DEFAULT;
+      if (traceOptionsPos > 0) {
+        String traceOptionsStr = headerStr.substring(traceOptionsPos + TRACE_OPTION_DELIMITER_SIZE);
+        if (traceOptionsStr.isEmpty()) {
+          throw new SpanContextParseException("Invalid TRACE_OPTIONS");
+        }
+        if (SAMPLED.equals(traceOptionsStr)) {
+          traceOptions = TraceOptions.builder().setIsSampled(true).build();
```
bogdandrutu: We can pre-build the sampled TraceOptions.

HailongWen: OK.

impl_core/src/main/java/io/opencensus/implcore/trace/propagation/CloudTraceFormat.java
HailongWen: I am using the `checkArgument` on `delimiterPos` to avoid out of range. But I guess your way should be more informative to user.

impl_core/src/main/java/io/opencensus/implcore/trace/propagation/CloudTraceFormat.java
```diff
@@ -0,0 +1,142 @@
...
+ *
+ * <p>Valid values:
+ *
+ * <ul>
+ *   <li>"105445aa7843bc8bf206b120001000/123;o=1"
+ *   <li>"105445aa7843bc8bf206b120001000/123"
+ *   <li>"105445aa7843bc8bf206b120001000/123;o=0"
+ * </ul>
+ */
+final class CloudTraceFormat extends TextFormat {
```
adriancole: curious.. this is package protected, and not used anywhere. Is this intending to be used somewhere? Also, this isn't a widely used format.. should this be in a different directory than core?

bogdandrutu: We will expose this to allow us to deprecate the old stackdriver libraries (this is important because a lot of the devs in this library have to maintain the old SD libraries as well). It is pretty used and is the official format supported for the moment by the Google Cloud (which may not be widely used, but still).

impl_core/src/main/java/io/opencensus/implcore/trace/propagation/CloudTraceFormat.java
```diff
@@ -0,0 +1,142 @@
...
+          headerStr.subSequence(
+              SPAN_ID_START_POS, traceOptionsPos < 0 ? headerStr.length() : traceOptionsPos);
+      SpanId spanId = longToSpanId(UnsignedLongs.parseUnsignedLong(spanIdStr.toString(), 10));
+      TraceOptions traceOptions = OPTIONS_NOT_SAMPLED;
+      if (traceOptionsPos > 0) {
+        String traceOptionsStr = headerStr.substring(traceOptionsPos + TRACE_OPTION_DELIMITER_SIZE);
+        if (traceOptionsStr.isEmpty()) {
+          throw new SpanContextParseException("Invalid TRACE_OPTIONS");
+        }
+        if (SAMPLED.equals(traceOptionsStr)) {
```
bogdandrutu: I think this is one thing that is wrong here. This can be int32 and the last bit is the sample bit. you need to read as an int and compare the last bit. I think that's what the format says, please double check.

HailongWen: Yes, I should have implemented it using bit compare. 
Just one more question, the spec defines it as a hex value, but the examples and the C++ implementation both treat it as a uint32 decimal. Which one is the latest?

adriancole:
Agree to support but wonder if core is the right place. For example, there
is no clone of cloud trace right? If it only exists as a service could it
exist somewhere not in the main code? If made public you will need to
deprecate to remove it easier from a support module.

Ps I am ok with this in core, just playing devils advocate.

You can anticipate if adding this, similar requests for other single-impl
formats, so need a good answer. Primary reason to facilitate moving off the
founding team's proprietary SDK sounds fair to me.

TL;DR is lgtm

impl_core/src/main/java/io/opencensus/implcore/trace/propagation/CloudTraceFormat.java
```diff
@@ -0,0 +1,141 @@
...
+      TraceId traceId = TraceId.fromLowerBase16(headerStr.subSequence(0, TRACE_ID_SIZE));
+      int traceOptionsPos = headerStr.indexOf(TRACE_OPTION_DELIMITER, SPAN_ID_DELIMITER);
+      CharSequence spanIdStr =
+          headerStr.subSequence(
+              SPAN_ID_START_POS, traceOptionsPos < 0 ? headerStr.length() : traceOptionsPos);
+      SpanId spanId = longToSpanId(UnsignedLongs.parseUnsignedLong(spanIdStr.toString(), 10));
+      TraceOptions traceOptions = OPTIONS_NOT_SAMPLED;
+      if (traceOptionsPos > 0) {
+        String traceOptionsStr = headerStr.substring(traceOptionsPos + TRACE_OPTION_DELIMITER_SIZE);
+        if ((UnsignedInts.parseUnsignedInt(traceOptionsStr, 10) & CLOUD_TRACE_IS_SAMPLED) != 0) {
```
bogdandrutu: Integer.parseInt(traceOptionsStr)

HailongWen: If using Integer.parseInt(), we'll have to an extra check to ensure it is bigger than 0.

bogdandrutu: Why? negative means bit 32 is set to 1, nothing wrong.

HailongWen: Just complying with the spec and current c++ impl, which also uses a uint32 version of parsing (`safe_strtou32_base`). In this case, negative value should not be accepted; besides, if users follow this spec to set 32 bit to 1, the decimal string will out of range of Integer.

I am OK with either, though, as long as we document it clearly.

bogdandrutu: fine to keep it this way.

HailongWen: Thanks!

bogdandrutu:
@adriancole put more thoughts into this and probably make sense to have only the TraceContext in the main API and have an contrib_http_util with support for SD, Zipkin, etc.

I agree that will make things easier to deprecate.

adriancole:
Only thing is B3 is widely used by clones, people doing just logging
(splunk etc) and so is a defacto format for several years. Still if you
prefer this is ok

adriancole:
If you chose, i would recommend doing it after trace-context exists in a
stable fashion as B3 is far more stable and supported by probably a hundred
libraries regardless of what export is doing. Otherwise we recommend
something guaranteed to break people.

Long story short i recommend keeping B3 and also whatever the most recent
trace-context format is.

PR #901, opened 2017-12-13
Support previous encoding versions.

songy23:
Do we want to keep backwards-compatible on encoding? I'm not sure about this but Go allows bytes in previous versions: https://github.com/census-instrumentation/opencensus-go/blob/master/tag/map_codec.go#L191


songy23:
@bogdandrutu previously `deserialize()` will reject bytes encoded in older binary versions. However if we decided to keep backwards-compatible on encoding versions, I think it's better for `deserialize()` code with higher version to decode bytes in former versions. 

On the other hand, if the bytes are encoded in a higher version than `deserialize()` code, `deserialize()` should still reject it.

PR #905, opened 2017-12-14
Increase exporting interval and fix warnings.

bogdandrutu:


bogdandrutu:
@HailongWen probably we can do that in the future.

adriancole:
yeah lets do this as zipkin folks are used to 1s delay max, and I've heard folks like dynatrace are 200ms or something. I guess this setting is an attempt to not overrun api quota at stackdriver?

PR #906, opened 2017-12-14
Add helper class to register views.

bogdandrutu:


contrib/grpc_metrics/src/test/java/io/opencensus/contrib/grpc/metrics/RpcViewsTest.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.Map;
+import java.util.Set;
+import javax.annotation.Nullable;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/** Test for {@link RpcViews}. */
+@RunWith(JUnit4.class)
+public class RpcViewsTest {
```
sebright: Suggestion: You could use reflection to test that the two sets of views in RpcViews.java contain all of the view constants in RpcViewConstants.java.

contrib/grpc_metrics/src/main/java/io/opencensus/contrib/grpc/metrics/RpcViews.java
```diff
@@ -0,0 +1,143 @@
...
+  static void registerIntervalViews(ViewManager viewManager) {
+    for (View view : RPC_INTERVAL_VIEWS_SET) {
+      viewManager.registerView(view);
+    }
+  }
+
+  /**
+   * Registers all views.
+   *
+   * <p>This must be called once at the beginning of the process.
```
songy23: nit: might also add a note here this method is equivalent to registerCumulativeViews + registerIntervalViews so users don't need to call them again.

contrib/grpc_metrics/src/main/java/io/opencensus/contrib/grpc/metrics/RpcViews.java
```diff
@@ -0,0 +1,143 @@
...
+          RpcViewConstants.RPC_SERVER_UNCOMPRESSED_RESPONSE_BYTES_HOUR_VIEW);
+
+  /**
+   * Registers all standard cumulative views.
+   *
+   * <p>This must be called once at the beginning of the process.
+   *
+   * @since 0.11.0
+   */
+  public static void registerCumulativeViews() {
```
sebright: I think `registerAllCumulativeViews()` and `registerAllIntervalViews()` would be more consistent with `registerAllViews()`.

contrib/grpc_metrics/src/main/java/io/opencensus/contrib/grpc/metrics/RpcViews.java
```diff
@@ -0,0 +1,143 @@
...
+          RpcViewConstants.RPC_SERVER_RESPONSE_BYTES_HOUR_VIEW,
+          RpcViewConstants.RPC_SERVER_REQUEST_COUNT_HOUR_VIEW,
+          RpcViewConstants.RPC_SERVER_RESPONSE_COUNT_HOUR_VIEW,
+          RpcViewConstants.RPC_SERVER_UNCOMPRESSED_REQUEST_BYTES_HOUR_VIEW,
+          RpcViewConstants.RPC_SERVER_UNCOMPRESSED_RESPONSE_BYTES_HOUR_VIEW);
+
+  /**
+   * Registers all standard cumulative views.
+   *
+   * <p>This must be called once at the beginning of the process.
```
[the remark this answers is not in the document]
HailongWen: Maybe we can rephrase it as a recommendation, rather than a limitation. It would also help if clarifying that stats recorded before register will not be available in the view.

contrib/grpc_metrics/src/main/java/io/opencensus/contrib/grpc/metrics/RpcViews.java
bogdandrutu: done.

contrib/grpc_metrics/src/main/java/io/opencensus/contrib/grpc/metrics/RpcViews.java
bogdandrutu: Done.

contrib/grpc_metrics/src/main/java/io/opencensus/contrib/grpc/metrics/RpcViews.java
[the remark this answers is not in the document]
bogdandrutu: done.

contrib/grpc_metrics/src/test/java/io/opencensus/contrib/grpc/metrics/RpcViewsTest.java
bogdandrutu: Added a TODO for the moment.

PR #907, opened 2017-12-14
Add a convenient method to create Stackdriver exporter with default settings

songy23:


songy23:
@sebright Yes I expect most ordinary users to use the default settings, and only advanced users should care about the optional configurations. The choices of supported MonitoredResource are in fact [quite limited](https://cloud.google.com/monitoring/custom-metrics/creating-metrics#which-resource), for users who don't deploy to GCE, GKE or AWS EC2, they probably won't use custom MonitoredResource. In addition, now that we have the opencensus_task label, the race condition of uploading stats is unlikely to happen even if all exporters use the same MonitoredResource.

Also, Stackdriver Trace exporter has a similar convenient method createAndRegister() with no arguments. I think it's good to keep consistent with that.

songy23:
Thanks!

Issue #359, opened 2017-06-15
Run Checker Framework null analysis as part of the Gradle build.


sebright:
See https://checkerframework.org/manual/#nullness-checker.  Null analysis would help detect bugs and also check the Nullable annotations that are part of the API.  There are instructions for using Gradle: https://checkerframework.org/manual/#gradle

g-easy:
I tried using this plugin: https://plugins.gradle.org/plugin/com.jaredsburrows.checkerframework
But it doesn't seem to play well with errorprone.

https://github.com/g-easy/opencensus-java/commit/105176f0730d3bae727a514fbbf104a7cc20c36b

I also start getting errors about "cannot find symbol AutoValue_*"

sebright:
Thanks for trying it out!  Maybe we can run Error Prone and the Checker Framework in separate build jobs, if they don't interact well.  The errors refer to classes generated by the AutoValue annotation processor, so it's possible that the plugin is overriding the list of annotation processors.

sebright:
This question looks like it might be related: https://groups.google.com/forum/#!topic/checker-framework-discuss/e5nt83iAIqo

sebright:
We may need to use stub files to add annotations for other libraries (https://checkerframework.org/releases/0.9.9/checkers-manual.html#stub-creating-and-using).  For example, we could specify that Preconditions.checkNotNull only returns normally when its argument is non-null.

PR #912, opened 2017-12-19
Add more null annotations (issue #359).

sebright:
This commit adds some Nullable annotations that are required by the Checker
Framework, but it doesn't change any other code.  It also suppresses some Error
Prone and FindBugs warnings that conflict with the Checker Framework, since the
three tools use different algorithms.

### DIFF
--- a/gradle/errorprone/experimental_suggestions
+++ b/gradle/errorprone/experimental_suggestions
@@ -1,6 +1,9 @@
 # FieldMissingNullable is turned off due to
 # https://github.com/google/error-prone/issues/823.
 
+# ReturnMissingNullable is turned off because it
+# conflicts with Checker Framework null analysis.
+
 errorProneExperimentalSuggestions = \
 -Xep:ConstantField:ERROR,\
 -Xep:EmptySetMultibindingContributions:ERROR,\
@@ -16,7 +19,7 @@ errorProneExperimentalSuggestions = \
 -Xep:PrivateConstructorForNoninstantiableModuleTest:ERROR,\
 -Xep:PrivateConstructorForUtilityClass:ERROR,\
 -Xep:RemoveUnusedImports:ERROR,\
--Xep:ReturnMissingNullable:ERROR,\
+-Xep:ReturnMissingNullable:OFF,\
 -Xep:ThrowsUncheckedException:ERROR,\
 -Xep:UngroupedOverloads:ERROR,\
 -Xep:UnnecessarySetDefault:ERROR,\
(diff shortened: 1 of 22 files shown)


impl_core/src/main/java/io/opencensus/implcore/trace/export/ExportComponentImpl.java
```diff
@@ -30,8 +30,8 @@
   private static final Duration EXPORTER_SCHEDULE_DELAY = Duration.create(5, 0);
 
   private final SpanExporterImpl spanExporter;
-  private final RunningSpanStoreImpl runningSpanStore;
-  private final SampledSpanStoreImpl sampledSpanStore;
+  @Nullable private final RunningSpanStoreImpl runningSpanStore;
```
HailongWen: Maybe I missed here when I was implementing the NoopRunningSpanStore and NoopSampledSpanStore - I should have changed them to return noop when `supportInProcessStores` is false. 

We can add `@Nullable` here now and remove them later in another PR.

gradle/errorprone/experimental_suggestions
```diff
@@ -1,6 +1,9 @@
 # FieldMissingNullable is turned off due to
 # https://github.com/google/error-prone/issues/823.
 
+# ReturnMissingNullable is turned off because it
+conflicts with Checker Framework null analysis.
```
HailongWen: Missing `#`?

HailongWen:
Thanks for adding!
BTW, have we enabled check framework already? I didn't see changes on gradle build file so I guess it will be added in another PR?

impl_core/src/main/java/io/opencensus/implcore/trace/export/ExportComponentImpl.java
sebright: Yes, I noticed that inconsistency when running the checker framework.  I tried to fix it quickly by returning no-op classes, but some code expected the values to have type RunningSpanStoreImpl and SampledSpanStoreImpl.  I'm not sure how hard it would be to refactor.

sebright:
Thanks!  This is just the start of enabling null analysis.  I need to make a lot more changes to the code, such as adding more null checks.

gradle/errorprone/experimental_suggestions
sebright: Done

impl_core/src/main/java/io/opencensus/implcore/trace/export/ExportComponentImpl.java
sebright: I opened an issue: #914

PR #913, opened 2017-12-19
Update contrib directory to pass Checker Framework null analysis (issue #359).

sebright:


contrib/zpages/src/main/java/io/opencensus/contrib/zpages/ZPageHandlers.java
```diff
@@ -149,7 +149,9 @@ public void run() {
   private static void stop() {
     synchronized (monitor) {
       // This should never happen because we register the shutdown hook only if we start the server.
-      checkState(server != null, "The HttpServer is already stopped.");
+      if (server == null) {
```
songy23: Why do we need to change this? The previous code looks simpler to me.

sebright: Guava doesn't use Checker Framework annotations, besides `@Nullable`, so the checker doesn't know that checkState only returns normally when the argument is true.  We could probably use stub files to specify annotations on other libraries (https://checkerframework.org/releases/0.9.9/checkers-manual.html#stub-creating-and-using).  I'll add a comment to the issue.

PR #921, opened 2017-12-28
Add Checker Framework stub file for Guava (fixes #920).

### LINKED ISSUE #920: Consider adding Checker Framework stub files to specify annotations for third-party libraries.
See https://checkerframework.org/releases/0.9.9/checkers-manual.html#stub-creating-and-using.  The stub files would allow us to specify nullness annotations for commonly used methods from Guava, such as `Preconditions.checkNotNull`.

sebright:
This commit adds a stub file as described in ~~https://checkerframework.org/releases/0.9.9/checkers-manual.html#stub-creating-and-using~~
 https://checkerframework.org/manual/#stub.  The initial file contains annotations for
com.google.common.base.Strings.isNullOrEmpty.

/cc @songy23

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -96,7 +96,9 @@ subprojects {
         }
         if (useCheckerFramework) {
             it.options.compilerArgs += [
-                '-processor', 'com.google.auto.value.processor.AutoValueProcessor,org.checkerframework.checker.nullness.NullnessChecker'
+                '-processor',
+		'com.google.auto.value.processor.AutoValueProcessor,org.checkerframework.checker.nullness.NullnessChecker',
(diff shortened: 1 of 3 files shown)


sebright:
@HailongWen Yes, I think we'll need to add more classes in the future.  I just wanted to set up the stub file directory so that we can easily add classes as needed.  Each file can contain multiple packages and classes.  What path do you think we should use?

sebright:
I moved the stub file from `stubs/` to `checker-framework/stubs/`.

sebright:
I'll leave the path to the stub files for now.  We can reorganize them as needed as we annotate more methods.

Issue #453, opened 2017-07-12
Add Javadocs for tag keys in RpcMeasureConstants.java.


sebright:
The RPC Measures and Views already have description fields, but it would help to have more documentation for the TagKeys:

https://github.com/census-instrumentation/opencensus-java/blob/4502e9b6164c1616f3448e9017a5b8331c986d85/core/src/main/java/io/opencensus/stats/RpcMeasureConstants.java#L25

PR #932, opened 2018-01-17
Add Javadocs for tag keys in RpcMeasureConstants.java

songy23:
Closes https://github.com/census-instrumentation/opencensus-java/issues/453


contrib/grpc_metrics/src/main/java/io/opencensus/contrib/grpc/metrics/RpcMeasureConstants.java
```diff
@@ -25,7 +25,9 @@
 public final class RpcMeasureConstants {
 
   // Rpc tag keys.
+  /** Tag key that represents a RPC canonical status. */
```
sebright: I think this should say gRPC, since it isn't a general RPC status.  It might help to add a link to the Status class.

songy23: grpc_metrics doesn't depend on gRPC, so I cannot directly link to Status.java. I added a link to gRPC status doc instead.

Issue #869, opened 2017-12-03
Stackdriver Monitoring exporter recieves Goaway after running an hour


rakyll:
After an hour later I start a server that exports to Stackdriver Monitoring, the worker thread throws the following exception:

```Dec 02, 2017 5:26:28 PM io.opencensus.exporter.stats.stackdriver.StackdriverExporterWorkerThread export
WARNING: Exception thrown when exporting TimeSeries.
com.google.api.gax.rpc.UnavailableException: io.grpc.StatusRuntimeException: UNAVAILABLE: HTTP/2 error code: NO_ERROR
Received Goaway
max_age
	at com.google.api.gax.rpc.ApiExceptionFactory.createException(ApiExceptionFactory.java:69)
	at com.google.api.gax.grpc.GrpcExceptionCallable$ExceptionTransformingFuture.setException(GrpcExceptionCallable.java:118)
	at com.google.api.gax.grpc.GrpcExceptionCallable$ExceptionTransformingFuture.onFailure(GrpcExceptionCallable.java:101)
	at com.google.api.core.ApiFutures$1.onFailure(ApiFutures.java:61)
	at com.google.common.util.concurrent.Futures$4.run(Futures.java:1123)
	at com.google.common.util.concurrent.MoreExecutors$DirectExecutor.execute(MoreExecutors.java:435)
	at com.google.common.util.concurrent.AbstractFuture.executeListener(AbstractFuture.java:900)
	at com.google.common.util.concurrent.AbstractFuture.complete(AbstractFuture.java:811)
	at com.google.common.util.concurrent.AbstractFuture.setException(AbstractFuture.java:675)
	at io.grpc.stub.ClientCalls$GrpcFuture.setException(ClientCalls.java:491)
	at io.grpc.stub.ClientCalls$UnaryStreamToFuture.onClose(ClientCalls.java:466)
	at io.grpc.ForwardingClientCallListener.onClose(ForwardingClientCallListener.java:41)
	at io.grpc.internal.CensusStatsModule$StatsClientInterceptor$1$1.onClose(CensusStatsModule.java:663)
	at io.grpc.ForwardingClientCallListener.onClose(ForwardingClientCallListener.java:41)
	at io.grpc.internal.CensusTracingModule$TracingClientInterceptor$1$1.onClose(CensusTracingModule.java:392)
	at io.grpc.internal.ClientCallImpl.closeObserver(ClientCallImpl.java:443)
[...]

PR #935, opened 2018-01-18
Silently skip NO_ERROR Goaway

songy23:
Closes https://github.com/census-instrumentation/opencensus-java/issues/869.

Based on the discussion in https://github.com/census-instrumentation/opencensus-java/issues/869, I think it's safe to silently skip status UNAVAILABLE with error code NO_ERROR.

### DIFF
--- a/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporterWorker.java
+++ b/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporterWorker.java
@@ -182,9 +182,10 @@ final class StackdriverExporterWorker implements Runnable {
         metricServiceClient.createTimeSeries(request);
         span.addAnnotation("Finish exporting TimeSeries.");
       } catch (com.google.api.gax.rpc.UnavailableException e) {
-        if (e.getLocalizedMessage()
-            .startsWith(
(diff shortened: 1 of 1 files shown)


exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporterWorker.java
```diff
@@ -181,13 +181,16 @@ void export() {
                 .build();
         metricServiceClient.createTimeSeries(request);
         span.addAnnotation("Finish exporting TimeSeries.");
+      } catch (com.google.api.gax.rpc.UnavailableException e) {
+        if (e.getLocalizedMessage()
+            .startsWith(
+                "io.grpc.StatusRuntimeException: UNAVAILABLE: HTTP/2 error code: NO_ERROR")) {
+          continue; // Silently skip NO_ERROR Goaway
```
sebright: Should the exporter retry in this case?

songy23: I'm hesitant on doing retry in this case, since I think NO_ERROR Goaway indicates server should have received and processed the request with no error. Retrying might generate duplicated stats and increase the burden of server.

songy23:
Talked offline. I'll close this PR for now, until I can make sure this failure has no impact on the uploaded stats.

Issue #925, opened 2018-01-04
Inconsistent naming for exporters


rakyll:
StackdriverStatsExporter contains stats in the name even though the trace exporter (StackdriverExporter) doesn't.

Names should consistent. Use io.opencensus.exporter.trace.stackdriver.StackdriverTraceExporter instead.

bogdandrutu:
Please make the change and deprecate the old function name.

HailongWen:
@bogdandrutu do we need to maintain backward compatibility for this class?

sebright:
I think we should leave the old class and deprecate it, though it could just call the methods on the new class.

bogdandrutu:
@HailongWen better safe then sorry. I think what @sebright proposed is correct.

bogdandrutu:
We still have inconsistency between method names:
CreateAndRegisterWithProjectId // trace
CreateWithProjectId // stats

We should fix this as well. if you fix in the trace the advantage is that we just change the names without deprecation because the new class was never released.

songy23:
Now all factory methods in StackdriverStatsExporter are also called [CreateAndRegister](https://github.com/census-instrumentation/opencensus-java/blob/master/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporter.java#L162)*.

CreateWithProjectId is an outdated term in StackdriverStatsExporter README. I'll update README to use the new method names (https://github.com/census-instrumentation/opencensus-java/issues/895).

PR #942, opened 2018-01-19
Rename trace exporters that have inconsistent naming.

HailongWen:
This PR contains two commits: 
1. rename old files 
2. add old files back. I didn't use one to implement another, just to avoid adding new logic.

The reason for this two-step renaming is to preserve history on the renamed files (old files will be deprecated and removed some day in the future).
Although the calculated diff shows that `**TraceExporter` are added files, they are just moved and renamed. Using `git blame` may give the correct change history.


bogdandrutu:
Why did you not rename the classes and add a deprecated class with the old names that just calls into the new class?

HailongWen:
I didn't make the old names call into the new class to minimize manual code. I can do that.

HailongWen:
@bogdandrutu PTAL.

exporters/trace/logging/src/test/java/io/opencensus/exporter/trace/logging/LoggingExporterTest.java
```diff
@@ -20,7 +20,7 @@
 import static org.mockito.Matchers.eq;
 import static org.mockito.Mockito.verify;
 
-import io.opencensus.exporter.trace.logging.LoggingExporter.LoggingExporterHandler;
```
bogdandrutu: This file can be removed.

exporters/trace/stackdriver/src/test/java/io/opencensus/exporter/trace/stackdriver/StackdriverExporterTest.java
```diff
@@ -45,9 +45,9 @@ public void registerUnregisterStackdriverExporter() {
     StackdriverExporter.register(spanExporter, handler);
     verify(spanExporter)
         .registerHandler(
-            eq("io.opencensus.exporter.trace.stackdriver.StackdriverExporter"), same(handler));
```
bogdandrutu: Same thing. You can remove this file.

HailongWen:
fixes #925.

PR #946, opened 2018-01-20
Update script in RELEASING.md and use them to update release version in all readmes and build files. (fixes #941)

### LINKED ISSUE #941: Update all readmes and examples to the latest version
* Once the 0.11.0 library is released we should update all readmes and examples code to point to the latest version.
* Add a section in the RELEASING.md about the necessity to update readmes and examples may a shell command to do that easly.

HailongWen:
I added sections in `RELEASING.md` on how to announce a new release and update release versions in readmes and build files.

This PR also contains the result of executing these added scripts. I've manually checked there are no more occurrences of old version "0.10.1".

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,14 +27,14 @@ For Maven add to your `pom.xml`:
   <dependency>
     <groupId>io.opencensus</groupId>
     <artifactId>opencensus-api</artifactId>
-    <version>0.11.0</version> <!-- LATEST_OPENCENSUS_RELEASE_VERSION -->
+    <version>0.11.0</version>
   </dependency>
 </dependencies>
 ```
 
 For Gradle add to your dependencies:
 ```gradle
-compile 'io.opencensus:opencensus-api:0.11.0' //LATEST_OPENCENSUS_RELEASE_VERSION
+compile 'io.opencensus:opencensus-api:0.11.0'
 ```
 
 ### Hello "OpenCensus" trace events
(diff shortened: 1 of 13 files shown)


RELEASING.md
```diff
@@ -189,6 +189,78 @@ Central (the staging repository will be destroyed in the process). You can see
...
+latest version.
+
+1. Update README.md and gradle/maven build files on `master` branch:
+
+```bash
+$ git checkout -b bump-document-version master
+$ RELEASE_VERSION_FILES=(
+  README.md
+  contrib/grpc_util/README.md
+  contrib/http_util/README.md
```
bogdandrutu: contrib/zpages must be updated as well

HailongWen: Do you mean contrib/zpages/README.md? It should be included in next line.

README.md
```diff
@@ -27,14 +27,14 @@ For Maven add to your `pom.xml`:
   <dependency>
     <groupId>io.opencensus</groupId>
     <artifactId>opencensus-api</artifactId>
-    <version>0.10.1</version>
+    <version>0.11.0</version> <!-- LATEST_OPENCENSUS_RELEASE_VERSION -->
```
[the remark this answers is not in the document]
HailongWen: Let me try that. Another option is to use a more complex regex. I think the motivation for adding comment is just to avoid substituting other components' version that happen to be the same as ours.

RELEASING.md
bogdandrutu: Clearly I need to go to an eye doctor 👍

HailongWen:
Thanks for the review!

Hi @sebright , PTAL.
I removed all `LATEST_OPENCENSUS_RELEASE_VERSION` comment in all `README.md`, while those in pom.xml and build.gradle are untouched.

For gradle build examples, I use `'s/\(\(compile\|runtime\).\+io\.opencensus:.\+:\)[0-9]\+\.[0-9]\+\.[0-9]\+/\1'$MAJOR.$MINOR.$PATCH'/'` to do the substitution.

For pom.xml examples, I use `'s/\(<version>\)[0-9]\+\.[0-9]\+\.[0-9]\+/\1'$MAJOR.$MINOR.$PATCH'/'`. Although it does not test whether the dependency is opencensus related, it should be enough for now, since we don't have other dependencies in `README.md`. If we do have that requirement in the future, we might consider other regex (though it will be a little bit tricky - sed only processes one line, so we might need to change all '\n' to some other temporary characters and change it back after substitution is done)

PR #950, opened 2018-01-23
Fixed bug in CloudTraceFormat that made it impossible to use short span id's

rlindooren:
The '/' char value is used to determine from where to start searching in order to find the trace options in the trace header.
I think that using this char value is done by mistake. Because the char value for '/' is 47 making it impossible to find the trace options in a trace header with a short span id (causing it to be shorter that 47).


rlindooren:
I signed it!

bogdandrutu:
Thanks for catching this subtle bug.

rlindooren:
@bogdandrutu no problem!
Do you know when a next release is planned?

bogdandrutu:
@rlindooren we will do a patch release probably today.

rlindooren:
@bogdandrutu that's fast, thanks.

HailongWen:
Thanks for catching that!

PR #957, opened 2018-01-23
Fix some typos in SignalFX exporter code examples

songy23:
/cc @mpetazzoni


mpetazzoni:
Thanks! That was missed when we refactored to introduce the configuration class.

PR #959, opened 2018-01-23
Keep artifacts dependencyConvergence-clean

bogdandrutu:
Maven Enforcer's dependencyConvergence is commonly used in Spring projects.

Currently we cannot do it for the SD exporters because of this https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2803

### DIFF
--- a/contrib/grpc_util/build.gradle
+++ b/contrib/grpc_util/build.gradle
@@ -8,8 +8,18 @@ apply plugin: 'java'
 }
 
 dependencies {
-    compile project(':opencensus-api'),
-            libraries.grpc_core
+    compile project(':opencensus-api')
+
+    compile (libraries.grpc_core) {
+        // Prefer library version.
+        exclude group: 'com.google.errorprone', module: 'error_prone_annotations'
+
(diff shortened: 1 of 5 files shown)


contrib/grpc_util/build.gradle
```diff
@@ -8,8 +8,18 @@ apply plugin: 'java'
...
+
+    compile (libraries.grpc_core) {
+        // Prefer library version.
+        exclude group: 'com.google.errorprone', module: 'error_prone_annotations'
+
+        // Prefer library version.
+        exclude group: 'com.google.code.findbugs', module: 'jsr305'
+
+        // We will always be more up to date.
+        exclude group: 'io.opencensus', module: 'opencensus-api'
```
sebright: How do we know that the opencensus version in this repository is always more up to date?  Is that because we always specify an exact version of grpc_core, which has already be released?

bogdandrutu: because it depends on HEAD so it will always be newer. (this repository in this sentence means opencensus project)

PR #964, opened 2018-01-27
Make `XxxSpanStoreImpl` abstract and add no-op implementation. (fixes #914)

### LINKED ISSUE #914: ExportComponentImpl.getRunningSpanStore() and getSampledSpanStore() can return null.
#679 made the `NoopExportComponent` return no-op implementations for the `RunningSpanStore` and `SampledSpanStore` and removed the `Nullable` annotations on their getters.  However, `ExportComponentImpl.createWithoutInProcessStores(...)` still creates an `ExportComponent` that has a null `RunningSpanStore` and `SampledSpanStore`.  We should make the implementation always return non-null values to be consistent with the no-op classes.

HailongWen:
Since `XxxSpanStoreImpl` have the `final` modifier, they need to be removed to `abstract` so we can add no-op implementation and pass nullness check.

This PR contains two commits. 
- The first one removes "final" modifier of the concrete implementations and rename them to `InProcessXxxSpanStoreImpl` since they serve in-process data collection purpose, it is a WIP PR and temporarily breaks the build.
- The second one rebuilds the old classes with `abstract` modifier and adds no-op implementation and related tests.

### DIFF
--- a/impl_core/src/main/java/io/opencensus/implcore/trace/export/ExportComponentImpl.java
+++ b/impl_core/src/main/java/io/opencensus/implcore/trace/export/ExportComponentImpl.java
@@ -21,7 +21,6 @@ import io.opencensus.implcore.internal.EventQueue;
 import io.opencensus.trace.export.ExportComponent;
 import io.opencensus.trace.export.RunningSpanStore;
 import io.opencensus.trace.export.SampledSpanStore;
-import javax.annotation.Nullable;
 
 /** Implementation of the {@link ExportComponent}. */
 public final class ExportComponentImpl extends ExportComponent {
@@ -30,8 +29,8 @@ public final class ExportComponentImpl extends ExportComponent {
   private static final Duration EXPORTER_SCHEDULE_DELAY = Duration.create(5, 0);
 
   private final SpanExporterImpl spanExporter;
-  @Nullable private final RunningSpanStoreImpl runningSpanStore;
-  @Nullable private final SampledSpanStoreImpl sampledSpanStore;
+  private final RunningSpanStoreImpl runningSpanStore;
+  private final SampledSpanStoreImpl sampledSpanStore;
 
   @Override
   public SpanExporterImpl getSpanExporter() {
--- a/impl_core/src/main/java/io/opencensus/implcore/trace/export/ExportComponentImpl.java
+++ b/impl_core/src/main/java/io/opencensus/implcore/trace/export/ExportComponentImpl.java
@@ -39,17 +39,11 @@ public final class ExportComponentImpl extends ExportComponent {
   }
 
   @Override
-  // TODO(#914): This method shouldn't be nullable.
-  @SuppressWarnings("nullness")
-  @Nullable
   public RunningSpanStoreImpl getRunningSpanStore() {
     return runningSpanStore;
   }
 
   @Override
-  // TODO(#914): This method shouldn't be nullable.
-  @SuppressWarnings("nullness")
-  @Nullable
   public SampledSpanStoreImpl getSampledSpanStore() {
     return sampledSpanStore;
   }
@@ -82,7 +76,13 @@ public final class ExportComponentImpl extends ExportComponent {
    */
   private ExportComponentImpl(boolean supportInProcessStores, EventQueue eventQueue) {
     this.spanExporter = SpanExporterImpl.create(EXPORTER_BUFFER_SIZE, EXPORTER_SCHEDULE_DELAY);
-    this.runningSpanStore = supportInProcessStores ? new RunningSpanStoreImpl() : null;
-    this.sampledSpanStore = supportInProcessStores ? new SampledSpanStoreImpl(eventQueue) : null;
+    this.runningSpanStore =
+        supportInProcessStores
+            ? new InProcessRunningSpanStoreImpl()
+            : RunningSpanStoreImpl.getNoopRunningSpanStoreImpl();
+    this.sampledSpanStore =
+        supportInProcessStores
+            ? new InProcessSampledSpanStoreImpl(eventQueue)
+            : SampledSpanStoreImpl.getNoopSampledSpanStoreImpl();
   }
 }
--- a/impl_core/src/main/java/io/opencensus/implcore/trace/export/InProcessRunningSpanStoreImpl.java
+++ b/impl_core/src/main/java/io/opencensus/implcore/trace/export/InProcessRunningSpanStoreImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017, OpenCensus Authors
+ * Copyright 2018, OpenCensus Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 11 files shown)


impl_core/src/main/java/io/opencensus/implcore/trace/export/ExportComponentImpl.java
```diff
@@ -39,17 +39,11 @@ public SpanExporterImpl getSpanExporter() {
   }
 
   @Override
-  // TODO(#914): This method shouldn't be nullable.
-  @SuppressWarnings("nullness")
```
songy23: You also need remove the `@Nullable` annotation on line 33.

impl_core/src/main/java/io/opencensus/implcore/trace/export/ExportComponentImpl.java
```diff
@@ -39,17 +39,11 @@ public SpanExporterImpl getSpanExporter() {
...
-  @Nullable
   public RunningSpanStoreImpl getRunningSpanStore() {
     return runningSpanStore;
   }
 
   @Override
-  // TODO(#914): This method shouldn't be nullable.
-  @SuppressWarnings("nullness")
```
songy23: ditto, line 34

impl_core/src/main/java/io/opencensus/implcore/trace/export/ExportComponentImpl.java
HailongWen: Thanks! Done.

impl_core/src/main/java/io/opencensus/implcore/trace/export/ExportComponentImpl.java
HailongWen: Done.

impl_core/src/test/java/io/opencensus/implcore/trace/export/NoopSampledSpanStoreImplTest.java
```diff
@@ -0,0 +1,105 @@
...
+  private final SpanContext spanContext =
+      SpanContext.create(
+          TraceId.generateRandomId(random), SpanId.generateRandomId(random), TraceOptions.DEFAULT);
+  private final TestClock testClock = TestClock.create(timestamp);
+  private final TimestampConverter timestampConverter = TimestampConverter.now(testClock);
+  private final EnumSet<Options> recordSpanOptions = EnumSet.of(Options.RECORD_EVENTS);
+  @Mock private StartEndHandler startEndHandler;
+  private SpanImpl spanImpl;
+  private ErrorFilter errorFilter = ErrorFilter.create(SPAN_NAME, null, 0);
+  private LatencyFilter latencyFilter = LatencyFilter.create(SPAN_NAME, 0, 0, 0);
```
sebright: I think it would be better to use filters that match the test spans.  maxSpansToReturn should be non-zero, and the latency range should include the test span latency.  The same applies to the filter in NoopRunningSpanStoreImplTest.

sebright: The two filter fields could be final.

impl_core/src/test/java/io/opencensus/implcore/trace/export/NoopSampledSpanStoreImplTest.java
```diff
@@ -0,0 +1,105 @@
...
+    assertThat(sampledSpanStoreImpl.getRegisteredSpanNamesForCollection()).isEmpty();
+    assertThat(sampledSpanStoreImpl.getErrorSampledSpans(errorFilter)).isEmpty();
+    assertThat(sampledSpanStoreImpl.getLatencySampledSpans(latencyFilter)).isEmpty();
+  }
+
+  @Test
+  public void noopImplementation() {
+    getMethodsReturnsEmpty();
+    // considerForSampling() does not affect the result.
+    sampledSpanStoreImpl.considerForSampling(spanImpl);
```
sebright: Shouldn't spanImpl be ended before considerForSampling is called?

impl_core/src/test/java/io/opencensus/implcore/trace/export/NoopSampledSpanStoreImplTest.java
```diff
@@ -0,0 +1,105 @@
...
+    getMethodsReturnsEmpty();
+    // registerSpanNamesForCollection() does not affect the result.
+    sampledSpanStoreImpl.registerSpanNamesForCollection(NAMES_FOR_COLLECTION);
```
sebright: I'm not sure I understand the order of method calls in this test.  I thought that registerSpanNamesForCollection should be called before considerForSampling, since that is what would be required to sample the span with the InProcess implementation.

impl_core/src/test/java/io/opencensus/implcore/trace/export/NoopSampledSpanStoreImplTest.java
HailongWen: I recall that when `maxSpansToReturn` is zero, the filter acts like [not capping the size of the result] (https://github.com/census-instrumentation/opencensus-java/blob/67b1f37cd91688cf42fe2ff1adc8a42ef32477d9/impl_core/src/main/java/io/opencensus/implcore/trace/export/RunningSpanStoreImpl.java#L77). Though this is not specified in the API. So yeah I guess yeah it is better to use a non-zero max value.

impl_core/src/test/java/io/opencensus/implcore/trace/export/NoopSampledSpanStoreImplTest.java
HailongWen: Good catch.

impl_core/src/test/java/io/opencensus/implcore/trace/export/NoopSampledSpanStoreImplTest.java
sebright: Sorry, you're right, 0 does mean unlimited.

impl_core/src/test/java/io/opencensus/implcore/trace/export/NoopSampledSpanStoreImplTest.java
HailongWen: Yes, you are right. Previously I was just thinking about testing test them one by one but didn't take the order into consideration.

impl_core/src/main/java/io/opencensus/implcore/trace/export/InProcessRunningSpanStoreImpl.java
```diff
@@ -0,0 +1,81 @@
+/*
+ * Copyright 2017, OpenCensus Authors
```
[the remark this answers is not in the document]
HailongWen: This file is actually renamed (not added), should we change the header as well?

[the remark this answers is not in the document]
songy23: I think so, it should be treated as a new file.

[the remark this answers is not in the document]
HailongWen: Got it, will do.

PR #965, opened 2018-01-29
Implement Prometheus Stats exporter

songy23:
Closes https://github.com/census-instrumentation/opencensus-java/issues/851.

Implement Prometheus Stats Exporter using a pull model, which is naturally supported by Prometheus client. Start with the minimal set of functionality - we have a collector that collects stats from OpenCensus library and converts them to Prometheus `Samples`, and users need to take care of registering this collector against the Prometheus backend. Once the collector is registered, `Samples` will be pulled every time when users query from Prometheus. This is consistent with the [OpenCensus-Go Prometheus exporter](https://github.com/census-instrumentation/opencensus-go/blob/master/exporter/prometheus/prometheus.go).

Here is an example that collects and exports stats from the [BigTable demo app](https://github.com/bogdandrutu/java-demo-apps). `Samples` are exported via a simple [Prometheus HTTPServer](https://github.com/prometheus/client_java/blob/master/simpleclient_httpserver/src/main/java/io/prometheus/client/exporter/HTTPServer.java).
prometheus-demo-app

### DIFF
--- a/exporters/stats/prometheus/README.md
+++ b/exporters/stats/prometheus/README.md
@@ -1,0 +1,58 @@
+# OpenCensus Prometheus Stats Exporter
+
+The *OpenCensus Prometheus Stats Exporter* is a stats exporter that exports data to 
+Prometheus. [Prometheus](https://prometheus.io/) is an open-source systems monitoring and alerting 
+toolkit originally built at [SoundCloud](https://soundcloud.com/).
+
+## Quickstart
+
+### Prerequisites
+
+To use this exporter, you need to install, configure and start Prometheus first. Follow the 
+instructions [here](https://prometheus.io/docs/introduction/first_steps/).
+
+### Hello "Prometheus Stats"
+
+#### Add the dependencies to your project
+
+TODO
+
+For Maven add to your `pom.xml`:
+
+For Gradle add to your dependencies:
+
+#### Register the exporter
+ 
+```java
+public class MyMainClass {
+  public static void main(String[] args) {
+    // Creates a PrometheusStatsCollector and registers it to the default Prometheus registry.
+    PrometheusStatsCollector.createAndRegister();
+    
+    // Uses a simple Prometheus HTTPServer to export metrics. 
+    // You can use a Prometheus PushGateway instead, though that's discouraged by Prometheus:
+    // https://prometheus.io/docs/practices/pushing/#should-i-be-using-the-pushgateway.
+    io.prometheus.client.exporter.HTTPServer server = 
+      new HTTPServer(/*host*/ "localhost", /*port*/  9091, /*daemon*/ true);
+    
+    // Your code here.
+    // ...
+  }
+}
+```
+
+In this example, you should be able to see all the OpenCensus Prometheus metrics by visiting 
+localhost:9091/metrics. Every time when you visit localhost:9091/metrics, the metrics will be 
+collected from OpenCensus library and refreshed.
+
+#### Exporting
+
+After collecting stats from OpenCensus, there are multiple options for exporting them. 
+See [Exporting via HTTP](https://github.com/prometheus/client_java#http), [Exporting to a Pushgateway](https://github.com/prometheus/client_java#exporting-to-a-pushgateway)
+and [Bridges](https://github.com/prometheus/client_java#bridges).
+
+#### Java Versions
+
+Java 7 or above is required for using this exporter.
+
+## FAQ
--- a/exporters/stats/prometheus/src/main/java/io/opencensus/exporter/stats/prometheus/PrometheusExportUtils.java
+++ b/exporters/stats/prometheus/src/main/java/io/opencensus/exporter/stats/prometheus/PrometheusExportUtils.java
@@ -22,6 +22,10 @@ import com.google.common.collect.Lists;
 import io.opencensus.common.Function;
 import io.opencensus.common.Functions;
 import io.opencensus.stats.Aggregation;
+import io.opencensus.stats.Aggregation.Count;
+import io.opencensus.stats.Aggregation.Distribution;
+import io.opencensus.stats.Aggregation.Mean;
+import io.opencensus.stats.Aggregation.Sum;
 import io.opencensus.stats.AggregationData;
 import io.opencensus.stats.AggregationData.CountData;
 import io.opencensus.stats.AggregationData.DistributionData;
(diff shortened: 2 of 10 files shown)


exporters/stats/prometheus/README.md
```diff
@@ -0,0 +1,58 @@
...
+### Prerequisites
+
+To use this exporter, you need to install, configure and start Prometheus first. Follow the 
+instructions [here](https://prometheus.io/docs/introduction/first_steps/).
+
+### Hello "Prometheus Stats"
+
+#### Add the dependencies to your project
+
+TODO
```
HailongWen: A kindly reminder: The readme files list in [RELEASING.md](https://github.com/census-instrumentation/opencensus-java/blob/master/RELEASING.md#update-release-versions-in-documentations-and-build-files) should be updated once this is released. :)

songy23: Thanks, I've created https://github.com/census-instrumentation/opencensus-java/issues/966 to keep track of this.

exporters/stats/prometheus/src/main/java/io/opencensus/exporter/stats/prometheus/PrometheusExportUtils.java
```diff
@@ -0,0 +1,169 @@
...
+import io.prometheus.client.Collector;
+import io.prometheus.client.Collector.MetricFamilySamples;
+import io.prometheus.client.Collector.MetricFamilySamples.Sample;
+import io.prometheus.client.Collector.Type;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map.Entry;
+
+/*>>>
```
bogdandrutu: What is this??

exporters/stats/prometheus/src/main/java/io/opencensus/exporter/stats/prometheus/PrometheusExportUtils.java
```diff
@@ -0,0 +1,169 @@
...
+          @Override
+          public Void apply(DistributionData arg) {
+            for (long bucketCount : arg.getBucketCounts()) {
+              samples.add(
+                  new MetricFamilySamples.Sample(
+                      name + "_bucket", labelNames, labelValues, bucketCount));
+            }
+            samples.add(
+                new MetricFamilySamples.Sample(
+                    name + "_count", labelNames, labelValues, arg.getCount()));
```
bogdandrutu: Make these special strings static constant.

exporters/stats/prometheus/src/main/java/io/opencensus/exporter/stats/prometheus/PrometheusExportUtils.java
songy23: This is used by our new nullness checker framework: https://github.com/census-instrumentation/opencensus-java/blob/master/build.gradle#L182

exporters/stats/prometheus/src/main/java/io/opencensus/exporter/stats/prometheus/PrometheusExportUtils.java
songy23: Updated.

exporters/stats/prometheus/src/main/java/io/opencensus/exporter/stats/prometheus/PrometheusExportUtils.java
sebright: It is also documented here: https://github.com/census-instrumentation/opencensus-java/blob/master/CONTRIBUTING.md#checker-framework-null-analysis

exporters/stats/prometheus/src/main/java/io/opencensus/exporter/stats/prometheus/PrometheusExportUtils.java
```diff
@@ -0,0 +1,178 @@
...
+        name, type, OPENCENSUS_HELP_MSG + view.getDescription(), Collections.<Sample>emptyList());
+  }
+
+  @VisibleForTesting
+  static Type getType(Aggregation aggregation, AggregationWindow window) {
+    if (!(window instanceof Cumulative)) {
+      return Type.UNTYPED;
+    }
+    return aggregation.match(
+        Functions.returnConstant(Type.UNTYPED), // SUM
```
sebright: How will untyped metrics be used?

exporters/stats/prometheus/src/main/java/io/opencensus/exporter/stats/prometheus/PrometheusStatsCollector.java
```diff
@@ -0,0 +1,98 @@
...
+  /**
+   * Creates a {@link PrometheusStatsCollector} and registers it to Prometheus {@link
+   * CollectorRegistry#defaultRegistry}.
+   *
+   * @throws IllegalStateException if a {@code PrometheusStatsCollector} has already been created
+   *     and registered.
+   */
+  public static void createAndRegister() {
+    Preconditions.checkState(
+        collector == null, "PrometheusStatsCollector has already been created and registered.");
```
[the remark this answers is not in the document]
songy23: That's right - if users try to register a second `PrometheusStatsCollector` to the same `CollectorRegistry`, they'll get an `IllegalArgumentException`: https://github.com/prometheus/client_java/blob/master/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java#L54. Do you think it's better to return an instance of `PrometheusStatsCollector`, and let users take care of registering it?

exporters/stats/prometheus/src/main/java/io/opencensus/exporter/stats/prometheus/PrometheusStatsCollector.java
```diff
@@ -0,0 +1,98 @@
...
+    collector = new PrometheusStatsCollector(Stats.getViewManager());
+    collector.register();
```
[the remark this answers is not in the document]
songy23: Nice catch - will update if we decided to keep `createAndRegister()` as it is.

exporters/stats/prometheus/src/main/java/io/opencensus/exporter/stats/prometheus/PrometheusStatsCollector.java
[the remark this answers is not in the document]
sebright: I meant that this library could still take care of registering the collector, but this method could be simplified to something like:

```
  public static void createAndRegister() {
    new PrometheusStatsCollector(Stats.getViewManager()).register();
  }
```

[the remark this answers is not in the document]
songy23: Sounds good, I've removed the static reference and simplified createAndRegister().

bogdandrutu:
I think this can be merged.

songy23:
@dinooliva would like to take a look today - will merge this PR by EOD today.

exporters/stats/prometheus/src/main/java/io/opencensus/exporter/stats/prometheus/PrometheusExportUtils.java
```diff
@@ -0,0 +1,178 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map.Entry;
+
+/*>>>
+import org.checkerframework.checker.nullness.qual.Nullable;
+*/
+
+/** Util methods to convert OpenCensus Stats data models to Prometheus data models. */
```
dinooliva: It would be good if you could add a comment (here or in PrometheusStatsCollector.java) describing the mapping of the OpenCensus stats data model to Prometheus - particularly an y restrictions.

songy23: Sure, I added some javadoc that described these conversions between data models.

exporters/stats/prometheus/src/main/java/io/opencensus/exporter/stats/prometheus/PrometheusExportUtils.java
```diff
@@ -0,0 +1,208 @@
...
+ * <p>{@link SumDataDouble}, {@link SumDataLong} and {@link CountData} will be converted to a single
+ * {@link Sample}. {@link MeanData} will be converted to two {@link Sample}s sum and count. {@link
+ * DistributionData} will be converted to a list of {@link Sample}s that have the sum, count and
+ * histogram buckets.
+ *
+ * <p>{@link TagKey} and {@link TagValue} will be converted to Prometheus {@code LabelName} and
+ * {@code LabelValue}. {@code Null} {@link TagValue} will be converted to an empty string.
+ *
+ * <p>Please note that Prometheus Metric and Label name can only have alphanumeric characters and
+ * underscore. All other characters will be sanitized by underscores.
```
dinooliva: This seems reasonable and consistent with what we've done in the past but it has also lead to some confusion in the past too.
1. Is this what the Census Go exporter does?
2. Does anyone have any concerns about this?

bogdandrutu: One more question:
How do we handle the case of duplicated entries after sanitization:
"grpc/io/test" -> "grpc_io_test"
"grpc.io/test" -> "grpc_io_test"

In this case I think we need to do a merge of the values.

songy23: Yes, this is consistent with the current Go implementation: https://github.com/census-instrumentation/opencensus-go/blob/master/exporter/prometheus/prometheus.go#L276.
@jcd2 had some concerns about the subtle corner cases in https://github.com/census-instrumentation/opencensus-go/issues/173, including the duplicated names after sanitization. 

In Java we're using the [standard sanitize method provided by Prometheus](https://github.com/prometheus/client_java/blob/master/simpleclient/src/main/java/io/prometheus/client/Collector.java#L181) and having the Prometheus client deal with the corner cases, just like any other direct users of Prometheus. For example, for the case of duplicated entries after sanitization, [Prometheus client will simply reject the second entry](https://github.com/prometheus/client_java/blob/master/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java#L54).

I'm not sure if we should wrap around such cases in our exporter though (like merging the values). To me it's like a hack around Prometheus restrictions, and might lead to confusions to the users. It might be better to just follow the Prometheus client library convention on these cases.

songy23: Discussed with @dinooliva and @ramonza , we decided to keep consistent with Go for the moment. I've created https://github.com/census-instrumentation/opencensus-specs/issues/50 to follow up on this.

songy23:
Thanks!

Issue #864, opened 2017-12-01
Consider adding @since tag to all public Javadocs.


sebright:
#794 added `ViewManager.getAllExportedViews()` with a `@since` tag.  We should probably add the tag to all other public methods, for consistency.  See https://github.com/census-instrumentation/opencensus-java/pull/794#discussion_r149854131.

sebright:
This issue is for updating all of the existing APIs.  We can also add a note to CONTRIBUTING.md for new code.

sebright:
This has been implemented for most APIs in #968, #975, #979, and #983.  The only APIs that are missing `@since` tags now are measure and view constants (issue #984) and the agent package (#978).  #985 adds a note to CONTRIBUTING.md.

PR #968, opened 2018-01-31
Add '@since' Javadoc tag to all stats and tagging APIs (issue #864).

sebright:
The only APIs that were missing `@since` tags were from the first stats/tagging API release.


sebright:
Yes, I think we should eventually add it to all API elements in all artifacts.

sebright:
Thanks!

PR #970, opened 2018-02-01
Checker Framework: 2.3.0 -> 2.3.2

sebright:
The new version fixes
https://github.com/typetools/checker-framework/issues/1712, so this commit also
removes the workaround.


sebright:
Thanks!

PR #971, opened 2018-02-02
Adds Eclipse plugin to build.gradle

CodingFabian:
The idea plugin is there already, this gives eclipse users a much easier start.


CodingFabian:
This is also a test-pr to see if my CLA signing worked out :)

PR #972, opened 2018-02-03
First proof of concept for Instana integration

CodingFabian:
Instana is happy to support OpenCensus with a proof of concept for its integration.

The target is to get this PR merged, but it can also used as basis for discussion for a while.

The exporter sends to an API that is intended for customers to report spans to, but is by far not as performant as several other internal APIs. This API also only supports a subset of the fields that Instana supports or would be desired for a OpenCensus integration.
Any Instana agent exposes this API via http. The agent then will handle offloading and authorization to the Instana backend(s).

The most significant limitation of Instanas data model compared to the OpenCencus one is that traceIds are only 64 bit wide. Instana is working on supporting wider ids in the near future, but currently this extension will just drop the lower 64bit off the trace id.

There are no 3rd party libraries used for generating the json and for posting it to the Instana agent.
Are there any recommendations for using libraries, maybe ones that might be readily available?

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 [![Coverage Status][codecov-image]][codecov-url]
 
 
-OpenCensus is a toolkit for collecting application performance and behavior data. It currently 
+OpenCensus is a toolkit for collecting application performance and behavior data. It currently
 includes 3 apis: stats, tracing and tags.
 
 The library is in alpha stage and the API is subject to change.
@@ -53,13 +53,13 @@ public final class MyClassWithTracing {
       doFinalWork();
     }
   }
-  
+
   private static void doInitialWork() {
     // ...
     tracer.getCurrentSpan().addAnnotation("Important.");
     // ...
   }
-  
+
   private static void doFinalWork() {
     // ...
     tracer.getCurrentSpan().addAnnotation("More important.");
@@ -74,7 +74,7 @@ TODO
 
 ## Quickstart for Applications
 
-Besides recording tracing/stats events the application also need to link the implementation, 
+Besides recording tracing/stats events the application also need to link the implementation,
 setup exporters, and debugging [Z-Pages](https://github.com/census-instrumentation/opencensus-java/tree/master/contrib/zpages).
 
 ### Add the dependencies to your project
@@ -105,6 +105,7 @@ runtime 'io.opencensus:opencensus-impl:0.11.1'
 ### How to setup exporters?
 
 #### Trace exporters
+* [Instana][TraceExporterInstana]
 * [Logging][TraceExporterLogging]
 * [Stackdriver][TraceExporterStackdriver]
 * [Zipkin][TraceExporterZipkin]
@@ -115,7 +116,7 @@ runtime 'io.opencensus:opencensus-impl:0.11.1'
 
 ### How to setup debugging Z-Pages?
 
-If the application owner wants to export in-process tracing and stats data via HTML debugging pages 
+If the application owner wants to export in-process tracing and stats data via HTML debugging pages
 see this [link](https://github.com/census-instrumentation/opencensus-java/tree/master/contrib/zpages#quickstart).
 
 [travis-image]: https://travis-ci.org/census-instrumentation/opencensus-java.svg?branch=master
@@ -130,6 +131,7 @@ see this [link](https://github.com/census-instrumentation/opencensus-java/tree/m
 [gitter-url]: https://gitter.im/census-instrumentation/lobby?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge
 [codecov-image]: https://codecov.io/gh/census-instrumentation/opencensus-java/branch/master/graph/badge.svg
 [codecov-url]: https://codecov.io/gh/census-instrumentation/opencensus-java/branch/master/
+[TraceExporterInstana]: https://github.com/census-instrumentation/opencensus-java/tree/master/exporters/trace/instana#quickstart
 [TraceExporterLogging]: https://github.com/census-instrumentation/opencensus-java/tree/master/exporters/trace/logging#quickstart
 [TraceExporterStackdriver]: https://github.com/census-instrumentation/opencensus-java/tree/master/exporters/trace/stackdriver#quickstart
 [TraceExporterZipkin]: https://github.com/census-instrumentation/opencensus-java/tree/master/exporters/trace/zipkin#quickstart
--- a/exporters/trace/instana/README.md
+++ b/exporters/trace/instana/README.md
@@ -1,0 +1,73 @@
+# OpenCensus Instana Trace Exporter
+[![Build Status][travis-image]][travis-url]
+[![Windows Build Status][appveyor-image]][appveyor-url]
+[![Maven Central][maven-image]][maven-url]
+
+The *OpenCensus Instana Trace Exporter* is a trace exporter that exports
+data to Instana. [Instana](http://www.instana.com/) Instana is a distributed
+tracing system.
+
+## Quickstart
+
+### Prerequisites
+
+[Instana](http://www.instana.com/) forwards traces exported by applications
+instrumented with Census to its backend using the Instana agent processes as proxy.
+If the agent is used on the same host as Census, please take care to deactivate
+automatic tracing.
+
+
+### Hello Stan
+
+#### Add the dependencies to your project
+
+For Maven add to your `pom.xml`:
+```xml
+<dependencies>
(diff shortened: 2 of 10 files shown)


exporters/trace/instana/src/main/java/io/opencensus/exporter/trace/instana/InstanaExporterHandler.java
```diff
@@ -0,0 +1,224 @@
...
+  private static String encodeSpanId(SpanId spanId) {
+    return BaseEncoding.base16().lowerCase().encode(spanId.getBytes());
+  }
+
+  private static String toSpanType(SpanData spanData) {
+    if (spanData.getParentSpanId() == null || Boolean.TRUE.equals(spanData.getHasRemoteParent())) {
+      return "ENTRY";
+    }
+
+    // This is a hack because the v2 API does not have SpanKind. When switch to v2 this will be
```
CodingFabian: I copied this from Zipkin. Do I understand it right that there will be a census api revision that offers SpanKind?
is the "Sent." prefix a convention or guaranteed by anything?

exporters/trace/instana/src/main/java/io/opencensus/exporter/trace/instana/InstanaExporterHandler.java
```diff
@@ -0,0 +1,224 @@
...
+    return attributeValue.match(
+        RETURN_STRING, RETURN_STRING, RETURN_STRING, Functions.<String>returnNull());
+  }
+
+  static String convertToJson(Collection<SpanData> spanDataList) {
+    StringBuilder sb = new StringBuilder();
+    sb.append('[');
+    for (SpanData span : spanDataList) {
+      if (span.getStatus() == null) {
+        // TODO: Instana expects completed spans. Will those come in again?
```
CodingFabian: I could not find any explanation of why a incomplete span is ever sent to exporter, and if it is sent again when complete

CodingFabian:
@mtwo as promised :)

README.md
```diff
@@ -105,6 +105,7 @@ runtime 'io.opencensus:opencensus-impl:0.11.1'
 ### How to setup exporters?
 
 #### Trace exporters
+* [Instana][TraceExporterInstana]
```
adriancole: alphabetical works in your favor.. for now.. :D

adriancole: zipkin will always have the last word!

exporters/trace/instana/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+[Instana](http://www.instana.com/) forwards traces exported by applications
+instrumented with Census to its backend using the Instana agent processes as proxy.
+If the agent is used on the same host as Census, please take care to deactivate
+automatic tracing.
+
+
+### Hello Stan
+
+#### Add the dependencies to your project
```
songy23: You can leave this section empty for now since it's not released to maven yet.

exporters/trace/instana/src/main/java/io/opencensus/exporter/trace/instana/InstanaExporterHandler.java
```diff
@@ -0,0 +1,224 @@
+/*
+ * Copyright 2017, OpenCensus Authors
```
HailongWen: s/2017/2018, same for other license headers.

exporters/trace/instana/src/main/java/io/opencensus/exporter/trace/instana/InstanaExporterHandler.java
```diff
@@ -0,0 +1,224 @@
...
+import java.util.logging.Logger;
+
+/*
+ * Exports to an Instana agent acting as proxy to the Instana backend (and handling authentication)
+ * Uses the Trace SDK documented:
+ * https://github.com/instana/instana-java-sdk#instana-trace-webservice
+ *
+ * Currently does a blocking export using HttpUrlConnection.
+ * Also uses a StringBuilder to build JSON.
+ * Both can be improved should 3rd party library usage not be a concern.
```
HailongWen: I think that won't be a concern. For http, we can use google-http-java-client since it is maintained by google and we've already instrumented our code in it. For json maybe we can use gson?

Any comments @bogdandrutu ?

exporters/trace/instana/src/main/java/io/opencensus/exporter/trace/instana/InstanaExporterHandler.java
HailongWen: Null span status means running span and that should not happen in exportation, since only ended spans will be added to the queue of span exporter. 

However, a check here would be good. We have [similar logic](https://github.com/census-instrumentation/opencensus-java/blob/3c74cea59b7c50c65e50401305e7f7feab614a90/impl_core/src/main/java/io/opencensus/implcore/trace/export/SampledSpanStoreImpl.java#L180) to ensure the library does not crash

exporters/trace/instana/src/main/java/io/opencensus/exporter/trace/instana/InstanaExporterHandler.java
```diff
@@ -0,0 +1,224 @@
...
+ *
+ * Major TODO is the limitation of Instana to only suport 64bit trace ids, which will be resolved.
+ * Until then it is crossing fingers and treating it as 50% sampler :).
+ */
+final class InstanaExporterHandler extends SpanExporter.Handler {
+
+  private static final Tracer tracer = Tracing.getTracer();
+  private static final Sampler probabilitySpampler = Samplers.probabilitySampler(0.0001);
```
HailongWen: Should we make the sample rate configurable?

CodingFabian: I do not mind doing this. Its a copy of the Zipkin Exporter, which if you want to have it changed, shoudl also be changed.

exporters/trace/instana/README.md
CodingFabian: I do not follow this conclusion. I assume that after its merged, it will be eventually released and then users will want to use it via maven? why should I leave this out and add this later?

exporters/trace/instana/src/main/java/io/opencensus/exporter/trace/instana/InstanaExporterHandler.java
CodingFabian: ok, thanks, will leave it in as a safety anyways.

exporters/trace/instana/README.md
songy23: Because v0.11.1 is our current version (already released), and does not have artifact of this exporter. Users might be confused when they read README, since opencensus-exporter-trace-instana is not included in io.opencensus v0.11.1.

You could use v0.12.0 instead, or you could leave a TODO here and we'll update this section once we've made a new release.

CodingFabian: Alright, then ill use 0.12 - do you have a procedure in place to update versions in readmes on release?

songy23: Yes, please add this README to RELEASING.md: https://github.com/census-instrumentation/opencensus-java/blob/master/RELEASING.md#update-release-versions-in-documentations-and-build-files. We'll update versions for all README file when we made new releases.

CodingFabian: Cool will do. The way the sed works would give preference to using 0.1.1.1 right now because the sed would catch it. I guess thats a first time integration issue that one has to take care for the first release.

exporters/trace/instana/src/main/java/io/opencensus/exporter/trace/instana/InstanaExporterHandler.java
HailongWen: OK, that is fine. We can leave it as it be.

exporters/trace/instana/src/main/java/io/opencensus/exporter/trace/instana/InstanaExporterHandler.java
HailongWen: AFAIK "Sent" is used conventional for client side spans and "Recv" is used for server side. At least this is the case in gRPC.

exporters/trace/instana/src/main/java/io/opencensus/exporter/trace/instana/InstanaExporterHandler.java
bogdandrutu: I think it is reasonable to use something like gson here.

PR #978, opened 2018-02-07
Add '@since' Javadoc tag to all agent APIs (issue #864).

sebright:
This commit also adds a missing Javadoc.

@ubschmidt2 Do you think that the agent classes should have `@since` tags?  I wasn't sure, but I added them while I was updating the other contrib classes.


sebright:
I'll merge this for now, but we can remove the tags later if they aren't necessary.

Issue #840, opened 2017-11-28
Batch export too many Timeseries might throw an Http2Exception


songy23:
When running the [E2E application](https://github.com/bogdandrutu/java-demo-apps) from @bogdandrutu, we've found that if we registered a lot of views and batch exported their stats to Stackdriver simultaneously, Stackdriver client will throw an exception:
`io.netty.handler.codec.http2.Http2Exception: Header size exceeded max allowed size (10240)`

This should be because the exporter tried to batch export too many `TimeSeries` in one RPC call. After some experiments I found batch export 3 `TimeSeries` in one call is the maximum.

PR #988, opened 2018-02-09
Increase batch export size of StackdriverStatsExporter to 200

songy23:
Previously we found an issue that if we batch too many `TimeSeries` in one call, Stackdriver metric service client might throw a `Http2Exception: Header size exceeded max allowed size (10240)` (https://github.com/census-instrumentation/opencensus-java/issues/840). I posted https://github.com/census-instrumentation/opencensus-java/pull/836 to fix this, which limit the batch size to 3. After running some E2E tests recently, I think this issue in Stackdriver client has been fixed, and it's safe to remove the batch limit.

Update the batch limit to 200, which is the maximum size allowed by Stackdriver client v3 (https://github.com/census-instrumentation/opencensus-go/issues/170).

/cc @jcd2

### DIFF
--- a/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporterWorker.java
+++ b/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporterWorker.java
@@ -60,7 +60,8 @@ final class StackdriverExporterWorker implements Runnable {
 
   private static final Logger logger = Logger.getLogger(StackdriverExporterWorker.class.getName());
 
-  @VisibleForTesting static final int MAX_BATCH_EXPORT_SIZE = 3;
+  // Stackdriver Monitoring v3 only accepts up to 200 TimeSeries per CreateTimeSeries call.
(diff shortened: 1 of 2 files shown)


exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporterWorker.java
```diff
@@ -60,7 +60,8 @@
 
   private static final Logger logger = Logger.getLogger(StackdriverExporterWorker.class.getName());
 
-  @VisibleForTesting static final int MAX_BATCH_EXPORT_SIZE = 3;
+  // Stackdriver Monitoring v3 only accepts up to 200 TimeSeries per CreateTimeSeries call.
+  @VisibleForTesting static final int MAX_BATCH_EXPORT_SIZE = 200;
```
sebright: Is there a StackDriver constant that we could use?

songy23: Didn't find one (not even in Stackdriver's documentation). I think they handle it in their backend.

songy23:
Thanks!

Issue #955, opened 2018-01-23
Update trace exporter examples to use new classes


HailongWen:
Since the exporters classes were renamed (not released), we should update the examples to use the new one after 0.12.0 is released.

songy23:
Should be fixed in https://github.com/census-instrumentation/opencensus-java/pull/1008.

PR #990, opened 2018-02-09
Add a configuration class to help to create StackdriverTraceExporter. (closes #949)

### LINKED ISSUE #949: Consider using a Configuration class to create StackdriverTraceExporter
Similar to [StackdriverStatsConfiguration](https://github.com/census-instrumentation/opencensus-java/blob/master/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsConfiguration.java) for StackdriverStatsExporter.

HailongWen:
Similar to [StackdriverStatsConfiguration](https://github.com/census-instrumentation/opencensus-java/blob/master/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsConfiguration.java).

Since we've just deprecated `StackdriverExporter` and rename it to `StackdriverTraceExporter` and it is not yet released, I simply removed those factory methods in the new class instead of deprecating them again.

### DIFF
--- a/exporters/trace/stackdriver/src/main/java/io/opencensus/exporter/trace/stackdriver/StackdriverTraceExporter.java
+++ b/exporters/trace/stackdriver/src/main/java/io/opencensus/exporter/trace/stackdriver/StackdriverTraceExporter.java
@@ -22,7 +22,6 @@ import com.google.auth.Credentials;
 import com.google.auth.oauth2.GoogleCredentials;
 import com.google.cloud.ServiceOptions;
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.MoreObjects;
 import io.opencensus.trace.Tracing;
 import io.opencensus.trace.export.SpanExporter;
 import io.opencensus.trace.export.SpanExporter.Handler;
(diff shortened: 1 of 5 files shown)


HailongWen:
Thanks, I will reuse #955.

exporters/trace/stackdriver/src/main/java/io/opencensus/exporter/trace/stackdriver/StackdriverTraceExporter.java
```diff
@@ -52,71 +56,32 @@
...
+      throws IOException {
     synchronized (monitor) {
       checkState(handler == null, "Stackdriver exporter is already registered.");
-      registerInternal(StackdriverV2ExporterHandler.create(ServiceOptions.getDefaultProjectId()));
+      Credentials credentials =
+          MoreObjects.<Credentials>firstNonNull(
+              configuration.getCredentials(), GoogleCredentials.getApplicationDefault());
+      String projectId =
+          MoreObjects.<String>firstNonNull(
+              configuration.getProjectId(), ServiceOptions.getDefaultProjectId());
```
sebright: It might be better to only call GoogleCredentials.getApplicationDefault() or ServiceOptions.getDefaultProjectId() when they are needed.  For example, the Javadoc for getApplicationDefault() says that it can throw an IOException in some environments.

HailongWen: Done. PTAL.

Issue #856, opened 2017-11-30
Add stats zpages


bogdandrutu:
We should add 2 zpages for stats:
* rpcz - where we display interval stats for all rpcs.
* statsz - where we display view data for all registered views (cumulative, interval, etc).

bogdandrutu:
What is the status of this?

songy23:
RpcZ is almost done. For now the Java RpcZ looks exactly the same as Go RpcZ (though some stats are not supported in Java):

Java RpcZ:
![rpcz](https://user-images.githubusercontent.com/10536136/36054671-8eac56d4-0dac-11e8-85c5-7c22b46f0979.png)

Go RpcZ:
![rpcz-go](https://user-images.githubusercontent.com/10536136/36054692-a6e6420a-0dac-11e8-8cf2-f107ab38002c.png)

bogdandrutu:
I would be happy if you can sort by method names.

songy23:
Sure, will do.

bogdandrutu:
Don't really know what your app looks like, but you may have a conversion problem in the sizes because sending 1MB to create a table seems a lot.

songy23:
You're right - fixed the conversion. The input/output stats become too small to be displayed though, we might consider changing input/output to kb/s or even bytes/s. 

Created https://github.com/census-instrumentation/opencensus-go/issues/437 to follow up.

bogdandrutu:
This is how I would see the "statsz" page:
1) Display all register views using something similar with the directory path in file systems:
**grpc.io/client/started_count/cumulative
grpc.io/server/started_count/cumulative
opencensus.io/http/client/started_count/cumulative**
For these set of views will initially display the root directories:
**grpc.io
opencensus.io**
When click on grpc.io we display:
**grpc.io/client
grpc.io/server**

2) Display all measures sorted by name and we include things like: name, description, units, type (optional). Optional: the number of measurements recorded for each measure.

songy23:
I think we need another table to display the stats for registered views. My thought is when we click on a view name (e.g. **grpc.io/client/roundtrip_latency/cumulative**), the stats on that view should be displayed in a table at the bottom of the page, e.g. tag values, sum/count/histogram.

PR #992, opened 2018-02-12
Add RpcZ Zpage for stats.

songy23:
Part of https://github.com/census-instrumentation/opencensus-java/issues/856.

Provides identical UI as [Go RpcZ page](https://github.com/census-instrumentation/opencensus-go/blob/master/zpages/rpcz.go). ~~Please note some stats are not supported in Java (cumulative RPC counts and max latency per minute/hour).~~

EDIT:
xwqwj404tnl

### DIFF
--- a/contrib/zpages/src/main/java/io/opencensus/contrib/zpages/RpczZPageHandler.java
+++ b/contrib/zpages/src/main/java/io/opencensus/contrib/zpages/RpczZPageHandler.java
@@ -1,0 +1,454 @@
+/*
+ * Copyright 2018, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.contrib.zpages;
+
+import static io.opencensus.contrib.grpc.metrics.RpcViewConstants.RPC_CLIENT_ERROR_COUNT_HOUR_VIEW;
+import static io.opencensus.contrib.grpc.metrics.RpcViewConstants.RPC_CLIENT_ERROR_COUNT_MINUTE_VIEW;
+import static io.opencensus.contrib.grpc.metrics.RpcViewConstants.RPC_CLIENT_ERROR_COUNT_VIEW;
+import static io.opencensus.contrib.grpc.metrics.RpcViewConstants.RPC_CLIENT_FINISHED_COUNT_HOUR_VIEW;
+import static io.opencensus.contrib.grpc.metrics.RpcViewConstants.RPC_CLIENT_FINISHED_COUNT_MINUTE_VIEW;
+import static io.opencensus.contrib.grpc.metrics.RpcViewConstants.RPC_CLIENT_REQUEST_BYTES_HOUR_VIEW;
+import static io.opencensus.contrib.grpc.metrics.RpcViewConstants.RPC_CLIENT_REQUEST_BYTES_MINUTE_VIEW;
+import static io.opencensus.contrib.grpc.metrics.RpcViewConstants.RPC_CLIENT_REQUEST_BYTES_VIEW;
+import static io.opencensus.contrib.grpc.metrics.RpcViewConstants.RPC_CLIENT_REQUEST_COUNT_HOUR_VIEW;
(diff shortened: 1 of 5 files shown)


bogdandrutu:
How can I play with the page?

contrib/zpages/src/main/java/io/opencensus/contrib/zpages/RpczZPageHandler.java
```diff
@@ -0,0 +1,454 @@
...
+  }
+
+  // Calculates the duration of the given CumulativeData in seconds.
+  private static double getDurationInSecs(CumulativeData cumulativeData) {
+    return toSecs(cumulativeData.getEnd().subtractTimestamp(cumulativeData.getStart()));
+  }
+
+  // Converts a Duration to seconds.
+  private static double toSecs(Duration duration) {
+    return duration.getNanos() / NANOS_PER_SECOND + duration.getSeconds();
```
bogdandrutu: Based on the definition of the getNanos this will always be equivalent to:

return duration.getSeconds();

Am I missing something?

contrib/zpages/src/main/java/io/opencensus/contrib/zpages/RpczZPageHandler.java
```diff
@@ -0,0 +1,454 @@
...
+    // We don't have a cumulative view for started/finished RPC counts in Java.
+    formatter.format("<td align=\"right\">%s</td><td></td>", NA);
+    formatter.format("<td align=\"right\">%.3f</td>", snapshot.avgLatencyMinute);
+    formatter.format("<td align=\"right\">%.3f</td>", snapshot.avgLatencyHour);
+    formatter.format("<td align=\"right\">%.3f</td><td></td>", snapshot.avgLatencyTotal);
+    // We don't record max/min values for minute/hour views in Java.
+    formatter.format("<td align=\"right\">%s</td>", NA);
+    formatter.format("<td align=\"right\">%s</td>", NA);
+    formatter.format("<td align=\"right\">%.3f</td><td></td>", snapshot.maxLatencyTotal);
+    formatter.format("<td align=\"right\">%.3f</td>", snapshot.countMinute / SECONDS_PER_MINUTE);
```
bogdandrutu: countMinute -> countLastMinute? countPerMinute? or do the same and directly calculate the rate similar to inputRateMinute

songy23: Should be countLastMinute, updated. Same for all the other minute/hour stats.

contrib/zpages/src/main/java/io/opencensus/contrib/zpages/RpczZPageHandler.java
songy23: I thought nanoseconds would be kept separately?

IIUC for example, consider 
`Duration duration = Duration.create(/* seconds */ 5, /* nanos */ 5 * 1e8);`. 
If we want to convert `duration` to seconds, it should be 5.5 seconds, but `duration.getSeconds()` will only return 5 seconds. Is that right?

bogdandrutu: Please explain that in the comment. Maybe also rename toDoubleSeconds.

contrib/zpages/src/main/java/io/opencensus/contrib/zpages/RpczZPageHandler.java
```diff
@@ -0,0 +1,472 @@
...
+    out.write("<tr>");
+    formatter.format("<td><b>%s</b></td>", method);
+    out.write("<td></td>");
+    formatter.format("<td align=\"right\">%d</td>", snapshot.countLastMinute);
+    formatter.format("<td align=\"right\">%d</td>", snapshot.countLastHour);
+    formatter.format("<td align=\"right\">%d</td><td></td>", snapshot.countTotal);
+    formatter.format("<td align=\"right\">%.3f</td>", snapshot.avgLatencyLastMinute);
+    formatter.format("<td align=\"right\">%.3f</td>", snapshot.avgLatencyLastHour);
+    formatter.format("<td align=\"right\">%.3f</td><td></td>", snapshot.avgLatencyTotal);
+    // We don't record max/min values for minute/hour views in Java.
```
bogdandrutu: Probably we should not have this column with max latency. We can add later a percentile column if needed.

contrib/zpages/src/main/java/io/opencensus/contrib/zpages/RpczZPageHandler.java
songy23: Sure, I added some comments on this.

contrib/zpages/src/main/java/io/opencensus/contrib/zpages/RpczZPageHandler.java
songy23: Saw your comments under https://github.com/census-instrumentation/opencensus-specs/issues/47, I've removed max latency for the moment and left a TODO to add percentiles support.

sebright:
LGTM, though I didn't do a very thorough review because @bogdandrutu already reviewed it.

songy23:
Thanks!

PR #993, opened 2018-02-12
Add cumulative views for started and finished RPCs.

songy23:
Since we're going to remove `Interval` window in the future, we'll need cumulative views for started and finished RPCs (currently there are only minute and hour views for them, which are to be removed).

### DIFF
--- a/contrib/grpc_metrics/src/main/java/io/opencensus/contrib/grpc/metrics/RpcViewConstants.java
+++ b/contrib/grpc_metrics/src/main/java/io/opencensus/contrib/grpc/metrics/RpcViewConstants.java
@@ -44,6 +44,7 @@ import static io.opencensus.contrib.grpc.metrics.RpcMeasureConstants.RPC_STATUS;
 import com.google.common.annotations.VisibleForTesting;
 import io.opencensus.common.Duration;
 import io.opencensus.stats.Aggregation;
+import io.opencensus.stats.Aggregation.Count;
 import io.opencensus.stats.Aggregation.Distribution;
 import io.opencensus.stats.Aggregation.Mean;
 import io.opencensus.stats.BucketBoundaries;
@@ -101,6 +102,7 @@ public final class RpcViewConstants {
 
   // Use Aggregation.Mean to record sum and count stats at the same time.
   @VisibleForTesting static final Aggregation MEAN = Mean.create();
+  @VisibleForTesting static final Aggregation COUNT = Count.create();
 
   @VisibleForTesting
   static final Aggregation AGGREGATION_WITH_BYTES_HISTOGRAM =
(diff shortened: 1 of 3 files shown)


contrib/grpc_metrics/src/main/java/io/opencensus/contrib/grpc/metrics/RpcViewConstants.java
```diff
@@ -248,6 +248,34 @@
...
+   * Cumulative {@link View} for started client RPCs.
+   *
+   * @since 0.12
+   */
+  public static final View RPC_CLIENT_STARTED_COUNT_CUMULATIVE_VIEW =
+      View.create(
+          View.Name.create("grpc.io/client/started_count/cumulative"),
+          "Number of started client RPCs",
+          RPC_CLIENT_STARTED_COUNT,
+          MEAN,
```
bogdandrutu: Why not SUM/COUNT?

songy23: Because we want to keep both sum and count for cumulative views.

bogdandrutu: Not sure I understand how Sum and Count will be different in this case?

songy23: We had a discussion with @jcd2 on this long time ago. Initially we decided to use Distribution for all views with histograms and Mean for all others, but I agree using a Count in this case will be more clear. I updated all started/finished count view to use COUNT.

songy23:
Thanks!

contrib/grpc_metrics/src/main/java/io/opencensus/contrib/grpc/metrics/RpcViewConstants.java
```diff
@@ -488,7 +546,7 @@
           View.Name.create("grpc.io/client/started_count/minute"),
           "Minute stats on the number of client RPCs started",
           RPC_CLIENT_STARTED_COUNT,
-          MEAN,
+          COUNT,
```
sebright: Are there any backwards compatibility issues with the changes to existing views?

songy23: Not in our library or gPRC, and it's unlikely to break user code unless they are doing a casting on AggreagtionData of these Views. Do you think we should deprecate the old views instead?

sebright: I was mainly concerned about gRPC, since I think that they are the only direct user of the RPC constants.  Since it doesn't affect gRPC, I think this is fine.

songy23: SGTM

PR #998, opened 2018-02-15
Add Nullable annotation to tag values.

songy23:
So that the nullness checker framework could catch dereference of TagValues.

### DIFF
--- a/api/src/main/java/io/opencensus/stats/ViewData.java
+++ b/api/src/main/java/io/opencensus/stats/ViewData.java
@@ -75,7 +75,7 @@ public abstract class ViewData {
    *
    * @since 0.8
    */
-  public abstract Map<List<TagValue>, AggregationData> getAggregationMap();
+  public abstract Map<List</*@Nullable*/ TagValue>, AggregationData> getAggregationMap();
 
   /**
    * Returns the {@link AggregationWindowData} associated with this {@link ViewData}.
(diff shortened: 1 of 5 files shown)


songy23:
Not 100% sure but I suspect it's because of [this method](https://github.com/census-instrumentation/opencensus-java/blob/master/impl_core/src/main/java/io/opencensus/implcore/stats/MutableViewData.java#L196):
`private static <T> Map<T, AggregationData> createAggregationMap(Map<T, MutableAggregation> tagValueAggregationMap, Measure measure)` and [this one](https://github.com/census-instrumentation/opencensus-java/blob/master/impl_core/src/main/java/io/opencensus/implcore/stats/MutableViewData.java#L465). 

`T` represents `List<TagValue>` here. Though `TagValue` is marked as `Nullable`, checker framework doesn't check **generic type inside generic type `T`** and assumes it's `List</*Nonnull*/ TagValue>`. So `ViewData.create()` will pass with possible null `TagValue`s. It's really a corner case.

sebright:
I'm not sure I understand.  Neither of those two methods inserts or modifies the lists of TagValues, so they couldn't have inserted null.  They only modify the maps' values.  I think we should report this if it's a Checker Framework bug.

songy23:
They didn't modify the lists, but using a generic type <T> somehow prevents checker framework from checking the nullness of list elements. If I modified the method signature like this:
`private static /*@Nullable*/ <E> Map<**List</*@Nullable*/ E>**, AggregationData> createAggregationMap(Map<**List</*@Nullable*/ E>**, MutableAggregation> tagValueAggregationMap, Measure measure)`

Then I think checker framework should be able to catch the nullable list elements.

songy23:
By the way I'm not sure if this is a bug in checker framework. To me it's more like a limitation of using Java generics.

sebright:
The Checker Framework should be able to handle generics, since it uses annotations on types within generic types.  Why is there a `/*@Nullable*/` before the `<E>` in your comment?  If an annotation is allowed in that position, I would expect it to apply to the whole return value, i.e., the whole Map.

songy23:
Sorry I got this wrong at first. I just did another experiment, it seems it's not because of these generic types, but rather the wildcard here: https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/stats/ViewData.java#L102.

With the `? extends` wildcard, checker framework won't be able to detect nullness of the list elements. If I removed it, checker framework could catch this case then.

For example:
`public ViewData create(Map<? extends List</*@NonNull*/ TagValue>, ? extends AggregationData> map);`
`ViewData.create(ImmutableMap.of(ImmutableList.of(null, null, null), aggregationData));`

Checker framework will pass without catching the violation here.

`public ViewData create(Map<List</*@NonNull*/ TagValue>, ? extends AggregationData> map);`
`ViewData.create(ImmutableMap.of(ImmutableList.of(null, null, null), aggregationData));`

Checker framework is able to catch this correctly.

sebright:
Yes, I can reproduce the issue with `? extends` in a smaller example, too.  I'll open a Checker Framework issue.

sebright:
I opened an issue: https://github.com/typetools/checker-framework/issues/1838

PR #1002, opened 2018-02-16
Add StatsZ Zpage for stats.

songy23:
Closes https://github.com/census-instrumentation/opencensus-java/issues/856.

Statsz page for displaying all exported views. (Currently only cumulative views are exported, and interval views will be removed soon.) Views are stored in a directory tree. Directory here refers to the namespace of a view. For example, view grpc.io/client/started_count/cumulative will be under directory structure root -> grpc.io -> client -> started_count.

For now the color and style of Statsz is similar to [Rpcz page](https://github.com/census-instrumentation/opencensus-java/blob/master/contrib/zpages/src/main/java/io/opencensus/contrib/zpages/RpczZPageHandler.java).

Some screenshots:
EDIT:
image

image

image

/cc @tags07

### DIFF
--- a/contrib/zpages/src/main/java/io/opencensus/contrib/zpages/StatszZPageHandler.java
+++ b/contrib/zpages/src/main/java/io/opencensus/contrib/zpages/StatszZPageHandler.java
@@ -148,7 +148,7 @@ final class StatszZPageHandler extends ZPageHandler {
     emitDirectoryTable(current, path, out, formatter);
     if (current != null && current.view != null) {
       ViewData viewData = viewManager.getView(current.view.getName());
-      emitViewData(viewData, out, formatter);
+      emitViewData(viewData, current.view.getName(), out, formatter);
     }
     emitMeasureTable(measures, out, formatter);
   }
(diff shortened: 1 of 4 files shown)


bogdandrutu:
Some high level feedback. I would change the header "Current Path" with "Views" and move the current path logic under that title. That way we have views/measures as headers.

songy23:
@bogdandrutu sure, updated. Also updated the screenshots.

songy23:
Thanks!

contrib/zpages/src/main/java/io/opencensus/contrib/zpages/StatszZPageHandler.java
```diff
@@ -0,0 +1,558 @@
...
+
+  /*
+   * TreeNode for storing the structure of views and directories that they're in. Think of this as
+   * file descriptors for view: non-leaf nodes are directories which may contain views or other
+   * directories, and leaf nodes are the ones with actual information on views. Each non-leaf node
+   * also has the number of views under its directory.
+   */
+  private static class TreeNode {
+    // Only leaf nodes have views.
+    @javax.annotation.Nullable final /*@Nullable*/ View view;
```
sebright: The commented annotation is redundant.

songy23: Created https://github.com/census-instrumentation/opencensus-java/pull/1005 to fix this.

PR #1006, opened 2018-02-17
Synchronize the access to the tree, views and measures for Statsz.

songy23:
To resolve the potential race condition when getting the Statsz page.

### DIFF
--- a/contrib/zpages/src/main/java/io/opencensus/contrib/zpages/StatszZPageHandler.java
+++ b/contrib/zpages/src/main/java/io/opencensus/contrib/zpages/StatszZPageHandler.java
@@ -72,6 +72,11 @@ final class StatszZPageHandler extends ZPageHandler {
 
   private final ViewManager viewManager;
 
+  // measures, cachedViews and root are created when StatszZPageHandler are initialized, and will
+  // be updated every time when there's a new View from viewManager.getAllExportedViews().
+  // viewManager.getAllExportedViews() will be called every time the StatsZ page is re-rendered,
+  // like refreshing or navigating to other paths.
+
   @GuardedBy("monitor")
   private final Map<String, Measure> measures = Maps.newTreeMap();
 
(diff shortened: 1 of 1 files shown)


contrib/zpages/src/main/java/io/opencensus/contrib/zpages/StatszZPageHandler.java
```diff
@@ -67,9 +68,17 @@
...
   private final ViewManager viewManager;
+
+  @GuardedBy("monitor")
   private final Map<String, Measure> measures = Maps.newTreeMap();
+
+  @GuardedBy("monitor")
   private final Set<View> cachedViews = Sets.newHashSet();
+
+  @GuardedBy("monitor")
   private final TreeNode root = new TreeNode();
```
sebright: It would help to have a comment explaining the purpose of `measures`, `cachedViews`, and `root` and how they are updated, though it doesn't need to be in this PR.

contrib/zpages/src/main/java/io/opencensus/contrib/zpages/StatszZPageHandler.java
```diff
@@ -193,61 +204,64 @@ private void emitDirectoryTable(
...
-            "<tr class=\"%s\"><td>Directory: <a href='?%s=%s'>%s</a> (%d %s)</td></tr>",
-            CLASS_LARGER_TR,
-            QUERY_PATH,
-            path + '/' + relativePath,
-            relativePath,
-            child.views,
-            child.views > 1 ? "views" : "view");
-      } else { // View node, emit a row for view.
-        String viewName = child.view.getName().asString();
+    synchronized (monitor) {
```
bogdandrutu: Why this? You call this function while you hold the monitor lock correct? You need to add an annotation on the method that is guarded by the monitor.

contrib/zpages/src/main/java/io/opencensus/contrib/zpages/StatszZPageHandler.java
```diff
@@ -193,61 +204,64 @@ private void emitDirectoryTable(
...
-      for (int i = 0; i < dirs.size(); i++) {
-        String dir = dirs.get(i);
-        if ("".equals(dir) && i == 0) {
-          continue; // Skip the first "", the path of root node.
-        }
-        if (!node.children.containsKey(dir)) {
-          return null;
-        } else {
-          node = node.children.get(dir);
+    synchronized (monitor) {
```
bogdandrutu: Same question here.

contrib/zpages/src/main/java/io/opencensus/contrib/zpages/StatszZPageHandler.java
songy23: Sure, I added some comments regarding that.

contrib/zpages/src/main/java/io/opencensus/contrib/zpages/StatszZPageHandler.java
songy23: Oh that's right - updated.

contrib/zpages/src/main/java/io/opencensus/contrib/zpages/StatszZPageHandler.java
songy23: Done.

songy23:
Make sense to me - I updated the header and add a sentence to explain this.

songy23:
Thanks!

PR #1012, opened 2018-02-20
Add a command on pushing tags for release.

songy23:


RELEASING.md
```diff
@@ -146,6 +146,7 @@ token](https://help.github.com/articles/creating-a-personal-access-token-for-the
     $ git merge --ff-only release
     $ git push upstream v$MAJOR.$MINOR.$PATCH
     $ git push upstream v$MAJOR.$MINOR.x
+    $ git push upstream --tags
```
sebright: Does --tags push all tags?  I think it would be better to specify only the release tag.

songy23: You're right, pushing all tags at once could be error-prone. Though for some reason sometimes `git push upstream v$MAJOR.$MINOR.$PATCH` doesn't work for me, and I have to explicitly push all tags.

songy23:
Closing - `git push upstream v$MAJOR.$MINOR.$PATCH` is preferred.

Issue #1014, opened 2018-02-22
Build fails on JDK9


al3xandru:
Please answer these questions before submitting a bug report.

### What version of OpenCensus are you using?

This is against master ca3e76f3 

### What JVM are you using (`java -version`)?

```
java version "9.0.4"
Java(TM) SE Runtime Environment (build 9.0.4+11)
Java HotSpot(TM) 64-Bit Server VM (build 9.0.4+11, mixed mode)
```

### What did you do?
If possible, provide a recipe for reproducing the error.

* check out the project
* `./gradlew build`

### What did you expect to see?

successful build :-)

### What did you see instead?

```
./gradlew build

> Task :opencensus-api:compileJava FAILED
warning: [options] bootstrap class path not set in conjunction with -source 1.6
warning: [options] source value 1.6 is obsolete and will be removed in a future release
warning: [options] target value 1.6 is obsolete and will be removed in a future release
warning: [options] To suppress warnings about obsolete options, use -Xlint:-options.
error: warnings found and -Werror specified


FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':opencensus-api:compileJava'.
> Compilation failed with exit code 1; see the compiler error output for details.
```

```
./gradlew build --stacktrace

> Task :opencensus-api:compileJava FAILED
warning: [options] bootstrap class path not set in conjunction with -source 1.6
warning: [options] source value 1.6 is obsolete and will be removed in a future release
warning: [options] target value 1.6 is obsolete and will be removed in a future release
warning: [options] To suppress warnings about obsolete options, use -Xlint:-options.
error: warnings found and -Werror specified


FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':opencensus-api:compileJava'.
> Compilation failed with exit code 1; see the compiler error output for details.

* Try:
Run with --info or --debug option to get more log output.

* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':opencensus-api:compileJava'.
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeActions(ExecuteActionsTaskExecuter.java:100)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:70)
[...]

PR #1016, opened 2018-02-22
Test with oraclejdk9 on Travis.

sebright:
This PR adds a Java 9 build job.  Previously, the build applied -Werror when the Java version was 8 or greater.  Since the Java 9 build currently has some warnings (issue #1014), I applied -Werror only with Java 8 and added a TODO.  I also fixed some Error Prone warnings.


sebright:
Thanks!

PR #1020, opened 2018-02-24
Initial HTTP instrumentation helper commit. (closes #980)

### LINKED ISSUE #980: Helper HTTP classes for instrumentation.
Create an HttpClientUtil and HttpServerUtil class that help with instrumenting http frameworks with tracing and stats.

We manually instrumented the Google cloud HTTP client (add attributes, determine the span name, record message events (determine ids), etc.) but for consistency between multiple HTTP implementations we should have helper classes to instrument clients and servers.

HailongWen:
This PR adds several helpers to instrument HTTP libraries/frameworks.

The helpers should:
1. provide default implementation so we can use them to instrument existing libraries/frameworks in a consistent way, and minimize the effort.
2. provide flexibility for users to customize and extend the functionalities.

The PR referenced the design of [zipkin http instrumentation helper](https://github.com/openzipkin/brave/blob/master/instrumentation/http), except for some minor simplifications:
1. this PR does not have `HttpSampler` - I am not sure if current version of Open Census can provide different samplers for different route.
2. only two kinds of classes - `HttpHandler` for OpenCensus to instrument framework (equivalent of `HttpHandler` in zipkin), and `HttpExtractor` as adaptor to help get information from framework specific classes (equivalent of `HttpAdaptor` and `HttpParser` in zipkin). 

The instrumentation of send and recv are separated. For sync client/server instrumentation, the send/recv instrumentation code should be used in pair. For async client/server instrumentation, the send/recv instrumentation code should be used separately, and users should write extra code to manage the span/context. 

Things to address:
- [x] use primitive int as statusCode.
- [x] remove `getRoute` since it is not an HTTP attribute, and is only available at server-side after the route mapping.
- [x] refactor instrumentation API - does not allow user to reuse an existing span as the HTTP span, and symmetrically remove the boolean `endSpan`.
- [x] add `HttpSpanCustomizer` for span creation customization.
- [x] extract the method of parsing status as a separate util.
- [x] extract the log message event behavior as `handleMessageSent` and `handleMessageReceived`.
- [x] doc enhancement (mention B3 format).
- [x] javadoc enhancement (arguments for Q and P).
- [x] modifier and annotation.
- [x] move mock test class to testing.
- [x] reuse the `MockableSpan` from #410 for test.
- [x] add integration examples.


adriancole:
Thanks.. couple things

Make http status primitive where zero is invalid. Prevents routine boxing

The route is usually not available until the response (at least later in
the request). You will want to parse a name with that at response time
(potentially overwriting an initial span name)

The boolean endspan is an interesting thought. Maybe show a comment where
you are at a point where you have a response or exception and dont want to
finish the span? There are some edge cases which i can re look at.. dont
recall if i felt the need to defer start or defer finish

contrib/http_util/README.md
```diff
@@ -32,6 +32,104 @@ compile 'io.opencensus:opencensus-api:0.12.0'
...
+// public class HttpRequest {
+//   ...
+// }
+//
+// // Http response entity in the library/framework.
+// public class HttpResponse {
+//   ...
+// }
+
+TextFormat mytextFormat = HttpPropagationUtil.getCloudTraceFormat();
```
bogdandrutu: We should explain that users can pick B3 format as well here.

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpClientHandler.java
```diff
@@ -0,0 +1,137 @@
...
+import io.opencensus.trace.Tracer;
+import io.opencensus.trace.propagation.TextFormat;
+import javax.annotation.Nullable;
+
+/**
+ * This helper class provides APIs for instrumenting HTTP clients.
+ *
+ * @since 0.13
+ */
+public class HttpClientHandler<Q, P> extends HttpHandler<Q, P> {
```
bogdandrutu: Please add javadoc for Q and P.

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpClientHandler.java
```diff
@@ -0,0 +1,137 @@
...
+   * @param method the name of the send method.
+   * @return a span that represents the request process.
+   * @since 0.13
+   */
+  public <C> Span handleSend(TextFormat.Setter<C> setter, C carrier, Q request, String method) {
+    checkNotNull(setter, "setter");
+    checkNotNull(carrier, "carrier");
+    checkNotNull(request, "request");
+    Span span = tracer.spanBuilder(getSpanName(request, method)).startSpan();
+    handleSend(setter, carrier, request, span);
```
bogdandrutu: You double check setter/carrier/request for null here.

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpClientHandler.java
bogdandrutu: Should this class be final?

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpHandler.java
```diff
@@ -0,0 +1,213 @@
...
+    this.extractor = extractor;
+  }
+
+  /**
+   * Returns the tracing component.
+   *
+   * @return the tracing component.
+   * @since 0.13
+   */
+  public Tracer getTracer() {
```
bogdandrutu: why is this public?

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpHandler.java
```diff
@@ -0,0 +1,213 @@
...
+ *
+ * <p>This class provides basic operations, such as recording {@link MessageEvent} and handling
+ * errors.
+ *
+ * @since 0.13
+ */
+public abstract class HttpHandler<Q, P> {
+
+  /** The Open Census tracing component. */
+  Tracer tracer;
```
bogdandrutu: Why not private final? Same for all the other members. If for test please use the annotation VisibleForTesting

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpClientHandler.java
bogdandrutu: I think we should split handleSent into 2 methods. handleStart + handleMessageSent similar for handleRecv.

I know that it means more calls for the person who instruments this, but to start a span you just need the method and that may be available earlier than the payload that you need for recording the message. Especially for the server side.

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpHandler.java
```diff
@@ -0,0 +1,213 @@
...
+    return sentMessageEventId.getAndIncrement();
+  }
+
+  /**
+   * Generate a sequential id for received {@link MessageEvent}.
+   *
+   * @return next id for received {@code MessageEvent}.
+   * @since 0.13
+   */
+  public long nextRecvMessageEventId() {
```
bogdandrutu: Why public? please make public only APIs that user must call. Also consider to make them final if you don't expect override.

contrib/http_util/src/test/java/io/opencensus/contrib/http/FakeSpan.java
```diff
@@ -0,0 +1,58 @@
...
+import io.opencensus.trace.EndSpanOptions;
+import io.opencensus.trace.Link;
+import io.opencensus.trace.MessageEvent;
+import io.opencensus.trace.Span;
+import io.opencensus.trace.SpanContext;
+import io.opencensus.trace.Status;
+import java.util.Map;
+
+/** A fake {@link Span} which allows user to set {@link SpanContext} upon initialization. */
+public class FakeSpan extends Span {
```
bogdandrutu: Please add this in the opencensus-testing package. See https://github.com/census-instrumentation/opencensus-java/pull/410 . I probably need to merge that.

contrib/http_util/src/test/java/io/opencensus/contrib/http/FakeHttpResponse.java
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.contrib.http;
+
+/** Fake HTTP response entity class. */
+public abstract class FakeHttpResponse {
```
bogdandrutu: If you want this to be public consider to put it in a package:
package io.opencensus.contrib.http.testing

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpHandler.java
```diff
@@ -0,0 +1,213 @@
...
+   *
+   * <p>Users can override this method to define their own parsing behaviors.
+   *
+   * @param response the HTTP response entity. It could be null if error is present.
+   * @param error the error occured during response transmition. It could be null if response is
+   *     present.
+   * @return the corresponding OpenCensus {@code Status}.
+   * @since 0.13
+   */
+  public Status parseResponseStatus(@Nullable P response, @Nullable Throwable error) {
```
bogdandrutu: I think we should make this an independent util function which accepts an int status and maybe an error.

adriancole:
I know this is somewhat routine here, but worth mentioning that depth-first merging has a few problems. What I mean by depth-first, is that instead of implementing instrumentation then showing it become better with an abstraction, we add abstractions and sometimes no instrumentation to use it. This makes it hard to grok especially edge cases or if the api really works or not.

there are a couple alternatives:
* merge instrumentation first, then refactor over
  * assumes some instrumentation lives here, or requires reviewing PRs based on a branch build
* make learning test with basic, but realistic instrumentation
  * ex literally use apache async http client, or async servlet to show that this works, in a test. this could possibly be removed later once formal support is done.

At any rate, some means to prevent "eyeball oks" would help folks make judgement calls as instrumentation of 3rd party apis can be subtle, and this is meant for 3rd party (not even google) apis I think.

sebright:
Did you mean to exclude the binary file?  https://github.com/HailongWen/opencensus-java/blob/8c2f90a2e360b43aafa3fa32443464bd2d50fea3/api/src/main/java/io/opencensus/trace/.Status.java.swp

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpHandler.java
HailongWen: It is public to be used by user to when recording a new message event. We could hide them if we expose something like `recordSentMessageEvent`/`recordRecvMessageEvent` which automatically specify id.
It is not final because I would like user to extend them if they need something other than sequential increment behaviors.

I know that composition is better than inheritance but here this case is too subtle, so I didn't create another class to hold this API.

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpClientHandler.java
HailongWen: No. Some of the methods are intended to be overridden by user if they want their own implementations.

contrib/http_util/src/test/java/io/opencensus/contrib/http/FakeSpan.java
bogdandrutu: For the moment mark this as internal so that we can remove it later. I don't want you to get block by my comment. Also if possible to make it package protected that is even better.

sebright:
@HailongWen Is this ready for review again?

contrib/http_util/README.md
```diff
@@ -32,6 +32,180 @@ compile 'io.opencensus:opencensus-api:0.12.2'
...
+HttpClientHandler<HttpRequest, HttpResponse> handler =
+    new HttpClientHandler<HttpRequest, HttpResponse>(tracer, myTextFormat, extractor, customizer);
+
+// Use #handleStart in client to start a new span.
+Span span = handler.handleStart(myTextFormatSetter, request, request);
+HttpResponse response = null;
+Throwable error = null;
+try {
+  // Do something to send the request, and get response code from the server
+  int responseCode = request.getResponseCode();
```
adriancole: remove this as it is confusing :) request usually doesn't have response code.

maybe use real servlet code in this example instead? Doesn't have to be async.. rather it helps it be more concrete.

Here's a chopped example of brave servlet filter that only works for async, but yeah can be used.
```java
    Span span = handler.handleReceive(extractor, httpRequest);
    Throwable error = null;
    try (Tracer.SpanInScope ws = tracer.withSpanInScope(span)) {
      chain.doFilter(httpRequest, httpResponse);
    } catch (IOException | ServletException | RuntimeException | Error e) {
      error = e;
      throw e;
    } finally {
        handler.handleSend(ADAPTER.adaptResponse(httpRequest, httpResponse), error, span);
   }
```

HailongWen: Thanks! The `getResponseCode` comes from `HttpURLConnection` and I think I kind of mixed them up.

I've written a small netty example here: https://github.com/census-instrumentation/opencensus-experiments/blob/master/integration/src/main/java/io/opencensus/interop/http/netty/OpenCensusClientInboundOutboundHandler.java

HailongWen:
@bogdandrutu @adriancole PTAL.
An example for Netty HTTP async server/client is in a separate PR #1105.

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpExtractor.java
```diff
@@ -0,0 +1,90 @@
...
+ *
+ * <p>Please refer to this <a
+ * href="https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md">document</a>
+ * for more information about the HTTP attributes recorded in Open Census.
+ *
+ * @param <Q> the HTTP request entity.
+ * @param <P> the HTTP response entity.
+ * @since 0.13
+ */
+public class HttpExtractor<Q, P> {
```
bogdandrutu: better to make this class abstract with abstract methods instead of returning default fake values.

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpSpanCustomizer.java
```diff
@@ -0,0 +1,101 @@
...
+   * <p>This method will set the status by default. Users can override this to implement their own
+   * parsing logic.
+   *
+   * @param response the HTTP response entity. {@code null} means invalid response.
+   * @param error the error occurs when processing the response.
+   * @param span the span to be customized.
+   * @param extractor the framework specific extractor to get information from the request.
+   * @since 0.13
+   */
+  public void customizeSpanEnd(
```
bogdandrutu: We now have a spec for what attributes we should record:
https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpHandler.java
```diff
@@ -0,0 +1,120 @@
...
+import javax.annotation.Nullable;
+
+/**
+ * Base class for {@link HttpClientHandler} and {@link HttpServerHandler}.
+ *
+ * @param <Q> the HTTP request entity.
+ * @param <P> the HTTP response entity.
+ * @since 0.13
+ */
+public abstract class HttpHandler<Q, P> {
```
bogdandrutu: Please mark all your new classes with: `@ExperimentalApi`

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpClientHandler.java
```diff
@@ -0,0 +1,92 @@
...
+   * @param <C> the type of the carrier.
+   * @param textFormat the {@code TextFormat} used in HTTP propagation.
+   * @param setter the setter used when injecting information to the {@code carrier}.
+   * @param carrier the entity that holds the HTTP information.
+   * @param request the request entity.
+   * @return a span that represents the request process.
+   * @since 0.13
+   */
+  public <C> Span handleStart(
+      TextFormat textFormat, TextFormat.Setter<C> setter, C carrier, Q request) {
```
bogdandrutu: Can we make handleStart accept only "Request" and move everything else to the ctor? Please do this only if it does not complicate things too much.

Same applies for the Server.

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpSpanCustomizer.java
```diff
@@ -0,0 +1,101 @@
...
+ * href="https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md">HTTP
+ * trace spec</a> for more information.
+ *
+ * <p>Users can extend this helper to customize builder and end status.
+ *
+ * @param <Q> the HTTP request entity.
+ * @param <P> the HTTP response entity.
+ * @since 0.13
+ */
+public class HttpSpanCustomizer<Q, P> {
```
bogdandrutu: Can we make this package protected and avoid mentioning it in the README until we find an use-case where we really need it?

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpHandler.java
```diff
@@ -0,0 +1,120 @@
...
+  @VisibleForTesting final Tracer tracer;
+
+  /** The {@link HttpExtractor} used to extract information from request/response. */
+  @VisibleForTesting final HttpExtractor<Q, P> extractor;
+
+  /** The {@link HttpSpanCustomizer} used to customize span behaviors. * */
+  @VisibleForTesting final HttpSpanCustomizer<Q, P> customizer;
+
+  /** Package-protected constructor to allow access from subclasses only. */
+  HttpHandler(Tracer tracer, HttpExtractor<Q, P> extractor, HttpSpanCustomizer<Q, P> customizer) {
```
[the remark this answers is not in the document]
HailongWen: It is the second case. I will update the comment.

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpExtractor.java
HailongWen: That makes sense. I will make it `abstract`.
My original thought is to reduce the amount of user code when instrumenting.
How about adding a `HttpExtractorAdaptor` that provides fake no-op?

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpSpanCustomizer.java
HailongWen: My understanding of the spec is that it is just "recommended" but instrumenters (probably not us) still can decide what attributes they want to add.
If that is the case I would like to make the customizer `abstract` and implement a `OpenCensusHttpSpanCustomizer` to add those routines defined in our specs.

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpClientHandler.java
HailongWen: Maybe.
The existence of `textFormat` is for users to decide what text format to use. Sometimes users might want to use request parameters to determine what text format to use (just as "q=textFormat" in our interop testing). If we move it to ctor, users can create multiple handlers, each for different `TextFormat`.
The `carrier` and `setter` is framework specific and should already be decided. So yes I think they can be placed in ctor.

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpSpanCustomizer.java
HailongWen: Yes. Since we have a canonical way to instrument a span in the specification, we can use it as the default one and hide the customizer.

bogdandrutu:
Please ping me when ready for review.

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpSpanCustomizer.java
Sebruck: In opencensus-scala we decided to have functions to set the basic http attributes as well since this is what you probably always want to set.

https://github.com/Sebruck/opencensus-scala/blob/master/http/src/main/scala/com/github/sebruck/opencensus/http/HttpAttributes.scala

contrib/http_util/src/main/java/io/opencensus/contrib/http/util/HttpTraceUtil.java
```diff
@@ -0,0 +1,94 @@
...
+   * canonical error code</a>, and the behavior is defined in <a
+   * href="https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md">OpenCensus
+   * Specs</a>.
+   *
+   * @param statusCode the HTTP response status code. {@code 0} means invalid response.
+   * @param error the error occured during response transmission (optional).
+   * @return the corresponding OpenCensus {@code Status}.
+   * @since 0.13
+   */
+  public static final Status parseResponseStatus(int statusCode, @Nullable Throwable error) {
```
Sebruck: I'd prefer to have a simpel `@Nullable String message` as second parameter. In Scala land  (or FP in general) we do not use Exceptions to signal "expected errors" (like a 404), instead we encode it into an ADT (like an Either monad). 

Having the need of an Throwable here would force everyone to use exceptions for expected errors.

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpServerHandler.java
```diff
@@ -0,0 +1,101 @@
...
+    checkNotNull(carrier, "carrier");
+    checkNotNull(request, "request");
+    SpanBuilder spanBuilder = null;
+    String spanName = customizer.getSpanName(request, extractor);
+    String parseError = null;
+    // de-serialize the context
+    try {
+      SpanContext spanContext = textFormat.extract(carrier, getter);
+      spanBuilder = tracer.spanBuilderWithRemoteParent(spanName, spanContext);
+    } catch (Exception e) {
```
Sebruck: Maybe it would make sense to be more explicit here what exceptions you catch.

`textFormat.extract` throws a `SpanContextParseException` when the input is invalid

`tracer.spanBuilderWithRemoteParent` throws a `NullPointerException` when the spanName is null.

I think those two error cases need explicit handling, especially when the second case happens the whole thing will blew up since `tracer.spanBuilder` will again throw a `NullPointerException`.

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpServerHandler.java
```diff
@@ -0,0 +1,101 @@
...
+      // record this exception
+      spanBuilder = tracer.spanBuilder(spanName);
+      parseError = e.getMessage() == null ? e.getClass().getSimpleName() : e.getMessage();
+    }
+
+    customizer.customizeSpanBuilder(request, spanBuilder, extractor);
+    Span span = spanBuilder.startSpan();
+    // log an annotation to indicate the error
+    if (parseError != null) {
+      span.addAnnotation(Annotation.fromDescription("Error parsing span context: " + parseError));
```
Sebruck: Should this really be an annotation on the span and just a simple info-log? Having a client not sending a span context is totally normal and expected. Especially when the service is the "first on in the row".

I would feel weird to have the same Annotation as an "error" (in fact it's not an error, but expected) on each span.

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpExtractor.java
Sebruck: FYI: We have a similar type class in opencensus-scala https://github.com/Sebruck/opencensus-scala/blob/master/http/src/main/scala/com/github/sebruck/opencensus/http/HttpAttributes.scala#L21

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpSpanCustomizer.java
```diff
@@ -0,0 +1,101 @@
...
+   *
+   * <p>Note that in OpenCensus, {@link Span} is immutable and cannot be renamed.
+   *
+   * @param request the HTTP request entity.
+   * @param extractor the framework specific extractor to get information from the request.
+   * @return a string that represents the name of the span.
+   */
+  public String getSpanName(Q request, HttpExtractor<Q, P> extractor) {
+    // default span name
+    return "/" + extractor.getPath(request);
```
Sebruck: how can you be sure that the path does not already start with a leading slash? In this case we would have a double `//` in the span name. E.g. in the Akka-Http framework `req.uri.path.toString()` returns the path with a leading slash.

Maybe it makes sense to have a look if `extractor.getPath(request)` already contains a leading slash?

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpClientHandler.java
```diff
@@ -0,0 +1,92 @@
...
+    checkNotNull(textFormat, "textFormat");
+    checkNotNull(setter, "setter");
+    checkNotNull(carrier, "carrier");
+    checkNotNull(request, "request");
+    String spanName = customizer.getSpanName(request, extractor);
+    SpanBuilder builder = tracer.spanBuilder(spanName);
```
Sebruck: It would be good to be able to also start a span with a parentContext on the client side. Since it could be that the client is just another service which already has a started span.

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpSpanCustomizer.java
HailongWen: Updated. Now it contains basic routine. "Route" is not added because it is usually framework specific.

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpHandler.java
HailongWen: Done.

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpClientHandler.java
HailongWen: Removed `textFormat` and `setter` (`getter` for server).

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpSpanCustomizer.java
HailongWen: Removed `public` modifier.

contrib/http_util/src/main/java/io/opencensus/contrib/http/util/HttpTraceUtil.java
HailongWen: Ack. I will leave it as it is now (i.e., null indicating no error).
For `Either` or `Option` or whatever ADT, this would require an extra transform.

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpServerHandler.java
HailongWen: Thanks. Refined the logic.
Currently we can't distinguish between parse error and missing context so I just ignore the exception. I added a TODO here.
As for the `NullPointerException`, we currently use a non-public customizer and it always return non-null span name. I added a TODO here as well.

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpServerHandler.java
HailongWen: Removed.

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpSpanCustomizer.java
HailongWen: Updated.

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpClientHandler.java
HailongWen: Added param `parent` for specifying parent span. `null` means using current span.

HailongWen:
@bogdandrutu @Sebruck PTAL.

contrib/http_util/README.md
```diff
@@ -32,6 +32,148 @@ compile 'io.opencensus:opencensus-api:0.13.2'
...
+Throwable error = null;
+try {
+  // Do something to send the request, and get response code from the server
+  response = getResponse(request);
+
+  // Optionally, use #handleMessageSent in client to log a SENT event and its size.
+  handler.handleMessageSent(span, sentId++, request.getContentLength());
+
+  // Optionally, use #handleMessageReceived in client to log a RECEIVED event and message size.
+  handler.handleMessageReceived(span, recvId++, response.getContentLength());
```
Sebruck: By just reading this documentation, it's not obvious to me why i need `sentId` and `recvId`

contrib/http_util/README.md
```diff
@@ -32,6 +32,148 @@ compile 'io.opencensus:opencensus-api:0.13.2'
...
+
+An example usage of the handler would be:
+
+```java
+HttpServerHandler<HttpRequest, HttpResponse> handler =
+    new HttpServerHandler<HttpRequest, HttpResponse, HttpRequest>(
+        tracer, extractor, customizer, myTextFormat, myTextFormatGetter);
+
+// Use #handleStart in server to start a new span.
+Span span = handler.handleStart(request, request);
```
Sebruck: Two time `request`?

contrib/http_util/README.md
```diff
@@ -32,6 +32,148 @@ compile 'io.opencensus:opencensus-api:0.13.2'
...
+### client
+
+Users can create a `HttpClientHandler` to help instrument client-side HTTP request/response.
+
+An example usage of the handler would be:
+
+```java
+HttpClientHandler<HttpRequest, HttpResponse, HttpRequest> handler =
+    new HttpClientHandler<HttpRequest, HttpResponse>(
+        tracer, extractor, customizer, myTextFormat, myTextFormatSetter);
```
Sebruck: The concept of the `customizer` is not explained in the doc.

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpClientHandler.java
```diff
@@ -0,0 +1,104 @@
...
+   *     request/response.
+   * @param customizer the {@link HttpSpanCustomizer} used to customize span behaviors.
+   * @param textFormat the {@code TextFormat} used in HTTP propagation.
+   * @param setter the setter used when injecting information to the {@code carrier}.
+   * @since 0.13
+   */
+  public HttpClientHandler(
+      Tracer tracer,
+      HttpExtractor<Q, P> extractor,
+      HttpSpanCustomizer<Q, P> customizer,
```
Sebruck: I think in most of the cases people do not want to pass a custom `customizer` since implementation of `HttpSpanCustomizer` already offers most of the things you usually need. I'd suggest to add a second constructor which does not have the `customizer` paramter and uses the current `HttpSpanCustomizer` as default.

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpHandler.java
```diff
@@ -0,0 +1,132 @@
...
+   * @param messageId an id for the message.
+   * @param messageSize the size of the message.
+   * @since 0.13
+   */
+  public final void handleMessageReceived(Span span, long messageId, long messageSize) {
+    checkNotNull(span, "span");
+    // record message size
+    recordMessageEvent(span, messageId, Type.RECEIVED, messageSize, 0L);
+  }
+
```
Sebruck: `handleMessageSent` and `handleMessageReceived` could be static as well, or?

Sebruck:
Except some small comments it looks quite good to me now. Good job :+1:

Most of the stuff in here, we've already done in the http instrumentations of [opencensus-scala](https://github.com/census-ecosystem/opencensus-scala). I will definitely try to replace large parts of it with this.

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpClientHandler.java
adriancole: Personally I think having a component like we do in Brave "HttpTracing" which includes these things is less maintenance and easier to manage. public constructors can be troublesome and this makes it less, moving the varied things to one spot.

contrib/http_util/README.md
HailongWen: Thanks for pointing out. We decided to hide that and won't expose it now. I'll change the example.

contrib/http_util/README.md
```diff
@@ -32,6 +32,148 @@ compile 'io.opencensus:opencensus-api:0.13.2'
...
+
+In asynchronous HTTP calls, message receiving and sending may happen in different
+threads. Users need to ensure the started span (as well as scope, if any) is
+closed or ended no matter the call is successful or not.
+
+To do that, store current scope and span somewhere, e.g. the context of the channel,
+and close them before the channel exits.
+
+Here is an [example](
+https://github.com/census-instrumentation/opencensus-java/tree/master/examples/src/main/java/io/opencensus/examples/http/netty
```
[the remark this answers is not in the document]
HailongWen: This is not an ideal example - too low level. I'll remove it for now.

contrib/http_util/README.md
```diff
@@ -32,6 +32,148 @@ compile 'io.opencensus:opencensus-api:0.13.2'
...
+
+    // Optionally, use #handleMessageSent in server to log a SENT message event and its message size.
+    handler.handleMessageSent(span, sentId++, response.getContentLength());
+  } catch (Throwable e) {
+    error = e;
+  } finally {
+    // Use #handleEnd in server to close the span.
+    handler.handleEnd(response, error, span);
+  }
+}
```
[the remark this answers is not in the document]
HailongWen: Agree. I will remove them - it is extracted from the netty example which probably contains too many details on the protocol.

contrib/http_util/README.md
HailongWen: My understanding is that this is a convention. In GRPC we separate the sentId and recvId.
As an example we could have used just one ID. @bogdandrutu

Sebruck: I’m still puzzled what I would pass when I Instrument a HTTP library. (Sorry i don’t know grpc good enough) Just an incrementing number? What is this good for, why do I need it? Is it meant for deduplication? 

I think others who instrument http libraries may ask the same questions, that’s why I think it might be good do document/explain.

langecode:
Hi,

I am looking into contributing a metrics filter for JAX-RS based services (and maybe also clients). Looks like it would be beneficial to use some of the work in this PR, e.g., for extraction of data from request and response.

I can see this PR has been laying still since 31 May - any plans on this?

HailongWen:
@Sebruck @langecode sorry for the delayed response. I am constantly interrupted by other high priority things. I'll set aside a block of time to get it done this week.

bogdandrutu:
Work restored in https://github.com/census-instrumentation/opencensus-java/pull/1537

PR #1021, opened 2018-02-25
Fix bugs, fix warnings and update disruptor.

bogdandrutu:
There are 2 bugs fixed in this PR:
1) The memory of SpanImpl, TagContext, MeasureMap was hold for all the objects in disruptor. Disruptor is configured with 8K objects so the memory for the past 8K recorded objects was hold.
2) The objects that are shared between threads via disruptor were not synchronized. Because the event is shared between producer and consumer thread all instances must be volatile.

### DIFF
--- a/impl/src/main/java/io/opencensus/impl/internal/DisruptorEventQueue.java
+++ b/impl/src/main/java/io/opencensus/impl/internal/DisruptorEventQueue.java
@@ -24,8 +24,6 @@ import com.lmax.disruptor.dsl.Disruptor;
 import com.lmax.disruptor.dsl.ProducerType;
 import io.opencensus.implcore.internal.DaemonThreadFactory;
 import io.opencensus.implcore.internal.EventQueue;
-import java.util.concurrent.Executors;
-import javax.annotation.Nullable;
 import javax.annotation.concurrent.ThreadSafe;
 
 /**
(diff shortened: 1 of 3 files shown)


impl_core/src/main/java/io/opencensus/implcore/internal/EventQueue.java
```diff
@@ -31,4 +31,12 @@
      */
     void process();
   }
+
+  /** No-op implementation of the {@link EventQueue.Entry}. */
+  enum NoopEntry implements Entry {
```
sebright: Why do we need a no-op Entry?

bogdandrutu: To fix the nullness warning. The entry is never going to be null.

bogdandrutu:
@sebright while thinking more about your question I found 2 bugs in the implementation of the Disruptor event queue. They are explained in the description of this PR.

impl/src/main/java/io/opencensus/impl/internal/DisruptorEventQueue.java
```diff
@@ -171,12 +175,17 @@ public DisruptorEvent newInstance() {
...
-    // TODO(sebright): Fix the Checker Framework warning.
-    @SuppressWarnings("nullness")
     public void onEvent(DisruptorEvent event, long sequence, boolean endOfBatch) {
-      event.getEntry().process();
+      Entry entry = event.getEntry();
+      if (entry != null) {
+        entry.process();
+      }
+      // Remove the reference to the previous entry to allow the memory to be gc'ed.
+      event.setEntry(null);
```
songy23: Just curious: `DisruptorEvent` will never be gc'ed?

bogdandrutu: That is correct and an intended behavior. `DisruptorEvent` is a small object and in total we hold 8KB memory here which should be fine.

PR #1022, opened 2018-02-26
Cleanup benchmarks and test both impl and impl-lite.

bogdandrutu:


benchmarks/README.md
```diff
@@ -3,5 +3,15 @@
...
 
 ```
-$ ./gradlew :benchmarks:jmh
+$ ./gradlew :opencensus-benchmarks:jmh
 ```
+
+## To debug compilation errors
+When you make incompatible changes in the Benchmarks classes you may get compilation errors which
+are related to the old code not being compatible with the new code. Run this commands to clean 
+the Gradle's cache:
```
sebright: Why doesn't Gradle detect that it needs to recompile?

benchmarks/src/jmh/java/io/opencensus/benchmarks/trace/BenchmarksUtil.java
```diff
@@ -0,0 +1,43 @@
...
+final class BenchmarksUtil {
+  private static final TraceComponentImplLite traceComponentImplLite = new TraceComponentImplLite();
+
+  static Tracer getTracer(String implementation) {
+    if (implementation.equals("impl")) {
+      // We can return the global tracer here because if impl is linked the global tracer will be
+      // the impl one.
+      // TODO(bdrutu): Make everything not be a singleton (disruptor, etc.) and use a new
+      // TraceComponentImpl similar to TraceComponentImplLite.
+      return Tracing.getTracer();
```
sebright: Should this method check that the tracer is the correct class in this case?

benchmarks/src/jmh/java/io/opencensus/benchmarks/trace/StartEndSpanBenchmark.java
```diff
@@ -16,33 +16,49 @@
...
-          .setSampler(Samplers.neverSample())
-          .startSpan();
 
-  @TearDown
-  public void doTearDown() {
-    rootSpan.end();
+  @State(Scope.Benchmark)
+  public static class Data {
+    private Tracer tracer;
+    private Span rootSpan = BlankSpan.INSTANCE;
```
sebright: Shouldn't `rootSpan` always be initialized by the setup method?

benchmarks/src/jmh/java/io/opencensus/benchmarks/trace/BenchmarksUtil.java
bogdandrutu: we do not have a different Tracer implementation, the only difference is the event-queue used (similar for stats). Once the TODO is resolved this will not be a problem.

benchmarks/src/jmh/java/io/opencensus/benchmarks/trace/StartEndSpanBenchmark.java
bogdandrutu: Yes, but the findbugs framework cannot determine that setup is always called before tear-down so I got an error that tear down may call uninitialized objects.

benchmarks/README.md
bogdandrutu: Few reasons that I found on internet:
1) Any plugin cannot delete the generated code (jmh generates code) because if the user configured the directory as the same as source code the plugin will delete users source code.
2) For some reasons which I couldn't find the right answer (or I did not understand exactly) after you run jmh  a demon will stay alive which may cache the  generated code in memory and generates the same code even if the files were changed (I am not clear why this happens because they should use some hash algorithm to determine if the file changed).

benchmarks/src/jmh/java/io/opencensus/benchmarks/trace/StartEndSpanBenchmark.java
sebright: I still think it would be safer to do a null check in the tear down method, or suppress the warning, to avoid the possibility of testing with the wrong span.

benchmarks/src/jmh/java/io/opencensus/benchmarks/trace/BenchmarksUtil.java
sebright: Right, and we can't access the TraceComponent to check its class.

benchmarks/README.md
sebright: I think it would help to give a little more detail about the reason in the readme.

bogdandrutu: Done.

benchmarks/src/jmh/java/io/opencensus/benchmarks/trace/BenchmarksUtil.java
bogdandrutu: Correct. We will fix this with the TODO.

benchmarks/src/jmh/java/io/opencensus/benchmarks/trace/StartEndSpanBenchmark.java
bogdandrutu: If we go a null check then we need to do that in all the benchmarks so not ideal. I will add the check that the value is different than the BlankSpan.

sebright: I don't understand, I thought that FindBugs only required the null check in the tear down method, not in the benchmark methods.  I thought about this more, and I think it might be better to suppress all of the FindBugs nullness warnings, since we are using the Checker Framework now.  The FindBugs null analysis is incomplete, and it sometimes conflicts with the Checker Framework.  I'll work on a PR for that.

PR #1023, opened 2018-02-26
Add traces exporter to Jaeger

### LINKED ISSUE #1003: Implement Jaeger Trace Exporter.

marccarre:
Fixes #1003.

**Description**:
This is an attempt to add support for Jaeger, inspired from:

- https://github.com/census-instrumentation/opencensus-go/blob/master/exporter/jaeger/jaeger.go
- https://github.com/jaegertracing/jaeger-client-java/blob/master/jaeger-core/src/main/java/com/uber/jaeger/reporters/protocols/JaegerThriftSpanConverter.java

**Disclaimer**:

- I'm new to both OpenCensus *and* Jaeger -- please be forgiving 🙂.
- ~I'm still confirming everything works end-to-end.
  I plan to share this soon, ideally in an automated/reproducible way.~ (done, see integration test)

~**The Ask**: 
Given this PR is still WIP, and especially since I'm new to OpenCensus and Jaeger, I'm looking for early feedback, and ideally, some guidance.~ (PR no longer WIP)

**Remaining tasks**:

- [x] Fix issues reported by CI.
- [x] Unit test(s) should verify all converted fields.
- [x] Unit test(s) should verify registration of the exporter.
- [x] Integration test(s) should prove that Jaeger properly receives exported traces.
- [x] `README.md` with introduction & instructions.
- [x] Pass code review 🤞 and/or address potential comments/feedback.

**Out of scope**:

- Add support for UDP sender (currently only supports HTTP+Thrift via port `14268`).


exporters/trace/jaeger/src/main/java/io/opencensus/exporter/trace/jaeger/JaegerExporterHandler.java
```diff
@@ -0,0 +1,273 @@
...
+    return spans;
+  }
+
+  private void exportToJaeger(final List<Span> spans) {
+    try {
+      sender.send(process, spans);
+    } catch (TException e) {
+      tracer
+          .getCurrentSpan()
+          .setStatus(
```
HailongWen: This will set the status of current span, which I think might not be owned by the exporter.
Better start your own span at the beginning of the method and end it before exit.

exporters/trace/jaeger/src/main/java/io/opencensus/exporter/trace/jaeger/JaegerExporterHandler.java
```diff
@@ -0,0 +1,273 @@
...
+        traceIdBuffer[9],
+        traceIdBuffer[10],
+        traceIdBuffer[11],
+        traceIdBuffer[12],
+        traceIdBuffer[13],
+        traceIdBuffer[14],
+        traceIdBuffer[15]);
+  }
+
+  private long spanIdToLong(final @Nullable SpanId traceId) {
```
HailongWen: The input argument is named `traceId`, which I think is misleading.

exporters/trace/jaeger/src/main/java/io/opencensus/exporter/trace/jaeger/JaegerExporterHandler.java
```diff
@@ -0,0 +1,273 @@
...
+    if (traceId == null) {
+      return 0L;
+    }
+    // Attempt to minimise allocations, since SpanId#getBytes currently creates a defensive copy:
+    traceId.copyBytesTo(spanIdBuffer, 0);
+    return Longs.fromByteArray(spanIdBuffer);
+  }
+
+  private int optionsToFlags(final TraceOptions traceOptions) {
+    // Attempt to minimise allocations, since SpanId#getBytes currently creates a defensive copy:
```
HailongWen: in the comment, s/SpanId/TraceOptions

exporters/trace/jaeger/src/main/java/io/opencensus/exporter/trace/jaeger/JaegerExporterHandler.java
marccarre: Sorry, bad copy-pasting, consider it fixed!

exporters/trace/jaeger/src/main/java/io/opencensus/exporter/trace/jaeger/JaegerExporterHandler.java
marccarre: Sorry, bad copy-pasting, consider it fixed!

exporters/trace/jaeger/src/main/java/io/opencensus/exporter/trace/jaeger/JaegerExporterHandler.java
```diff
@@ -0,0 +1,273 @@
...
+        @Override
+        public Tag apply(final Object value) {
+          final Tag tag = new Tag();
+          tag.setVType(TagType.STRING);
+          tag.setVStr(value.toString());
+          return tag;
+        }
+      };
+
+  // Re-usable buffers to avoid too much memory allocation during conversions.
```
marccarre: When exporting trace ID, span ID and options, these buffers allow us to avoid allocating too much memory for each exported span, however, we still copy to these buffers. Would it be better to move the conversion to the `TraceId`, `SpanId` and `TraceOptions` classes? Or would this be pre-mature optimisations at the cost of polluting their APIs?

exporters/trace/jaeger/src/main/java/io/opencensus/exporter/trace/jaeger/JaegerExporterHandler.java
marccarre: Thanks, this makes sense!
I guess I'll also override the default sampler with a probability sampler (with a low probability) to avoid potential infinite loops, like for other exporters.

marccarre:
FYI: When accessing the Jaeger instance started by the integration test, one can see the following:
image

exporters/trace/jaeger/src/main/java/io/opencensus/exporter/trace/jaeger/JaegerExporterHandler.java
```diff
@@ -0,0 +1,309 @@
+/*
+ * Copyright 2017, OpenCensus Authors
```
HailongWen: s/2017/2018. Others as well.

exporters/trace/jaeger/src/main/java/io/opencensus/exporter/trace/jaeger/JaegerExporterHandler.java
HailongWen: I understand the performance concerns behind that but I am still not sure how big the impact would be, given that the spans are already sampled and the total occupation is really small. Are you suggesting to expose something like `unsafeGetBytes()` to return the underlying buffer instead of a copied one in `TraceId`, `SpanId` and `TraceOptions`, or adding the similar kind of BufferManager logic to reuse a fixed size of buffer? I am fine with the former one as long as invoker knows what they are doing. The latter one seems to me a little bit pre-mature at least for now, though I do see the possible necessity of doing so in some certain scenarios, e.g. a fast streaming processing system.

/cc @bogdandrutu

marccarre: Again, this may be premature optimisation, so feel free to just reply with "_meh, YAGNI_" 😄.

FWIW, I've run a small benchmark, profiled it, and though none of the below calls stands out as a clear bottleneck or performance killer, the ones appearing at the relative "top" are:

- `AttributeValue#match(Function, Function, Function, Function)`
- `AutoValue_Annotation.getAttributes()` (from `annotationEventsToLogs`)
- `TraceId.copyBytesTo(byte[], int)`
- `Ints.fromByteArray(byte[])` (from `optionsToFlags`)
- `Lists.newArrayListWithExpectedSize(int)` (from various calls)

so I guess my intuition wasn't completely off.
Note that the benchmark was specifically on the export, not the entire lifecycle of an application with tracing enabled: this might not matter in the grand scheme of things.

Assuming users sample with low probability, it is probably not important.
However, given I'm new to all this, can we assume that?

In any case, if we end up caring, I can think of the following solutions:
1. Keep it as it is.
    - **Pro**: doesn't pollute the API ; "_OK_" performance as per benchmark.
    - **Con**: array-copies on each sample exported, but "_meh_".
2. Expose something like `unsafeGetBytes`.
    - **Pro**: allow exporter code to get the bytes w/o a copy. 
    - **Con**: pollutes the API a bit.
3. Expose something like `idAsJaegerLong()` (and for `TraceId`: `idAsJaegerHighLong()`/`idAsJaegerLowLong()`).
    - **Pro**: safer than 2. 
    - **Con**: pollutes the API a lot more than 2 ; opens the door to more pollution for all the other exporters.

so I'd probably for go 2. as well @HailongWen.
But I'll defer that call to you and  @bogdandrutu! 🙂
I'm not sure I understand the alternative with the "*BufferManager*" though, sorry.

HailongWen: About the `BufferManager`: never mind, it is a little bit off the original topic. I was thinking if you want to move the buffer logic you wrote in the exporter to OpenCensus library as a util or something. It turns out it is not the case.

About sampling: yes, usually the sampling ratio would be small but it all depends on users' scenario. `alwaysSample` can be used sometimes, e.g. if users don't want to miss a rare exception and can afford the overheads. In such case, I will be more concerned about exportation I/O.

About the benchmark: thanks for adding that. I am interested in the details, e.g. what the program is like, what is measured (time? memory? I/O?), what percentage of resources these functions consume.

marccarre: Thanks for the clarifications.

Regarding the benchmark, I used [YourKit](https://www.yourkit.com/) and measured CPU time, while running a modified version of `JaegerExporterHandlerTest`, exporting many times, re-using the same span (which means it is probably cached):
```java
    for (int i = 0; i < 5; i++) {
      handler.export(Collections.nCopies(1000000, spanData));
    }
```
Results aren't exactly identical when re-running it today, but still similar-ish:

image

HailongWen:
@marccarre PR merged. Thanks for the work!

HailongWen:
An integration test failed. @songy23 prepared a fix for that, see #1063 .

Issue #1017, opened 2018-02-22
Upgrade AutoValue to 1.5.3 or later.


sebright:
The Java 9 build currently produces warnings that could be fixed by upgrading to AutoValue 1.5.3.  See the discussion starting at https://github.com/census-instrumentation/opencensus-java/issues/1014#issuecomment-367601410.  However, AutoValue 1.5 removes support for generating code that can be built with Java 7 (https://github.com/google/auto/issues/449).  When we upgrade we will probably need to remove the Java 7 build job but ensure that we are still testing for compatibility with Android.

bogdandrutu:
We can still have the job for java7. We will compile the code using java8+ then we will just run tests on java7.

PR #1031, opened 2018-02-27
Reorganize warning suppression and enable -Werror with Java 7.

sebright:
#### Reorganize warning suppression and enable -Werror with Java 7.

This commit suppresses specific warnings for Java 7 and 9.  This allows us to
enable -Werror with Java 7, though there is still a warning with Java 9, due to
issue #1017.


sebright:
Thanks!

PR #1032, opened 2018-02-27
Suppress all FindBugs warnings about NullPointerExceptions.

sebright:
The warnings are redundant with the Checker Framework's warnings, and they
sometimes conflict.


sebright:
Thanks!

PR #1039, opened 2018-03-01
Add a QuickStart example.

songy23:
Provides an (almost) identical quick start example to [OpenCensus-Go](https://github.com/census-instrumentation/opencensus-go/pull/507). As suggested from the documentation meeting, we'd like to have one consistent quick start for all languages.

/cc @rakyll

### DIFF
--- a/examples/src/main/java/io/opencensus/examples/helloworld/QuickStart.java
+++ b/examples/src/main/java/io/opencensus/examples/helloworld/QuickStart.java
@@ -69,7 +69,7 @@ public final class QuickStart {
       Collections.singletonList(FRONTEND_KEY),
       Cumulative.create());
 
-
+  /** Main launcher for the QuickStart example. */
   public static void main(String[] args) {
     TagContextBuilder tagContextBuilder =
         tagger.currentBuilder().put(FRONTEND_KEY, TagValue.create("mobile-ios9.3.5"));
--- a/examples/src/main/java/io/opencensus/examples/helloworld/QuickStart.java
+++ b/examples/src/main/java/io/opencensus/examples/helloworld/QuickStart.java
@@ -60,6 +60,8 @@ public final class QuickStart {
       "my.org/measure/video_size", "size of processed videos", "MBy");
 
   // Create view to see the processed video size distribution broken down by frontend.
+  // The view has bucket boundaries (0, 256, 65536) that will group measure values into
+  // histogram buckets.
   private static final View.Name VIDEO_SIZE_VIEW_NAME = View.Name.create("my.org/views/video_size");
   private static final View VIDEO_SIZE_VIEW = View.create(
       VIDEO_SIZE_VIEW_NAME,
@@ -97,6 +99,7 @@ public final class QuickStart {
       logger.severe(e.getMessage());
     }
 
+    logger.info("Wait longer than the reporting duration...");
     // Wait for a duration longer than reporting duration (5s) to ensure spans are exported.
     Thread.sleep(5100);
     ViewData viewData = viewManager.getView(VIDEO_SIZE_VIEW_NAME);
--- a/examples/src/main/java/io/opencensus/examples/helloworld/QuickStart.java
+++ b/examples/src/main/java/io/opencensus/examples/helloworld/QuickStart.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-17, OpenCensus Authors
+ * Copyright 2018, OpenCensus Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/examples/src/main/java/io/opencensus/examples/helloworld/QuickStart.java
+++ b/examples/src/main/java/io/opencensus/examples/helloworld/QuickStart.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright 2016-17, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.examples.helloworld;
+
+import io.opencensus.common.Scope;
+import io.opencensus.stats.Aggregation;
+import io.opencensus.stats.BucketBoundaries;
+import io.opencensus.stats.Measure.MeasureLong;
+import io.opencensus.stats.Stats;
+import io.opencensus.stats.StatsRecorder;
+import io.opencensus.stats.View;
+import io.opencensus.stats.View.AggregationWindow.Cumulative;
+import io.opencensus.stats.ViewData;
+import io.opencensus.stats.ViewManager;
+import io.opencensus.tags.TagContextBuilder;
+import io.opencensus.tags.TagKey;
+import io.opencensus.tags.TagValue;
+import io.opencensus.tags.Tagger;
+import io.opencensus.tags.Tags;
+import io.opencensus.trace.SpanBuilder;
+import io.opencensus.trace.Status;
+import io.opencensus.trace.Tracer;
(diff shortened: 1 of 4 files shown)


examples/src/main/java/io/opencensus/examples/helloworld/QuickStart.java
```diff
@@ -0,0 +1,98 @@
+/*
+ * Copyright 2016-17, OpenCensus Authors
```
rakyll: -18

Btw, you can only put the current year.

examples/src/main/java/io/opencensus/examples/helloworld/QuickStart.java
```diff
@@ -0,0 +1,98 @@
...
+    TagContextBuilder tagContextBuilder =
+        tagger.currentBuilder().put(FRONTEND_KEY, TagValue.create("mobile-ios9.3.5"));
+    SpanBuilder spanBuilder = tracer.spanBuilder("my.org/ProcessVideo");
+    viewManager.registerView(VIDEO_SIZE_VIEW);
+
+    // Process video.
+    // Record the processed video size.
+    try (Scope scopedTags = tagContextBuilder.buildScoped();
+        Scope scopedSpan = spanBuilder.startScopedSpan()) {
+      tracer.getCurrentSpan().addAnnotation("Start processing video.");
```
rakyll: I will modify the Go example to have some annotations like this.

examples/src/main/java/io/opencensus/examples/helloworld/QuickStart.java
songy23: Done.

examples/src/main/java/io/opencensus/examples/helloworld/QuickStart.java
```diff
@@ -0,0 +1,98 @@
...
+    } catch (Exception e) {
+      tracer.getCurrentSpan().addAnnotation("Exception thrown when processing video.");
+      tracer.getCurrentSpan().setStatus(Status.UNKNOWN);
+      logger.severe(e.getMessage());
+    }
+
+    ViewData viewData = viewManager.getView(VIDEO_SIZE_VIEW_NAME);
+    logger.info(
+        String.format("Recorded stats for %s:\n %s", VIDEO_SIZE_VIEW_NAME.asString(), viewData));
+  }
```
HailongWen: Should we output the collected spans (e.g. using `LoggingTraceExproter`)? Or at least add a comment to pretend we are doing so? The caveats are that we need to specify a proper sampling policy, as well as `Thread.sleep()` for some time to wait for exportation. Both of these add to the complexity of the code.

examples/src/main/java/io/opencensus/examples/helloworld/QuickStart.java
```diff
@@ -0,0 +1,98 @@
...
+import io.opencensus.trace.SpanBuilder;
+import io.opencensus.trace.Status;
+import io.opencensus.trace.Tracer;
+import io.opencensus.trace.Tracing;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Random;
+import java.util.logging.Logger;
+
+/** Simple program that collects data for video size. */
```
HailongWen: Do we have documents for the designated behaviors of this example?

songy23: We have some ideas on what kind of examples we want. Just shared the doc with you.

examples/src/main/java/io/opencensus/examples/helloworld/QuickStart.java
songy23: SG, I added a LoggingExporter to print spans to log. The only downside is we need to [wait for >5s](https://github.com/census-instrumentation/opencensus-java/blob/master/impl_core/src/main/java/io/opencensus/implcore/trace/export/ExportComponentImpl.java#L29) before spans are exporter, while in the Go example we only wait for [2s](https://github.com/census-instrumentation/opencensus-go/blob/master/examples/helloworld/main.go#L86).

HailongWen: If you are not hurry to merge this change, I can create a PR to add a shutdownhook to export and flush the unexported data.

songy23: That sounds great, I can hold off this PR till you make that change.

examples/src/main/java/io/opencensus/examples/helloworld/QuickStart.java
```diff
@@ -0,0 +1,106 @@
...
+  private static final Tracer tracer = Tracing.getTracer();
+
+  // frontendKey allows us to break down the recorded data
+  private static final TagKey FRONTEND_KEY = TagKey.create("my.org/keys/frontend");
+
+  // videoSize will measure the size of processed videos.
+  private static final MeasureLong VIDEO_SIZE = MeasureLong.create(
+      "my.org/measure/video_size", "size of processed videos", "MBy");
+
+  // Create view to see the processed video size distribution broken down by frontend.
```
dinooliva: Expand this comment to describe the histogram buckets.

examples/src/main/java/io/opencensus/examples/helloworld/QuickStart.java
```diff
@@ -0,0 +1,106 @@
...
+      tracer.getCurrentSpan().addAnnotation("Exception thrown when processing video.");
+      tracer.getCurrentSpan().setStatus(Status.UNKNOWN);
+      logger.severe(e.getMessage());
+    }
+
+    // Wait for a duration longer than reporting duration (5s) to ensure spans are exported.
+    Thread.sleep(5100);
+    ViewData viewData = viewManager.getView(VIDEO_SIZE_VIEW_NAME);
+    logger.info(
+        String.format("Recorded stats for %s:\n %s", VIDEO_SIZE_VIEW_NAME.asString(), viewData));
```
dinooliva: Is there a way to log the trace annotations?

examples/src/main/java/io/opencensus/examples/helloworld/QuickStart.java
songy23: Added some explanations.

examples/src/main/java/io/opencensus/examples/helloworld/QuickStart.java
songy23: The `LoggingTraceExporter` will log `SpanData` to command line [every 5 seconds](https://github.com/census-instrumentation/opencensus-java/blob/master/impl_core/src/main/java/io/opencensus/implcore/trace/export/ExportComponentImpl.java#L29), and that should include all the information on these annotations (the message and time stamp). That's also why we need to wait 5.1s before exiting.

I added one more log to explain this, which matches [the Go example](https://github.com/census-instrumentation/opencensus-go/blob/master/examples/helloworld/main.go#L85).

examples/src/main/java/io/opencensus/examples/helloworld/QuickStart.java
songy23: Per our documentation meeting today, we would like to have this example ready ASAP. I'll leave a TODO to remove the 5.1s gap once we have the export shutdownhook.

songy23:
Thanks!

examples/src/main/java/io/opencensus/examples/helloworld/QuickStart.java
```diff
@@ -0,0 +1,110 @@
...
+  // The view has bucket boundaries (0, 256, 65536) that will group measure values into
+  // histogram buckets.
+  private static final View.Name VIDEO_SIZE_VIEW_NAME = View.Name.create("my.org/views/video_size");
+  private static final View VIDEO_SIZE_VIEW = View.create(
+      VIDEO_SIZE_VIEW_NAME,
+      "processed video size over time",
+      VIDEO_SIZE,
+      Aggregation.Distribution.create(
+          BucketBoundaries.create(
+              Arrays.asList(0.0, (double) (1 << 8), (double) (1 << 16)))),
```
odeke-em: Just a stray comment/question :)

This looks verbose, would autoboxing help(int to double) or is BucketBoundaries' parameterized type `Double`?

songy23: BucketBoundaries takes a List<Double>, so I have to do an explicit casting here. 

Maybe using 256.0 and 65536.0 instead of all these bit shifting will make the code cleaner - I'll post another PR to update this.

odeke-em: Great plan!

Issue #599, opened 2017-09-08
Add an example of stats/trace/tags


bogdandrutu:
Probably we should have a gRPC example where we show all:
* Z-Pages
* Exporting to: 
   Stackdriver;
   Zipkin and one open-source monitoring;
* Maybe include gRPC instrumentation service.

PR #1044, opened 2018-03-03
Add a gRPC Hello World example.

songy23:
Provides an (almost) identical gRPC Hello World example to [OpenCensus-C++](https://github.com/census-instrumentation/opencensus-cpp/tree/master/examples/grpc).

For now this example supports Maven and Gradle build. I'll post another PR to add support for bazel to avoid making this PR too huge (adding bazel build for this example introduces a huge amount of new dependencies).

EDIT: closes https://github.com/census-instrumentation/opencensus-java/issues/599.

~~For now this example only supports Maven build. I'll post another PR to add support for gradle and bazel to avoid making this PR too huge.~~

/cc @g-easy 

~~Maven building is working - I'm working on getting gradle/bazel work as well~~

~~TODO:~~
~~- [ ] Fix gradle and bazel build.~~
~~- [ ] Read port number from args.~~
~~- [ ] Add user tutorial.~~

### DIFF
--- a/examples/build.gradle
+++ b/examples/build.gradle
@@ -10,16 +10,17 @@ buildscript {
     }
     dependencies {
         classpath "gradle.plugin.com.github.sherter.google-java-format:google-java-format-gradle-plugin:0.6"
+	classpath 'com.google.protobuf:protobuf-gradle-plugin:0.8.3'
     }
 }
 
 apply plugin: 'checkstyle'
 apply plugin: 'idea'
 apply plugin: 'java'
+apply plugin: 'com.google.protobuf'
 // Plugins that require java8
 if (JavaVersion.current().isJava8Compatible()) {
     apply plugin: 'com.github.sherter.google-java-format'
-}
 
 repositories {
     mavenCentral()
@@ -30,6 +31,8 @@ group = "io.opencensus"
 version = "0.13.0-SNAPSHOT" // CURRENT_OPENCENSUS_VERSION
 
 def opencensusVersion = "0.12.2" // LATEST_OPENCENSUS_RELEASE_VERSION
+def grpcVersion = "1.9.0" // CURRENT_GRPC_VERSION
+def prometheusVersion = "0.3.0"
 
 checkstyle {
     configFile = file("$rootDir/../buildscripts/checkstyle.xml")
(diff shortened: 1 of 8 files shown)


examples/src/main/java/io/opencensus/examples/grpc/helloworld/HelloWorldServer.java
```diff
@@ -0,0 +1,189 @@
...
+    server.start();
+    server.blockUntilShutdown();
+  }
+
+  static class GreeterImpl extends GreeterGrpc.GreeterImplBase {
+
+    @Override
+    public void sayHello(HelloRequest req, StreamObserver<HelloReply> responseObserver) {
+      // Trace all incoming RPCs.
+      Tracing.getTraceConfig().updateActiveTraceParams(
```
HailongWen: I am not quite sure if we should do this in a multi-threaded RPC handler. Our current implementation of `getTraceConfig()` returns the [global, not thread-local configuration](https://github.com/census-instrumentation/opencensus-java/blob/ee1fd0259f46e21b1a90100ed35926a528d20ee0/impl_core/src/main/java/io/opencensus/implcore/trace/config/TraceConfigImpl.java#L23). 

The only way I can think of to make it sampled now is to invoke `setRecordEvents()` or pre-register the span for collection. @sebright and @bogdandrutu for confirmation.

sebright: I agree, the RPC shouldn't enable 100% sampling globally.  SpanBuilder allows setting a sampler for the new span, though.

HailongWen: Sorry I made a mistake in previous comment. They are for local span collection, not for sampling. We can call `SpanBuilder.setSampler()` to apply the sampler to the span.

examples/src/main/java/io/opencensus/examples/grpc/helloworld/HelloWorldClient.java
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  public void shutdown() throws InterruptedException {
+    channel.shutdown().awaitTermination(5, TimeUnit.SECONDS);
+  }
+
+  /** Say hello to server. */
+  public void greet(String name) {
+    // Trace all outgoing RPCs.
+    Tracing.getTraceConfig().updateActiveTraceParams(
```
HailongWen: Same as below. Prefer to use `setSampler()`.

songy23: Done. This one is actually redundant.

examples/src/main/java/io/opencensus/examples/grpc/helloworld/HelloWorldServer.java
songy23: Done, though I've seen cpp is setting global trace params: https://github.com/census-instrumentation/opencensus-cpp/blob/master/examples/grpc/hello_client.cc#L55.

examples/build.gradle
```diff
@@ -10,12 +10,14 @@ buildscript {
     }
     dependencies {
         classpath "gradle.plugin.com.github.sherter.google-java-format:google-java-format-gradle-plugin:0.6"
+	classpath 'com.google.protobuf:protobuf-gradle-plugin:0.8.3'
```
HailongWen: Tab here.

examples/build.gradle
```diff
@@ -30,6 +32,8 @@ group = "io.opencensus"
 version = "0.13.0-SNAPSHOT" // CURRENT_OPENCENSUS_VERSION
 
 def opencensusVersion = "0.12.2" // LATEST_OPENCENSUS_RELEASE_VERSION
+def grpcVersion = "1.9.0" // CURRENT_GRPC_VERSION
```
HailongWen: We've upgraded grpc to 1.10.0.

examples/pom.xml
```diff
@@ -11,23 +11,64 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <!-- change to the version you want to use. -->
     <opencensus.version>0.12.2</opencensus.version><!-- LATEST_OPENCENSUS_RELEASE_VERSION -->
+    <grpc.version>1.9.0</grpc.version><!-- CURRENT_GRPC_VERSION -->
```
HailongWen: The same.

examples/build.gradle
songy23: Done, seems a bad merge.

examples/build.gradle
songy23: Actually we have to use gRPC v1.9.0 for this example now, because it's using the latest released version (v0.12.2), which depends on gRPC 1.9.0.

I have tested with gRPC v1.10.0 before - they have made some breaking changes in CensusModule.java, and running the example with v1.10.0 won't work.

examples/pom.xml
songy23: Same reason, have to use v1.9.0 until we update the example to use opencensus v0.13.0.

songy23:
Thanks!

Issue #1015, opened 2018-02-22
Sampler not working


grv0908:
Please answer these questions before submitting a bug report.

### What version of OpenCensus are you using?
0.12.0

### What JVM are you using (`java -version`)?
java version "1.8.0_151"
Java(TM) SE Runtime Environment (build 1.8.0_151-b12)
Java HotSpot(TM) 64-Bit Server VM (build 25.151-b12, mixed mode)

### What did you do?

Tried this [example](https://github.com/census-instrumentation/opencensus-java/blob/master/examples/src/main/java/io/opencensus/examples/trace/MultiSpansTracing.java) with zipkinTraceExporter
`ZipkinTraceExporter.createAndRegister("http://127.0.0.1:9411/api/v2/spans", "service1");`
also used 
`private static final Sampler sampler = Samplers.alwaysSample();`
`Span rootSpan = tracer.spanBuilderWithExplicitParent("RootSpan", null).setSampler(sampler).startSpan();`

### What did you expect to see?
Spans in Zipkin UI

### What did you see instead?
No Span

But if I generate around 10k spans 
`for(int i= 0; i<10000; i++) {
             doWork();
        }`

Spans can be seen in zipkin ui probably only default sampler is working

bogdandrutu:
@grv0908 do you see 1 span or >100 spans? I am asking because we may have a different problem than the sampler which is the fact that if the application finishes fast enough we may not be able to send the data out of the task. This is still a bug but I need to know what is the behavior.

grv0908:
@bogdandrutu Yes I am getting lot more than 100 spans.

Yes you are correct. After removing for loop I have added a `Thread.sleep()` after
`doWork()`  so now Application is getting enough time to export that single span and able to see it in Zipkin.

bogdandrutu:
I think this can be closed now.

Issue #842, opened 2017-11-29
Consider refining trace examples and README


HailongWen:
Currently we have 3 trace examples: `MultiSpansTracing`, `MultiSpansContextTracing` and `MultiSpansScopedTracing`. Execute these examples just yield nothing. The examples themselves are correct, but there is no sign of any success or failure in the output - which is confusing to users.

2 issues are found related to this issue:
1. the interval for collecting batch span export(`EXPORTER_SCHEDULE_DELAY_MS`) is by default 2s and the examples run pretty fast so no spans actually got exported. In test, sleeping for some time helped.
1. default sampler set sampling probability to 1e-4. The span added in the examples are rarely sampled. In test, explicitly changing the parameter helped.

IMHO, we should either add instruction to clear the expected behavior and known issues, or refine the examples so that they can yield meaningful output.

songy23:
Examples have been moved to opencensus.io.

Issue #862, opened 2017-12-01
Switch to a push model for stats exporters.


sebright:
This issue requires ViewManager to support registering callbacks for views.  Then ViewManager can push stats to the StackDriver exporter.

bogdandrutu:
This is more important now that we have more exporters duplicating the worker thread code.

songy23:
Closing, since we will move to the metrics data model for exporters, which is a push model in essence.

PR #1050, opened 2018-03-05
Add shutdown hook to trace exporter.

HailongWen:
This PR adds a class `ShutdownHookManager` to manage registering the shutdown hooks, and use it in `SpanExporterImpl` to flush unexported data before exit. 

GAE Java 7 is not supported for now. Although tests shows that using `LifecycleManager` did help resolve the compatibility issue, the added appengine-api-sdk-1.0 is a big artifact, and it implicitly updates the version of the errorprone plugin which requires some extra code change. So for now I prefer not to include that change and leave it as a TODO in the future. For GAE Java 8, the compatibility is tested using grpc-interop-testing. (we should add our own interop-testing in the future.)

This PR might potentially resolve https://github.com/census-instrumentation/opencensus-java/issues/1015 so no sleep will not be necessary. However, a test is needed to verify that. 

A known issue is that this PR does not resolve the unexported data in `LoggingTraceExporter`, as reported in https://github.com/census-instrumentation/opencensus-java/pull/1039#discussion_r171713621 and https://github.com/census-instrumentation/opencensus-java/issues/842. The `java.util.logging` component registers its own shutdown hook, which closes all handlers upon shutdown signal and this hook could be executed before our registered one. Other logging system might not suffer this (I've tested log4j and it worked well). The same applies to all other user-registered shutdown hook, which is mentioned in the javadoc.

### DIFF
--- a/impl_core/src/main/java/io/opencensus/implcore/internal/ShutdownHookManager.java
+++ b/impl_core/src/main/java/io/opencensus/implcore/internal/ShutdownHookManager.java
@@ -30,19 +30,25 @@ import javax.annotation.concurrent.GuardedBy;
 import javax.annotation.concurrent.ThreadSafe;
 
 /**
- * A utility class to help with registering shutdown {@link Hook}s to the runtime.
+ * A utility class to help with registering shutdown {@link Hook}s to the runtime, influenced by
+ * Apache Hadoop's <a
+ * href="https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/ShutdownHookManager.java">ShutdownHookManager</a>.
+ * It is used for OpenCensus to clean things up in a deterministic order. Possible usages are: 1.
+ * close running spans. 2. Dump unexported spans for stat/tracing. 3. Allow users to have their own
+ * shutdown hooks (not supported yet, still in discussion).
  *
  * <p>A {@link Hook} can have its own name and priority. The execution order among registered hooks
  * will be determined by their priorities first (descending order) and then registration time
  * (ascending order). That is to say, higher priority {@code Hook} will be executed first, and
  * earlier registered {@code Hook} will be executed first given the same priority.
  *
- * <p>Caveat: 1. GAE Java 7 environment is not supported for now. 2. User can still use other method
- * to register a shutdown hook, e.g. the native {@link Runtime#addShutdownHook}. The execution order
- * of registered {@link Hook}s and other hooks depends on the JVM implementation and thus can not be
- * guaranteed, which might sometimes leads to unwanted behaviors. An example is that {@code
- * java.util.logging.Logger} has its own shutdown hook to reset all handlers, and once invoked, it
- * will prevent all future logging in other shutdown hooks.
+ * <p>Caveat: 1. GAE Java 7 environment is not supported for now. 2. This manager only manages
+ * OpenCensus related shutdown hooks. User can still use other methods to register a shutdown hook,
+ * e.g. the native {@link Runtime#addShutdownHook}, or the Hadoop's {@code ShutdownHookManager}. The
+ * execution order of registered {@link Hook}s and hooks registered by other systems depends on the
+ * JVM implementation and thus can not be guaranteed, which might sometimes leads to unwanted
+ * behaviors. An example is that {@code java.util.logging.Logger} has its own shutdown hook to reset
+ * all handlers, and once invoked, it will prevent all future logging in other shutdown hooks.
  */
 @ThreadSafe
 public final class ShutdownHookManager {
--- a/impl_core/src/main/java/io/opencensus/implcore/internal/ShutdownHookManager.java
+++ b/impl_core/src/main/java/io/opencensus/implcore/internal/ShutdownHookManager.java
@@ -30,7 +30,7 @@ import javax.annotation.concurrent.GuardedBy;
 import javax.annotation.concurrent.ThreadSafe;
 
 /**
- * An utility class to help with registering shutdown {@link Hook}s to the runtime.
+ * A utility class to help with registering shutdown {@link Hook}s to the runtime.
  *
  * <p>A {@link Hook} can have its own name and priority. The execution order among registered hooks
  * will be determined by their priorities first (descending order) and then registration time
@@ -54,7 +54,7 @@ public final class ShutdownHookManager {
 
   private final Object monitor = new Object();
 
-  // A list to store registered {@link Hook}s.
+  // A list to store registered hooks.
   @GuardedBy("monitor")
   private final List<Hook> hooks = new ArrayList<Hook>();
 
@@ -194,7 +194,7 @@ public final class ShutdownHookManager {
    * @param hook the {@code Hook}
    */
   public void addShutdownHook(Hook hook) {
-    checkNotNull(hook);
+    checkNotNull(hook, "hook");
     synchronized (monitor) {
       if (!isShuttingDown) {
         hooks.add(hook);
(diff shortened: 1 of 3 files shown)


impl_core/src/main/java/io/opencensus/implcore/internal/ShutdownHookManager.java
```diff
@@ -0,0 +1,215 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.GuardedBy;
+import javax.annotation.concurrent.ThreadSafe;
+
+/**
+ * An utility class to help with registering shutdown {@link Hook}s to the runtime.
```
songy23: nit: s/An/A

impl_core/src/main/java/io/opencensus/implcore/internal/ShutdownHookManager.java
```diff
@@ -0,0 +1,215 @@
...
+public final class ShutdownHookManager {
+
+  private static final Logger logger = Logger.getLogger(ShutdownHookManager.class.getName());
+
+  // The singlelon instance.
+  private static final ShutdownHookManager instance = new ShutdownHookManager();
+
+  private final Object monitor = new Object();
+
+  // A list to store registered {@link Hook}s.
```
songy23: nit: `@link` only works for javadoc.

impl_core/src/main/java/io/opencensus/implcore/internal/ShutdownHookManager.java
```diff
@@ -0,0 +1,215 @@
...
+ * java.util.logging.Logger} has its own shutdown hook to reset all handlers, and once invoked, it
+ * will prevent all future logging in other shutdown hooks.
+ */
+@ThreadSafe
+public final class ShutdownHookManager {
+
+  private static final Logger logger = Logger.getLogger(ShutdownHookManager.class.getName());
+
+  // The singlelon instance.
+  private static final ShutdownHookManager instance = new ShutdownHookManager();
```
songy23: Should this be guarded by `monitor` as well?

impl_core/src/main/java/io/opencensus/implcore/internal/ShutdownHookManager.java
```diff
@@ -0,0 +1,215 @@
...
+    }
+  }
+
+  /**
+   * Adds a {@link Hook} to the manager.
+   *
+   * @param hook the {@code Hook}
+   */
+  public void addShutdownHook(Hook hook) {
+    checkNotNull(hook);
```
songy23: nit: checkNotNull(hook, "hook");

impl_core/src/main/java/io/opencensus/implcore/internal/ShutdownHookManager.java
HailongWen: Yes. Thanks for pointing out that.

HailongWen: On a second thought, in this case the instance is initialized when the class is loaded so I think it should be fine here.

adriancole:
no offense, but this looks like code that might have been copy/pasted from somewhere else (notably all the edge cases). If that was the case and it was open source, maybe mention from where it is influenced?

HailongWen:
/cc @adriancole , sure, none taken. :)

The idea of registering shutdownhooks with priorities is influenced by the ones used in hadoop ecosystem, e.g. [hadoop](https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/ShutdownHookManager.java), [spark](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/util/ShutdownHookManager.scala) and [hive](https://github.com/apache/hive/blob/master/common/src/java/org/apache/hive/common/util/ShutdownHookManager.java) (maybe there are more), without knowing the implementation details. So the credit goes to the projects and I should have mentioned it in the javadoc.

I did read the code of native `Runtime.addShutdownHook()`. Knowing that the order is determined by the hash of the hook itself in the implementation, I decided to use a single thread to execute hooks one by one to at least ensure the order of our own registered ones, so as to have a deterministic behavior. Turns out it is the same strategy already been implemented by [hadoop](https://github.com/apache/hadoop/blob/99ab511cbac570bea9d31a55898b95590a8e3159/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/ShutdownHookManager.java#L55), plus it has a more thorough solution (using threadpool, enforcing timeout).

I admit re-inventing something like this is not a good idea. Do you have suggestions on what opensource libraries we could rely on?

cwensel:
My only suggestion would be to consider:
 
- making the shutdown hook registration optional (#registerFlushShutdownHook())
- exposing the #flush() safely/publicly

this way other parties can either control ordering of the registration, or use their own shutdown hook calls.

I recognize #flush() can be abused, but ordering can be hard. to use the Hadoop example, HDFS client magically registers its own hook, good news the instance is kept in a static private field allowing the caller to reflectively grab it, un-register it, and re-register it last.

HailongWen:
@cwensel Thanks for the suggestion! The suggestion makes sense if we expose the util for third party to use, which is not what I intended to.

Currently this util is internal for OpenCensus implementation itself only, visible for contributors of OpenCensus `impl_core`/`impl`/`impl_lite`. It controls the behaviors of shutdown hooks in OpenCensus, execute them in a deterministic way. Components that may use it are: `SpanExporterImpl`, StatsExporter if https://github.com/census-instrumentation/opencensus-java/issues/862 is implemented, and possibly for tracer to close all running span and set proper status (this should happen before exporting).

I didn't add it in the public API because it is really not related to the stats/traces operations. And just as mentioned, other libraries have their own shutdown hook management and I just don't want to mix with them and make things messy.

cwensel:
One could argue this is related to the trace api. 

First, OpenCensus provides a static initializer to start (#start()) it. In the name of symmetry, there is no API call to shut it down gracefully (no #flush() or #stop() that i’ve seen).

Second, complex systems may need to order the shutdowns. I’m not asking you expose the manager, I’m asking that it (again in the name of symmetry) the user have a choice as to when then shutdown hook is registered (as to make it last?, first?), or have the option to not register the OpenCensus shutdown hook (as it behaves badly with the current app) and let the user manage the shutdown on their own (via their own hook manager).

Case in point, I have a batch app I want to trace, the jvm lifespan is the root span. I want to guarantee the last thing to happen the the trace shutdown. So I want to manage the shutdown hook to guarantee that.

I’m happy to provide a PR at a later date. These are just suggestions to consider.

HailongWen:
Thanks @cwensel, your case is a good one, and I think this topic touches the lifecycle management of OpenCensus and probably needs more discussion. /cc @bogdandrutu

cwensel:
fwiw, I would like to see this patch merged, and yes @HailongWen on further discussion.

cwensel:
So I went ahead and gave this patch a twirl, and I am definitely seeing race conditions with this hook and my hooks.

If I could disable/not-register the open-census shutdown hook, and use my own shutdown api calls to order the flush, things would be very fine.

HailongWen:
Hi @cwensel, given the jvm case you mentioned above, is the race conditions refer to something like 'exporters flush before rootSpan ends'?

cwensel:
Yes, exactly.

fyi, using the log extractor also conflates things as it _looks like_ the logger handlers are removed before the worker#flush() as well - so no logging of spans to console. 

but I am definitely having worker#flush() hit before my shutdown hook gets to ending the root span.

HailongWen:
Thanks.

For `java.util.logging.Logger`, it is a known issue and I've already documented it in the javadoc. There does exist a work around (overriding `LogManager` to take control of the shutdown) but I guess it is better to resolve it by switching to other logging systems.

For your case, I think we should add a public API `SpanExporter#flush()` for users to call, and they should decide when and where to call it (i.e., I am leaning to revert the `ShutdownHookManager` change). This might seem a better way to resolve the issue.

HailongWen:
Close this for now. I'll prepare another PR to expose the `flush()`.

PR #1051, opened 2018-03-06
Dependency update

songy23:


songy23:
Upgrade to Prometheus v0.3.0 doesn't work with JDK 8 and 9 build - holding off upgrading Prometheus for now.

songy23:
Thanks!

PR #1052, opened 2018-03-06
Add bazel build for gRPC Hello World example.

songy23:
As a follow up for PR https://github.com/census-instrumentation/opencensus-java/pull/1044. This PR depends on https://github.com/census-instrumentation/opencensus-java/pull/1044.

It's nice to have all 3 build systems supported for the gRPC example - however, given the significant amount of new dependencies (and complexity) it introduced to BUILD.bazel file (mostly netty runtime dependencies, which cannot be resolved by the [migration-tool](https://github.com/bazelbuild/migration-tooling)), I'm not sure if it's a good idea to support this example with Bazel.


songy23:
@HailongWen not that I knew of. All the netty and google cloud dependencies are transitive dependencies and are auto-imported to opencensus_workspace.bzl as native maven jars. Once they are loaded into WORKSPACE, they cannot be defined and need to be used as-is in BUILD.bazel.

Defining a separate gRPC build target is possible, but that won't simplify BUILD a lot, since 
`opencensus_examples` and the gRPC example share many common deps.

bogdandrutu:
One option is to split examples into multiple projects:
examples/helloworld/...
examples/quickstart/...
etc.

songy23:
Yes, the examples are already split to different packages and build targets. We can also split them into different BUILD files, but that will also introduce a lot of duplication, since all these targets share a lot of common dependencies.

RELEASING.md
```diff
@@ -257,6 +257,13 @@ $ sed -i 's/\(<version>\)[0-9]\+\.[0-9]\+\.[0-9]\+/\1'$MAJOR.$MINOR.$PATCH'/' \
     --artifact=io.opencensus:opencensus-contrib-zpages:$MAJOR.$MINOR.$PATCH \
     --artifact=io.opencensus:opencensus-exporter-trace-logging:$MAJOR.$MINOR.$PATCH \
     --artifact=io.opencensus:opencensus-impl:$MAJOR.$MINOR.$PATCH \
+    --artifact=io.opencensus:opencensus-contrib-grpc-metrics:$MAJOR.$MINOR.$PATCH \
+    --artifact=io.opencensus:opencensus-exporter-trace-stackdriver:$MAJOR.$MINOR.$PATCH \
+    --artifact=io.opencensus:opencensus-exporter-stats-stackdriver:$MAJOR.$MINOR.$PATCH \
+    --artifact=io.opencensus:opencensus-exporter-stats-prometheus:$MAJOR.$MINOR.$PATCH \
```
sebright: Suggestion: Sorting the artifacts might help readability.

examples/BUILD.bazel
```diff
@@ -67,4 +137,22 @@ java_binary(
     runtime_deps = [
         ":opencensus_examples",
     ],
+)
+
+java_binary(
+    name = "HelloWorldClient",
+    testonly = 1,
```
sebright: Why are the new build targets `testonly`?

RELEASING.md
songy23: Done.

examples/BUILD.bazel
songy23: Done. I copied this from gRPC java build: https://github.com/grpc/grpc-java/blob/master/examples/BUILD.bazel#L85. Not sure why they specify this as test-only though.

songy23:
Merging for now. As I can see, the BUILD file for grpc-java example also needs to explicitly depend on a lot of deps: https://github.com/grpc/grpc-java/blob/master/examples/BUILD.bazel#L62. Our example needs even more dependencies because of exporting to Stackdriver (and thus tremendous transitive dependencies on Google Cloud client libraries).

Will update the BUILD file if I found a better way to manage the dependencies.

Issue #740, opened 2017-10-29
Update README.md with tags/stats examples similar with tracing.


bogdandrutu:
Add a quickstart section for tags/stats on the main readme.

PR #1053, opened 2018-03-06
Add quick start example to README

songy23:
Fixes https://github.com/census-instrumentation/opencensus-java/issues/740.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,8 @@ the same way.
 
 ```java
 public final class MyClassWithTracing {
+  private static final Tracer tracer = Tracing.getTracer();
+
   public static void doWork() {
     // Create a child Span of the current Span.
     try (Scope ss = tracer.spanBuilder("MyChildWorkSpan").startScopedSpan()) {
@@ -85,6 +87,9 @@ For the complete example, see
 
 ```java
 public final class QuickStart {
+  private static final Tagger tagger = Tags.getTagger();
+  private static final ViewManager viewManager = Stats.getViewManager();
+  private static final StatsRecorder statsRecorder = Stats.getStatsRecorder();  
 
   // frontendKey allows us to break down the recorded data
   private static final TagKey FRONTEND_KEY = TagKey.create("my.org/keys/frontend");
--- a/README.md
+++ b/README.md
@@ -15,9 +15,11 @@ The library is in alpha stage and the API is subject to change.
 Please join [gitter](https://gitter.im/census-instrumentation/Lobby) for help or feedback on this
 project.
 
-## Instrumentation Quickstart
+## OpenCensus Quickstart
 
 Integrating OpenCensus with a new library means recording stats or traces and propagating context.
+
+TODO: add a link to Java quick start documentation on opencensus.io once it's ready.
 
 ### Add the dependencies to your project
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -70,7 +72,45 @@ public final class MyClassWithTracing {
 
 ### Hello "OpenCensus" stats events
 
-TODO
+Here's an example on defining TagKey, Measure and View, registering view, recording stats and get
+ViewData. For the complete example, see
```
sebright: Suggestion: It might be clearer to put the steps in a bulleted list and avoid two levels of commas.

README.md
```diff
@@ -70,7 +72,45 @@ public final class MyClassWithTracing {
...
+      VIDEO_SIZE_VIEW_NAME,
+      "processed video size over time",
+      VIDEO_SIZE,
+      Aggregation.Distribution.create(BucketBoundaries.create(Arrays.asList(0.0, 256.0, 65536.0))),
+      Collections.singletonList(FRONTEND_KEY),
+      Cumulative.create());
+
+  public static void main(String[] args) throws InterruptedException {
+    TagContextBuilder tagContextBuilder =
+      tagger.currentBuilder().put(FRONTEND_KEY, TagValue.create("mobile-ios9.3.5"));
```
sebright: Suggestion: Creating the builder within the try-with-resources could help show that the builder is normally only used once, on the line that creates the TagContext.

README.md
```diff
@@ -70,7 +72,45 @@ public final class MyClassWithTracing {
...
+    viewManager.registerView(VIDEO_SIZE_VIEW);
+    // Process video.
+    // Record the processed video size.
+    try (Scope scopedTags = tagContextBuilder.buildScoped()) {
+      statsRecorder.newMeasureMap().put(VIDEO_SIZE, 25648).record();
+    }
+    ViewData viewData = viewManager.getView(VIDEO_SIZE_VIEW_NAME);
```
sebright: Suggestion: This line could print the ViewData.

README.md
```diff
@@ -70,7 +72,45 @@ public final class MyClassWithTracing {
...
+      Cumulative.create());
+
+  public static void main(String[] args) throws InterruptedException {
+    TagContextBuilder tagContextBuilder =
+      tagger.currentBuilder().put(FRONTEND_KEY, TagValue.create("mobile-ios9.3.5"));
+    viewManager.registerView(VIDEO_SIZE_VIEW);
+    // Process video.
+    // Record the processed video size.
+    try (Scope scopedTags = tagContextBuilder.buildScoped()) {
+      statsRecorder.newMeasureMap().put(VIDEO_SIZE, 25648).record();
```
sebright: Should there be a method call after this line, such as "processVideo()"?

README.md
```diff
@@ -70,7 +72,45 @@ public final class MyClassWithTracing {
...
+      Collections.singletonList(FRONTEND_KEY),
+      Cumulative.create());
+
+  public static void main(String[] args) throws InterruptedException {
+    TagContextBuilder tagContextBuilder =
+      tagger.currentBuilder().put(FRONTEND_KEY, TagValue.create("mobile-ios9.3.5"));
+    viewManager.registerView(VIDEO_SIZE_VIEW);
+    // Process video.
+    // Record the processed video size.
+    try (Scope scopedTags = tagContextBuilder.buildScoped()) {
```
sebright: I think it would be clearer to move the code that deals with one video to a separate method, which could be called on each request, similar to the tracing example.  It would also help distinguish between code used for initialization and stats recording.

README.md
songy23: Done.

README.md
songy23: Done.

README.md
songy23: Make sense, I split this block of code to a separate method `processVideo()`.

README.md
songy23: Moved to `processVideo()`.

README.md
songy23: Done.

README.md
```diff
@@ -98,16 +105,26 @@ public final class QuickStart {
...
+      // ...
+
+      // Record the processed video size.
       statsRecorder.newMeasureMap().put(VIDEO_SIZE, 25648).record();
     }
+  }
+
+  public static void main(String[] args) throws InterruptedException {
+    viewManager.registerView(VIDEO_SIZE_VIEW);
+    processVideo();
```
sebright: I still think that calling processVideo() directly from main() is slightly confusing, because it looks like processVideo() is only ever called once, instead of once per request.  It would help to either explain that it is a simplification or separate it from main, for example,

- Add an intermediate method call here, such as "handleRequests()", which is assumed to call processVideo, or
- Create three independent methods, such as "processVideo", "initialize" (registers view), and "printStats" (calls getView).

songy23: Calling from main() here is kind of a simplification. I updated the example with the 2nd option, hopefully it makes it more clear that main() is like a simulated job.

README.md
```diff
@@ -70,7 +72,71 @@ public final class MyClassWithTracing {
...
+    }
+  }
+
+  private static void printStats() {
+    ViewData viewData = viewManager.getView(VIDEO_SIZE_VIEW_NAME);
+    System.out.println(
+      String.format("Recorded stats for %s:\n %s", VIDEO_SIZE_VIEW_NAME.asString(), viewData));
+  }
+
+  public static void main(String[] args) {
```
HailongWen: I agree with @sebright comment that the `main()` somehow implies the program just run once.
Besides, IIUC, this snippet is just for highlighting the usage of APIs of the full example. I think it is not necessary to add the `main()` to make it executable.

songy23: Make sense to me. I noticed the tracing code example also did not have main(), so I just removed main() from this code snippet.

README.md
```diff
@@ -70,7 +72,65 @@ public final class MyClassWithTracing {
...
+      Cumulative.create());
+
+  private static void initialize() {
+    // ...
+    viewManager.registerView(VIDEO_SIZE_VIEW);
+  }
+
+  private static void processVideo() {
+    try (Scope scopedTags =
+           tagger
```
sebright: I just realize that `viewManager` and `tagger` aren't declared in the example.

songy23: Done. Same for tracer.

songy23:
Thanks!

PR #1058, opened 2018-03-15
Change MBy to By units.

g-easy:
This reflects the default gRPC views which use bytes as units.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,14 +89,14 @@ For the complete example, see
 public final class QuickStart {
   private static final Tagger tagger = Tags.getTagger();
   private static final ViewManager viewManager = Stats.getViewManager();
-  private static final StatsRecorder statsRecorder = Stats.getStatsRecorder();  
+  private static final StatsRecorder statsRecorder = Stats.getStatsRecorder();
 
   // frontendKey allows us to break down the recorded data
   private static final TagKey FRONTEND_KEY = TagKey.create("my.org/keys/frontend");
 
   // videoSize will measure the size of processed videos.
   private static final MeasureLong VIDEO_SIZE = MeasureLong.create(
-      "my.org/measure/video_size", "size of processed videos", "MBy");
+      "my.org/measure/video_size", "size of processed videos", "By");
 
   // Create view to see the processed video size distribution broken down by frontend.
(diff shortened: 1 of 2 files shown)


examples/src/main/java/io/opencensus/examples/helloworld/QuickStart.java
```diff
@@ -52,12 +52,12 @@
...
   private static final Tracer tracer = Tracing.getTracer();
 
-  // frontendKey allows us to break down the recorded data
+  // frontendKey allows us to break down the recorded data.
   private static final TagKey FRONTEND_KEY = TagKey.create("my.org/keys/frontend");
 
   // videoSize will measure the size of processed videos.
   private static final MeasureLong VIDEO_SIZE = MeasureLong.create(
-      "my.org/measure/video_size", "size of processed videos", "MBy");
+      "my.org/measure/video_size", "size of processed videos", "By");
```
g-easy: Not sure if it's worth changing the video size on L92 or the buckets on L70.

songy23: I think it makes more sense to change the video size.

g-easy: Changed to MiB range. How does this look?

README.md
```diff
@@ -89,14 +89,14 @@ For the complete example, see
 public final class QuickStart {
   private static final Tagger tagger = Tags.getTagger();
   private static final ViewManager viewManager = Stats.getViewManager();
-  private static final StatsRecorder statsRecorder = Stats.getStatsRecorder();  
+  private static final StatsRecorder statsRecorder = Stats.getStatsRecorder();
```
g-easy: Instead of duplicating the code in the README, can we just add a hyperlink to the *.java file?

songy23: The code example in README is slightly different from QuickStart.java though. QuickStart.java should be made executable so it has a main(), while README is more like a real class whose methods could be invoked multiple times.

g-easy: Ok, I'll leave it as-is.

g-easy:
After all the expensive builds finish, I get dinged on my git history. :^)

sebright:
I made a PR that moves the git history check earlier: #1062

PR #1059, opened 2018-03-15
Make "./gradlew goJF" work in examples/.

g-easy:


sebright:
I think that we should try to keep the example as simple as possible and only include the plugins necessary to build and run it.  It might not be clear to users which parts of the build file are only needed for formatting.  I think it would be better to enforce format from outside of the examples directory, for example, if we could add the check to the top level build.gradle.  (It would also be nice to move the license check out of examples.)

Issue #188, opened 2017-03-31
Split RpcConstants into two files.


zhangkun83:
gRPC core only uses the MeasurementDescriptors, not the ViewDescriptors. Having all of them in RpcConstants forces JVM to load all related classes when loading RpcConstants, even if they are not used. Android is sensitive to class loading time. We'd better put the ViewDescriptors in a separate file.

zhangkun83:
cc @songy23

songy23:
Needs to remove deprecated RpcConstants.java and RpcConstantsTest.java once no code depends on them.

songy23:
Fixed in https://github.com/census-instrumentation/opencensus-java/pull/389

PR #1060, opened 2018-03-15
Add utility class to expose default HTTP measures, tags and view, and register default views.

HailongWen:
Same as the grpc-metrics. The only difference is that we expose HTTP measures, tags and views for 3rd-party libraries to access, so they can reuse existing ones.

The spec for this PR is in https://github.com/census-instrumentation/opencensus-specs/blob/master/http/stats.md. And I referenced [Go's implementation](https://github.com/census-instrumentation/opencensus-go/pull/534) for most of the description.


contrib/http_util/src/main/java/io/opencensus/contrib/http/util/HttpStatsUtil.java
```diff
@@ -0,0 +1,435 @@
...
+   * @since 0.13
+   */
+  public static final MeasureLong HTTP_CLIENT_RESPONSE_BYTES =
+      Measure.MeasureLong.create(
+          getName("client", "response_bytes"),
+          "Client-side HTTP response body size (uncompressed)",
+          UNIT_SIZE);
+
+  /**
+   * {@link Measure} for the client-side end-to-end latency.
```
songy23: nit: s/end-to-end/roundtrip? (consistent with gRPC metric)

contrib/http_util/src/main/java/io/opencensus/contrib/http/util/HttpStatsUtil.java
```diff
@@ -0,0 +1,435 @@
...
+   * @since 0.13
+   */
+  public static final MeasureLong HTTP_SERVER_RESPONSE_BYTES =
+      Measure.MeasureLong.create(
+          getName("server", "response_bytes"),
+          "Server-side HTTP response body size (uncompressed)",
+          UNIT_SIZE);
+
+  /**
+   * {@link Measure} for the server-side end-to-end latency.
```
songy23: Does "end-to-end" apply to server?

contrib/http_util/src/test/java/io/opencensus/contrib/http/util/HttpStatsUtilTest.java
```diff
@@ -0,0 +1,143 @@
...
+    assertThat(HttpStatsUtil.getName("client", "request_count"))
+        .isEqualTo("opencensus.io/http/client/request_count");
+    assertThat(HttpStatsUtil.getName("server", "request_bytes"))
+        .isEqualTo("opencensus.io/http/server/request_bytes");
+  }
+
+  @Test
+  public void constants() {
+    // Test Tags
+    assertThat(HttpStatsUtil.HTTP_STATUS_CODE).isNotNull();
```
songy23: nit: may be worth verifying their values.

contrib/http_util/src/main/java/io/opencensus/contrib/http/util/HttpStatsUtil.java
```diff
@@ -0,0 +1,435 @@
...
+   * @since 0.13
+   */
+  public static final TagKey HTTP_METHOD = TagKey.create("http.method");
+
+  /**
+   * {@link View} for count of client-side HTTP requests started.
+   *
+   * @since 0.13
+   */
+  public static final View HTTP_CLIENT_REQUEST_COUNT_VIEW =
```
songy23: Consider splitting HTTP measures and views into two files, since it's likely that other libraries only need to load and record against HTTP measures.

Similar issue to gRPC constants: https://github.com/census-instrumentation/opencensus-java/issues/188.

contrib/http_util/src/main/java/io/opencensus/contrib/http/util/HttpStatsUtil.java
HailongWen: Thanks! Maybe not. The specs for client latency is "Time between first byte of request headers sent to last byte of response received, or terminal error", so symmetrically the server latency would be something like "Time between first byte of request headers *received* to last byte of response *sent*, or terminal error". So compared to the client one, the server latency does not include time spend on receiving/sending data from/to client. So it is not a true e2e.

contrib/http_util/src/main/java/io/opencensus/contrib/http/util/HttpStatsUtil.java
HailongWen: SG to me.

contrib/http_util/src/main/java/io/opencensus/contrib/http/util/HttpStatsUtil.java
songy23: Yes that's also what I was thinking - maybe we can just remove the word "end-to-end" here.

contrib/http_util/src/test/java/io/opencensus/contrib/http/util/HttpViewConstantsTest.java
```diff
@@ -0,0 +1,168 @@
...
+            ((Distribution) HttpViewConstants.LATENCY_DISTRIBUTION)
+                .getBucketBoundaries()
+                .getBoundaries())
+        .containsExactly(
+            0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 8.0, 10.0, 13.0, 16.0, 20.0, 25.0, 30.0, 40.0, 50.0,
+            65.0, 80.0, 100.0, 130.0, 160.0, 200.0, 250.0, 300.0, 400.0, 500.0, 650.0, 800.0,
+            1000.0, 2000.0, 5000.0, 10000.0, 20000.0, 50000.0, 100000.0)
+        .inOrder();
+
+    // Test view descriptors.
```
songy23: nit:s/view descriptors/views

contrib/http_util/src/test/java/io/opencensus/contrib/http/util/HttpMeasureConstantsTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+  @Test
+  public void constants() {
+    // Test Tags
+    assertThat(HttpMeasureConstants.HTTP_STATUS_CODE).isEqualTo(TagKey.create("http.status"));
+    assertThat(HttpMeasureConstants.HTTP_METHOD).isEqualTo(TagKey.create("http.method"));
+    assertThat(HttpMeasureConstants.HTTP_PATH).isEqualTo(TagKey.create("http.path"));
+    assertThat(HttpMeasureConstants.HTTP_HOST).isEqualTo(TagKey.create("http.host"));
+
+    // Test measurement descriptors.
```
songy23: nit: s/measurement descriptors/measures

contrib/http_util/src/test/java/io/opencensus/contrib/http/util/HttpMeasureConstantsTest.java
```diff
@@ -0,0 +1,65 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/** Test for {@link HttpMeasureConstants}. */
+@RunWith(JUnit4.class)
+public class HttpMeasureConstantsTest {
+
+  @Test
+  public void constants() {
+    // Test Tags
```
songy23: nit: s/Tags/TagKeys

contrib/http_util/src/main/java/io/opencensus/contrib/http/util/HttpMeasureConstants.java
```diff
@@ -0,0 +1,162 @@
...
+  private static final String UNIT_COUNT = "1";
+  private static final String UNIT_SIZE = "By";
+  private static final String UNIT_LATENCY = "ms";
+
+  /**
+   * {@link Measure} for the number of client-side HTTP requests started.
+   *
+   * @since 0.13
+   */
+  public static final MeasureLong HTTP_CLIENT_REQUEST_COUNT =
```
bogdandrutu: Why is this needed? If this is recorded at the beginning of the request. Maybe rename it to started_count to be in sync with grpc?

contrib/http_util/src/main/java/io/opencensus/contrib/http/util/HttpMeasureConstants.java
```diff
@@ -0,0 +1,162 @@
...
+          UNIT_LATENCY);
+
+  /**
+   * {@link Measure} for the number of server-side HTTP requests started.
+   *
+   * @since 0.13
+   */
+  public static final MeasureLong HTTP_SERVER_REQUEST_COUNT =
+      Measure.MeasureLong.create(
+          "opencensus.io/http/server/request_count",
```
bogdandrutu: Same comment here.

contrib/http_util/src/main/java/io/opencensus/contrib/http/util/HttpViewConstants.java
```diff
@@ -0,0 +1,241 @@
...
+   *
+   * @since 0.13
+   */
+  public static final View HTTP_CLIENT_REQUEST_BYTES_VIEW =
+      View.create(
+          View.Name.create("opencensus.io/http/client/request_bytes"),
+          "Size distribution of client-side HTTP request body",
+          HTTP_CLIENT_REQUEST_BYTES,
+          SIZE_DISTRIBUTION,
+          Collections.<TagKey>emptyList());
```
bogdandrutu: I would have here:
http.method
http.path

Also in all bytes + latency views.

sebright:
I noticed that these classes are in the http_util library, but the corresponding gRPC metrics are in grpc_metrics.  Does it make sense to put this in a new http_metrics library, for consistency?  grpc_utils and grpc_metrics are separate libraries because grpc_utils has a dependency on gRPC.  Is there any chance that http_utils will also require more heavyweight dependencies in the future?

contrib/http_util/src/main/java/io/opencensus/contrib/http/util/HttpMeasureConstants.java
```diff
@@ -0,0 +1,162 @@
...
+
+/**
+ * A helper class which holds OpenCensus's default HTTP {@link Measure}s and {@link TagKey}s.
+ *
+ * <p>{@link Measure}s and {@link TagKey}s in this class are all public for other
+ * libraries/frameworks to reference and use.
+ *
+ * @since 0.13
+ */
+public class HttpMeasureConstants {
```
sebright: The three non-test classes could be final.

contrib/http_util/src/main/java/io/opencensus/contrib/http/util/HttpMeasureConstants.java
```diff
@@ -0,0 +1,162 @@
...
+  /**
+   * {@link TagKey} for the value of the HTTP host header.
+   *
+   * @since 0.13
+   */
+  public static final TagKey HTTP_HOST = TagKey.create("http.host");
+
+  /**
+   * {@link TagKey} for the numeric HTTP response status code. If a transport error occurred and no
+   * status code was read, use "error" as the {@code TagValue}.
```
sebright: Should we also provide the error tag value?

contrib/http_util/src/main/java/io/opencensus/contrib/http/util/HttpMeasureConstants.java
```diff
@@ -0,0 +1,162 @@
...
+ *
+ * @since 0.13
+ */
+public class HttpMeasureConstants {
+
+  private HttpMeasureConstants() {}
+
+  private static final String UNIT_COUNT = "1";
+  private static final String UNIT_SIZE = "By";
+  private static final String UNIT_LATENCY = "ms";
```
sebright: Suggestion: It might be clearer to include the units in the constant names, for example, UNIT_SIZE_BYTE or 
UNIT_LATENCY_MS.

HailongWen:
Thanks @sebright for pointing out that.
Initially I didn't put this into `http_metrics` like the grpc one because I noticed that `grpc_metrics` is claimed to be moved into grpc in the future. For HTTP there is no such library to move into.

I am totally fine with moving it into a new artifact.

contrib/http_util/src/main/java/io/opencensus/contrib/http/util/HttpMeasureConstants.java
```diff
@@ -0,0 +1,162 @@
...
+   * {@link Measure} for the client-side HTTP request body size if set as ContentLength
+   * (uncompressed).
+   *
+   * @since 0.13
+   */
+  public static final MeasureLong HTTP_CLIENT_REQUEST_BYTES =
+      Measure.MeasureLong.create(
+          "opencensus.io/http/client/request_bytes",
+          "Client-side HTTP request body size if set as ContentLength (uncompressed)",
+          UNIT_SIZE);
```
sebright: Should we specify that the size is uncompressed in the measure and constant names?  I'm not suggesting making a change in this PR, though, since it's from opencensus-specs.

sebright:
> Thanks @sebright for pointing out that.
Initially I didn't put this into http_metrics like the grpc one because I noticed that grpc_metrics is claimed to be moved into grpc in the future. For HTTP there is no such library to move into.
> 
> I am totally fine with moving it into a new artifact.

@HailongWen I forgot that grpc_metrics is moving into gRPC, so I don't think that we need to split http_utils for consistency.  I only think we should split it if there could be a large difference in their dependencies or purpose, similar to the difference between grpc_utils and grpc_metrics.

contrib/http_util/src/main/java/io/opencensus/contrib/http/util/HttpMeasureConstants.java
bogdandrutu: Please use the new tags/views defined:
https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/HTTP.md

HailongWen: Will do.

HailongWen:
Thanks!

PR #1061, opened 2018-03-15
Improve StatszZPageHandler.TreeNode

songy23:
Update TreeNode to store view names instead of the whole view, since for getting stats all we need is view name. ~~Also use AutoValue to implement TreeNode.~~

Hopefully this could reduce the memory usage of StatsZ a little bit, especially when there are a huge amount of exported views.

### DIFF
--- a/contrib/zpages/src/main/java/io/opencensus/contrib/zpages/StatszZPageHandler.java
+++ b/contrib/zpages/src/main/java/io/opencensus/contrib/zpages/StatszZPageHandler.java
@@ -160,9 +160,9 @@ final class StatszZPageHandler extends ZPageHandler {
       String path = queryMap.get(QUERY_PATH);
       TreeNode current = findNode(path);
       emitDirectoryTable(current, path, out, formatter);
-      if (current != null && current.view != null) {
-        ViewData viewData = viewManager.getView(current.view.getName());
-        emitViewData(viewData, current.view.getName(), out, formatter);
+      if (current != null && current.viewName != null) {
+        ViewData viewData = viewManager.getView(current.viewName);
+        emitViewData(viewData, current.viewName, out, formatter);
       }
       emitMeasureTable(measures, out, formatter);
     }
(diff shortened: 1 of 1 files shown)


contrib/zpages/src/main/java/io/opencensus/contrib/zpages/StatszZPageHandler.java
```diff
@@ -554,22 +559,31 @@ private StatszZPageHandler(ViewManager viewManager) {
    * directories, and leaf nodes are the ones with actual information on views. Each non-leaf node
    * also has the number of views under its directory.
    */
-  private static class TreeNode {
+  @AutoValue
+  // Suppress Checker Framework warning about missing @Nullable in generated equals method.
+  @AutoValue.CopyAnnotations
+  @SuppressWarnings("nullness")
+  abstract static class TreeNode {
```
sebright: Is TreeNode immutable?  It looks like `children` isn't populated in the factory method.

songy23: No it cannot be immutable, since we can register views after StatsZ is initialized, and that will update all the parent nodes of the newly registered view. (Specifically, root will be updated every time when there's a new view.)

songy23: Hmmm in this case I shouldn't use AutoValue for TreeNode... I've reverted the last 2 commits.

sebright:
> > I thought that statz only referenced views that were already registered in the ViewManager.
> 
> That's right, and statsz only needs to know the names of registered views.

I meant that whether statsz stores the views or view names shouldn't affect memory usage, since the full views are already stored in the ViewManager.  But I think that this change simplifies the code by only referencing the part of the view that is needed.

songy23:
Thanks!

PR #1062, opened 2018-03-15
Move Travis jobs for example projects to the end of the build matrix.

sebright:
This change allows build jobs that are affected by more pull requests to run
first, to increase the probability of failures appearing in the first set of
concurrent job runs.


sebright:
Thanks!

PR #1063, opened 2018-03-16
Fix a failing test with Jaeger exporter

songy23:
Looks like the length of Jaeger span id is 15 with JDK7 and 16 with JDK8.

/cc @marccarre


HailongWen:
The spanId is generated by OpenCensus and exported to jaeger, and then the integration test will fetch it from jaeger (encoded in json). I think the spanId's length might vary. 

Better check the jaeger implementation.

songy23:
SGTM, agree that the leading or trailing 0s are probably to blame.

marccarre:
Indeed, this looks like leading zeros are to blame.

Note that if the trace/span ID is `0L`, then Jaeger simply returns `"0"`.
Behaviour is similar for any small `long` value for the trace/span IDs (e.g. `1L` -> `"1"`).

Looking at `SpanId#generateRandomId`/`SpanId#INVALID` and `TraceId#generateRandomId`/`TraceId#INVALID`, it seems `0L` is a forbidden value which is actively filtered out, but anything else has the potential to happen.

If we want to make the tests even more robust to such extreme (but possible) values, we could change the regexes to be `{1,32}` and `{1,16}` respectively. WDYT @songy23 and @HailongWen?

songy23:
@marccarre That sounds reasonable to me. IIUC it's possible that we generate `TraceId` or `SpanId` with just a few digits, and having a wilder range for the regexes should make the tests more stable.

/cc @bogdandrutu

PR #1067, opened 2018-03-16
Disable Jaeger integration test by default. (closes #1066)

### LINKED ISSUE #1066: JaegerExporterHandlerIntegrationTest is fluctuant in continuous build
For example https://ci.appveyor.com/project/opencensusjavateam/opencensus-java/build/1.0.2174.

/cc @marccarre 

> io.opencensus.exporter.trace.jaeger.JaegerExporterHandlerIntegrationTest > exportToJaeger FAILED
    org.junit.runners.model.TestTimedOutException: test timed out after 30000 milliseconds
        at java.io.WinNTFileSystem.getBooleanAttributes(Native Method)
        at java.io.File.isFile(File.java:882)
        at io.opencensus.exporter.trace.jaeger.JaegerExporterHandlerIntegrationTest.isDocker(JaegerExporterHandlerIntegrationTest.java:225)
        at io.opencensus.exporter.trace.jaeger.JaegerExporterHandlerIntegrationTest.dockerIsInstalledAndExecutable(JaegerExporterHandlerIntegrationTest.java:215)
        at io.opencensus.exporter.trace.jaeger.JaegerExporterHandlerIntegrationTest.exportToJaeger(JaegerExporterHandlerIntegrationTest.java:70)
3 tests completed, 1 failed
:opencensus-exporter-trace-jaeger:test FAILED

HailongWen:
The integration test will try to lookup "docker" in the "PATH" and this operation might timeout and fail the windows build.

This PR add a system property "enable.jaeger.integration.test" to make it option.
Users can use "gradlew -Denable.jaeger.integration.test=true" to execute the integration test.

### DIFF
--- a/exporters/trace/jaeger/build.gradle
+++ b/exporters/trace/jaeger/build.gradle
@@ -16,3 +16,7 @@ dependencies {
 
     signature "org.codehaus.mojo.signature:java16:+@signature"
 }
+
+test {
+  systemProperty "enable.jaeger.integration.test", System.getProperty("enable.jaeger.integration.test")
+}
--- a/exporters/trace/jaeger/src/test/java/io/opencensus/exporter/trace/jaeger/JaegerExporterHandlerIntegrationTest.java
+++ b/exporters/trace/jaeger/src/test/java/io/opencensus/exporter/trace/jaeger/JaegerExporterHandlerIntegrationTest.java
@@ -67,6 +67,7 @@ public class JaegerExporterHandlerIntegrationTest {
 
   @Test(timeout = 30000)
   public void exportToJaeger() throws InterruptedException, IOException {
+    assumeThat("enable jaeger integration test", enableJaegerIntegrationTest(), is(true));
     assumeThat("docker is installed and executable", dockerIsInstalledAndExecutable(), is(true));
 
     final Process jaeger =
@@ -202,6 +203,10 @@ public class JaegerExporterHandlerIntegrationTest {
     } finally {
       jaeger.destroy();
     }
+  }
+
+  private static boolean enableJaegerIntegrationTest() {
+    return "true".equals(System.getProperty("enable.jaeger.integration.test"));
   }
 
   private static boolean dockerIsInstalledAndExecutable() {


exporters/trace/jaeger/README.md
```diff
@@ -40,27 +40,27 @@ For Maven add to your `pom.xml`:
   <dependency>
     <groupId>io.opencensus</groupId>
     <artifactId>opencensus-api</artifactId>
-    <version>0.13.0</version>
+    <version>0.12.2</version>
```
songy23: I don't think this needs to be updated, since Jaeger exporter is not included in v0.12.2 release, instead it's planned to be released in v0.13.0.

sebright:
I'm not sure I understand this change.  If the test is disabled by default, don't we need to explicitly enable it in .travis.yml to run it in CI?  Can we increase the timeout so that it can also run on AppVeyor?

exporters/trace/jaeger/README.md
HailongWen: Ack.

HailongWen:
@sebright. Good point! 

Initially I am just trying to avoid the random build failure. Increasing the timeout to a relatively large number (say, an hour?) should probably work. It is just that there is no guarantee.

I agree it would be better if we can use travis to run the integration test. IMHO, we may have more integration tests that may temporarily deploy a standalone server in the future. How about we move this test out and create a separate task to check it (and future integration test)? /cc @songy23

songy23:
I think we will have some kind of integration test framework for all these integration tests (phase 2 plan for https://github.com/census-instrumentation/opencensus-experiments/tree/master/integration). For now I think it's fine to either bump up the time out or disable it.

sebright:
I'd rather increase the test time limit so that we are running each test in at least one CI build job.  The time limit for each Travis job is 50 minutes, so it looks like we're still well below the limit in the latest builds on master.  I'm also fine with splitting integration tests into a separate job.

HailongWen:
Closed by #1077 .

Issue #1068, opened 2018-03-16
OpenCensus API depends on Guava collections


ericgribkoff:
This (unnecessarily?) increases APK size and method count for gRPC Android users. The `com.google.common.collect` dependencies brought in by OpenCensus are slightly larger than the `io.opencensus` code itself that gets included in classes.dex after proguard is run, and 10% of the size of the gRPC library itself, and introduce ~400 new method references.

It looked (at a quick glance) like the use of `com.google.common.collect` in opencensus-api is fairly shallow and wouldn't be too hard to remove, which would help downstream gRPC Android users. Would removing the use of `com.google.common.collect` introduce any problems for the OpenCensus library?

**Edit** gRPC Java does depend on Guava, but avoids use of collections, which lets those get stripped out. Updated the issue title/description to reflect that it's only half of the total impact of Guava on the APK size which would be better avoided. See https://github.com/grpc/grpc-java/pull/2208 for prior art in grpc-java around removing use of Guava collections.

songy23:
Fixed by https://github.com/census-instrumentation/opencensus-java/pull/1069.

PR #1069, opened 2018-03-17
Remove usages of guava collections in api/

ericgribkoff:
This allows proguard to strip out the com.google.common.collect.* classes on Android, and is aimed at resolving https://github.com/census-instrumentation/opencensus-java/issues/1068

On Android, making this change saves 325 method references and 7.8KB in the APK size. I don't have numbers for an app that directly depends on OpenCensus, but for gRPC (grpc-core depends on opencensus-api) this is a 6.5% decrease in method references and a 3% decrease in APK size (percentages based on the increase an app will see from adding a dependency on gRPC Java and Protobuf Lite).

Dropping `Multiset` from `TagContext#equals` is the only change that looks potentially significant. But, if I understand correctly, `TagContextImpl` instances are used where the equality check may need to be fast. If this means that the comparison is always between two `TagContextImpl` objects, then the implementation of `equals` in the super class isn't invoked, and the changes here shouldn't incur any penalties in actual use of OpenCensus.

### DIFF
--- a/api/src/main/java/io/opencensus/stats/AggregationData.java
+++ b/api/src/main/java/io/opencensus/stats/AggregationData.java
@@ -20,8 +20,8 @@ import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.auto.value.AutoValue;
-import com.google.common.collect.Lists;
 import io.opencensus.common.Function;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import javax.annotation.concurrent.Immutable;
@@ -291,7 +291,7 @@ public abstract class AggregationData {
       }
 
       checkNotNull(bucketCounts, "bucket counts should not be null.");
-      List<Long> bucketCountsCopy = Collections.unmodifiableList(Lists.newArrayList(bucketCounts));
+      List<Long> bucketCountsCopy = Collections.unmodifiableList(new ArrayList<Long>(bucketCounts));
       for (Long bucket : bucketCountsCopy) {
         checkNotNull(bucket, "bucket should not be null.");
       }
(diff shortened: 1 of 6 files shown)


bogdandrutu:
@ericgribkoff do you have any tool/script to check that we do not add a dependency on guava.collect in the future in the API artifact?

bogdandrutu:
@sebright can you comment on the Tag change. Is that fine?

ericgribkoff:
Thanks for the reviews.

@bogdandrutu We don't have any automation around this (AFAIK), it's just enforced manually by the team. gRPC Java now monitors APK size and method count increases for a Helloworld app as part of its Github CI. This doesn't specifically check for guava.collect, but will notice the effect (large increase in method count). If OpenCensus was planning to add an example app for Android users, a similar Android buildscript could be useful - although the specific case that motivated this PR is a bit odd, in that it's targeted at apps that are not actually using OpenCensus but include a dependency on its API.

songy23:
Shall we also do a minor release for this? @bogdandrutu

sebright:
> @ericgribkoff do you have any tool/script to check that we do not add a dependency on guava.collect in the future in the API artifact?

Another solution is to completely remove the Guava dependency from the API library, like we did previously in #13.  I'm not sure how difficult that would be, though.

ericgribkoff:
> Shall we also do a patch release for this? @bogdandrutu 

From gRPC's side, there's no immediate pressure for this to make it into a release. I'll bump gRPC's dependency once this does make it (it looks like this should make it into opencensus 0.13?), but I don't think there's a need to disrupt the regular release cadence just for this change.

songy23:
Sorry I was about to say a patch release (i.e v0.12.3 release).

If this is not urgent we can just include it in our next minor release (v0.13.0).

PR #1073, opened 2018-03-21
Simplify example Gradle build file by moving license and format checks to the Travis script.

sebright:
This change simplifies the example build file by removing plugins not necessary to build the project.  It adds two new build jobs, one for each check.  The google-java-format check is under allow_failures so that contributors aren't required to download and run the tool, but it is easy to see the status of the formatting.

I tested the change with license and formatting errors: https://travis-ci.org/sebright/opencensus-java/builds/356175781

A future improvement would be to move the two checks to the top-level build.gradle.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,11 +60,6 @@ matrix:
   - env: TASK=BUILD
     os: osx
 
-  # Allowing failures because not everyone has downloaded the google-java-format tool.
-  - jdk: oraclejdk8
-    env: TASK=CHECK_EXAMPLES_FORMAT
-    os: linux
-
 before_install:
   - git log --oneline --decorate --graph -30
   - if \[ "$TASK" == "BUILD" \]; then
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,10 @@ matrix:
     env: TASK=BUILD_EXAMPLES_BAZEL
     os: linux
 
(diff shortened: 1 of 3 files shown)


g-easy:
Does https://travis-ci.org/sebright/opencensus-java/jobs/356175791 mean that MultiSpansTracing.java had formatting errors?

g-easy:
"tested the change with license and formatting errors" according to the first comment. :)

bogdandrutu:
I would make the formatting required. We will never fix this unless is required.

g-easy:
Adding to what Bogdan said, I found out the hard way (compile error, but only on Windows) that I had a line too long in examples/

sebright:
Thanks!  I made both checks required.

@g-easy What compile error did you get?  I thought that the line length restriction was only enforced by google-java-format, and possibly checkstyle.

g-easy:
Sorry, I misread it at the time. It was indeed checkstyle, but it fired on AppVeyor and not Travis or my workstation:

```
:checkstyleMain[ant:checkstyle] [ERROR] C:\projects\opencensus-java\examples\src\main\java\io\opencensus\examples\helloworld\QuickStart.java:72: Line is longer than 100 characters (found 108). [LineLength]
```

https://ci.appveyor.com/project/opencensusjavateam/opencensus-java/build/1.0.2152

PR #1074, opened 2018-03-21
Add InternalUtils to get probability from ProbabilitySampler.

songy23:
Similar to [Tags.InternalUtils](https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/tags/InternalUtils.java).


songy23:
Closing for now - will use some other approach.

PR #1076, opened 2018-03-21
Move BaseMessageEventUtil to trace package.

songy23:
`BaseMessageEventUtil` cannot be under internal package, because in some build systems this introduces circular dependency (internal depends on trace while trace depends on internal).


api/src/main/java/io/opencensus/trace/BaseMessageEventUtil.java
```diff
@@ -14,17 +14,18 @@
...
+import io.opencensus.common.Internal;
 
 /**
  * Helper class to convert/cast between for {@link MessageEvent} and {@link NetworkEvent}.
  *
  * @since 0.12
  */
+@Internal
 @SuppressWarnings("deprecation")
 public final class BaseMessageEventUtil {
```
sebright: Could this class go into a new `io.opencensus.trace.internal` package?

songy23: Sounds good, updated.

api/src/main/java/io/opencensus/trace/internal/BaseMessageEventUtil.java
```diff
@@ -14,39 +14,46 @@
...
 
-import io.opencensus.trace.MessageEvent;
+import io.opencensus.common.Internal;
 
 /**
- * Helper class to convert/cast between for {@link MessageEvent} and {@link NetworkEvent}.
+ * Helper class to convert/cast between for {@link io.opencensus.trace.MessageEvent} and {@link
+ * io.opencensus.trace.NetworkEvent}.
  *
  * @since 0.12
```
sebright: I think the since tag should be removed, or updated to 0.13 for the new location.

api/src/main/java/io/opencensus/trace/internal/BaseMessageEventUtil.java
```diff
@@ -14,39 +14,46 @@
...
+  public static io.opencensus.trace.MessageEvent asMessageEvent(
+      io.opencensus.trace.BaseMessageEvent event) {
     checkNotNull(event);
-    if (event instanceof MessageEvent) {
-      return (MessageEvent) event;
+    if (event instanceof io.opencensus.trace.MessageEvent) {
+      return (io.opencensus.trace.MessageEvent) event;
+    }
+    if (!(event instanceof io.opencensus.trace.NetworkEvent)) { // Work around FindBugs.
+      throw new IllegalArgumentException("NetworkEvent expected.");
```
sebright: Could you please update findbugs-exclude.xml instead? https://github.com/census-instrumentation/opencensus-java/blob/master/findbugs-exclude.xml#L20

api/src/main/java/io/opencensus/trace/internal/BaseMessageEventUtil.java
```diff
@@ -65,9 +72,12 @@ public static MessageEvent asMessageEvent(io.opencensus.trace.BaseMessageEvent e
     if (event instanceof io.opencensus.trace.NetworkEvent) {
       return (io.opencensus.trace.NetworkEvent) event;
     }
-    MessageEvent messageEvent = (MessageEvent) event;
+    if (!(event instanceof io.opencensus.trace.MessageEvent)) { // Work around FindBugs.
+      throw new IllegalArgumentException("MessageEvent expected.");
```
sebright: Same here

api/src/main/java/io/opencensus/trace/internal/BaseMessageEventUtil.java
songy23: I just removed it since it's only for internal usage.

api/src/main/java/io/opencensus/trace/internal/BaseMessageEventUtil.java
songy23: Good catch, updated.

api/src/main/java/io/opencensus/trace/internal/BaseMessageEventUtil.java
songy23: Done.

PR #1078, opened 2018-03-22
Jaeger exporter test is now more robust to small IDs.

marccarre:
If the trace/span ID is `0L` (resp. `1L`), then Jaeger simply returns `"0"` (resp. `"1"`) via its REST API.

Looking at `TraceId#generateRandomId`/`TraceId#INVALID` and `SpanId#generateRandomId`/`SpanId#INVALID`, it seems `0L` is a forbidden value which is actively filtered out, but anything else has the potential to happen.

Therefore, regexes for trace IDs and span IDs should be `{1,32}` and `{1,16}` respectively.

Improves on #1063.


HailongWen:
Thanks for the investigation and fix!

PR #1082, opened 2018-03-23
Remove an unused local variable

songy23:


impl_core/src/test/java/io/opencensus/implcore/trace/export/NoopRunningSpanStoreImplTest.java
```diff
@@ -66,17 +66,16 @@
...
-            spanContext,
-            recordSpanOptions,
-            SPAN_NAME,
-            null,
-            false,
-            TraceParams.DEFAULT,
-            startEndHandler,
-            timestampConverter,
-            testClock);
+    SpanImpl.startSpan(
```
sebright: I'm not sure I understand this change.  Does creating the span have a side effect?  I think it is more likely that the new span should be assigned to `spanImpl`, which is null.

I also wonder if RunningSpanStoreImpl.onStart and onEnd should have rejected the null `spanImpl`.

songy23: I don't have context on this test. The local variable `SpanImpl span` is never used and failed the build when import. 

@HailongWen should this be assigned to `private SpanImpl spanImpl;` instead?

bogdandrutu: I think so.

songy23: Done.

songy23:
Thanks!

PR #1085, opened 2018-03-23
More robust integration test for Jaeger exporter.

### LINKED ISSUE #1066: JaegerExporterHandlerIntegrationTest is fluctuant in continuous build

marccarre:
Changelog:

- Wait for Jaeger to be ready by polling it over HTTP rather than waiting an arbitrary amount of time.
- Add unit to all times and durations.
- Fix/improve assertion on span duration.

Fixes #1066, particularly [point 2. in this comment](https://github.com/census-instrumentation/opencensus-java/issues/1066#issuecomment-375554352), i.e. should avoid build failures like https://travis-ci.org/census-instrumentation/opencensus-java/jobs/357147554.

### DIFF
--- a/exporters/trace/jaeger/src/test/java/io/opencensus/exporter/trace/jaeger/JaegerExporterHandlerIntegrationTest.java
+++ b/exporters/trace/jaeger/src/test/java/io/opencensus/exporter/trace/jaeger/JaegerExporterHandlerIntegrationTest.java
@@ -47,6 +47,7 @@ import io.opencensus.trace.Tracing;
 import io.opencensus.trace.samplers.Samplers;
 import java.io.IOException;
 import java.net.ConnectException;
+import java.net.SocketException;
 import java.util.Random;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -242,6 +243,15 @@ public class JaegerExporterHandlerIntegrationTest {
       } catch (ConnectException e) {
         logger.log(Level.INFO, "Jaeger is not yet ready, waiting a bit...");
         Thread.sleep(10L);
+      } catch (SocketException e) {
+        if (e.getMessage().contains("Unexpected end of file from server")) {
+          // Jaeger seems to accept connections even though it is not yet ready to handle HTTP
+          // requests.
+          logger.log(Level.INFO, "Jaeger is still not yet ready, waiting a bit more...", e);
+          Thread.sleep(10L);
+        } else {
+          throw e;
+        }
       }
     }
   }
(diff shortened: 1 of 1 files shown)


sebright:
@HailongWen Why will the integration test sometimes break the build?  I thought that #1084 would make the test more stable by avoiding automatically upgrading Jaeger.  I think that adding a separate build job or moving the test to a different CI setup makes sense, though.

sebright:
There are two test failures, though I don't know whether they are the failures that should be fixed by #1084.

marccarre:
> There are two test failures, though I don't know whether they are the failures that should be fixed by #1084.

They seem to be variants of the connectivity issues I saw on my local machine.
Not sure if Jaeger crashed or if something else happened, though.
They are unrelated to #1084.

marccarre:
@sebright, retrying on the `Connection reset` [seems to work](https://travis-ci.org/census-instrumentation/opencensus-java/builds/357690114), so I suppose it is just Jaeger transiently acting out then.

HailongWen:
@sebright My main concern is that we require a successful build to commit code, the unstable build (mostly due to setting) will sometimes cause trouble, especially when some one is not familiar with the test.

sebright:
> In the long term we may consider moving this test to be part of https://github.com/census-instrumentation/opencensus-experiments/tree/master/integration, but for now I think it's fine to have it in Java CI.

I agree, I think we should require the test to pass, unless there is a specific issue with the test, and then we can temporarily move it to allow_failures.

sebright:
I'll merge this for now, since splitting the test into a separate build job could be a different PR.

sebright:
I opened #1096 for the separate build job.

PR #1087, opened 2018-03-24
Set unit '1' when the aggregation type is Count.

bogdandrutu:


exporters/stats/stackdriver/src/test/java/io/opencensus/exporter/stats/stackdriver/StackdriverExportUtilsTest.java
```diff
@@ -142,6 +142,15 @@ public void createValueType() {
         .isEqualTo(MetricDescriptor.ValueType.DISTRIBUTION);
   }
 
+  @Test
+  public void createUnit() {
```
sebright: I think we should also have a test that calls createMetricDescriptor with a count view, to check that createUnit is called in the right place.

bogdandrutu: Done. Though we do not have these kind of tests for all the createXXX helpers.

sebright: I thought it was important to test this change, since count is a special case, and it would be easy to accidentally use the original units.

sebright:
LGTM

PR #1097, opened 2018-03-26
Add unit test for HTTP header name validation.

HailongWen:
This PR added `HttpUtils` for `TextFormat` propagation tests. It validates the characters of header names (#1072).

The validation logic is inspired by [netty](https://github.com/netty/netty/blob/fc3b145cbbcad4a257f92f0640eb4ded19b44afe/codec-http/src/main/java/io/netty/handler/codec/http/DefaultHttpHeaders.java#L381:25), however it does not fully implement the spec.

There are two specs for the header and they are actually the same. The only difference is that [RFC-2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec2.html#sec2.1) defines the charset by excluding invalid characters, while [RFC-7230](https://tools.ietf.org/html/rfc7230#page-22) defines it by listing all allowed characters.

### DIFF
--- a/contrib/http_util/src/test/java/io/opencensus/contrib/http/util/CloudTraceFormatTest.java
+++ b/contrib/http_util/src/test/java/io/opencensus/contrib/http/util/CloudTraceFormatTest.java
@@ -286,7 +286,7 @@ public final class CloudTraceFormatTest {
   @Test
   public void fieldsShouldBeValid() {
     for (String header : cloudTraceFormat.fields()) {
-      assertThat(HttpUtils.validateHeaderName(header)).isEqualTo(true);
+      HttpUtils.assertHeaderNameIsValid(header);
     }
   }
 
--- a/contrib/http_util/src/test/java/io/opencensus/contrib/http/util/CloudTraceFormatTest.java
+++ b/contrib/http_util/src/test/java/io/opencensus/contrib/http/util/CloudTraceFormatTest.java
@@ -24,6 +24,7 @@ import static io.opencensus.contrib.http.util.CloudTraceFormat.SPAN_ID_DELIMITER
 import static io.opencensus.contrib.http.util.CloudTraceFormat.TRACE_OPTION_DELIMITER;
 
 import com.google.common.primitives.UnsignedLong;
+import io.opencensus.testing.propagation.HttpUtils;
 import io.opencensus.trace.SpanContext;
 import io.opencensus.trace.SpanId;
 import io.opencensus.trace.TraceId;
@@ -280,6 +281,13 @@ public final class CloudTraceFormatTest {
   @Test
   public void fieldsShouldMatch() {
     assertThat(cloudTraceFormat.fields()).containsExactly(HEADER_NAME);
+  }
+
+  @Test
+  public void fieldsShouldBeValid() {
+    for (String header : cloudTraceFormat.fields()) {
+      assertThat(HttpUtils.validateHeaderName(header)).isEqualTo(true);
+    }
   }
 
   @Test
--- a/impl_core/src/test/java/io/opencensus/implcore/trace/propagation/B3FormatTest.java
+++ b/impl_core/src/test/java/io/opencensus/implcore/trace/propagation/B3FormatTest.java
@@ -23,6 +23,7 @@ import static io.opencensus.implcore.trace.propagation.B3Format.X_B3_SAMPLED;
 import static io.opencensus.implcore.trace.propagation.B3Format.X_B3_SPAN_ID;
 import static io.opencensus.implcore.trace.propagation.B3Format.X_B3_TRACE_ID;
 
+import io.opencensus.testing.propagation.HttpUtils;
 import io.opencensus.trace.SpanContext;
 import io.opencensus.trace.SpanId;
 import io.opencensus.trace.TraceId;
@@ -217,5 +218,8 @@ public class B3FormatTest {
     assertThat(b3Format.fields())
         .containsExactly(
             X_B3_TRACE_ID, X_B3_SPAN_ID, X_B3_PARENT_SPAN_ID, X_B3_SAMPLED, X_B3_FLAGS);
+    for (String header : b3Format.fields()) {
+      assertThat(HttpUtils.validateHeaderName(header)).isEqualTo(true);
+    }
   }
 }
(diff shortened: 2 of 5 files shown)


adriancole:
Since this is a test.. why not rename validateX to assertX and throw an
assertion error with a message instead of false.

testing/src/test/java/io/opencensus/testing/propagation/HttpUtilsTest.java
```diff
@@ -0,0 +1,93 @@
...
+ * separators = "(" | ")" | "&lt;" | "&gt;" | "{@literal @}" | "," | ";" | ":" | "\" | &lt;"&gt;
+ *            | "/" | "[" | "]" | "?" | "=" | "{" | "}" | SP | HT
+ * SP         = &lt;US-ASCII SP, space (32)&gt;
+ * HT         = &lt;US-ASCII HT, horizontal-tab (9)&gt;
+ * </pre>
+ */
+@RunWith(JUnit4.class)
+public final class HttpUtilsTest {
+
+  private final Set<Character> disallowed = new HashSet<Character>();
```
songy23: nit: consider using an `ImmutableSet` and initialize it statically.

testing/src/main/java/io/opencensus/testing/propagation/HttpUtils.java
```diff
@@ -0,0 +1,77 @@
...
+   * <p>This method is also inspired by
+   * https://github.com/netty/netty/blob/fc3b145cbbcad4a257f92f0640eb4ded19b44afe/codec-http/src/main/java/io/netty/handler/codec/http/DefaultHttpHeaders.java#L381:25
+   *
+   * @param name the name of the header.
+   * @return {@code false} if a header name contains invalid characters, otherwise {@code true}.
+   */
+  public static boolean validateHeaderName(String name) {
+    for (int i = 0; i < name.length(); i++) {
+      char c = name.charAt(i);
+      switch (c) {
```
songy23: nit: consider using an `ImmutableSet` with all these allowed characters and check if c is in the set.

HailongWen: It is implemented with an allowed set of characters, so I want to test it in another way (testing the disallowed ones).

contrib/http_util/src/test/java/io/opencensus/contrib/http/util/CloudTraceFormatTest.java
```diff
@@ -282,6 +283,13 @@ public void fieldsShouldMatch() {
     assertThat(cloudTraceFormat.fields()).containsExactly(HEADER_NAME);
   }
 
+  @Test
+  public void fieldsShouldBeValid() {
+    for (String header : cloudTraceFormat.fields()) {
+      assertThat(HttpUtils.validateHeaderName(header)).isEqualTo(true);
```
sebright: Suggestion: This line could use `assertTrue`.

impl_core/src/test/java/io/opencensus/implcore/trace/propagation/B3FormatTest.java
```diff
@@ -217,5 +218,8 @@ public void fields_list() {
     assertThat(b3Format.fields())
         .containsExactly(
             X_B3_TRACE_ID, X_B3_SPAN_ID, X_B3_PARENT_SPAN_ID, X_B3_SAMPLED, X_B3_FLAGS);
+    for (String header : b3Format.fields()) {
+      assertThat(HttpUtils.validateHeaderName(header)).isEqualTo(true);
```
sebright: Same here

testing/src/main/java/io/opencensus/testing/propagation/HttpUtils.java
```diff
@@ -0,0 +1,77 @@
...
+ */
+
+package io.opencensus.testing.propagation;
+
+/**
+ * Utilities for HTTP testing.
+ *
+ * @since 0.13
+ */
+public final class HttpUtils {
```
sebright: Is this class also meant to be used by users of OpenCensus?  If not, I think it would be better to put it in a separate library.  I think opencensus-testing is meant to provide utilities for users of OpenCensus to test their use of the API, such as fake implementations.

/cc @bogdandrutu

testing/src/main/java/io/opencensus/testing/propagation/HttpUtils.java
```diff
@@ -0,0 +1,77 @@
...
+   *
+   * <pre>
+   * field-name = token
+   * token      = 1*tchar
+   * tchar      = "!" / "#" / "$" / "%" / "&amp;" / "'" / "*" / "+" / "-" / "." / "^" / "_"
+   *            / "`" / "|" / "~" / DIGIT / ALPHA
+   * </pre>
+   *
+   * <p>This method is also inspired by
+   * https://github.com/netty/netty/blob/fc3b145cbbcad4a257f92f0640eb4ded19b44afe/codec-http/src/main/java/io/netty/handler/codec/http/DefaultHttpHeaders.java#L381:25
```
sebright: Is there any way that we can reuse the method from Netty (or another library), since this is only used in tests, where we have fewer restrictions on dependencies?

HailongWen: The Netty one does not fully follow the spec, though it should be enough to serve the purpose of preventing similar issues. Though the method itself is private, we can construct a `DefaultHttpHeaders` if we want to trigger the validation logic.

I am fine with using Netty if dependency is not a concern.

testing/src/main/java/io/opencensus/testing/propagation/HttpUtils.java
HailongWen: I don't think it will be used by users. But it would be used by impl_core and contrib_http_util. Since it is test related, I put it in opencensus-testing, not opencensus-api. Not sure if there are better places.

testing/src/main/java/io/opencensus/testing/propagation/HttpUtils.java
sebright: I can't think of any problems with adding Netty as a test-only dependency, except possibly slowing down the build.

Using Netty would also work around the issue of needing to share code between two test suites.  We'll probably still need to find a way to share code at some point, though.

HailongWen:
I am looking for a proper library to provide such functionality - the netty-codec-http has a bunch of secondary dependencies and is not an ideal one for this small purpose.

impl_core/src/test/java/io/opencensus/implcore/trace/propagation/B3FormatTest.java
```diff
@@ -217,5 +218,8 @@ public void fields_list() {
     assertThat(b3Format.fields())
         .containsExactly(
             X_B3_TRACE_ID, X_B3_SPAN_ID, X_B3_PARENT_SPAN_ID, X_B3_SAMPLED, X_B3_FLAGS);
+    for (String header : b3Format.fields()) {
+      HttpUtils.assertHeaderNameIsValid(header);
```
odeke-em: It's great that we are testing that `X_B3*` are valid, however, it might suffice for us to get the originally offending header and add that to assert a failure, thus preventing any future regressions.

sebright:
Closing for now, since this PR isn't active.

Issue #1081, opened 2018-03-23
Remove Guava from opencensus-api?


saturnism:
Please answer these questions before submitting a bug report.

### What version of OpenCensus are you using?
0.12.x

### What did you expect to see?
opencensus-api has dependency on Guava. I don't know how Guava is actually used by the API, but it's strange for an API module to have explicit dependency on Guava that may actually introduce dependency conflicts for users.

In fact, if Guava is not used extensively, it's probably better not to use it at all, or, shade it to avoid issues.

saturnism:
Also, here is a dependency tree on using OpenCensus. I wonder if they are all necessary to expose to the end user.

```
[INFO] com.example:census-example:jar:1.0-SNAPSHOT
[INFO] +- io.opencensus:opencensus-api:jar:0.12.2:compile
[INFO] |  +- com.google.errorprone:error_prone_annotations:jar:2.2.0:compile
[INFO] |  +- com.google.code.findbugs:jsr305:jar:3.0.1:compile
[INFO] |  +- io.grpc:grpc-context:jar:1.9.0:compile
[INFO] |  \- com.google.guava:guava:jar:19.0:compile
[INFO] +- io.opencensus:opencensus-impl:jar:0.12.2:runtime
[INFO] |  +- io.opencensus:opencensus-impl-core:jar:0.12.2:runtime
[INFO] |  \- com.lmax:disruptor:jar:3.3.9:runtime
[INFO] +- io.opencensus:opencensus-exporter-stats-prometheus:jar:0.12.2:compile
[INFO] +- io.prometheus:simpleclient:jar:0.3.0:compile
[INFO] +- io.prometheus:simpleclient_hotspot:jar:0.3.0:compile
[INFO] +- io.prometheus:simpleclient_httpserver:jar:0.3.0:compile
[INFO] |  \- io.prometheus:simpleclient_common:jar:0.3.0:compile
[INFO] \- io.opencensus:opencensus-exporter-trace-zipkin:jar:0.12.2:compile
[INFO]    +- io.zipkin.reporter2:zipkin-reporter:jar:2.3.2:compile
[INFO]    |  \- io.zipkin.zipkin2:zipkin:jar:2.4.6:compile
[INFO]    \- io.zipkin.reporter2:zipkin-sender-urlconnection:jar:2.3.2:compile
```

sebright:
[...]

PR #1098, opened 2018-03-26
Make Error Prone and FindBugs annotations 'compileOnly' dependencies.

sebright:
This is part of issue #1081.


sebright:
Thanks!

Issue #1094, opened 2018-03-26
Deprecate Mean Aggregation


songy23:
As part of https://github.com/census-instrumentation/opencensus-specs/issues/56.

PR #1101, opened 2018-03-27
Deprecate Mean and MeanData

songy23:
Fixes https://github.com/census-instrumentation/opencensus-java/issues/1094

/cc @dinooliva

### DIFF
--- a/contrib/grpc_metrics/src/main/java/io/opencensus/contrib/grpc/metrics/RpcViewConstants.java
+++ b/contrib/grpc_metrics/src/main/java/io/opencensus/contrib/grpc/metrics/RpcViewConstants.java
@@ -46,7 +46,6 @@ import io.opencensus.common.Duration;
 import io.opencensus.stats.Aggregation;
 import io.opencensus.stats.Aggregation.Count;
 import io.opencensus.stats.Aggregation.Distribution;
-import io.opencensus.stats.Aggregation.Mean;
 import io.opencensus.stats.BucketBoundaries;
 import io.opencensus.stats.View;
 import java.util.Arrays;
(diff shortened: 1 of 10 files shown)


contrib/grpc_metrics/src/main/java/io/opencensus/contrib/grpc/metrics/RpcViewConstants.java
```diff
@@ -99,7 +98,7 @@
               4096.0, 8192.0, 16384.0, 32768.0, 65536.0));
 
   // Use Aggregation.Mean to record sum and count stats at the same time.
-  @VisibleForTesting static final Aggregation MEAN = Mean.create();
+  @VisibleForTesting static final Aggregation MEAN = Aggregation.Mean.create();
```
sebright: How will we handle the existing Mean RPC views?

songy23: Mean is only used in interval views, and those views are not exported to any backends. I think we can remove those views along with Mean and Window in the future.

sebright: The error count views also use Mean, though I think they also need to be updated to use different Measures, so we can update the Aggregation type at the same time.

songy23: Oh correct, the error count views should probably be updated to use Count instead.

songy23:
Looks like AppVeyor is having some difficulties connecting to maven. I saw another AppVeyor failure in another PR. I'll do a dummy push to run AppVeyor test again later today.

songy23:
Finally got CI working...

PR #1102, opened 2018-03-27
Suppress Checker Framework warnings in AutoValue generated classes (fixes #917).

### LINKED ISSUE #917: Suppress Checker Framework warnings in AutoValue generated classes.
~~https://github.com/census-instrumentation/opencensus-java/pull/916/commits/18cf999ddea65775db2a482e00cea05d52961077~~ bbe6b10f7c6b622b4ffd18e98f58d96a8eba3854 used `SuppressWarnings("nullness")` on all AutoValue abstract classes to suppress warnings in the generated subclasses.  The warnings were caused by missing Nullable annotations on the parameter to `equals`.  It would be better to find a way to only suppress warnings in the generated classes.  Another solution would be to make AutoValue add the correct annotations.

sebright:
#### Remove @SuppressWarnings("nullness") from all AutoValue classes.

This change allows the Checker Framework to check the classes annotated with
`@AutoValue`, even though their generated subclasses are skipped.  This commit
also fixes some previously suppressed warnings in the ViewData class.

### DIFF
--- a/api/src/main/java/io/opencensus/common/Duration.java
+++ b/api/src/main/java/io/opencensus/common/Duration.java
@@ -34,9 +34,6 @@ import javax.annotation.concurrent.Immutable;
  */
 @Immutable
 @AutoValue
-// Suppress Checker Framework warning about missing @Nullable in generated equals method.
-@AutoValue.CopyAnnotations
-@SuppressWarnings("nullness")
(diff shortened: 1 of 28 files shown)


sebright:
I think the AppVeyor build failed because it couldn't download the jmhreport Gradle plugin.  I'm retrying it.

PR #1105, opened 2018-03-30
[WIP] Add HTTP Netty example and gradle build.

HailongWen:
Part of #980. Still need to resolve maven and bazel build.


HailongWen:
@adriancole Thanks. I can add another apache-httpclient example just as brave does.

HailongWen:
@chen56 thanks. 
I would close it for now since it is not a good high-level example and the code base is far behind master.

PR #1106, opened 2018-03-30
Turn off Error Prone warning MutableMethodReturnType.

sebright:
MutableMethodReturnType suggests returning Guava immutable collections from API
methods, but OpenCensus doesn't expose Guava in its API. See
https://github.com/google/error-prone/issues/982.


sebright:
Thanks!

PR #1108, opened 2018-03-30
Sort TagKeys in View.create(...) (fixes #1080).

### LINKED ISSUE #1080: Simplify Views by ignoring the order of the input TagKeys.
Since the TagKey order doesn't affect the aggregation of stats, it would probably be simpler to sort the TagKeys in the View factory method.  Consistently sorting the keys would prevent Views from differing only by the order of the keys, which would reduce the likelihood of a View being rejected from `registerView` because of a conflict with an existing view.

We still need to store the TagKeys in a list, instead of a set, since the order is used to match TagValues in a ViewData with TagKeys in the corresponding View.

sebright:
This PR also contains some cleanup changes in ViewTest.java.

### DIFF
--- a/api/src/test/java/io/opencensus/stats/ViewTest.java
+++ b/api/src/test/java/io/opencensus/stats/ViewTest.java
@@ -18,7 +18,6 @@ package io.opencensus.stats;
 
 import static com.google.common.truth.Truth.assertThat;
 
-import com.google.common.collect.ImmutableList;
 import com.google.common.testing.EqualsTester;
 import io.opencensus.common.Duration;
 import io.opencensus.stats.Aggregation.Mean;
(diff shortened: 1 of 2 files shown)


api/src/test/java/io/opencensus/stats/ViewTest.java
```diff
@@ -70,7 +70,7 @@ public void testIntervalView() {
   public void testViewEquals() {
     new EqualsTester()
         .addEqualityGroup(
-            View.create(NAME, DESCRIPTION, MEASURE, MEAN, KEYS, Cumulative.create()),
+            View.create(NAME, DESCRIPTION, MEASURE, MEAN, KEYS),
             View.create(NAME, DESCRIPTION, MEASURE, MEAN, KEYS, Cumulative.create()))
```
songy23: Should this `Cumulative.create()` be removed too?

sebright: I left it in order to assert that the new and old factory methods give the same result for cumulative views.

sebright:
Thanks!

PR #1109, opened 2018-04-02
Update README based on customers' feedback

songy23:


README.md
```diff
@@ -50,8 +50,13 @@ public final class MyClassWithTracing {
   private static final Tracer tracer = Tracing.getTracer();
 
   public static void doWork() {
-    // Create a child Span of the current Span.
-    try (Scope ss = tracer.spanBuilder("MyChildWorkSpan").startScopedSpan()) {
+    // Create a child Span of the current Span. Always record events for this span and force it to 
+    // be sampled.
```
sebright: Should the paragraph above explain why the example forces sampling?  IIUC, forcing sampling makes the example easier to try out, but it isn't the most common use case.

songy23: Sounds good, I added some explanation here.

songy23:
Thanks!

README.md
```diff
@@ -50,8 +50,14 @@ public final class MyClassWithTracing {
...
 
   public static void doWork() {
-    // Create a child Span of the current Span.
-    try (Scope ss = tracer.spanBuilder("MyChildWorkSpan").startScopedSpan()) {
+    // Create a child Span of the current Span. Always record events for this span and force it to 
+    // be sampled. This makes it easier to try out the example, but unless you have a clear use 
+    // case, you don't need to explicitly set record events or sampler.
+    try (Scope ss = 
+         tracer.spanBuilder("MyChildWorkSpan")
+           .setRecordEvents(true)
```
adriancole: maybe we can special-case and when sampler is always, imply event recording? Seems a bit cluttery and the intent is probably clear. Otherwise, maybe we just need to be more clear about when you'd *no* record events, yet always sample.

songy23: /cc @bogdandrutu

Issue #977, opened 2018-02-07
Several public methods are annotated with @VisibleForTesting but not @Internal.


sebright:
I don't think we should annotate methods with `VisibleForTesting` without adding other documentation indicating that they shouldn't be used.  `VisibleForTesting` doesn't appear in Javadocs so it isn't enough to show that the element isn't part of the API.  We should probably either add `@Internal`, or remove `@VisibleForTesting` for each of these methods:

`SampledSpanStore.getRegisteredSpanNamesForCollection()`
This method can't be completely hidden from the API, because it is abstract, and SampledSpanStore subclasses must implement it.
https://github.com/census-instrumentation/opencensus-java/blob/b848a65bd6d76aac8e59f8365746846bee405896/api/src/main/java/io/opencensus/trace/export/SampledSpanStore.java#L127-L128

`Settings(Config)`
https://github.com/census-instrumentation/opencensus-java/blob/b848a65bd6d76aac8e59f8365746846bee405896/contrib/agent/src/main/java/io/opencensus/contrib/agent/Settings.java#L34-L35

`Status.CanonicalCode.toStatus()` (issue #976)
https://github.com/census-instrumentation/opencensus-java/blob/b848a65bd6d76aac8e59f8365746846bee405896/api/src/main/java/io/opencensus/trace/Status.java#L179-L180

sebright:
This issue was fixed by #1196, #1197, and #1299.

PR #1111, opened 2018-04-03
Remove more usages of Guava from opencensus-api (issue #1081).

sebright:
After this PR, opencensus-api only uses Guava in two places: `SpanId` and `TraceId` both use `com.google.common.io.BaseEncoding` for base 16 encoding.  This PR also adds an internal `Util` class that replaces some of the commonly used functionality, such as argument checking.

### DIFF
--- a/api/src/main/java/io/opencensus/internal/Utils.java
+++ b/api/src/main/java/io/opencensus/internal/Utils.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright 2018, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.internal;
+
+/*>>>
+import org.checkerframework.checker.nullness.qual.NonNull;
+*/
+
+/** General internal utility methods. */
+public final class Utils {
+
+  private Utils() {}
+
+  /**
+   * Throws an {@link IllegalArgumentException} if the argument is false. This method is similar to
+   * {@code Preconditions.checkArgument(boolean, Object)} from Guava.
+   *
+   * @param isValid whether the argument check passed.
+   * @param message the message to use for the exception.
+   */
+  public static void checkArgument(boolean isValid, String message) {
+    if (!isValid) {
+      throw new IllegalArgumentException(message);
+    }
+  }
+
+  /**
+   * Throws an {@link IllegalStateException} if the argument is false. This method is similar to
+   * {@code Preconditions.checkState(boolean, Object)} from Guava.
+   *
+   * @param isValid whether the state check passed.
+   * @param message the message to use for the exception.
+   */
+  public static void checkState(boolean isValid, String message) {
+    if (!isValid) {
+      throw new IllegalStateException(message);
+    }
+  }
+
+  /**
+   * Validates an index in an array or other container. This method throws an {@link
+   * IllegalArgumentException} if the size is negative and throws an {@link
+   * IndexOutOfBoundsException} if the index is negative or greater than or equal to the size. This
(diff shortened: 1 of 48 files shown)


api/src/main/java/io/opencensus/stats/Aggregation.java
```diff
@@ -221,7 +220,7 @@ public static Mean create() {
      * @since 0.8
      */
     public static Distribution create(BucketBoundaries bucketBoundaries) {
-      checkNotNull(bucketBoundaries, "bucketBoundaries should not be null.");
+      Utils.checkNotNull(bucketBoundaries, "bucketBoundaries should not be null.");
```
HailongWen: Maybe we can use static import to minimize all these "Utils.xxxx" changes?

api/src/main/java/io/opencensus/tags/Tags.java
```diff
@@ -91,7 +90,7 @@ public static void setState(TaggingState state) {
   }
 
   // Any provider that may be used for TagsComponent can be added here.
-  @VisibleForTesting
+  // Visible for testing.
```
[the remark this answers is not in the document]
sebright: Sure.  I also split it into DefaultVisibilityForTesting and PublicForTesting to try to make the meaning clearer.  PublicForTesting should be removed by issue #977, though.

api/src/main/java/io/opencensus/stats/Aggregation.java
sebright: I'd rather qualify these methods to distinguish them from the methods in Preconditions, especially since we still use Guava in other parts of the project.

api/src/main/java/io/opencensus/internal/Utils.java
```diff
@@ -0,0 +1,126 @@
...
+
+  /**
+   * Throws a {@link NullPointerException} if the argument is null. This method is similar to {@code
+   * Preconditions.checkNotNull(Object, Object)} from Guava.
+   *
+   * @param arg the argument to check for null.
+   * @param message the message to use for the exception.
+   * @return the argument, if it passes the null check.
+   */
+  public static <T /*>>> extends @NonNull Object*/> T checkNotNull(T arg, String message) {
```
bogdandrutu: do we need this annotation that extends NonNull here?

api/src/main/java/io/opencensus/stats/Aggregation.java
bogdandrutu: I think we all agree to no longer use Guava, so I think it is fine to use static imports.

impl_core/src/main/java/io/opencensus/implcore/trace/propagation/B3Format.java
```diff
@@ -58,7 +62,8 @@
   }
 
   @Override
-  public <C> void inject(SpanContext spanContext, C carrier, Setter<C> setter) {
+  public <C /*>>> extends @NonNull Object*/> void inject(
+      SpanContext spanContext, C carrier, Setter<C> setter) {
     checkNotNull(spanContext, "spanContext");
```
bogdandrutu: Should we also remove guava here an use our own implementation?

api/src/main/java/io/opencensus/stats/Aggregation.java
sebright: Do you mean that you also want to remove the Guava dependencies from the implementation and exporters?  I think that should be a separate change.  I haven't looked at how we use Guava in the other artifacts, but if we use it heavily, it could be better to shade it than remove it.

api/src/main/java/io/opencensus/internal/Utils.java
sebright: Yes, this method shouldn't be called on a `@Nullable` argument.

impl_core/src/main/java/io/opencensus/implcore/trace/propagation/B3Format.java
sebright: Same here, I think that should be a separate change.

sebright:
I rebased my branch, and I'll merge it now.  Thanks for the review!

adriancole:
I think removing guava as an impl dep is good. We cant control all
exporters, but definitely we should not use guava in ones we write unless
it is exceptional

sebright:
Do you think we should also avoid shading Guava in the implementation and exporters?

adriancole:
>
> Do you think we should also avoid shading Guava in the implementation and
> exporters?
>
when I've tried to shade guava, I've noticed it can lead to bloat as often
the types end up connected to eachother. This can make the jar larger which
can be problematic for those sensitive to it.  Also, shading makes it easy
to bloat even more later as you are less aware of use.

PR #1112, opened 2018-04-04
Checker Framework: 2.3.2 -> 2.4.0

sebright:
Checker Framework 2.4.0 is the second latest version.  The latest, 2.5.0, ends support for annotations in comments (https://github.com/typetools/checker-framework/releases/tag/checker-framework-2.5.0), so we probably can't upgrade further until we can use Java 8 syntax.

This PR also enables `-AinvariantArrays`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -206,7 +206,7 @@ subprojects {
 
     dependencies {
         if (useCheckerFramework) {
-            ext.checkerFrameworkVersion = '2.3.2'
+            ext.checkerFrameworkVersion = '2.4.0'
             ext.jdkVersion = 'jdk8'
             checkerFrameworkAnnotatedJDK "org.checkerframework:${jdkVersion}:${checkerFrameworkVersion}"
             checkerFrameworkJavac "org.checkerframework:compiler:${checkerFrameworkVersion}"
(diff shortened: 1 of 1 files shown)


sebright:
Thanks!

wmdietl:
@sebright A comment about annotations-in-comments: you can continue to use annotations-in-comments by using the 2.4.0 jsr308-langtools compiler. JDK 8 isn't changing much anymore, so we decided to not update that module anymore and simplify the documentation.
Annotations-in-comments is purely a compiler change, so by using the 2.4.0 compiler with a current Checker Framework you will get both annotations-in-comments and the current Checker Framework.

But of course, updating to Java 8 would be better.

sebright:
@wmdietl Thanks for the advice!  I'm not completely sure I understand the relationship between the different components of the Checker Framework.  Does that mean that we can use 
version 2.4.0 of `org.checkerframework:compiler` but upgrade all other artifacts, i.e., the annotated JDK, `org.checkerframework:checker`, and `org.checkerframework:checker-qual`?  I upgraded to 2.5.0 (#1136), and it seemed to work.

wmdietl:
Yes, you can use the 2.4.0 compiler with newer Checker Framework artifacts, at least while there are no major changes to the compiler.
Please let us know if you run into any issues with this setup. Annotations-in-comments are still there for existing users, but people should start going to Java 8 :-)

PR #1114, opened 2018-04-05
 Move toMillis(Duration) to Utils so that it can be reused.

songy23:
There are multiple classes that need the `toMillis(Duration)` method. I moved it to `Uitls` so that it can be reused.


songy23:
Yes that could be possible, and if `Utils` ends up being too big in the future, we can split it into multiple classes, e.g StringUtils, PreconditionUtils, TimeUtils, etc.

songy23:
The second approach sounds better. I can hold off this PR until we have the internal utility subproject.

sebright:
Here is the branch: https://github.com/sebright/opencensus-java/tree/shadow-plugin

I've never used the shadow plugin before, so I'm not sure if this is correct.  It adds a subproject with one utility method that is shared by api, impl_core, and impl.

sebright:
I accidentally imported the wrong Utils class in my branch, so I fixed it in a new commit.

songy23:
Just found that `toMillis(Duration)` cannot be part of the internal-utils because it introduced circular dependency.

songy23:
@sebright Yes there will be circular dependencies between `internal-utils` and `api` if I add `toMillis()` to internal-utils. As I see `api` already depends on `internal-utils`: https://github.com/sebright/opencensus-java/blob/shadow-plugin/api/build.gradle#L6.

sebright:
I was just experimenting with the shadow plugin in that branch.  The checkNotNull method in isn't needed after #1111, so the api library doesn't need to depend on it.

songy23:
Got it, I was thinking about making a PR to your branch initially.

I copied internal utils and moved `toMillis()` to it instead.

exporters/stats/signalfx/build.gradle
```diff
@@ -19,3 +20,10 @@ dependencies {
 
     signature "org.codehaus.mojo.signature:java17:+@signature"
 }
+
+shadowJar {
```
bogdandrutu: Why do we need this?

songy23: Because we don't want implementation and exporters to access opencensus-api's internal package. Instead we want to use a shaded dependency by the other artifacts, as suggested by @sebright in https://github.com/census-instrumentation/opencensus-java/pull/1114#pullrequestreview-109828652.

bogdandrutu: If that is the case then it should be in common directory and implementations can use it.

songy23: Sounds reasonable, I think we can make [common.TimeUtils](https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/common/TimeUtils.java) public and have this method be part of it.

bogdandrutu:
@sebright ping on this.

sebright:
Sorry for the delay.  I'm fine with making toMillis public, if we add a warning about overflow.  I'd rather add it as a non-static method on Duration, though.

songy23:
Sounds good, I added some java doc about the potential overflow and precision loss. I also moved `toMillis()` to `Duration.java` next to `fromMillis()`.

api/src/main/java/io/opencensus/common/Duration.java
```diff
@@ -78,6 +79,20 @@ public static Duration fromMillis(long millis) {
     return Duration.create(seconds, nanos);
   }
 
+  /**
+   * Converts a {@link Duration} to milliseconds.
+   *
+   * <p>Note that there could be overflow or loss of precision by making this conversion. See {@link
+   * TimeUnit#convert(long, TimeUnit)} for details.
```
sebright: I think it would be better to describe the overflow behavior without mentioning TimeUnit, since TimeUnit is an implementation detail, and its documentation is more general.  Also, it looks like TimeUnit.toMillis returns Long.MIN_VALUE or Long.MAX_VALUE instead of overflowing.  I don't think there is any loss of precision, besides the unavoidable conversion from nanoseconds to milliseconds.

api/src/main/java/io/opencensus/common/Duration.java
```diff
@@ -78,6 +79,20 @@ public static Duration fromMillis(long millis) {
...
+   *
+   * @param duration a {@code Duration}.
+   * @return the milliseconds representation of this {@code Duration}.
+   */
+  public static long toMillis(Duration duration) {
```
sebright: I would rather make this method non-static, for consistency with getSeconds and getNanos.

api/src/main/java/io/opencensus/common/Duration.java
songy23: Done - I copied the description from TimeUnit.toMillis here. I'm not sure if we should mention that converting nanoseconds to milliseconds will cause precision loss though (seems users should have known that).

api/src/main/java/io/opencensus/common/Duration.java
```diff
@@ -78,6 +79,20 @@ public static Duration fromMillis(long millis) {
...
   }
 
+  /**
+   * Converts a {@link Duration} to milliseconds.
+   *
+   * <p>Note that there could be overflow or loss of precision by making this conversion. See {@link
+   * TimeUnit#convert(long, TimeUnit)} for details.
+   *
+   * @param duration a {@code Duration}.
+   * @return the milliseconds representation of this {@code Duration}.
```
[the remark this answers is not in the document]
songy23: Good catch, done.

api/src/main/java/io/opencensus/common/Duration.java
songy23: Done.

api/src/main/java/io/opencensus/common/Duration.java
sebright: > I'm not sure if we should mention that converting nanoseconds to milliseconds will cause precision loss though (seems users should have known that).

I agree that we don't need to mention the precision loss.

PR #1115, opened 2018-04-06
Add new RPC measure and view constants, deprecate old ones.

songy23:
Update RPC constants according to [specs](https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/gRPC.md) and our meeting today.

Equivalent changes to OC C++: https://github.com/census-instrumentation/opencensus-cpp/pull/138.

PR #1118, opened 2018-04-10
Fix issues in RPC constants

songy23:
As a follow up for https://github.com/census-instrumentation/opencensus-java/pull/1115.


songy23:
Thanks!

PR #1120, opened 2018-04-10
Makes the trace and span ID fields mandatory in binary format

adriancole:
Before, we didn't error when these were missing.

### DIFF
--- a/impl_core/src/main/java/io/opencensus/implcore/trace/propagation/BinaryFormatImpl.java
+++ b/impl_core/src/main/java/io/opencensus/implcore/trace/propagation/BinaryFormatImpl.java
@@ -87,13 +87,15 @@ final class BinaryFormatImpl extends BinaryFormat {
   static final int TRACE_OPTION_FIELD_ID_OFFSET = SPAN_ID_OFFSET + SpanId.SIZE;
 
   private static final int TRACE_OPTIONS_OFFSET = TRACE_OPTION_FIELD_ID_OFFSET + ID_SIZE;
-  private static final int FORMAT_LENGTH =
-      4 * ID_SIZE + TraceId.SIZE + SpanId.SIZE + TraceOptions.SIZE;
+  /** Version, Trace and Span IDs are required fields */
+  private static final int REQUIRED_FORMAT_LENGTH = 3 * ID_SIZE + TraceId.SIZE + SpanId.SIZE;
+  /** Use {@link TraceOptions#DEFAULT} unless its optional field is present */
+  private static final int ALL_FORMAT_LENGTH = REQUIRED_FORMAT_LENGTH + ID_SIZE + TraceOptions.SIZE;
 
   @Override
   public byte[] toByteArray(SpanContext spanContext) {
     checkNotNull(spanContext, "spanContext");
-    byte[] bytes = new byte[FORMAT_LENGTH];
+    byte[] bytes = new byte[ALL_FORMAT_LENGTH];
     bytes[VERSION_ID_OFFSET] = VERSION_ID;
     bytes[TRACE_ID_FIELD_ID_OFFSET] = TRACE_ID_FIELD_ID;
     spanContext.getTraceId().copyBytesTo(bytes, TRACE_ID_OFFSET);
@@ -110,7 +112,7 @@ final class BinaryFormatImpl extends BinaryFormat {
     if (bytes.length == 0 || bytes[0] != VERSION_ID) {
       throw new SpanContextParseException("Unsupported version.");
     }
-    if (bytes.length < FORMAT_LENGTH - ID_SIZE - TraceOptions.SIZE) {
+    if (bytes.length < REQUIRED_FORMAT_LENGTH) {
       throw new SpanContextParseException("Invalid input: truncated");
     }
     // TODO: the following logic assumes that fields are written in ID order. The spec does not say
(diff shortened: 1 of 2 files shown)


impl_core/src/main/java/io/opencensus/implcore/trace/propagation/BinaryFormatImpl.java
```diff
@@ -65,13 +65,13 @@
...
-  private static final int TRACE_ID_FIELD_ID_OFFSET = VERSION_ID_OFFSET + ID_SIZE;
+  static final int TRACE_ID_FIELD_ID_OFFSET = VERSION_ID_OFFSET + ID_SIZE;
   private static final int TRACE_ID_OFFSET = TRACE_ID_FIELD_ID_OFFSET + ID_SIZE;
   private static final byte SPAN_ID_FIELD_ID = 1;
-  private static final int SPAN_ID_FIELD_ID_OFFSET = TRACE_ID_OFFSET + TraceId.SIZE;
+  static final int SPAN_ID_FIELD_ID_OFFSET = TRACE_ID_OFFSET + TraceId.SIZE;
   private static final int SPAN_ID_OFFSET = SPAN_ID_FIELD_ID_OFFSET + ID_SIZE;
   private static final byte TRACE_OPTION_FIELD_ID = 2;
-  private static final int TRACE_OPTION_FIELD_ID_OFFSET = SPAN_ID_OFFSET + SpanId.SIZE;
+  static final int TRACE_OPTION_FIELD_ID_OFFSET = SPAN_ID_OFFSET + SpanId.SIZE;
```
bogdandrutu: Use `@DefaultVisibilityForTesting`

adriancole:
>
> Use @DefaultVisibilityForTesting
>
done

impl_core/src/main/java/io/opencensus/implcore/trace/propagation/BinaryFormatImpl.java
```diff
@@ -96,25 +106,36 @@ public SpanContext fromByteArray(byte[] bytes) throws SpanContextParseException
...
+    if (bytes[pos] == SPAN_ID_FIELD_ID) {
+      spanId = SpanId.fromBytes(bytes, pos + ID_SIZE);
+      pos += ID_SIZE + SpanId.SIZE;
+    } else {
+      throw new SpanContextParseException("Invalid input: expected span ID at offset " + pos);
+    }
+    // The trace options field is optional. However, when present, it should be valid.
+    if (bytes.length > pos) {
+      if (bytes[pos] != TRACE_OPTION_FIELD_ID) {
+        throw new SpanContextParseException(
```
bogdandrutu: I would not throw an exception in this case. If in the future we want to extend the fields this let's us do that. What do you think?

adriancole:
You mean to say you want to allow skipping option field for a yet undefined
field? Like field number 4.

Or do you mean out of order fields..

Either way seems to imply a version bump,

If so wouldnt that require a version bump? unless version 0 is extensible
which is ok but we should clarify

bogdandrutu:
Yes I mean like field id number 4. I think we are consider version 0 extensible in that regard. Where do you want to clarify?

adriancole:
>
> Yes I mean like field id number 4. I think we are consider version 0
> extensible in that regard. Where do you want to clarify?
>
ok so the current thing breaks if field 1 isn't followed by field 2. I
think we should clarify.

are fields 0 and 1 are fixed position? or can they be out of order?
fields that can be skipped (ex 2) imply a field number 3 can happen this
seeing field 2. Is that right?
do we allow only skipping? or out-of-order

Basically if we have a hybrid offset type of format we should be clear
about that in the spec and quite specific about it.

The code before didn't really operate in a way that suggests anything
besides fixed offset, so if we are changing it we should be clear about
what flexibility so that we can write tests etc.

adriancole:
most concretely, I want to be able to complete this change which was
intended to be a quick bit of volunteer help.

https://github.com/census-instrumentation/opencensus-specs/blob/master/encodings/BinaryEncoding.md#deserialization-rules
says all things are optional, which this code doesnt expect. It doesn't
describe if out of order in a format is ok. I will leave this for someone
else.

I will just skip the options field for now and make a comment.

adriancole:
added another commit with a bunch of TODOs for follow up.

As I'm trying to work on portability, and this stuff came up during
testing, and likely there aren't others working on non-census code
reading census data.. would be nice if we clear this up somehow.

impl_core/src/main/java/io/opencensus/implcore/trace/propagation/BinaryFormatImpl.java
```diff
@@ -96,25 +110,38 @@ public SpanContext fromByteArray(byte[] bytes) throws SpanContextParseException
     if (bytes.length == 0 || bytes[0] != VERSION_ID) {
       throw new SpanContextParseException("Unsupported version.");
     }
-    TraceId traceId = TraceId.INVALID;
-    SpanId spanId = SpanId.INVALID;
+    if (bytes.length < FORMAT_LENGTH - ID_SIZE - TraceOptions.SIZE) {
```
bogdandrutu: Maybe use something like:
private static final int REQUIRED_FORMAT_LENGHT = 3 * ID_SIZE + TraceId.SIZE + SpanId.SIZE;
private static final int ALL_FORMAT_LENGHT = REQUIRED_FORMAT_LENGHT + ID_SIZE + TraceOption.SIZE;

adriancole: sgtm doing this now

adriancole: pushed. looks nicer and easier to understand now. good call

adriancole:
thanks for the review folks

PR #1136, opened 2018-04-17
Checker Framework: 2.4.0 -> 2.5.0

sebright:
This commit upgrades the Checker Framework but continues to use version 2.4.0 of
the Checker Framework compiler, since 2.4.0 is the last version that supports
annotations in comments.  It should continue to work with newer versions of the
Checker Framework, though.  See
https://github.com/census-instrumentation/opencensus-java/pull/1112#issuecomment-381366366.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -206,10 +206,17 @@ subprojects {
 
     dependencies {
         if (useCheckerFramework) {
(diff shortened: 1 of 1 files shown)


sebright:
Thanks!

Issue #1113, opened 2018-04-05
Consider removing dependency on Guava from the implementations and exporters.


sebright2:
Removing the dependency on Guava would help avoid version incompatibilities.  It would also let us avoid shading Guava, which would increase jar size.  This issue is a followup to the comments in #1111, especially https://github.com/census-instrumentation/opencensus-java/pull/1111#issuecomment-378809890.

koalalam:
Any ETA on this?

In short term, do you consider using a newer version of Guava? Version "29.0-android" is a bit old and comes with vulnerability.

https://mvnrepository.com/artifact/com.google.guava/guava
https://github.com/census-instrumentation/opencensus-java/blob/master/build.gradle#L165

PR #1137, opened 2018-04-18
Control package dependencies with Checkstyle.

sebright:
This commit uses the Checkstyle ImportControl feature
(http://checkstyle.sourceforge.net/config_imports.html#ImportControl) to specify
the allowed imports for each package.  Specifying imports has several benefits:

- It makes some parts of the design more concrete.  For example, it shows that
  `stats` can depend on `tags`, but `tags` shouldn't depend on `stats` or
  `trace`.

- We can use it to restrict access to `internal` packages.

- It can help us remove the dependency on Guava incrementally.  For example,
  this commit specifies that the API can only depend on
  `com.google.common.io.BaseEncoding`, which should help us avoid accidentally
  adding more uses of Guava to the API.

One downside is that the check seems to only apply to import statements, not
qualified class names.

This commit specifies imports positively, using "allow" XML elements.  If the
lists of imports are too difficult to maintain, especially for packages such as
exporters that have many imports, we could switch to using "disallow" elements
in some places.

__________________________________________________________________

This PR also contains some simple refactoring to fix two existing issues.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -113,8 +113,9 @@ be included in .class files and Javadocs.
 This project uses Checkstyle to specify the allowed dependencies between
 packages, using its ImportControl feature
 (http://checkstyle.sourceforge.net/config_imports.html#ImportControl).
-`buildscripts/import-control.xml` specifies the allowed imports. An error
-messsage such as
+`buildscripts/import-control.xml` specifies the allowed imports and contains
+some guidelines on OpenCensus' inter-package dependencies. An error messsage
+such as
 `Disallowed import - edu.umd.cs.findbugs.annotations.SuppressFBWarnings. [ImportControl]`
 could mean that `import-control.xml` needs to be updated.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -108,6 +108,16 @@ applied to types. However, it uses `javax.annotation.Nullable` in API method
 signatures whenever possible, so that the annotations can be uncommented and
 be included in .class files and Javadocs.
 
+### Checkstyle import control
+
+This project uses Checkstyle to specify the allowed dependencies between
+packages, using its ImportControl feature
+(http://checkstyle.sourceforge.net/config_imports.html#ImportControl).
+`buildscripts/import-control.xml` specifies the allowed imports. An error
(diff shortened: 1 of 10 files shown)


impl_core/src/main/java/io/opencensus/implcore/trace/propagation/BinaryFormatImpl.java
```diff
@@ -18,7 +18,7 @@
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
-import io.opencensus.internal.DefaultVisibilityForTesting;
+import com.google.common.annotations.VisibleForTesting;
```
HailongWen: I thought we would eventually remove guava from api and impl?

sebright: Yes, we'll probably remove Guava from the implementation if everyone agrees in #1113, but I don't think we've started removing any uses yet.  I think that finding a way to replace all uses of common utilities, like VisibleForTesting, that works across libraries should be a separate change.  I'll open an issue for that.

CONTRIBUTING.md
```diff
@@ -108,6 +108,16 @@ applied to types. However, it uses `javax.annotation.Nullable` in API method
...
+### Checkstyle import control
+
+This project uses Checkstyle to specify the allowed dependencies between
+packages, using its ImportControl feature
+(http://checkstyle.sourceforge.net/config_imports.html#ImportControl).
+`buildscripts/import-control.xml` specifies the allowed imports. An error
+messsage such as
+`Disallowed import - edu.umd.cs.findbugs.annotations.SuppressFBWarnings. [ImportControl]`
+could mean that `import-control.xml` needs to be updated.
+
```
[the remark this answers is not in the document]
sebright: Sure.  I added some guidelines to import-control.xml, and referenced it in CONTRIBUTING.md.

impl_core/src/main/java/io/opencensus/implcore/trace/propagation/BinaryFormatImpl.java
sebright: I opened #1138.

PR #1141, opened 2018-04-20
Adds Tracing.getExportComponent().flushAndShutdown() for use within application shutdown hooks.

cwensel:
This allows a developer to force a flush from within a shutdown hook or other means.

Unfortunately the underlying Disruptor instance only provides a #shutdown() call, not a flush, or a public method for testing for backlog. Thus _shutdown_ has propagated up to the above api call.

I did not add a test case, happy to discuss how this would be implemented reliably.


api/src/main/java/io/opencensus/trace/export/SampledSpanStore.java
```diff
@@ -132,6 +132,13 @@ static SampledSpanStore newNoopSampledSpanStore() {
   @PublicForTesting
   public abstract Set<String> getRegisteredSpanNamesForCollection();
 
+  /**
+   * This forces any underlying event queue to flush any pending events and shutdown and handlers.
+   *
+   * @since 0.13
+   */
+  public abstract void shutdown();
```
HailongWen: I am not so sure if we should expose this API in `RunningSpanStore`.
For me, adding a `shutdown()` in `ExportComponent` is fine - that could be the entry for users and it hides all the details.

sebright: What would a shutdown method do in RunningSpanStore?  I don't think it has a separate thread.

HailongWen: Maybe make it protected?

HailongWen: @sebright sorry, I meant `SampledSpanStore` in the first place. But for `RunningSpanStore`, a `shutdown` does make sense. For a `RunningSpanStore`, one might want to close, set proper status and sample option and export some important running spans before the program exit unexpectedly - I made that up :) 

Back to the topic, what I originally meant was that we should probably only expose `ExportComponent.shutdown()`. Other subsequent `shutdown` methods should be hide from users (by making them protected, or mark as `Internal`).

sebright: Yes, I like that idea.  Users probably wouldn't need to call shutdown on only one part of the TraceComponent.  I think that the new methods on SampledSpanStore and SpanExporter are only called from the implementation anyway, so they could be completely removed from the superclasses under api/.

impl/src/main/java/io/opencensus/impl/internal/DisruptorEventQueue.java
```diff
@@ -143,6 +143,17 @@ public void enqueue(Entry entry) {
     }
   }
 
+  /**
+   * Shuts down the underlying disruptor.
+   *
+   * <p>Unfortunately there is no underlying public flush mechanism, without it there is a race
+   * condition in the ring buffer where it can hold events into the jvm shutdown.
```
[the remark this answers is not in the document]
cwensel: I removed the comment. It was lamenting the lack of a flush vs a shutdown, unclearly.

api/src/main/java/io/opencensus/trace/export/SampledSpanStore.java
```diff
@@ -132,6 +132,13 @@ static SampledSpanStore newNoopSampledSpanStore() {
   @PublicForTesting
   public abstract Set<String> getRegisteredSpanNamesForCollection();
 
+  /**
+   * This forces any underlying event queue to flush any pending events and shutdown and handlers.
```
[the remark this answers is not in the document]
cwensel: the underlying implementation is simply shutting down the event queue. there are no handlers.

impl/src/main/java/io/opencensus/impl/internal/DisruptorEventQueue.java
```diff
@@ -143,6 +143,17 @@ public void enqueue(Entry entry) {
...
+   */
+  @Override
+  public void shutdown() {
+    disruptor.shutdown();
```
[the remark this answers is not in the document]
cwensel: In theory, but this would require a check on every publish/enqueue call from what I can tell, unless we swap something out with a Noop impl. 

but this is a shutdown, its intended to be run during the jvm shutdown, so new spans should stop at some point naturally. thus this continuing to accept spans until there are no more might be considered a virtue. 

if the jvm never shuts down, there is clearly an issue elsewhere (the thing accepting requests that spawns spans is not itself shutdown).

impl_core/src/main/java/io/opencensus/implcore/trace/export/SpanExporterImpl.java
```diff
@@ -188,5 +193,18 @@ public void run() {
         }
       }
     }
+
+    void flush() {
```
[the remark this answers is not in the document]
cwensel: ok, i left this out for now. looks like there are three options

- copy the test class, and change the delay (removing the other tests)
- add a method on SpanExporter to change the delay via a test visible method, change the delay in the new test
- refactor the test class so the setup is in setup() and subclass to change the delay and other params

i’m sure you have something better in mind, so let me know.

api/src/main/java/io/opencensus/trace/export/SampledSpanStore.java
cwensel: i’ll wait on guidance after you have review the new commit.

cwensel:
pushed changes per some of the comments. 

still needs a test, see above.

unsure if we should still hide the shutdown call chain. waiting on guidance.

impl/src/main/java/io/opencensus/impl/internal/DisruptorEventQueue.java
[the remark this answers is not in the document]
sebright: I meant that I don't think that this class should continue to enqueue Entries on the Disruptor after the Disruptor has been shut down.  That could cause the thread that recorded the instrumentation data to block forever.  I think it would be better to discard the data that cannot be processed, and log an error, to avoid interfering with the application.

I'm not sure how we could make this class continue to accept data after shutdown without a method supporting similar behavior on the Disruptor.

/cc @bogdandrutu

api/src/main/java/io/opencensus/trace/export/SampledSpanStore.java
sebright: I would prefer to avoid exposing new APIs until they are needed.  Did you only add the methods so that they could be called by `ExportComponent.shutdown`, or is there a use case for only calling `shutdown` or `flush` on a subset of the components?

cwensel:
ok, added the test

impl_core/src/test/java/io/opencensus/implcore/trace/export/SpanExporterImplTest.java
```diff
@@ -170,8 +194,40 @@ public void exportSpansToMultipleServices() {
...
+    // Need to check this because otherwise the variable span1 is unused, other option is to not
+    // have a span1 variable.
+    assertThat(exported).doesNotContain(span1.toSpanData());
+    assertThat(exported).containsExactly(span2.toSpanData());
+  }
+
+  @Test
+  public void exportNotSampledSpansFlushed() {
+    // Set the export delay to a very long value in order to confirm the #flush() below works
+    SpanExporterImpl spanExporter = SpanExporterImpl.create(4, Duration.create(Long.MAX_VALUE, 0));
```
sebright: It looks like `Duration.create` has a limit on the number of seconds, and it uses zero when the value is greater than 315576000000L:

https://github.com/census-instrumentation/opencensus-java/blob/773c7638f448bdb6d33d1c30bc8ec4ba5450da18/api/src/main/java/io/opencensus/common/Duration.java#L55-L56

I created an issue to improve the API: #1179

impl_core/src/test/java/io/opencensus/implcore/trace/export/SpanExporterImplTest.java
```diff
@@ -170,8 +194,40 @@ public void exportSpansToMultipleServices() {
...
+    StartEndHandler startEndHandler =
+        new StartEndHandlerImpl(spanExporter, runningSpanStore, null, new SimpleEventQueue());
+
+    spanExporter.registerHandler("test.service", serviceHandler);
+
+    SpanImpl span1 = createNotSampledEndedSpan(startEndHandler, SPAN_NAME_1);
+    SpanImpl span2 = createSampledEndedSpan(startEndHandler, SPAN_NAME_2);
```
sebright: I think this test could be simplified by removing the non-sampled span, since that was tested in `exportNotSampledSpans`.

cwensel:
cleaned up test.

if looks good, I can squash and rebase to resolve any conflicts.

sebright:
The test looks good.  I don't mind if you rebase.

I think that there are only two remaining review comments, related to removing the public methods on SpanExporter and SampledSpanStore (https://github.com/census-instrumentation/opencensus-java/pull/1141#discussion_r185329346), and preventing threads from blocking when enqueueing events after Disruptor shutdown (https://github.com/census-instrumentation/opencensus-java/pull/1141#discussion_r185327270).

impl/src/main/java/io/opencensus/impl/internal/DisruptorEventQueue.java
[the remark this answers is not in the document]
sebright: I think that the number of entries is limited by the buffer size, but I want to avoid having the call to `enqueue` not return when the buffer is full.  I prefer your suggestion to swap the disruptor for a no-op version.  The no-op `enqueue` method could also log a warning the first time it is called.

cwensel:
pushed a noop handler for the enqueuing. let me know if I missed anything else.

impl/src/main/java/io/opencensus/impl/internal/DisruptorEventQueue.java
```diff
@@ -99,6 +103,8 @@
   // Ring Buffer for the {@link Disruptor} that underlies the queue.
   private final RingBuffer<DisruptorEvent> ringBuffer;
 
+  private DisruptorEnqueuer enqueuer = null;
```
sebright: This field should be volatile, since it could be accessed by multiple threads.

impl/src/main/java/io/opencensus/impl/internal/DisruptorEventQueue.java
```diff
@@ -134,13 +154,31 @@ public static DisruptorEventQueue getInstance() {
...
+  @Override
+  public void shutdown() {
+    enqueuer =
+        new DisruptorEnqueuer() {
+          boolean logged = false;
+
+          @Override
+          public void enqueue(Entry entry) {
+            if (!logged) {
+              logged = true;
```
sebright: This access to `logged` isn't synchronized.  I'm not sure whether it would be better to use a lock, a volatile variable, or AtomicBoolean.  Only using volatile would allow duplicate log statements, but that probably isn't an issue.

cwensel:
rebased, didn’t squash.

went w/ AtomicBoolean, since it isn’t initialized unless `shutdown()` is called and seems a little cleaner.

CHANGELOG.md
```diff
@@ -1,13 +1,15 @@
 ## Unreleased
 
 ## 0.13.2 - 2018-05-08
+- Adds Tracing.getExportComponent().shutdown() for use within application shutdown hooks.
```
sebright: 0.13.2 was already released, so this should go under unreleased.

cwensel: ugh, sorry, wasn’t paying attention. fixed.

impl/src/main/java/io/opencensus/impl/internal/DisruptorEventQueue.java
cwensel: ah, missed this one. fixed.

CHANGELOG.md
```diff
@@ -1,13 +1,15 @@
...
 - Map http attributes to Stackdriver format (fix [#1153](https://github.com/census-instrumentation/opencensus-java/issues/1153)).
 
 ## 0.13.1 - 2018-05-02
 - Fix a typo on displaying Aggregation Type for a View on StatsZ page.
 - Set bucket bounds as "le" labels for Prometheus Stats exporter.
 
 ## 0.13.0 - 2018-04-27
+=======
```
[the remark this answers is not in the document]
cwensel: fixed.

sebright:
Thanks!  I'll merge it when the build passes.  Do you mind if I squash it?

cwensel:
please do squash it. i didn’t for fear of losing track of all the pending change comments.

cwensel:
ok, if i change `@SuppressWarnings({"unchecked", "nullness”})` back to `@SuppressWarnings({"unchecked”})` the verGJF check passes.

i’m using both the intellij google-java-format plugin and google java format style, which have slightly different results, but neither fixes the above issue.

i’m unsure how to make verGJF more informative on the failure.

sebright:
I'm surprised that `./gradlew verGJF` is giving different results in CI.  I thought it would behave consistently, since the Gradle version is fixed, and build.gradle specifies the version for google-java-format.  Maybe we should run `./gradlew goGJF` in CI and then print the diff to help with debugging.

I tried running `./gradlew goGJF` locally, and the diff didn't include the line with SuppressWarnings:

```diff
diff --git a/impl/src/main/java/io/opencensus/impl/internal/DisruptorEventQueue.java b/impl/src/main/java/io/opencensus/impl/internal/DisruptorEventQueue.java
index 95c5d14d1..5145ca3b7 100644
--- a/impl/src/main/java/io/opencensus/impl/internal/DisruptorEventQueue.java
+++ b/impl/src/main/java/io/opencensus/impl/internal/DisruptorEventQueue.java
@@ -157,13 +157,11 @@ public final class DisruptorEventQueue implements EventQueue {
   @Override
   public void enqueue(Entry entry) {
     enqueuer.enqueue(entry);
   }
 
-  /**
-   * Shuts down the underlying disruptor.
-   */
+  /** Shuts down the underlying disruptor. */
   @Override
   public void shutdown() {
     enqueuer =
         new DisruptorEnqueuer() {
           final AtomicBoolean logged = new AtomicBoolean(false);
@@ -188,12 +186,11 @@ public final class DisruptorEventQueue implements EventQueue {
   // An event in the {@link EventQueue}. Just holds a reference to an EventQueue.Entry.
   private static final class DisruptorEvent {
 
     // TODO(bdrutu): Investigate if volatile is needed. This object is shared between threads so
     // intuitively this variable must be volatile.
-    @Nullable
-    private volatile Entry entry = null;
+    @Nullable private volatile Entry entry = null;
 
     // Sets the EventQueueEntry associated with this DisruptorEvent.
     void setEntry(@Nullable Entry entry) {
       this.entry = entry;
     }
```

cwensel:
ah, goGJF, new to me. will try that.

PR #1149, opened 2018-04-22
trace exporter for elasticsearch

malike:
javadocs javadocs

compilejava compilejava

gojf gojf

@bogdandrutu I'll be glad if you could help me identify the issues.

### DIFF
--- a/exporters/stats/stackdriver/README.md
+++ b/exporters/stats/stackdriver/README.md
@@ -70,9 +70,9 @@ public class MyMainClass {
 #### Set Monitored Resource for exporter
 
 By default, Stackdriver Stats Exporter will try to automatically detect the environment if your 
-application is running on GCE or GKE, and generate a corresponding Stackdriver GCE/GKE monitored 
-resource. For GKE particularly, you may want to set up some environment variables so that Exporter 
-can correctly identify your pod, cluster and container. Follow the Kubernetes instruction 
+application is running on GCE, GKE or AWS EC2, and generate a corresponding Stackdriver GCE/GKE/EC2 
+monitored resource. For GKE particularly, you may want to set up some environment variables so that 
+Exporter can correctly identify your pod, cluster and container. Follow the Kubernetes instruction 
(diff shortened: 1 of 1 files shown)


exporters/trace/elasticsearch/src/main/java/io/opencensus/exporter/trace/elasticsearch/ElasticsearchTraceExporter.java
```diff
@@ -0,0 +1,243 @@
...
+          elasticsearchConfiguration.getElasticsearchType() != null,
+          "Elasticsearch type not specified");
+      checkState(
+          elasticsearchConfiguration.getElasticsearchUrl() != null,
+          "Elasticsearch URL not specified");
+      registerElasticsearchHandler(new ElasticsearchTraceHandler(elasticsearchConfiguration));
+    }
+  }
+
+  private static void registerElasticsearchHandler(Handler elasticsearchHandler) {
```
bogdandrutu: To fix this, the quick solution I can think of:
@SuppressWarnings("nullness")

@sebright any better solution? I think the checker is wrong here.

sebright: I don't think that the checker looks at synchronization, so it doesn't know that "handler" can't be changed by another thread between lines 105 and 106.  I would register "elasticsearchHandler" instead of "handler" to make it clear that this method is always registering the input handler.

malike: oh ok, makes sense

bogdandrutu: I think personally that this function can be removed:
1) In the current code is called only once.
2) It is error prone, for example currently you synchronized twice which is not a big problem because in java synchronized blocks are reentrant, but I don't think it is a good practice.

sebright:
@malike Let us know when you're ready for us to review the rest of the code.

Issue #785, opened 2017-11-07
Check for null status/annotations/etc in Span


bogdandrutu:
We should check for nulls in the Span (blankspan, noopspan, etc.);

sebright:
[NullPointerTester](https://static.javadoc.io/com.google.guava/guava-testlib/19.0/com/google/common/testing/NullPointerTester.html) might be useful for checking all of the classes in the API.

dvfeinblum:
Hey @sebright. I'm looking to get under opencensus's hood and was wondering if this would be a good first issue for a would-be contributor! Thanks.

sebright:
Yes, I think this is a good first issue.  Let us know if you have any questions!

dvfeinblum:
As it so happens, I do have a couple questions:

1. Is it safe to assume that adding null checking to the abstract `Span` class will trickle down and cover the actual Spans themselves (blankspan, noopspan, etc.). Basically, is this the right place to add [this](https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/trace/Span.java) checking?

2. When the original comment says that "[w]e should check for nulls in the Span," what are they specifically referring to? Is this the sort of thing that a few `@NonNull` tags in the implementations could rectify or is there something more complex implied here?

Hope these questions make sense; I'm a bit unfamiliar with the codebase and the original comment is a little sparse with information. Any other helpful hints you have are more than welcome, too. Thanks!

sebright:
> Is it safe to assume that adding null checking to the abstract Span class will trickle down and cover the actual Spans themselves (blankspan, noopspan, etc.). Basically, is this the right place to add this checking?

I think that the null checks should be added in all of the methods that implement the methods in Span.  For example, [Span.putAttribute](https://github.com/census-instrumentation/opencensus-java/blob/8ff7ff9abec6b7abe95391fb450d99866c6a2c3a/api/src/main/java/io/opencensus/trace/Span.java#L95) is implemented in Span, BlankSpan, and SpanImpl, so all three implementations should check that the key and value are not null.

> When the original comment says that "[w]e should check for nulls in the Span," what are they specifically referring to? Is this the sort of thing that a few `@NonNull` tags in the implementations could rectify or is there something more complex implied here?

[...]

PR #1150, opened 2018-04-24
#785 - Check for null status/annotations/etc in Span

dvfeinblum:
This PR adds null checking to classes that implement Span. Specifically,

- `BlankSpan`
- `NoopSpan`
- `Span`
- `SpanImpl`

For the latter, I had to use `Preconditions.checkNotNull` because `io.opencensus.internal` shouldn't be imported into `ImplCore`.

Let me know if you have any questions!

### DIFF
--- a/impl_core/src/main/java/io/opencensus/implcore/trace/SpanImpl.java
+++ b/impl_core/src/main/java/io/opencensus/implcore/trace/SpanImpl.java
@@ -20,6 +20,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Preconditions.checkState;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
 import com.google.common.collect.EvictingQueue;
 import io.opencensus.common.Clock;
 import io.opencensus.implcore.internal.CheckerFrameworkUtils;
@@ -281,6 +282,7 @@ public final class SpanImpl extends Span implements Element<SpanImpl> {
 
   @Override
   public void putAttributes(Map<String, AttributeValue> attributes) {
+    Preconditions.checkNotNull(attributes, "attributes");
     if (!getOptions().contains(Options.RECORD_EVENTS)) {
       return;
     }
@@ -295,6 +297,8 @@ public final class SpanImpl extends Span implements Element<SpanImpl> {
 
   @Override
   public void addAnnotation(String description, Map<String, AttributeValue> attributes) {
+    Preconditions.checkNotNull(description, "description");
+    Preconditions.checkNotNull(attributes, "attribute");
     if (!getOptions().contains(Options.RECORD_EVENTS)) {
       return;
     }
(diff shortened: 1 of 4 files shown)


impl_core/src/main/java/io/opencensus/implcore/trace/SpanImpl.java
```diff
@@ -507,6 +516,8 @@ private AttributesWithCapacity(int capacity) {
     // Users must call this method instead of put to keep count of the total number of entries
     // inserted.
     private void putAttribute(String key, AttributeValue value) {
+      Utils.checkNotNull(key, "key");
```
sebright: I don't think that the null checks in private methods are necessary, since the arguments should already have been validated in the public methods.

impl_core/src/main/java/io/opencensus/implcore/trace/SpanImpl.java
```diff
@@ -313,6 +319,7 @@ public void addAnnotation(String description, Map<String, AttributeValue> attrib
 
   @Override
   public void addAnnotation(Annotation annotation) {
+    Utils.checkNotNull(annotation, "annotation");
```
sebright: There was already a null check later in the method, though I think that validating arguments at the start of the method is clearer.

impl_core/src/main/java/io/opencensus/implcore/trace/SpanImpl.java
```diff
@@ -348,6 +355,7 @@ public void addNetworkEvent(io.opencensus.trace.NetworkEvent networkEvent) {
 
   @Override
   public void addLink(Link link) {
+    Utils.checkNotNull(link, "link");
```
sebright: Same here

buildscripts/import-control.xml
```diff
@@ -139,6 +139,7 @@ General guidelines on imports:
     <allow pkg="com.google.common"/>
     <allow pkg="io.opencensus.common"/>
     <allow pkg="io.opencensus.implcore"/>
+    <allow pkg="io.opencensus.internal"/>
```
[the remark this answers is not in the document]
dvfeinblum: Ah okay. I'll remove this and switch to `Preconditions` instead for that class.

impl_core/src/main/java/io/opencensus/implcore/trace/SpanImpl.java
dvfeinblum: Agree completely. I avoided privates and deprecated methods but missed this one!

impl_core/src/main/java/io/opencensus/implcore/trace/SpanImpl.java
dvfeinblum: Oooh good catch. I'll remove the second one.

impl_core/src/main/java/io/opencensus/implcore/trace/SpanImpl.java
dvfeinblum: Ditto.

impl_core/src/main/java/io/opencensus/implcore/trace/SpanImpl.java
sebright: Did you remove the wrong null check?  The null checks in the two private methods are still there.

dvfeinblum: Ouf. Sorry about that. I could sworn I changed all of them but definitely missed these two. I just force pushed some changes that should rectify that issue!

PR #1183, opened 2018-05-04
Map http attributes to the stackdriver format, resolves #1153

Sebruck:
@sebright  WDYT?

PR #1185, opened 2018-05-08
Map http attributes to the stackdriver format, resolves #1153 (#1183)

### LINKED ISSUE #1153: HTTP Attributes are not mapped in Stackdriver exporter

0.12.3

8

### What did you do?
Adding http attributes according to the [spec](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md) to a span (http.host, http.method, ...), see here: https://github.com/Sebruck/opencensus-scala/blob/master/akka-http/src/main/scala/com/github/sebruck/opencensus/akka/http/trace/HttpAttributes.scala#L16

### What did you expect to see?
The spec defines that those attributes should be mapped when exporting to stackdriver, e.g.:
`http.host` -> `/http/host`

But this seems not to happen, in Stackdriver I still see the initial value.

I also can't find any code doing this in the `opencensus-exporter-trace-stackdriver` module.

### What did you see instead?
The attributes should be mapped as defined in the specs.

Is this a missing feature or a bug?

songy23:
(cherry picked from commit a1a6cf16ec5e6ec3c1e08e96b1e8dd9a45dce13e)

/cc @Sebruck

### DIFF
--- a/exporters/trace/stackdriver/src/main/java/io/opencensus/exporter/trace/stackdriver/StackdriverV2ExporterHandler.java
+++ b/exporters/trace/stackdriver/src/main/java/io/opencensus/exporter/trace/stackdriver/StackdriverV2ExporterHandler.java
@@ -23,6 +23,7 @@ import com.google.auth.Credentials;
 import com.google.cloud.trace.v2.TraceServiceClient;
 import com.google.cloud.trace.v2.TraceServiceSettings;
 import com.google.common.annotations.VisibleForTesting;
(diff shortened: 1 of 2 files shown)


songy23:
Ping @googlebot

songy23:
I just realized branch `v0.13.x` doesn't have Kokoro build scripts yet. I created https://github.com/census-instrumentation/opencensus-java/pull/1186 to cherry-pick the build scripts first.

PR #1186, opened 2018-05-08
Cherry-pick Kokoro build scripts for v0.13.x branch

songy23:


songy23:
Talked with @sebright, currently the Kokoro build has the exact same check as travis-ci, so it's probably not necessary on the release branch (it should be more useful on the master branch).

PR #1192, opened 2018-05-12
Remove warning suppression in DisruptorEventQueue.

sebright:
This PR fixes a nullness warning by splitting the constructor into a factory method and a constructor, to ensure that the DisruptorEventQueue is fully initialized before any events are enqueued.  It also fixes a varargs warning by creating an explicit array.


sebright:
Thanks!

PR #1196, opened 2018-05-14
Remove PublicForTesting from Status.CanonicalCode.toStatus() (fixes #976).

### LINKED ISSUE #976: Should we remove @VisibleForTesting from Status.CanonicalCode.toStatus()?
`Status.CanonicalCode.toStatus()` is already used by the Stackdriver stats exporter to convert status types, so it may be better to treat it as part of the API.

https://github.com/census-instrumentation/opencensus-java/blob/67b1f37cd91688cf42fe2ff1adc8a42ef32477d9/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporterWorker.java#L138-L142

https://github.com/census-instrumentation/opencensus-java/blob/67b1f37cd91688cf42fe2ff1adc8a42ef32477d9/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporterWorker.java#L186-L190

If we don't remove `@VisibleForTesting`, then I think we should also add `@Internal`.  `@Internal` would appear in Javadocs and warn users that the method isn't part of the API.

/cc @bogdandrutu

sebright:
This method is already used by the Stackdriver stats exporter to convert status
types, so it should be part of the API.

### DIFF
--- a/api/src/main/java/io/opencensus/trace/Status.java
+++ b/api/src/main/java/io/opencensus/trace/Status.java
@@ -16,7 +16,6 @@
 
 package io.opencensus.trace;
 
-import io.opencensus.internal.PublicForTesting;
 import io.opencensus.internal.Utils;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


sebright:
Thanks!

PR #1198, opened 2018-05-15
Throw IllegalArgumentException when Duration.create receives invalid arguments.

### LINKED ISSUE #1179: Duration.create fails silently when given invalid arguments.
`Duration.create` returns a "zero" value when the arguments are invalid:

https://github.com/census-instrumentation/opencensus-java/blob/773c7638f448bdb6d33d1c30bc8ec4ba5450da18/api/src/main/java/io/opencensus/common/Duration.java#L54-L65

I think this API would be easier to use if it either truncated the values to give a value closer to what the user probably wanted or threw an IllegalArgumentException.  Here are two possible ways to change the behavior:

1. If it is important to avoid throwing exceptions, we could truncate the seconds and nanos to their minimum and maximum values.  We could throw an IllegalArgumentException if the signs of the seconds and nanos are inconsistent, since it would be unclear what sign the user wanted.

2. We could throw an IllegalArgumentException if the seconds or nanos are out of bounds or have inconsistent sign.  This is more consistent with other parts of the API that enforce all class invariants with `checkArgument(...)`.

sebright:
Fixes #1179.  Throwing IllegalArgumentException for invalid arguments is more
consistent with the rest of the opencensus-java API.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Unreleased
 - Adds Tracing.getExportComponent().shutdown() for use within application shutdown hooks.
+- `Duration.create` now throws an `IllegalArgumentException` instead of
+  returning a zero `Duration` when the arguments are invalid.
 
 ## 0.13.2 - 2018-05-08
 - Map http attributes to Stackdriver format (fix [#1153](https://github.com/census-instrumentation/opencensus-java/issues/1153)).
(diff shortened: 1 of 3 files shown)


sebright:
Rebased, since CI was stuck.

sebright:
Thanks!

PR #1199, opened 2018-05-15
Throw IllegalArgumentException when Timestamp.create receives invalid arguments.

sebright:
Throwing IllegalArgumentException for invalid arguments is more consistent with
the rest of the opencensus-java API.  This commit also fixes a test that created
an invalid Timestamp.

___________________________________________________________________

This change is similar to #1198, but applied to Timestamp.  @songy23 is this the correct fix for StackdriverExportUtilsTest?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 - Adds Tracing.getExportComponent().shutdown() for use within application shutdown hooks.
 - `Duration.create` now throws an `IllegalArgumentException` instead of
   returning a zero `Duration` when the arguments are invalid.
+- `Timestamp.create` now throws an `IllegalArgumentException` instead of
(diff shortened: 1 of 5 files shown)


songy23:
Let me try sending a negative timestamp to Stackdriver backend.

songy23:
Stackdriver doesn't allow negative timestamps (actually they don't allow timestamp older than a day from current time). Can you add a check in `StackdriverExportUtils.convertTimestamp` so that it throws IllegalArgumentException for negative case?

sebright:
If Stackdriver can't handle negative timestamps, shouldn't the exporter continue to convert the negative timestamps to zero, as on master?  I thought that the only problem was that the test used a negative value for nanoseconds, when negative timestamps should still use positive nanoseconds.

sebright:
Thanks!

PR #1208, opened 2018-05-23
Suggest avoiding force pushing to pull requests in CONTRIBUTING.md.

sebright:


rghetia:
is it fair to say that each PR should only contain related changes. If there are two or more unrelated changes then create a separate PR. This way two unrelated change will not get squashed into 1.

sebright:
I think we should allow pull requests to contain commits for multiple independent code changes that are part of the same feature or bug fix.  For example, someone could refactor something in one commit to prepare for a behavior change in the next commit, where the refactoring doesn't make sense without the following change.  I think it is important to keep those commits separate, for a more informative history on `master`.  When deciding whether to squash a pull request, I think we should err on the side of not squashing, to avoid accidentally combining independent changes, especially if the author has put a lot of effort into separating out distinct code changes and describing each one in the commit message.  I think we should just add a note to CONTRIBUTING.md that we may squash the pull request if it doesn't have a clean history.

bogdandrutu:
If you suggest people to not force push then if I have a comment to fix a typo I will create and push that commit. That adds 0 value to the history, more than that it makes history very hard to follow.

Also I disagree with you because if that is the case (the one you described) then the user should create different PRs one for cleanup and one for the feature. I think this is the correct thing to do.

adriancole:
I think the first part of this change is fine if folks want to nag people
to make sure they don't force push etc.

otoh, change should be well organized once it is merged. I often ask users
to create a coherent commit (or in rare cases commits) before a change is
completed.

very long change (where review takes weeks and lots of things) are what
this is optimizing for, and there's a lot of change which I think it
doesn't really help too much, but not hurt either.

A spam of commits to fix typo vs a force push hurts most when there's a lot
of code in the PR iotw. Smaller changes meh.. almost rather folks force
push.

adriancole:
tl;dr; I hope this isn't disabling force push, rather a suggestion

adriancole:
in case not obvious.. the force push allows you to teach folks how to
write a good commit message.. and have them do it.. vs you doing it
for them. It also allows committers to fixup things before things are
ready to merge.

multiple commits often should be squashed into one, and so that's why
sometimes squash is good on our side, but sometimes rebase is better
if there are two changes.

volunteers are not always willing to pull other changes out to a
different pull request (it is busy work), so that's one reason why
I'll often force push a user's branch to help them with that..
basically use 2 commits in this case. Other times, the code is in
nowhere near a state that a diff helps with (think formatting), so
force push is significantly better than a commit that changes every
line, or where people accidentally changed the wrong files.

sebright:
@adriancole I only meant it as a suggestion, and I would be happy to reword it.  I also wasn't thinking of the other cases where force pushing makes sense, such as fixing typos, fixing commit messages, or refactoring the whole change.  The main reason I made the PR was to suggest making a new commit when the best way to review the new changes is by looking at the diff from the old changes, for example, when the author changes 10% of a medium to large PR.

I also don't want to discourage anyone from cleaning up the branch history after the review, though I'm not sure of the best way to avoid having a maintainer merge the PR before it is ready.

> volunteers are not always willing to pull other changes out to a
different pull request (it is busy work), so that's one reason why
I'll often force push a user's branch to help them with that..

Do you mean that you rebase the author's branch to clean up history before merging it yourself?  Does that work well?  That sounds like a good solution if the author doesn't clean up the history and there are some commits that shouldn't be combined with the Github squash option.

adriancole:
>
> @adriancole <https://github.com/adriancole> I only meant it as a
> suggestion, and I would be happy to reword it. I also wasn't thinking of
> the other cases where force pushing makes sense, such as fixing typos,
> fixing commit messages, or refactoring the whole change. The main reason I
> made the PR was to suggest making a new commit when the best way to review
> the new changes is by looking at the diff from the old changes, for
> example, when the author changes 10% of a medium to large PR.
>
> I also don't want to discourage anyone from cleaning up the branch history
> after the review, though I'm not sure of the best way to avoid having a
> maintainer merge the PR before it is ready.
>
Cool. I think we are probably on the same page.. and agree yeah medium to
large PR it is actually somewhat antisocial to always force push (something
I am guilty of and should do this better)


> volunteers are not always willing to pull other changes out to a
> different pull request (it is busy work), so that's one reason why
> I'll often force push a user's branch to help them with that..
>
> Do you mean that you rebase the author's branch to clean up history before
> merging it yourself? Does that work well? That sounds like a good solution
> if the author doesn't clean up the history and there are some commits that
> shouldn't be combined with the Github squash option.
>
What I've noticed is many still aren't great at git. It is subtle, but
sometimes you can tell someone is at the end of their rope, or that they
only have 30 minutes more on a change and fighting git isn't a great place
to spend that time. So, sometimes I will help cleanup code (by adding to
their change!) so that they can finish the last mile with less effort. Not
sure this makes sense or not as a general practice.

sebright:
I realized that it would be difficult to come up with general, uncontroversial guidelines on when an author should make a new commit or modify existing commits, so I think it would be better to not specify anything.  I'll close this PR for now, though I'm still open to adding guidelines if anyone disagrees and/or has ideas on the wording.

adriancole:
thanks!

PR #1211, opened 2018-05-24
Starts adding a metrics package.

songy23:
Add package info first, will send separate PRs that implement each of the data types in [metrics.proto](https://github.com/census-instrumentation/opencensus-proto/blob/master/opencensus/proto/stats/metrics/metrics.proto).

### DIFF
--- a/api/src/main/java/io/opencensus/stats/metrics/package-info.java
+++ b/api/src/main/java/io/opencensus/stats/metrics/package-info.java
@@ -27,4 +27,5 @@
  * https://github.com/census-instrumentation/opencensus-proto/blob/master/opencensus/proto/stats/metrics/metrics.proto
  * for more details.
  */
+@io.opencensus.common.ExperimentalApi
 package io.opencensus.stats.metrics;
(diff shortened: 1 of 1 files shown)


api/src/main/java/io/opencensus/stats/metrics/package-info.java
```diff
@@ -0,0 +1,30 @@
...
+ * <p>Currently all the public classes under this package are marked as {@link
+ * io.opencensus.common.ExperimentalApi}. This data model may eventually become the wire format for
+ * metrics.
+ *
+ * <p>Please see
+ * https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/Metrics.md and
+ * https://github.com/census-instrumentation/opencensus-proto/blob/master/opencensus/proto/stats/metrics/metrics.proto
+ * for more details.
+ */
+package io.opencensus.stats.metrics;
```
sebright: Should this package be marked with ExperimentalApi, too?

songy23: That's right, added.

bogdandrutu:
Is this the right moment to start splitting the exporter package out of api? I think we should do that.

songy23:
Do you mean splitting the metrics package to a separate artifact?

songy23:
Merging this PR first - I'll create a following PR to move the metrics package to an independent artifact, and we can continue discussion there.

PR #1212, opened 2018-05-25
Start adding log correlation for Stackdriver Logging.

sebright:
/cc @g-easy @bogdandrutu 

This is a rough PR for early feedback.  It adds a library for inserting tracing data into Stackdriver log entries, through the `google-cloud-logging` library.  The tests aren't complete, and it has some TODOs.

I also included a demo application.  It contains SLF4J log statements and OpenCensus tracing instrumentation. It configures logging with a Logback XML configuration and exports logs using the Stackdriver Logging Logback appender. It also exports traces using `opencensus-exporter-trace-stackdriver`, so that Stackdriver can show the log entries associated with each trace.

The demo requires a Google Cloud project with Stackdriver Logging enabled.  It can be run with

```
$ ./gradlew :opencensus-contrib-log-correlation-stackdriver-demo:installDist
$ ./contrib/log_correlation/stackdriver_demo/build/install/opencensus-contrib-log-correlation-stackdriver-demo/bin/OpenCensusTraceLoggingEnhancerDemo
```

Here is a screenshot of a trace with log statements in Stackdriver:
log-correlation-in-stackdriver

The demo is currently a subproject of opencensus-java, but I'm planning to move it.


sebright:
I also mentioned that the log correlation feature is experimental in the Javadoc.

I'm planning to move the demo to https://github.com/census-ecosystem/opencensus-experiments before I merge this PR.

contrib/log_correlation/stackdriver/src/main/java/io/opencensus/contrib/logcorrelation/stackdriver/OpenCensusTraceLoggingEnhancer.java
```diff
@@ -0,0 +1,58 @@
...
+    SpanContext span = tracer.getCurrentSpan().getContext();
+    builder.setTrace(formatTraceId(span.getTraceId()));
+
+    // TODO(sebright): Find the correct way to add span ID and sampling decision.
+    builder.addLabel(SPAN_ID_KEY, span.getSpanId().toLowerBase16());
+    builder.addLabel(SAMPLED_KEY, Boolean.toString(span.getTraceOptions().isSampled()));
+  }
+
+  private static String formatTraceId(TraceId traceId) {
+    // TODO(sebright): Should we cache the project ID, for efficiency?
```
bogdandrutu: Yes please. You should probably cache `"projects/" + projectId + "/traces/"`

contrib/log_correlation/stackdriver/src/main/java/io/opencensus/contrib/logcorrelation/stackdriver/OpenCensusTraceLoggingEnhancer.java
```diff
@@ -0,0 +1,58 @@
...
+    String projectId = ServiceOptions.getDefaultProjectId();
```
bogdandrutu: I would accept the projectId as a configuration argument for this LogginEnhancer. Reason is because users may configure a different projectId for trace so they should use the same value here. And if not set use default.

contrib/log_correlation/stackdriver/src/main/java/io/opencensus/contrib/logcorrelation/stackdriver/OpenCensusTraceLoggingEnhancer.java
```diff
@@ -0,0 +1,58 @@
...
+  private static final String SAMPLED_KEY = "sampled";
+
+  private static final Tracer tracer = Tracing.getTracer();
+
+  /** Constructor to be called by Stackdriver logging. */
+  public OpenCensusTraceLoggingEnhancer() {}
+
+  @Override
+  public void enhanceLogEntry(LogEntry.Builder builder) {
+    SpanContext span = tracer.getCurrentSpan().getContext();
```
Sebruck: Is there any possibility to access the args of a log message within an enhancer and take the span from there (if present)? This would allow people who do not use scoped spans to use this as well.

In Scala we can't use scoped spans in most of the cases since the Future based code is executed on arbitrary threads.

sebright: @Sebruck Thanks for the feedback!

> Is there any possibility to access the args of a log message within an enhancer and take the span from there (if present)? This would allow people who do not use scoped spans to use this as well.

I don't see any getters on the builder, so I think that would require building the LogEntry within enhanceLogEntry to call the LogEntry's getters.  How would the span be set on the log message before it is passed to enhanceLogEntry?

> In Scala we can't use scoped spans in most of the cases since the Future based code is executed on arbitrary threads.

What library runs the Futures?  The opencensus-java agent (https://github.com/census-instrumentation/opencensus-java/tree/master/contrib/agent) propagates the context to threads or executors, though I don't know whether it could be extended to work with other libraries.

sebright:
I updated the PR to set the LogEntry spanId field after https://github.com/GoogleCloudPlatform/google-cloud-java/pull/3332 was released.  Now the log messages are shown under the correct spans:

log-correlation-in-stackdriver-2

sebright:
@g-easy suggested that I merge this PR as is, since it is already large.  I added TODOs for the two code review comments, added a few new TODOs, removed the new artifact from the list of released artifacts, and removed the demo application, so it should be ready to merge now.

sebright:
Thanks!

PR #1216, opened 2018-05-29
Metrics: Move Metrics to a separate artifact.

songy23:
I haven't added this artifact to `build.gradle` or `setting.gradle` since it doesn't have any classes yet, and the `javaDoc` task will fail (I haven't found a way to disable this task).

### DIFF
--- a/buildscripts/import-control.xml
+++ b/buildscripts/import-control.xml
@@ -40,7 +40,6 @@ General guidelines on imports:
   </subpackage>
   <subpackage name="metrics">
     <allow pkg="io.opencensus.common"/>
-    <allow pkg="io.opencensus.internal"/>
     <allow pkg="io.opencensus.stats"/>
     <allow pkg="io.opencensus.tags"/>
   </subpackage>
--- a/buildscripts/import-control.xml
+++ b/buildscripts/import-control.xml
@@ -38,6 +38,12 @@ General guidelines on imports:
     <allow pkg="io.opencensus.internal"/>
     <allow pkg="io.opencensus.tags"/>
   </subpackage>
+  <subpackage name="metrics">
+    <allow pkg="io.opencensus.common"/>
+    <allow pkg="io.opencensus.internal"/>
+    <allow pkg="io.opencensus.stats"/>
+    <allow pkg="io.opencensus.tags"/>
+  </subpackage>
   <subpackage name="stats">
     <allow pkg="io.opencensus.common"/>
     <allow pkg="io.opencensus.internal"/>
--- a/metrics/README
+++ b/metrics/README
@@ -1,0 +1,7 @@
+OpenCensus Metrics
+======================================================
+
+* The Metrics data model used by Stats exporters. This data model may eventually become the wire
+format for metrics.
+* Currently all the the public classes under this package are marked as `ExperimentalApi`.
+* Java 7 and compatible.
--- a/metrics/build.gradle
+++ b/metrics/build.gradle
@@ -1,0 +1,11 @@
+description = 'OpenCensus Metrics Data Model'
+
+dependencies {
+    compile project(':opencensus-api')
+
+    compileOnly libraries.auto_value
+
+    testCompile project(':opencensus-api')
+
+    signature "org.codehaus.mojo.signature:java16:+@signature"
+}
--- a/metrics/build.gradle
+++ b/metrics/build.gradle
@@ -1,4 +1,4 @@
-description = 'OpenCensus Metrics Data Model'
+description = 'OpenCensus Metrics'
 
 dependencies {
     compile project(':opencensus-api')
--- a/build.gradle
+++ b/build.gradle
@@ -386,6 +386,8 @@ subprojects {
                  'opencensus-impl-core',
                  'opencensus-impl-lite',
                  'opencensus-impl',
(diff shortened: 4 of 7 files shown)


sebright:
Why is this a separate artifact?  Should we also move views and view registration into this artifact?

songy23:
I think @bogdandrutu mentioned that the exporter package should be out of api (https://github.com/census-instrumentation/opencensus-java/pull/1211#issuecomment-391833200). I guess the reason is to have exporters explicitly depend on the new Metrics artifact rather than APIs.

bogdandrutu:
I was thinking of moving all exporters out of the api package if possible. Not just metrics, but span exporter as well. What do you think?

songy23:
Make sense to me. However I would suggest to make Metrics and SpanData two independent artifacts since any exporter would only need to depend on one of them.

sebright:
@bogdandrutu Which classes do you think we should move out of opencensus-api?  Should we move the io.opencensus.trace.export package for trace, and ViewManager, View, and the metrics package for stats?  I think that it would be good to separate those parts of the API since they have very different uses and they should probably have different stability guarantees.  I'm not completely sure how to split the API when the implementation for both parts is tightly coupled, though.  I assume we won't split the implementation artifacts.

bogdandrutu:
@sebright I would start only with io.opencensus.trace.export and metrics initially. The implementation artifact can depend on both api and "export" packages.

buildscripts/import-control.xml
```diff
@@ -38,6 +38,12 @@ General guidelines on imports:
     <allow pkg="io.opencensus.internal"/>
     <allow pkg="io.opencensus.tags"/>
   </subpackage>
+  <subpackage name="metrics">
+    <allow pkg="io.opencensus.common"/>
+    <allow pkg="io.opencensus.internal"/>
```
sebright: metrics shouldn't use the internal package if it is in a separate artifact.

metrics/README
```diff
@@ -0,0 +1,7 @@
+OpenCensus Metrics
+======================================================
+
+* The Metrics data model used by Stats exporters. This data model may eventually become the wire
+format for metrics.
+* Currently all the the public classes under this package are marked as `ExperimentalApi`.
```
sebright: There is an extra "the".

metrics/build.gradle
```diff
@@ -0,0 +1,11 @@
+description = 'OpenCensus Metrics Data Model'
```
sebright: I think "data model" is more descriptive of the metrics Protocol Buffers, so it might be better to just call this artifact "OpenCensus Metrics".

metrics/README
```diff
@@ -0,0 +1,7 @@
+OpenCensus Metrics
```
sebright: I think this file should be README.md, for consistency with the other readmes.

buildscripts/import-control.xml
songy23: Done.

metrics/README
songy23: Good catch, done.

metrics/README
songy23: Removed.

metrics/build.gradle
songy23: Make sense, done.

bogdandrutu:
Should we name it exporter instead of metrics?

songy23:
Done, though I think it should be named `export` since we already have a [exporter](https://github.com/census-instrumentation/opencensus-java/tree/master/exporters) directory that has the actual exporters.

export/src/main/java/io/opencensus/export/metrics/package-info.java
```diff
@@ -28,4 +28,4 @@
  * for more details.
  */
 @io.opencensus.common.ExperimentalApi
-package io.opencensus.stats.metrics;
+package io.opencensus.export.metrics;
```
bogdandrutu: Should this be export.stats.metrics to match with the proto package?

songy23: Make sense, done.

sebright:
> Should we name it exporter instead of metrics?

I think metrics is more descriptive, because the library will do more than export.  It will also provide an API for reporting pre-aggregated values.

I also thought that metrics would be a separate artifact from the tracing export artifact, since each exporter would only need to depend on one of metrics or tracing, as @songy23 mentioned.

songy23:
Maybe we can have two artifacts `metrics` and `spandata` under directory `export`, similar to [exporters](https://github.com/census-instrumentation/opencensus-java/tree/master/exporters) (where stats and trace exporters are also under separate directories)?

sebright:
I would rather put `metrics` and `spandata` at the top level, since they are core artifacts for opencensus-java.  I also think that it could be confusing to have `export` and `exporters` directories.

songy23:
I'm more inclined to @sebright 's idea of keeping Metrics and SpanData separate. Anyway I think we can start with adding the Metrics artifact first. I created https://github.com/census-instrumentation/opencensus-java/issues/1239 and we can continue discussion there.

songy23:
Merging this PR to unblock following work. This artifact is not included in the uploadArchives for now, until we have a consensus on how the export artifact should look like (https://github.com/census-instrumentation/opencensus-java/issues/1239).

Issue #1218, opened 2018-05-30
exporter/stats/Stackdriver: spamming from ExportStatsToStackdriver in traces


odeke-em:
Idly running a Java server instrumented with OpenCensus and Stackdriver stats spams the traces with a bunch of outlier and high latency traces `ExportStatsToStackdriver` which highly distort the trace views when included with other traces.

For example, for traces that normally are in the 1ms range see what this looks like, the usual >100ms `ExportStatsToStackdriver` distort the output but also spam it
![screen shot 2018-05-30 at 12 12 55 pm](https://user-images.githubusercontent.com/4898263/40742290-e2f1ae0e-6402-11e8-9f61-15a5c29494e2.png)

Idly running:
![screen shot 2018-05-30 at 12 09 48 pm](https://user-images.githubusercontent.com/4898263/40742153-6d564d80-6402-11e8-808f-8bc1941dd7de.png)
![screen shot 2018-05-30 at 12 10 07 pm](https://user-images.githubusercontent.com/4898263/40742166-74c23e4e-6402-11e8-910b-bf1a816c32ee.png)

You can reproduce this by running this simple project https://github.com/orijtech/opencensus-for-grpc-java-developers
```shell
GOOGLE_APPLICATION_CREDENTIALS=~/Downloads/census-demos-237a8e1e41df.json mvn exec:java -Dexec.mainClass=io.ocgrpc.capitalize.CapitalizeServer
```

songy23:
@odeke-em thanks for catching this. I added the span to instrument Stackdriver stats exporter in https://github.com/census-instrumentation/opencensus-java/pull/898, by that time the trace `ExportStatsToStackdriver` should have default sampling probability. However we changed some logic later and now all the `ExportStatsToStackdriver` traces will be sampled. I'll post a PR to restore it to default sampling probability.

odeke-em:
Great, thanks @songy23 for the quick response!

Also I just noticed that there is one too for Prometheus `ExportStatsToPrometheus`
![screen shot 2018-05-30 at 12 31 40 pm](https://user-images.githubusercontent.com/4898263/40743123-80010a76-6405-11e8-8487-9adf0139452e.png)

Also doesn't the Scoped span in the try-with-resources in that PR mean that no support for Java 6 and below?

songy23:
[...]

PR #1219, opened 2018-05-30
 Overwrite the sampling probability to default for stats exporters.

songy23:
Fixes https://github.com/census-instrumentation/opencensus-java/issues/1218.

So that the span that instruments stats exporter should have the same sampling probability to that of [trace exporters](https://github.com/census-instrumentation/opencensus-java/blob/master/exporters/trace/stackdriver/src/main/java/io/opencensus/exporter/trace/stackdriver/StackdriverV2ExporterHandler.java#L72), instead of always being sampled.

/cc @odeke-em


songy23:
Thanks!

odeke-em:
@songy23 will y'all be making some sort of release to fix this? Or a snapshot?

songy23:
@odeke-em Yes I just updated the `0.14.0-SNAPSHOT` release, for now please try with that version.

PR #1222, opened 2018-05-30
Update grpc and jagger.

bogdandrutu:


songy23:
Please fix the build. Looks like there's method signature change in Jaeger.

PR #1223, opened 2018-05-31
Add Span.Kind to the trace API.

bogdandrutu:
Partially fix https://github.com/census-instrumentation/opencensus-java/issues/1054


api/src/main/java/io/opencensus/trace/export/SpanData.java
```diff
@@ -49,6 +50,41 @@
 @AutoValue
 public abstract class SpanData {
 
+  /**
+   * Returns a new immutable {@code SpanData}.
+   *
+   * @deprecated Use the version that contains the {@code Kind}.
```
songy23: suggestion: you can add a Javadoc link to the new method instead.

api/src/main/java/io/opencensus/trace/export/SpanData.java
```diff
@@ -111,6 +149,7 @@ public static SpanData create(
         parentSpanId,
         hasRemoteParent,
         name,
+        kind,
```
songy23: Should we checkNotNull() here (since the arg `Kind kind` of this method is not marked `Nullable`)?

impl_core/src/main/java/io/opencensus/implcore/trace/SpanBuilderImpl.java
```diff
@@ -234,4 +239,10 @@ public SpanBuilderImpl setRecordEvents(boolean recordEvents) {
     this.recordEvents = recordEvents;
     return this;
   }
+
+  @Override
+  public SpanBuilderImpl setSpanKind(Kind kind) {
+    this.kind = kind;
```
songy23: Same, checkNotNull() here?

impl_core/src/main/java/io/opencensus/implcore/trace/SpanImpl.java
```diff
@@ -213,6 +217,15 @@ public boolean getSampleToLocalSpanStore() {
     }
   }
 
+  /**
+   * Returns the kind of this {@code Span}.
+   *
+   * @return the kind of this {@code Span}.
+   */
+  public Kind getKind() {
```
songy23: `@Nullable`?

api/src/main/java/io/opencensus/trace/Span.java
```diff
@@ -259,4 +259,23 @@ public final SpanContext getContext() {
...
+   * parent/child relationship.
+   *
+   * @since 0.14
+   */
+  public enum Kind {
+    /**
+     * Indicates that the span covers server-side handling of an RPC or other remote network
+     * request.
+     */
+    SERVER,
```
adriancole: /me wants PRODUCER, CONSUMER for christmas, santa

bogdandrutu: We should continue our discussion here https://github.com/census-instrumentation/opencensus-specs/issues/54

api/src/main/java/io/opencensus/trace/export/SpanData.java
bogdandrutu: done.

api/src/main/java/io/opencensus/trace/export/SpanData.java
bogdandrutu: Made it @Nullable. Actually null is a valid value.

impl_core/src/main/java/io/opencensus/implcore/trace/SpanImpl.java
bogdandrutu: done.

bogdandrutu:
PTAL.

api/src/main/java/io/opencensus/trace/Span.java
sebright: The enum constants should also have `@since` tags, for when we add more.

api/src/main/java/io/opencensus/trace/export/SpanData.java
sebright: Should SpanBuilder.setSpanKind also use `@Nullable`?

api/src/main/java/io/opencensus/trace/SpanBuilder.java
```diff
@@ -133,12 +133,22 @@
...
-   * @param recordEvents New value determining if this {@code Span} should have events recorded.
+   * @param recordEvents new value determining if this {@code Span} should have events recorded.
    * @return this.
    * @since 0.5
    */
   public abstract SpanBuilder setRecordEvents(boolean recordEvents);
 
+  /**
+   * Sets the {@link Span.Kind} for the newly created {@code Span}. If not called, the
+   * implementation will provide a default.
```
sebright: Can the default be null?  It looks like this implementation doesn't set a non-null default.

api/src/main/java/io/opencensus/trace/SpanBuilder.java
```diff
@@ -133,12 +133,22 @@
...
+   *
+   * @param spanKind the kind of the newly created {@code Span}.
+   * @return this.
+   * @since 0.14
+   */
+  public abstract SpanBuilder setSpanKind(Span.Kind spanKind);
```
sebright: Should this method be non-abstract, to avoid breaking existing subclasses?

api/src/main/java/io/opencensus/trace/SpanBuilder.java
bogdandrutu: Yes, null is a valid option. we can do unspecified if really is important but I think null is fine.

PR #1229, opened 2018-06-01
Temporarily disable example-bazel build for Travis CI.

songy23:
Bazel just released a new version today and that broke our build. Disable bazel build target temporarily, we can re-enable example-bazel build using a stable version later (tracked in https://github.com/census-instrumentation/opencensus-java/issues/1228).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,9 +43,10 @@ matrix:
     env: TASK=BUILD_EXAMPLES_MAVEN
     os: linux
 
-  - jdk: oraclejdk8
-    env: TASK=BUILD_EXAMPLES_BAZEL
-    os: linux
+  # TODO(songya): re-enable this build once bazel is stable.
+  # - jdk: oraclejdk8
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -43,9 +43,10 @@ matrix:
     env: TASK=BUILD_EXAMPLES_MAVEN
     os: linux
 
-  - jdk: oraclejdk8
-    env: TASK=BUILD_EXAMPLES_BAZEL
-    os: linux
+  # TODO(songya): re-enable this build once bazel is stable.
+  # - jdk: oraclejdk8
+  #   env: TASK=BUILD_EXAMPLES_BAZEL
+  #   os: linux
```
sebright: Can we just move it to "allow_failures"?

songy23: Sure, done.

sebright: I think you also need to leave the original entry.

songy23: I see, done.

PR #1237, opened 2018-06-05
Update release versions for all readme and build files.

songy23:


songy23:
Also updated the gRPC version for maven and bazel example build, to resolve the version conflict (we upgraded to gRPC v0.10.1 in OC v0.14.0).

songy23:
Thanks!

PR #1243, opened 2018-06-07
Stackdriver Trace Exporter: add resource labels to Spans.

songy23:
Resource labels are consistent with [Specs](https://github.com/census-instrumentation/opencensus-specs/blob/master/utils/MonitoredResource.md). The format of label key is `g.co/r/<resource_type>/<label>`, for example `g.co/r/aws_ec2_instance/aws_account`.

### DIFF
--- a/exporters/trace/stackdriver/src/main/java/io/opencensus/exporter/trace/stackdriver/StackdriverV2ExporterHandler.java
+++ b/exporters/trace/stackdriver/src/main/java/io/opencensus/exporter/trace/stackdriver/StackdriverV2ExporterHandler.java
@@ -66,8 +66,10 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 
 /*>>>
 import org.checkerframework.checker.nullness.qual.Nullable;
@@ -100,6 +102,9 @@ final class StackdriverV2ExporterHandler extends SpanExporter.Handler {
   @VisibleForTesting @javax.annotation.Nullable
   static final MonitoredResource RESOURCE = MonitoredResourceUtils.getDefaultResource();
 
+  // Only initialize once.
+  private static final Map<String, AttributeValue> RESOURCE_LABELS = getResourceLabels(RESOURCE);
+
   private final String projectId;
   private final TraceServiceClient traceServiceClient;
   private final ProjectName projectName;
@@ -131,7 +136,7 @@ final class StackdriverV2ExporterHandler extends SpanExporter.Handler {
   }
 
   @VisibleForTesting
-  Span generateSpan(SpanData spanData, @javax.annotation.Nullable MonitoredResource resource) {
+  Span generateSpan(SpanData spanData) {
     SpanContext context = spanData.getContext();
     final String traceIdHex = encodeTraceId(context.getTraceId());
     final String spanIdHex = encodeSpanId(context.getSpanId());
@@ -144,7 +149,7 @@ final class StackdriverV2ExporterHandler extends SpanExporter.Handler {
             .setDisplayName(
                 toTruncatableStringProto(toDisplayName(spanData.getName(), spanData.getKind())))
             .setStartTime(toTimestampProto(spanData.getStartTimestamp()))
-            .setAttributes(toAttributesProto(spanData.getAttributes(), resource))
+            .setAttributes(toAttributesProto(spanData.getAttributes()))
             .setTimeEvents(
                 toTimeEventsProto(spanData.getAnnotations(), spanData.getMessageEvents()));
     io.opencensus.trace.Status status = spanData.getStatus();
@@ -229,13 +234,14 @@ final class StackdriverV2ExporterHandler extends SpanExporter.Handler {
 
   // These are the attributes of the Span, where usually we may add more attributes like the agent.
   private static Attributes toAttributesProto(
-      io.opencensus.trace.export.SpanData.Attributes attributes,
-      @javax.annotation.Nullable MonitoredResource resource) {
+      io.opencensus.trace.export.SpanData.Attributes attributes) {
     Attributes.Builder attributesBuilder =
         toAttributesBuilderProto(
             attributes.getAttributeMap(), attributes.getDroppedAttributesCount());
     attributesBuilder.putAttributeMap(AGENT_LABEL_KEY, AGENT_LABEL_VALUE);
-    putResourceLabels(attributesBuilder, resource);
+    for (Entry<String, AttributeValue> entry : RESOURCE_LABELS.entrySet()) {
+      attributesBuilder.putAttributeMap(entry.getKey(), entry.getValue());
+    }
     return attributesBuilder.build();
   }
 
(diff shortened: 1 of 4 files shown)


exporters/trace/stackdriver/src/main/java/io/opencensus/exporter/trace/stackdriver/StackdriverV2ExporterHandler.java
```diff
@@ -239,6 +249,90 @@ private static Attributes toAttributesProto(
...
+      return;
+    }
+    ResourceType resourceType = resource.getResourceType();
+    switch (resourceType) {
+      case AWS_EC2_INSTANCE:
+        AwsEc2InstanceMonitoredResource awsEc2InstanceMonitoredResource =
+            (AwsEc2InstanceMonitoredResource) resource;
+        attributesBuilder.putAttributeMap(
+            createResourceLabelKey(resourceType, "aws_account"),
+            toStringAttributeValueProto(awsEc2InstanceMonitoredResource.getAccount()));
```
sebright: >         attributesBuilder.putAttributeMap(
>             createResourceLabelKey(resourceType, "aws_account"),
>             toStringAttributeValueProto(awsEc2InstanceMonitoredResource.getAccount()));

Suggestion: The code that adds an attribute to the builder could probably be factored out into a helper method.

exporters/trace/stackdriver/src/test/java/io/opencensus/exporter/trace/stackdriver/StackdriverV2ExporterHandlerProtoTest.java
```diff
@@ -258,6 +262,35 @@ public void generateSpan() {
         .isEqualTo(Int32Value.newBuilder().setValue(CHILD_SPAN_COUNT).build());
   }
 
+  private static void verifyResourceLabels(Map<String, AttributeValue> attributeMap) {
+    if (StackdriverV2ExporterHandler.RESOURCE == null) {
+      return;
+    }
+    ResourceType resourceType = StackdriverV2ExporterHandler.RESOURCE.getResourceType();
+    switch (resourceType) {
```
sebright: It looks like it is possible for this method to not test the logic of putResourceLabels if the ResourceType isn't detected.  Is there a way to refactor the code so that the ResourceType is passed in to generateSpan, or putResourceLabels is tested directly?  Then we could easily test the code for all three cases, independent of the environment.

songy23: Done. The constructor of `MonitoredResource` is not public so I used some mocks here.

exporters/trace/stackdriver/src/main/java/io/opencensus/exporter/trace/stackdriver/StackdriverV2ExporterHandler.java
```diff
@@ -239,6 +250,119 @@ private static Attributes toAttributesProto(
...
+    }
+  }
+
+  private static void putResourceAttributeMapToBuilder(
+      Attributes.Builder attributesBuilder,
+      ResourceType resourceType,
+      String attributeName,
+      String attributeValue) {
+    attributesBuilder.putAttributeMap(
+        createResourceLabelKey(resourceType, attributeName),
```
bogdandrutu: We should cache all of these work.
private static final java.util.Map<java.lang.String, com.google.devtools.cloudtrace.v2.AttributeValue> GCE_INSTANCE_LABELS;


attributesBuilder.putAllAttributeMap(GCE_INSTANCE_LABELS);

songy23: Make sense to me, though making static map makes it difficult to test. 

If we really want to test the resource labels under all 3 cases, we may need to use something like PowerMock to mock `MonitoredResourceUtils.getDefaultResource()`. @sebright any idea on this?

sebright: If it is better to cache the whole map, couldn't we just pass RESOURCE_LABELS into generateSpan, instead of passing in RESOURCE?

songy23: Ahh you're right, updated. I also left a TODO to test the parsing.

exporters/trace/stackdriver/src/test/java/io/opencensus/exporter/trace/stackdriver/StackdriverV2ExporterHandlerProtoTest.java
```diff
@@ -262,33 +281,97 @@ public void generateSpan() {
...
+            messageEvents,
+            links,
+            CHILD_SPAN_COUNT,
+            status,
+            endTimestamp);
+    Span span = handler.generateSpan(spanData, mockAwsEc2Resource);
+    Map<String, AttributeValue> attributeMap = span.getAttributes().getAttributeMapMap();
+    assertThat(attributeMap).containsKey(createResourceLabelKey(AWS_EC2_INSTANCE, "aws_account"));
+    assertThat(attributeMap).containsKey(createResourceLabelKey(AWS_EC2_INSTANCE, "instance_id"));
+    assertThat(attributeMap).containsKey(createResourceLabelKey(AWS_EC2_INSTANCE, "region"));
```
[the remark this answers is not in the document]
songy23: Good catch, done.

exporters/trace/stackdriver/src/main/java/io/opencensus/exporter/trace/stackdriver/StackdriverV2ExporterHandler.java
```diff
@@ -239,6 +249,90 @@ private static Attributes toAttributesProto(
     return attributesBuilder;
   }
 
+  @SuppressWarnings("unchecked")
```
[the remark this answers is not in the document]
songy23: You're right, removed.

exporters/trace/stackdriver/src/test/java/io/opencensus/exporter/trace/stackdriver/StackdriverV2ExporterHandlerProtoTest.java
```diff
@@ -258,6 +271,39 @@ public void generateSpan() {
         .isEqualTo(Int32Value.newBuilder().setValue(CHILD_SPAN_COUNT).build());
   }
 
+  @Test
+  public void generateSpan_WithResourceLabels() {
```
sebright: This change looks good, though I think we should also test the other two ResourceTypes, as in 57765e8cea76cf612fda1251a8f99bba750c12ff.

songy23: Done. All these tests tested the same code path though, I'll update this to test parsing MonitoredResource instead, after https://github.com/census-instrumentation/opencensus-java/pull/1250 is merged.

PR #1247, opened 2018-06-08
Examples: add runtime dependency on netty-tcnative-boringssl-static.

songy23:
gRPC no longer shades the netty dependency `netty-tcnative-boringssl-static` in their latest release, but it is required by the cloud client libraries. We need to explicitly add runtime dependency of it to make example work.

Thanks @zhengyr for reporting this.


songy23:
Bazel already has it: https://github.com/census-instrumentation/opencensus-java/blob/master/examples/BUILD.bazel#L83.

For bazel we need to explicitly add all the transitive dependencies, while gradle/maven can figure out implicit transitive dependencies on their own.

PR #1250, opened 2018-06-12
Expose the factory methods of MonitoredResource.

songy23:
To make the unit tests of other libraries that depend on `monitored-resource-utils` easier to write.

### DIFF
--- a/contrib/monitored_resource_util/src/main/java/io/opencensus/contrib/monitoredresource/util/MonitoredResource.java
+++ b/contrib/monitored_resource_util/src/main/java/io/opencensus/contrib/monitoredresource/util/MonitoredResource.java
@@ -102,9 +102,29 @@ public abstract class MonitoredResource {
      */
     public abstract String getRegion();
 
-    static AwsEc2InstanceMonitoredResource create() {
+    /**
+     * Returns an {@link AwsEc2InstanceMonitoredResource}.
+     *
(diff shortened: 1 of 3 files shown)


rghetia:
LGTM

contrib/monitored_resource_util/src/main/java/io/opencensus/contrib/monitoredresource/util/MonitoredResource.java
```diff
@@ -102,9 +102,29 @@ public ResourceType getResourceType() {
...
+          account, instanceId, region);
+    }
+
+    /**
+     * Returns an {@link AwsEc2InstanceMonitoredResource} with the default parameters.
+     *
+     * @return an {@code AwsEc2InstanceMonitoredResource}.
+     * @since 0.15
+     */
+    public static AwsEc2InstanceMonitoredResource create() {
```
sebright: Could we continue hiding the no-arg factory methods, to reduce the size of the API?  I also think that the no-arg methods could be confusing, because users might expect the factory methods in AutoValue classes to be deterministic, but these depend on the environment.

songy23: Make sense to me, done.

songy23:
Thanks!

PR #1251, opened 2018-06-12
Stats: add MeasureMap.withSpanContext()

songy23:
EDIT:
Blocked by https://github.com/census-instrumentation/opencensus-specs/pull/122.

### DIFF
--- a/api/src/main/java/io/opencensus/stats/MeasureMap.java
+++ b/api/src/main/java/io/opencensus/stats/MeasureMap.java
@@ -19,6 +19,7 @@ package io.opencensus.stats;
 import io.opencensus.stats.Measure.MeasureDouble;
 import io.opencensus.stats.Measure.MeasureLong;
 import io.opencensus.tags.TagContext;
+import io.opencensus.trace.SpanContext;
 import javax.annotation.concurrent.NotThreadSafe;
 
 /**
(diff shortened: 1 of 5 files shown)


api/src/main/java/io/opencensus/stats/MeasureMap.java
```diff
@@ -19,6 +19,7 @@
 import io.opencensus.stats.Measure.MeasureDouble;
 import io.opencensus.stats.Measure.MeasureLong;
 import io.opencensus.tags.TagContext;
+import io.opencensus.trace.SpanContext;
```
dinooliva: I'm concerned about adding a direct dependency on trace from stats (I can easily see wanting to add in the other direction too). Is there any way to accomplish the same thing in a more abstract way?

api/src/main/java/io/opencensus/stats/MeasureMap.java
```diff
@@ -51,6 +52,15 @@
...
   public abstract MeasureMap put(MeasureLong measure, long value);
 
+  /**
+   * Associate measurements with a {@link SpanContext}.
+   *
+   * @param spanContext the {@link SpanContext}
+   * @return this
+   * @since 0.15
+   */
+  public abstract MeasureMap withSpanContext(SpanContext spanContext);
```
dinooliva: Can you give more details of what this is actually supposed to do (in the pr description).

api/src/main/java/io/opencensus/stats/MeasureMap.java
songy23: The only solution I can think of is to extract SpanContext (or create an equivalent class) to the common package, so that it can be shared between trace and stats.

api/src/main/java/io/opencensus/stats/MeasureMap.java
songy23: This is supposed to record a SpanContext when recording measurements. Please refer to the Specs PR: https://github.com/census-instrumentation/opencensus-specs/pull/122.

songy23:
Closing - I'll update the APIs based on https://github.com/census-instrumentation/opencensus-specs/pull/122, and post another PR.

PR #1254, opened 2018-06-13
Stackdriver Trace Exporter: improve unit tests on resource labels.

songy23:


exporters/trace/stackdriver/src/test/java/io/opencensus/exporter/trace/stackdriver/StackdriverV2ExporterHandlerProtoTest.java
```diff
@@ -21,6 +21,7 @@
 import static io.opencensus.contrib.monitoredresource.util.ResourceType.GCP_GCE_INSTANCE;
 import static io.opencensus.contrib.monitoredresource.util.ResourceType.GCP_GKE_CONTAINER;
 import static io.opencensus.exporter.trace.stackdriver.StackdriverV2ExporterHandler.createResourceLabelKey;
+import static io.opencensus.exporter.trace.stackdriver.StackdriverV2ExporterHandler.getResourceLabels;
```
sebright: Suggestion: I think the static import makes the test harder to read, because it's hard to see whether this is a non-test method or a test helper method without looking at the imports.

exporters/trace/stackdriver/src/test/java/io/opencensus/exporter/trace/stackdriver/StackdriverV2ExporterHandlerProtoTest.java
```diff
@@ -336,9 +349,9 @@ private void generateSpan_WithResourceLabels(Map<String, AttributeValue> resourc
             CHILD_SPAN_COUNT,
             status,
             endTimestamp);
-    Span span = handler.generateSpan(spanData, resourceLabels);
+    Span span = handler.generateSpan(spanData, getResourceLabels(resource));
```
sebright: I'm not sure if this would work, but I was wondering if it would be better to combine the original versions of this helper method and the three tests above into one test for `generateSpan` (only testing one of the MonitoredResource types), and write separate tests for `getResourceLabels` that test all three MonitoredResource types.

songy23:
Thanks!

PR #1257, opened 2018-06-13
Error Prone: 2.2.0 -> 2.3.1

sebright:
This commit contains several other changes as part of the upgrade:

- Use -XepAllDisabledChecksAsWarnings to enable all available warnings so
  that we don't need to enable each warning explicitly.

- Remove explicit suppression of warnings in generated code, since
  -XepDisableWarningsInGeneratedCode is now sufficient to suppress all warnings
  in AutoValue and Protocol Buffer generated classes.

- Remove a suppression of "ConstructorLeaksThis", since
  https://github.com/google/error-prone/pull/789 was fixed.

- Fix a few occurrences of "FieldCanBeFinal".


sebright:
Thanks!

PR #1258, opened 2018-06-14
Refactor OpenCensusTraceLoggingEnhancerTest.

sebright:
This commit removes some duplicated code and refactors the test helper method so
that it is easier to add more tests.


sebright:
Thanks!

PR #1259, opened 2018-06-14
Add ALL_SPANS/NO_SPANS/SAMPLED_SPANS property to OpenCensusTraceLoggingEnhancer.

sebright:
The full name of the property is
`io.opencensus.contrib.logcorrelation.stackdriver.OpenCensusTraceLoggingEnhancer.spanSelection`.
It can be set in logging.properties or set as a system property.

Allowed values for spanSelection:

NO_SPANS: disables the log correlation feature
ALL_SPANS: adds tracing data to all log entries
SAMPLED_SPANS: adds tracing data to log entries when the current span is sampled

SAMPLED_SPANS is the default.

_______________________________________________________________________

I'm not sure if this is the best way to allow configuration of the `spanSelection` field.  My goal is to make it work with as many logging frameworks as possible that could be used in combination with google-cloud-logging, though I've only tested it with Logback and java.util.logging so far.  With the current design, `spanSelection` could be set with

```xml
<property scope="system" name="io.opencensus.contrib.logcorrelation.stackdriver.OpenCensusTraceLoggingEnhancer.spanSelection" value="ALL_SPANS" />
```

in logback.xml or

```properties
io.opencensus.contrib.logcorrelation.stackdriver.OpenCensusTraceLoggingEnhancer.spanSelection=ALL_SPANS
```

in logging.properties.

/cc @songy23 @adriancole @ramonza @HailongWen

### DIFF
--- a/contrib/log_correlation/stackdriver/src/main/java/io/opencensus/contrib/logcorrelation/stackdriver/OpenCensusTraceLoggingEnhancer.java
+++ b/contrib/log_correlation/stackdriver/src/main/java/io/opencensus/contrib/logcorrelation/stackdriver/OpenCensusTraceLoggingEnhancer.java
@@ -24,35 +24,117 @@ import io.opencensus.trace.SpanContext;
 import io.opencensus.trace.TraceId;
 import io.opencensus.trace.Tracer;
 import io.opencensus.trace.Tracing;
+import java.util.logging.LogManager;
 
 /**
  * Stackdriver {@link LoggingEnhancer} that adds OpenCensus tracing data to log entries.
  *
  * <p>This feature is currently experimental.
  */
-
-// TODO(sebright): Allow configuring the LoggingEnhancer with Java properties. One property should
-// control the types of spans to correlate, with these three choices:
-// - off
-// - only add tracing data for sampled spans
-// - add tracing data for all spans
 @ExperimentalApi
 public final class OpenCensusTraceLoggingEnhancer implements LoggingEnhancer {
-  private static final String SAMPLED_KEY = "sampled";
+  private static final String SAMPLED_LABEL_KEY = "sampled";
+  private static final SpanSelection DEFAULT_SPAN_SELECTION = SpanSelection.ALL_SPANS;
+
+  /** Name of the property that defines the {@link SpanSelection}. The value is {@value}. */
+  public static final String SPAN_SELECTION_PROPERTY_NAME =
+      "io.opencensus.contrib.logcorrelation.stackdriver."
+          + "OpenCensusTraceLoggingEnhancer.spanSelection";
 
   private static final Tracer tracer = Tracing.getTracer();
 
-  /** Constructor to be called by Stackdriver logging. */
-  public OpenCensusTraceLoggingEnhancer() {}
+  private final SpanSelection spanSelection;
 
+  /** How to decide whether to add tracing data from the current span to a log entry. */
+  public enum SpanSelection {
+
+    /**
+     * Never add tracing data to log entries. This constant disables the log correlation feature.
+     */
+    NO_SPANS,
+
+    /** Add tracing data to a log entry iff the current span is sampled. */
+    SAMPLED_SPANS,
+
+    /**
+     * Always add tracing data to log entries, even when the current span is not sampled. This is
+     * the default.
+     */
+    ALL_SPANS
+  }
+
+  /**
+   * Constructor to be called by Stackdriver logging.
+   *
+   * <p>This constructor looks up the {@link SpanSelection} from the environment using the property
+   * name {@value SPAN_SELECTION_PROPERTY_NAME}. It looks for a {@link java.util.logging} property
+   * or a system property, giving preference to the logging property.
+   */
+  public OpenCensusTraceLoggingEnhancer() {
+    this(lookUpSpanSelectionProperty());
+  }
+
+  /** Constructor used for testing. */
+  OpenCensusTraceLoggingEnhancer(SpanSelection spanSelection) {
+    this.spanSelection = spanSelection;
+  }
+
+  private static SpanSelection lookUpSpanSelectionProperty() {
+    String spanSelectionProperty = lookUpProperty(SPAN_SELECTION_PROPERTY_NAME);
+    return spanSelectionProperty == null || spanSelectionProperty.isEmpty()
+        ? DEFAULT_SPAN_SELECTION
+        : parseSpanSelection(spanSelectionProperty);
+  }
+
+  private static SpanSelection parseSpanSelection(String spanSelection) {
+    try {
+      return SpanSelection.valueOf(spanSelection);
+    } catch (IllegalArgumentException e) {
+      return DEFAULT_SPAN_SELECTION;
+    }
+  }
+
+  // An OpenCensusTraceLoggingEnhancer property can be set with a logging property or a system
+  // property.
+  private static String lookUpProperty(String name) {
+    String property = LogManager.getLogManager().getProperty(name);
+    return property == null || property.isEmpty() ? System.getProperty(name) : property;
+  }
+
+  /**
+   * Returns the {@code SpanSelection} setting for this instance.
+   *
+   * @return the {@code SpanSelection} setting for this instance.
+   */
+  public SpanSelection getSpanSelection() {
+    return spanSelection;
+  }
+
+  // This method avoids getting the current span when the feature is disabled, for efficiency.
   @Override
   public void enhanceLogEntry(LogEntry.Builder builder) {
-    SpanContext span = tracer.getCurrentSpan().getContext();
+    switch (spanSelection) {
+      case NO_SPANS:
+        return;
+      case SAMPLED_SPANS:
+        SpanContext span = tracer.getCurrentSpan().getContext();
+        if (span.getTraceOptions().isSampled()) {
+          addTracingData(span, builder);
+        }
+        return;
+      case ALL_SPANS:
+        addTracingData(tracer.getCurrentSpan().getContext(), builder);
+        return;
+    }
+    throw new AssertionError("Unknown spanSelection: " + spanSelection);
+  }
+
+  private static void addTracingData(SpanContext span, LogEntry.Builder builder) {
     builder.setTrace(formatTraceId(span.getTraceId()));
     builder.setSpanId(span.getSpanId().toLowerBase16());
 
     // TODO(sebright): Find the correct way to add the sampling decision.
-    builder.addLabel(SAMPLED_KEY, Boolean.toString(span.getTraceOptions().isSampled()));
+    builder.addLabel(SAMPLED_LABEL_KEY, Boolean.toString(span.getTraceOptions().isSampled()));
   }
 
   private static String formatTraceId(TraceId traceId) {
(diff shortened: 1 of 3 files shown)


contrib/log_correlation/stackdriver/src/main/java/io/opencensus/contrib/logcorrelation/stackdriver/OpenCensusTraceLoggingEnhancer.java
```diff
@@ -24,35 +24,114 @@
...
+  /** How to decide whether to add tracing data from the current span to a log entry. */
+  public enum SpanSelection {
+
+    /**
+     * Never add tracing data to log entries. This constant disables the log correlation feature.
+     */
+    NO_SPANS,
+
+    /** Add tracing data to a log entry iff the current span is sampled. This is the default. */
+    SAMPLED_SPANS,
```
adriancole: what's the use case for this btw? I've personally not seen this arise (only sampled spans), and folks like @nicmunroe @bplotnick tend to suggest always trace ids. attaching nic's latest neat blog on it.

https://medium.com/nikeengineering/hit-the-ground-running-with-distributed-tracing-core-concepts-ff5ad47c7058

contrib/log_correlation/stackdriver/src/main/java/io/opencensus/contrib/logcorrelation/stackdriver/OpenCensusTraceLoggingEnhancer.java
```diff
@@ -24,35 +24,114 @@
...
-// TODO(sebright): Allow configuring the LoggingEnhancer with Java properties. One property should
-// control the types of spans to correlate, with these three choices:
-// - off
-// - only add tracing data for sampled spans
-// - add tracing data for all spans
 @ExperimentalApi
 public final class OpenCensusTraceLoggingEnhancer implements LoggingEnhancer {
-  private static final String SAMPLED_KEY = "sampled";
+  private static final String SAMPLED_LABEL_KEY = "sampled";
+  private static final SpanSelection DEFAULT_SPAN_SELECTION = SpanSelection.SAMPLED_SPANS;
```
rghetia: are all trace exported typically? OR only the sampled one? 
If it is all then default should be ALL_SPANS.

contrib/log_correlation/stackdriver/src/test/java/io/opencensus/contrib/logcorrelation/stackdriver/OpenCensusTraceLoggingEnhancerTest.java
```diff
@@ -105,6 +177,86 @@ private static LogEntry getEnhancedLogEntry(String projectId, Span span) {
...
+  @Test
+  public void spanSelectionDefaultIsSampledSpans() {
+    assertThat(new OpenCensusTraceLoggingEnhancer().getSpanSelection())
+        .isEqualTo(SpanSelection.SAMPLED_SPANS);
+  }
+
+  @Test
+  @SuppressWarnings("TruthConstantAsserts")
+  public void spanSelectionPropertyName() {
+    assertThat(OpenCensusTraceLoggingEnhancer.SPAN_SELECTION_PROPERTY_NAME)
```
rghetia: Why do we need to test this?

contrib/log_correlation/stackdriver/src/test/java/io/opencensus/contrib/logcorrelation/stackdriver/OpenCensusTraceLoggingEnhancerTest.java
```diff
@@ -49,53 +55,119 @@
...
+                    TraceOptions.builder().setIsSampled(true).build())));
+    assertContainsNoTracingData(logEntry);
+  }
+
+  @Test
+  public void enhanceLogEntry_DoNotAddNonSampledSpanToLogEntry() {
+    LogEntry logEntry =
+        getEnhancedLogEntry(
+            "my-test-project-4",
+            new OpenCensusTraceLoggingEnhancer(),
```
g-easy: IMO be explicit about SpanSelection here.

contrib/log_correlation/stackdriver/src/test/java/io/opencensus/contrib/logcorrelation/stackdriver/OpenCensusTraceLoggingEnhancerTest.java
g-easy: Makes sure the class name matches up. :)

contrib/log_correlation/stackdriver/src/test/java/io/opencensus/contrib/logcorrelation/stackdriver/OpenCensusTraceLoggingEnhancerTest.java
```diff
@@ -105,6 +177,86 @@ private static LogEntry getEnhancedLogEntry(String projectId, Span span) {
...
+    }
+  }
+
+  @Test
+  public void setSpanSelectionWithLoggingProperty() throws IOException {
+    try {
+      LogManager.getLogManager()
+          .readConfiguration(
+              CharSource.wrap(
+                      OpenCensusTraceLoggingEnhancer.SPAN_SELECTION_PROPERTY_NAME + "=ALL_SPANS")
```
g-easy: Please change this to NO_SPANS if you change the default. :)

contrib/log_correlation/stackdriver/src/main/java/io/opencensus/contrib/logcorrelation/stackdriver/OpenCensusTraceLoggingEnhancer.java
sebright: Thanks for the link!  That is really helpful for understanding the use case of exporting tracing data primarily through logs.  I added the SAMPLED_SPANS option for the case where users are already using Stackdriver Trace and want to use log messages similarly to annotations, where the tracing data for non-sampled spans would be less useful.  It sounds like ALL_SPANS would be a better default, though.  I updated the PR.

contrib/log_correlation/stackdriver/src/main/java/io/opencensus/contrib/logcorrelation/stackdriver/OpenCensusTraceLoggingEnhancer.java
sebright: I changed the default to ALL_SPANS.

contrib/log_correlation/stackdriver/src/test/java/io/opencensus/contrib/logcorrelation/stackdriver/OpenCensusTraceLoggingEnhancerTest.java
sebright: I agree, I updated that in the commit that changed the default.

contrib/log_correlation/stackdriver/src/test/java/io/opencensus/contrib/logcorrelation/stackdriver/OpenCensusTraceLoggingEnhancerTest.java
sebright: Yes, I had to use a String literal in the non-test code so that I could display it in the Javadoc, and I wanted to check that it matched the class name.  This is also important to test because users will need to specify it in their logging configurations.

contrib/log_correlation/stackdriver/src/test/java/io/opencensus/contrib/logcorrelation/stackdriver/OpenCensusTraceLoggingEnhancerTest.java
sebright: Do you mean change it to something other that the default?  I changed it to SAMPLED_SPANS when I changed the default.

contrib/log_correlation/stackdriver/src/test/java/io/opencensus/contrib/logcorrelation/stackdriver/OpenCensusTraceLoggingEnhancerTest.java
rghetia: why not initialize using class.getName().

```
public static final String SPAN_SELECTION_PROPERTY_NAME =
  OpenCensusTraceLoggingEnhancer.class.getName() + ".spanSelection";
```

sebright: I wanted to display the value with a `@value` tag in the Javadoc, so I had to use a String literal.

rghetia: got it. thx.

contrib/log_correlation/stackdriver/src/test/java/io/opencensus/contrib/logcorrelation/stackdriver/OpenCensusTraceLoggingEnhancerTest.java
g-easy: Yep, that's what I was after. SAMPLED_SPANS is fine.

contrib/log_correlation/stackdriver/src/main/java/io/opencensus/contrib/logcorrelation/stackdriver/OpenCensusTraceLoggingEnhancer.java
```diff
@@ -24,35 +24,117 @@
...
-// TODO(sebright): Allow configuring the LoggingEnhancer with Java properties. One property should
-// control the types of spans to correlate, with these three choices:
-// - off
-// - only add tracing data for sampled spans
-// - add tracing data for all spans
 @ExperimentalApi
 public final class OpenCensusTraceLoggingEnhancer implements LoggingEnhancer {
-  private static final String SAMPLED_KEY = "sampled";
+  private static final String SAMPLED_LABEL_KEY = "sampled";
+  private static final SpanSelection DEFAULT_SPAN_SELECTION = SpanSelection.ALL_SPANS;
```
dinooliva: Shouldn't this be sampled_spans?

contrib/log_correlation/stackdriver/src/test/java/io/opencensus/contrib/logcorrelation/stackdriver/OpenCensusTraceLoggingEnhancerTest.java
```diff
@@ -49,53 +55,119 @@
...
     LogEntry logEntry =
         getEnhancedLogEntry(
-            "my-test-project-2",
+            "my-test-project-3",
+            new OpenCensusTraceLoggingEnhancer(SpanSelection.ALL_SPANS),
+            new TestSpan(
+                SpanContext.create(
+                    TraceId.fromLowerBase16("4c6af40c499951eb7de2777ba1e4fefa"),
+                    SpanId.fromLowerBase16("de52e84d13dd232d"),
+                    TraceOptions.builder().setIsSampled(true).build())));
```
dinooliva: Shouldn't you set this to false in order to check that it's collected regardless?

dinooliva: Misread the name - ignore.

PR #1260, opened 2018-06-15
Allow overriding the Project ID used by OpenCensusTraceLoggingEnhancer.

sebright:
The project ID used for formatting trace IDs can be overridden with the property
"`io.opencensus.contrib.logcorrelation.stackdriver.OpenCensusTraceLoggingEnhancer.projectId`".
The property can be set as a java.util.logging property or a system property.
OpenCensusTraceLoggingEnhancer continues to use the google-cloud-logging default
project ID when the projectId property is not set.

This commit also caches the project ID on the OpenCensusTraceLoggingEnhancer
instance instead of looking it up for every log entry.

### DIFF
--- a/contrib/log_correlation/stackdriver/src/main/java/io/opencensus/contrib/logcorrelation/stackdriver/OpenCensusTraceLoggingEnhancer.java
+++ b/contrib/log_correlation/stackdriver/src/main/java/io/opencensus/contrib/logcorrelation/stackdriver/OpenCensusTraceLoggingEnhancer.java
@@ -54,6 +54,9 @@ public final class OpenCensusTraceLoggingEnhancer implements LoggingEnhancer {
   private final String projectId;
   private final SpanSelection spanSelection;
 
+  // This field caches the prefix used for the LogEntry.trace field and is derived from projectId.
+  private final String tracePrefix;
+
   /** How to decide whether to add tracing data from the current span to a log entry. */
   public enum SpanSelection {
 
@@ -91,6 +94,7 @@ public final class OpenCensusTraceLoggingEnhancer implements LoggingEnhancer {
   OpenCensusTraceLoggingEnhancer(@Nullable String projectId, SpanSelection spanSelection) {
     this.projectId = projectId == null ? "" : projectId;
     this.spanSelection = spanSelection;
+    this.tracePrefix = "projects/" + this.projectId + "/traces/";
   }
 
   private static String lookUpProjectId() {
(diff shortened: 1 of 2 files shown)


contrib/log_correlation/stackdriver/src/main/java/io/opencensus/contrib/logcorrelation/stackdriver/OpenCensusTraceLoggingEnhancer.java
```diff
@@ -36,13 +37,20 @@
   private static final String SAMPLED_LABEL_KEY = "sampled";
   private static final SpanSelection DEFAULT_SPAN_SELECTION = SpanSelection.ALL_SPANS;
 
+  /**
+   * Name of the property that overrides the default Stackdriver Project ID. The value is {@value}.
```
g-easy: Just to be super clear, could you please say that the "default" is actually fetched from the environment / is based on the project id that the app is running as. I realize the comment on the constructor repeats this but it's worth repeating.

rghetia:
LGTM

contrib/log_correlation/stackdriver/src/main/java/io/opencensus/contrib/logcorrelation/stackdriver/OpenCensusTraceLoggingEnhancer.java
```diff
@@ -66,19 +74,31 @@
...
+   * environment. It uses the project ID returned from {@code
+   * com.google.cloud.ServiceOptions.getDefaultProjectId()}, unless the ID is overridden by the
+   * property {@value PROJECT_ID_PROPERTY_NAME}. It looks up the {@code SpanSelection} using the
+   * property {@value SPAN_SELECTION_PROPERTY_NAME}. Each property can be specified with a {@link
+   * java.util.logging} property or a system property, with preference given to the logging
+   * property.
    */
   public OpenCensusTraceLoggingEnhancer() {
-    this(lookUpSpanSelectionProperty());
+    this(lookUpProjectId(), lookUpSpanSelectionProperty());
```
g-easy: Optional: instead of just projectID, cache: `"projects/" + projectId + "/traces/"`

contrib/log_correlation/stackdriver/src/main/java/io/opencensus/contrib/logcorrelation/stackdriver/OpenCensusTraceLoggingEnhancer.java
sebright: I tried to make the wording more consistent between the property name and constructor, and I referenced the method that returns the default cloud project ID in both.  I couldn't add a direct link to the `ServiceOptions.getDefaultProjectId()` Javadoc, though, since it's in another library.

contrib/log_correlation/stackdriver/src/main/java/io/opencensus/contrib/logcorrelation/stackdriver/OpenCensusTraceLoggingEnhancer.java
sebright: Thanks!  I forgot I could also cache the whole prefix.

sebright:
Thanks!  I'll rebase and then merge.

PR #1262, opened 2018-06-18
Add opencensus-contrib-log-correlation-stackdriver to the list of released artifacts.

sebright:


songy23:
s/opencensus-contrib-monitored-resource-util/opencensus-contrib-log-correlation-stackdriver for the first commit message.

sebright:
Good catch! Thanks!

PR #1263, opened 2018-06-18
Metrics:  Add Value and Point.

songy23:
~~Blocked by https://github.com/census-instrumentation/opencensus-java/pull/1261.~~

From [metrics.proto](https://github.com/census-instrumentation/opencensus-proto/blob/master/opencensus/proto/stats/metrics/metrics.proto):
```
// A timestamped measurement.
message Point {
  // The moment when this point was recorded.
  google.protobuf.Timestamp timestamp = 1; // required

  // The actual point value.
  oneof value {
    // A 64-bit integer.
    int64 int64_value = 2;

    // A 64-bit double-precision floating-point number.
    double double_value = 3;

    // A distribution value.
    DistributionValue distribution_value = 4;

    // TODO: Add support for Summary type. This is an aggregation that produces
    // percentiles directly.
    //
    // See also: https://prometheus.io/docs/concepts/metric_types/#summary
  }
}

// Distribution contains summary statistics for a population of values. It
// optionally contains a histogram representing the distribution of those
// values across a set of buckets.
message DistributionValue {
  // The number of values in the population. Must be non-negative. This value
  // must equal the sum of the values in bucket_counts if a histogram is
  // provided.
  int64 count = 1;

  // The arithmetic mean of the values in the population. If count is zero
  // then this field must be zero.
  double mean = 2;

  // The sum of squared deviations from the mean of the values in the
  // population. For values x_i this is:
  //
  //     Sum[i=1..n]((x_i - mean)^2)
  //
  // Knuth, "The Art of Computer Programming", Vol. 2, page 323, 3rd edition
  // describes Welford's method for accumulating this sum in one pass.
  //
  // If count is zero then this field must be zero.
  double sum_of_squared_deviation = 3;

  // The range of the population values.
  message Range {
    // The minimum of the population values.
    double min = 1;

    // The maximum of the population values.
    double max = 2;
  }

  // If specified, contains the range of the population values. The field
  // must not be present if the count is zero.
  Range range = 4;

  // A Distribution may optionally contain a histogram of the values in the
  // population. The bucket boundaries for that histogram are described by
  // bucket_bounds. This defines size(bucket_bounds) + 1 (= N)
  // buckets. The boundaries for bucket index i are:
  //
  // (-infinity, bucket_bounds[i]) for i == 0
  // [bucket_bounds[i-1], bucket_bounds[i]) for 0 < i < N-2
  // [bucket_bounds[i-1], +infinity) for i == N-1
  //
  // i.e. an underflow bucket (number 0), zero or more finite buckets (1
  // through N - 2, and an overflow bucket (N - 1), with inclusive lower
  // bounds and exclusive upper bounds.
  //
  // If bucket_bounds has no elements (zero size), then there is no
  // histogram associated with the Distribution. If bucket_bounds has only
  // one element, there are no finite buckets, and that single element is the
  // common boundary of the overflow and underflow buckets. The values must
  // be monotonically increasing.
  //
  // Don't change bucket boundaries within a timeseries if your backend
  // doesn't support this.
  repeated double bucket_bounds = 5;

  message Bucket {
    // The number of values in each bucket of the histogram, as described in
    // bucket_bounds.
    int64 count = 1;

    // TODO: Add support for exemplars.
  }

  // If the distribution does not have a histogram, then omit this field.
  // If there is a histogram, then the sum of the values in the Bucket counts
  // must equal the value in the count field of the distribution.
  repeated Bucket buckets = 6;
}

```

### DIFF
--- a/metrics/src/main/java/io/opencensus/metrics/Value.java
+++ b/metrics/src/main/java/io/opencensus/metrics/Value.java
@@ -37,7 +37,7 @@ import javax.annotation.concurrent.Immutable;
  *
  * <p>Each {@link Point} contains exactly one of the three {@link Value} types.
  *
- * @since 0.15
+ * @since 0.16
  */
 @ExperimentalApi
 @Immutable
@@ -48,7 +48,7 @@ public abstract class Value {
   /**
    * Applies the given match function to the underlying data type.
    *
-   * @since 0.15
+   * @since 0.16
    */
   public abstract <T> T match(
       Function<? super DoubleValue, T> p0,
@@ -59,7 +59,7 @@ public abstract class Value {
   /**
    * A 64-bit double-precision floating-point {@link Value}.
    *
-   * @since 0.15
+   * @since 0.16
    */
   @AutoValue
   @Immutable
@@ -81,7 +81,7 @@ public abstract class Value {
      *
      * @param value the value in double.
      * @return a {@code DoubleValue}.
-     * @since 0.15
+     * @since 0.16
      */
     public static DoubleValue create(double value) {
       return new AutoValue_Value_DoubleValue(value);
@@ -91,7 +91,7 @@ public abstract class Value {
      * Returns the double value.
      *
      * @return the double value.
-     * @since 0.15
+     * @since 0.16
      */
     public abstract double getValue();
   }
@@ -99,7 +99,7 @@ public abstract class Value {
   /**
    * A 64-bit integer {@link Value}.
    *
-   * @since 0.15
+   * @since 0.16
    */
   @AutoValue
   @Immutable
@@ -121,7 +121,7 @@ public abstract class Value {
      *
      * @param value the value in long.
      * @return a {@code LongValue}.
-     * @since 0.15
+     * @since 0.16
      */
     public static LongValue create(long value) {
       return new AutoValue_Value_LongValue(value);
@@ -131,7 +131,7 @@ public abstract class Value {
      * Returns the long value.
      *
      * @return the long value.
-     * @since 0.15
+     * @since 0.16
      */
     public abstract long getValue();
   }
@@ -140,7 +140,7 @@ public abstract class Value {
    * {@link DistributionValue} contains summary statistics for a population of values. It optionally
    * contains a histogram representing the distribution of those values across a set of buckets.
    *
-   * @since 0.15
+   * @since 0.16
    */
   @AutoValue
   @Immutable
@@ -167,7 +167,7 @@ public abstract class Value {
      * @param bucketBoundaries bucket boundaries of a histogram.
      * @param buckets {@link Bucket}s of a histogram.
      * @return a {@code DistributionValue}.
-     * @since 0.15
+     * @since 0.16
      */
     public static DistributionValue create(
         double mean,
@@ -214,7 +214,7 @@ public abstract class Value {
      * Returns the aggregated mean.
      *
      * @return the aggregated mean.
-     * @since 0.15
+     * @since 0.16
      */
     public abstract double getMean();
 
@@ -222,7 +222,7 @@ public abstract class Value {
      * Returns the aggregated count.
      *
      * @return the aggregated count.
-     * @since 0.15
+     * @since 0.16
      */
     public abstract long getCount();
 
@@ -240,7 +240,7 @@ public abstract class Value {
      * <p>If count is zero then this field must be zero.
      *
      * @return the aggregated sum of squared deviations.
-     * @since 0.15
+     * @since 0.16
      */
     public abstract double getSumOfSquaredDeviations();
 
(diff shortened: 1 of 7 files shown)


metrics/src/main/java/io/opencensus/metrics/Value.java
```diff
@@ -0,0 +1,372 @@
...
+  public abstract <T> T match(
+      Function<? super DoubleValue, T> p0,
+      Function<? super LongValue, T> p1,
+      Function<? super DistributionValue, T> p2,
+      Function<? super Value, T> defaultFunction);
+
+  /**
+   * A 64-bit double-precision floating-point {@link Value}.
+   *
+   * @since 0.16
```
dinooliva: Shouldn't the abstract class name be a prefix?

songy23: In the [proto](https://github.com/census-instrumentation/opencensus-proto/blob/master/opencensus/proto/stats/metrics/metrics.proto#L150) all the sub-types are suffixed: `int64_value`, `double_value`, `distribution_value`. Do you think we should rename these to `ValueDouble`, `ValueLong` and `ValueDistribution` in Java?

songy23: Sorry I didn't notice the match() method for [AttributeValue](https://github.com/census-instrumentation/opencensus-java/blob/660e8f375bb483a3eb817940b3aa8534f86da314/api/src/main/java/io/opencensus/trace/AttributeValue.java#L31). I think this approach is better.

metrics/src/main/java/io/opencensus/metrics/Value.java
```diff
@@ -173,10 +174,19 @@ public static DistributionValue create(
...
+        @Nullable Range range,
         List<Double> bucketBoundaries,
         List<Bucket> buckets) {
       Utils.checkArgument(count >= 0, "count should be non-negative.");
+      Utils.checkArgument(
+          sumOfSquaredDeviations >= 0, "sum of squared deviations should be non-negative.");
+      if (count == 0) {
+        Utils.checkArgument(range == null, "range should not be present if count is 0.");
+        Utils.checkArgument(
+            sumOfSquaredDeviations == 0, "sum of squared deviations should be 0 if count is 0.");
```
sebright: I think you could also check that `sum` is zero here.

metrics/src/main/java/io/opencensus/metrics/Value.java
```diff
@@ -229,14 +239,6 @@ public static DistributionValue create(
     /**
      * Returns the aggregated sum of squared deviations.
      *
-     * <p>The sum of squared deviations from the mean of the values in the population. For values
-     * x_i this is:
-     *
-     * <p>Sum[i=1..n]((x_i - mean)^2)
```
sebright: It is fine to define the "sum of squared deviations", I just thought that we shouldn't specify that it is calculated in one pass.

metrics/src/main/java/io/opencensus/metrics/Value.java
songy23: Done.

metrics/src/main/java/io/opencensus/metrics/Value.java
songy23: Make sense to me, and with this approach our APIs are more consistent with the proto definition (where we only have the Distribution as a separate class). WDYT @dinooliva ?

dinooliva: I'm good with having match handle Long and Double. Still on the fence about ValueDistribution vs DistributionValue but I prefer to stay consistent with our naming conventions (i.e. ValueDistribution).

On possibility would be to add a static factory method distributionValue() similar to what is done in AttributeValue.

songy23: SG, I renamed all the subclasses to be Value* (prefixed). Now the subclasses are package-private, and the top-level `Value` class exposes factory methods (`doubleValue()`, `longValue()`, `distributionValue()`) to create each subclass, which is the same to [AttributeValue](https://github.com/census-instrumentation/opencensus-java/blob/660e8f375bb483a3eb817940b3aa8534f86da314/api/src/main/java/io/opencensus/trace/AttributeValue.java#L31).

In addition, I added a top-level `Distribution` class in order to expose its internal stats.

metrics/src/main/java/io/opencensus/metrics/Value.java
```diff
@@ -0,0 +1,196 @@
...
+
+/**
+ * The actual point value for a {@link Point}.
+ *
+ * <p>Currently there are three types of {@link Value}:
+ *
+ * <ul>
+ *   <li>{@link ValueDouble}
+ *   <li>{@link ValueLong}
+ *   <li>{@link ValueDistribution}
```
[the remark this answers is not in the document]
songy23: Make sense, done.

songy23:
Thanks!

PR #1264, opened 2018-06-19
Metrics: Add TimeSeries.

songy23:
~~Blocked by https://github.com/census-instrumentation/opencensus-java/pull/1263.~~

[Proto definition](https://github.com/census-instrumentation/opencensus-proto/blob/master/opencensus/proto/stats/metrics/metrics.proto#L107-L132):

```
// A collection of data points that describes the time-varying values
// of a gauge metric.
message GaugeTimeSeries {
  // The set of label values that uniquely identify this timeseries. Applies to
  // all points. The order of label values must match that of label keys in the
  // metric descriptor.
  repeated LabelValue label_values = 1;

  // The data points of this timeseries. Point type MUST match the MetricDescriptor.type.
  repeated Point points = 2;
}

// A collection of data points that describes the time-varying values
// of a cumulative metric.
message CumulativeTimeSeries {
  // The time that the cumulative value was reset to zero.
  google.protobuf.Timestamp start_time = 1; // required

  // The set of label values that uniquely identify this timeseries. Applies to
  // all points. The order of label values must match that of label keys in the
  // metric descriptor.
  repeated LabelValue label_values = 2;

  // The data points of this timeseries. Point type MUST match the MetricDescriptor.type.
  repeated Point points = 3;
}
```

### DIFF
--- a/metrics/src/main/java/io/opencensus/metrics/TimeSeries.java
+++ b/metrics/src/main/java/io/opencensus/metrics/TimeSeries.java
@@ -99,6 +99,8 @@ public abstract class TimeSeries {
       // Fail fast on null lists to prevent NullPointerException when copying the lists.
       Utils.checkNotNull(labelValues, "labelValues");
       Utils.checkNotNull(points, "points");
+      Utils.checkListElementNotNull(labelValues, "labelValue");
+      Utils.checkListElementNotNull(points, "point");
       return new AutoValue_TimeSeries_TimeSeriesGauge(
           Collections.unmodifiableList(new ArrayList<LabelValue>(labelValues)),
           Collections.unmodifiableList(new ArrayList<Point>(points)));
@@ -140,6 +142,8 @@ public abstract class TimeSeries {
       // Fail fast on null lists to prevent NullPointerException when copying the lists.
       Utils.checkNotNull(labelValues, "labelValues");
       Utils.checkNotNull(points, "points");
+      Utils.checkListElementNotNull(labelValues, "labelValue");
+      Utils.checkListElementNotNull(points, "point");
       return new AutoValue_TimeSeries_TimeSeriesCumulative(
           Collections.unmodifiableList(new ArrayList<LabelValue>(labelValues)),
           Collections.unmodifiableList(new ArrayList<Point>(points)),
--- a/metrics/src/main/java/io/opencensus/metrics/TimeSeries.java
+++ b/metrics/src/main/java/io/opencensus/metrics/TimeSeries.java
@@ -1,0 +1,151 @@
+/*
+ * Copyright 2018, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


metrics/src/main/java/io/opencensus/metrics/TimeSeries.java
```diff
@@ -0,0 +1,157 @@
...
+     *
+     * @param labelValues the {@code LabelValue}s that uniquely identify this {@code TimeSeries}.
+     * @param points the data {@code Point}s of this {@code TimeSeries}.
+     * @return a {@code TimeSeriesGauge}.
+     * @since 0.16
+     */
+    public static TimeSeriesGauge create(List<LabelValue> labelValues, List<Point> points) {
+      // Fail fast on null lists to prevent NullPointerException when copying the lists.
+      Utils.checkNotNull(labelValues, "labelValues");
+      Utils.checkNotNull(points, "points");
```
sebright: This method should also check the list elements for null.

metrics/src/main/java/io/opencensus/metrics/TimeSeries.java
```diff
@@ -0,0 +1,157 @@
...
+     * @param labelValues the {@code LabelValue}s that uniquely identify this {@code TimeSeries}.
+     * @param points the data {@code Point}s of this {@code TimeSeries}.
+     * @param startTimestamp the start {@code Timestamp} of this {@code TimeSeriesCumulative}.
+     * @return a {@code TimeSeriesCumulative}.
+     * @since 0.16
+     */
+    public static TimeSeriesCumulative create(
+        List<LabelValue> labelValues, List<Point> points, Timestamp startTimestamp) {
+      // Fail fast on null lists to prevent NullPointerException when copying the lists.
+      Utils.checkNotNull(labelValues, "labelValues");
```
sebright: Same here

metrics/src/test/java/io/opencensus/metrics/TimeSeriesTest.java
```diff
@@ -0,0 +1,118 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/** Unit tests for {@link TimeSeries}. */
+@RunWith(JUnit4.class)
+public class TimeSeriesTest {
+
+  private static final LabelValue LABEL_VALUE_1 = LabelValue.create("value1");
+  private static final LabelValue LABEL_VALUE_2 = LabelValue.create("value1");
```
sebright: Should LABEL_VALUE_2 have a different value from LABEL_VALUE_1?

metrics/src/main/java/io/opencensus/metrics/TimeSeries.java
songy23: Done. Added a Util method to check list element.

metrics/src/main/java/io/opencensus/metrics/TimeSeries.java
songy23: Done.

metrics/src/test/java/io/opencensus/metrics/TimeSeriesTest.java
songy23: Good catch, done.

songy23:
Thanks!

Issue #1265, opened 2018-06-19
Avoid unnecessary calls to .toString in ViewData.checkAggregation


steveniemitz:
Please answer these questions before submitting a bug report.

### What version of OpenCensus are you using?
0.12.2, but present up to current version

### What JVM are you using (`java -version`)?
java version "1.8.0_111"
Java(TM) SE Runtime Environment (build 1.8.0_111-b14)
Java HotSpot(TM) 64-Bit Server VM (build 25.111-b14, mixed mode)

I ran into this randomly running a CPU profiler on a live instance running OpenCensus 0.12.2.  We use a custom exporter to export OpenCensus metrics into our own system.  I noticed that a lot of CPU time was spent in ViewData.checkAggregation (https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/stats/ViewData.java#L252).  Specific to our use case, we create `Distribution` objects with ~10,000 buckets to get high precision on our percentile calculations.   `createErrorMessageForAggregation` is called for each View instance, regardless if the error message will be used or not.  This then calls .toString on the Distribution instance, which calls toString on the bucket boundaries, which in my case was very large.

sebright:
I agree that we shouldn't generate the error message if it isn't needed.  We should add a `checkArgument` method that takes an object that generates the message String lazily.

songy23:
Thanks for investigating and reporting this issue. I'll send a PR to fix this soon.

steveniemitz:
Thank you!

PR #1267, opened 2018-06-19
Stats: Only include the simple class name in the error message for ViewData.check methods.

songy23:
Fixes https://github.com/census-instrumentation/opencensus-java/issues/1265.

When there's a mismatch between `Aggregation` - `AggregationData` or `Window` - `WindowData`, all we care about is their types. Only include the subclass name in the error message instead of calling the `toString` method to reduce overhead.

/cc @steveniemitz

### DIFF
--- a/api/src/main/java/io/opencensus/stats/ViewData.java
+++ b/api/src/main/java/io/opencensus/stats/ViewData.java
@@ -21,7 +21,6 @@ import io.opencensus.common.Duration;
 import io.opencensus.common.Function;
 import io.opencensus.common.Functions;
 import io.opencensus.common.Timestamp;
-import io.opencensus.internal.Utils;
 import io.opencensus.stats.Aggregation.Count;
 import io.opencensus.stats.Aggregation.Distribution;
 import io.opencensus.stats.Aggregation.LastValue;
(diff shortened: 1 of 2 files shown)


api/src/main/java/io/opencensus/stats/ViewData.java
```diff
@@ -232,9 +232,9 @@ private static String createErrorMessageForWindow(
       View.AggregationWindow window, AggregationWindowData windowData) {
     return "AggregationWindow and AggregationWindowData types mismatch. "
         + "AggregationWindow: "
-        + window
+        + window.getClass().getSimpleName()
         + " AggregationWindowData: "
-        + windowData;
+        + windowData.getClass().getSimpleName();
```
rghetia: May be create another checkArgument() which takes generic Object obj and calls obj.getClass().getSimpleName() only if validation fails.

songy23: Done, though this might be an overkill since getClass().getSimpleName() should only introduce very little overhead.

sebright:
It looks like some of the tests in ViewDataTest are currently only testing a prefix of this error message.  Should we test the whole message now that it is simpler?

sebright:
Everything else looks good.

songy23:
> It looks like some of the tests in ViewDataTest are currently only testing a prefix of this error message. Should we test the whole message now that it is simpler?

Done.

songy23:
Thanks!

PR #1268, opened 2018-06-19
Stackdriver Stats Exporter: allow user to force update existing Metric with new schema.

songy23:
Make the Stackdriver Stats Exporter more robust by allowing schema change in Stackdriver `MetricDescriptor`. The default config is not to do force update. If a user set `StackdriverStatsConfiguration.overrideExistingMetrics` to true, when they update existing views to a different schema, the exporter will delete the existing `MetricDescriptor` from Stackdriver and create one with the new schema.

This new config will be also useful if we want to remove the opencensus-task label from existing `MetricDescriptor`s, which is a backwards-incompatible schema change.

/cc @ramonza

### DIFF
--- a/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporterWorker.java
+++ b/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporterWorker.java
@@ -24,6 +24,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Lists;
 import com.google.monitoring.v3.CreateMetricDescriptorRequest;
 import com.google.monitoring.v3.CreateTimeSeriesRequest;
+import com.google.monitoring.v3.MetricDescriptorName;
 import com.google.monitoring.v3.ProjectName;
 import com.google.monitoring.v3.TimeSeries;
 import io.opencensus.common.Duration;
@@ -71,6 +72,7 @@ final class StackdriverExporterWorker implements Runnable {
   private final MetricServiceClient metricServiceClient;
   private final ViewManager viewManager;
   private final MonitoredResource monitoredResource;
+  private final boolean overrideExistingMetrics;
   private final Map<View.Name, View> registeredViews = new HashMap<View.Name, View>();
 
   private static final Tracer tracer = Tracing.getTracer();
@@ -81,13 +83,15 @@ final class StackdriverExporterWorker implements Runnable {
       MetricServiceClient metricServiceClient,
       Duration exportInterval,
       ViewManager viewManager,
-      MonitoredResource monitoredResource) {
+      MonitoredResource monitoredResource,
+      boolean overrideExistingMetrics) {
     this.scheduleDelayMillis = exportInterval.toMillis();
     this.projectId = projectId;
     projectName = ProjectName.newBuilder().setProject(projectId).build();
     this.metricServiceClient = metricServiceClient;
     this.viewManager = viewManager;
     this.monitoredResource = monitoredResource;
+    this.overrideExistingMetrics = overrideExistingMetrics;
 
     Tracing.getExportComponent()
         .getSampledSpanStore()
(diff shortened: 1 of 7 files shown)


exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporterWorker.java
```diff
@@ -105,16 +109,23 @@ boolean registerView(View view) {
...
         return true;
       } else {
-        // If we upload a view that has the same name with a registered view but with different
-        // attributes, Stackdriver client will throw an exception.
-        logger.log(
-            Level.WARNING,
-            "A different view with the same name is already registered: " + existing);
-        return false;
+        if (overrideExistingMetrics) {
+          deleteMetricDescriptor(view);
```
rghetia: can it fail without deleting the view? If so, what is the expectation?

songy23: If this failed, the subsequent call to create MetricDescriptor would fail too. Both failures will be logged and silently skipped, and stats with the changed MetricDescriptor will not be exported. This is also the current behavior of exporter.

songy23: I updated the logic here to fail fast if `deleteMetricDescriptor` failed, since if `deleteMetricDescriptor` failed, `createMetricDescriptor` will fail for sure.

ramonza:
What happens to the old data when we do this? If it just gets deleted then I am -1 on providing this option at all. I don't think we should cause data-loss automatically, the user should go and do that in the UI.

songy23:
The old `TimeSeries` will be deleted: https://github.com/census-instrumentation/opencensus-java/pull/1268/commits/728b01daea390f29a498c431256ff95d4b0d288f#diff-472b46d90df6809e393649362aa48c23R148. The default option is not to do force update, and I want to make sure users are aware of the risk if they explicitly enable this.

I still think we should provide this functionality because:
1. If users don't want the opencensus-task label for metrics that are already uploaded, they will still end up deleting all of them manually, and that's a lot of work;
2. Updating an existing `View` schema is in our V2 plan and is likely to be supported in the future. If we add that functionality and users begin to update their `View`s, they will still need to delete the corresponding metrics manually.
3. We have updated RPC and HTTP constants in the past (e.g https://github.com/census-instrumentation/opencensus-specs/pull/78), and if this happened again, there will the same problem that users need to manually delete the created RPC/HTTP metrics.

ramonza:
I agree that it's painful. We should file an issue with Stackdriver for them to make it less painful e.g. by not caring so much if a label is deleted.

I still don't think the exporter should delete historical data under any circumstances. 

If you really want to provide this functionality right now, I think we should provide a separate utility class to explicitly check the schema and this class can have an option to perform overwrites.

In future, once we have defined a text view configuration format (which I believe @rakyll is working on as part of the opencensus-service), we could instead provide a command line tool that operates on the view definitions to reconcile them with the backend.

ramonza:
Another idea might be for us to provide a hook for users to customize the behavior on schema mismatch. For labels that have been added that we don't know about (like opencensus_task), we could allow the user to specify a default value. For labels that have been removed, we could (and maybe we should by default) not throw an error but just log a warning and drop the extra tag.

songy23:
> I still don't think the exporter should delete historical data under any circumstances.

Sorry I was wrong at first. Looked at the Stackdriver documentation and I found this (https://cloud.google.com/monitoring/custom-metrics/creating-metrics#deleting_metrics): 

> You cannot delete the time series data from your custom metric, but the data expires and is deleted according to the data retention policy.

So there won't be loss of historical data even if we delete the metrics.

ramonza:
The commend below says:
> Caution: At present, deleting a metric descriptor makes its existing time series data inaccessible, but if you were to recreate the metric descriptor exactly as it was before deletion, then you might regain access to the data. Do not depend on this behavior.

So this isn't really any different from deleting the data from the perspective of potential data loss, you are just not *guaranteed( to have your data deleted!

songy23:
OK, I'll close this PR for now and see if there's really a need for this functionality from users.

PR #1269, opened 2018-06-20
Add contrib/log_correlation/stackdriver/README.md to RELEASING.md.

sebright:
The readme contains build file snippets that need to be updated for each
release.


sebright:
Yes, I'll update that next.

PR #1279, opened 2018-06-22
Examples: Record some stats and spans for ZPageTester.

songy23:
So that there will be some useful information on the z pages.

**Before**:
Tracez
zpagetester-tracez-old
Statsz
zpagetester-statsz-old
Rpcz
zpagetester-rpcz-old

**After**:
Tracez
zpagetester-tracez-new
Statsz
zpagetester-statsz-new
Rpcz
zpagetester-rpcz-new

/cc @adamgarza

### DIFF
--- a/examples/src/main/java/io/opencensus/examples/zpages/ZPagesTester.java
+++ b/examples/src/main/java/io/opencensus/examples/zpages/ZPagesTester.java
@@ -16,14 +16,82 @@
 
 package io.opencensus.examples.zpages;
 
+import io.opencensus.common.Scope;
+import io.opencensus.contrib.grpc.metrics.RpcMeasureConstants;
+import io.opencensus.contrib.grpc.metrics.RpcViews;
 import io.opencensus.contrib.zpages.ZPageHandlers;
+import io.opencensus.stats.MeasureMap;
+import io.opencensus.stats.Stats;
+import io.opencensus.stats.StatsRecorder;
+import io.opencensus.tags.TagValue;
+import io.opencensus.tags.Tagger;
+import io.opencensus.tags.Tags;
+import io.opencensus.trace.SpanBuilder;
+import io.opencensus.trace.Tracer;
+import io.opencensus.trace.Tracing;
+import io.opencensus.trace.samplers.Samplers;
+import java.util.Collections;
 
 /** Testing only class for the UI. */
 public class ZPagesTester {
+
   private ZPagesTester() {}
+
+  private static final Tagger tagger = Tags.getTagger();
+  private static final Tracer tracer = Tracing.getTracer();
+  private static final StatsRecorder statsRecorder = Stats.getStatsRecorder();
+
+  private static final String SPAN_NAME = "SampleSpan";
 
   /** Main method. */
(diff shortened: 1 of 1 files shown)


examples/src/main/java/io/opencensus/examples/zpages/ZPagesTester.java
```diff
@@ -16,14 +16,87 @@
...
 /** Testing only class for the UI. */
 public class ZPagesTester {
+
   private ZPagesTester() {}
 
+  private static final Tagger tagger = Tags.getTagger();
+  private static final Tracer tracer = Tracing.getTracer();
+  private static final StatsRecorder statsRecorder = Stats.getStatsRecorder();
+
+  private static final String SPAN_NAME = "SampleSpan";
```
sebright: Suggestion: "ExampleSpan" might be better, to avoid confusion with "sampling".

examples/src/main/java/io/opencensus/examples/zpages/ZPagesTester.java
```diff
@@ -16,14 +16,87 @@
...
+          tagger
+              .currentBuilder()
+              .put(RpcMeasureConstants.RPC_STATUS, TagValue.create("UNKNOWN"))
+              .build());
+
+      Thread.sleep(200); // sleep for fake work.
+      tracer.getCurrentSpan().addAnnotation("Finish recording.");
+    }
+
     ZPageHandlers.startHttpServerAndRegisterAll(8080);
```
sebright: Shouldn't the zpages be started at the beginning of the method, since they are usually registered at program initialization?

examples/src/main/java/io/opencensus/examples/zpages/ZPagesTester.java
```diff
@@ -16,14 +16,87 @@
...
 
+  private static final Tagger tagger = Tags.getTagger();
+  private static final Tracer tracer = Tracing.getTracer();
+  private static final StatsRecorder statsRecorder = Stats.getStatsRecorder();
+
+  private static final String SPAN_NAME = "SampleSpan";
+
   /** Main method. */
   public static void main(String[] args) throws Exception {
+    Tracing.getExportComponent()
```
sebright: Suggestion: It might be better to put this section of new code in a separate method, such as "createExampleData", to make it clear why the span and stats are being recorded.

examples/src/main/java/io/opencensus/examples/zpages/ZPagesTester.java
songy23: SG, done.

examples/src/main/java/io/opencensus/examples/zpages/ZPagesTester.java
songy23: Done.

examples/src/main/java/io/opencensus/examples/zpages/ZPagesTester.java
songy23: That's right, done.

songy23:
Thanks!

PR #1284, opened 2018-06-27
Stats: Add new MeasureMap.record() APIs that take a string map for recording exemplars.

songy23:
Add two new record() APIs based on the discussion in https://github.com/census-instrumentation/opencensus-specs/pull/122.

To do next (in another PR): add the Exemplar class, record and store Exemplars in impl.

/cc @rghetia


songy23:
Chatted offline, we thought the APIs would be easier to use if we add the attachments with a builder pattern instead of overloading the record() methods. Closing, will update and post a new PR using the builder approach.

PR #1285, opened 2018-06-27
Stats: Add API MeasureMap.putAttachment() for recording exemplars.

songy23:
Add a builder API that supports attaching contextual information of exemplars to measure maps. This should make the API easier to use, compared to overloading and having multiple versions of record (https://github.com/census-instrumentation/opencensus-java/pull/1284).

To do next (in another PR): add the Exemplar class, record and store Exemplars in impl.

/cc @rghetia

### DIFF
--- a/api/src/main/java/io/opencensus/stats/MeasureMap.java
+++ b/api/src/main/java/io/opencensus/stats/MeasureMap.java
@@ -19,6 +19,7 @@ package io.opencensus.stats;
 import io.opencensus.stats.Measure.MeasureDouble;
 import io.opencensus.stats.Measure.MeasureLong;
 import io.opencensus.tags.TagContext;
+import java.util.Map;
 import javax.annotation.concurrent.NotThreadSafe;
 
 /**
@@ -50,6 +51,17 @@ public abstract class MeasureMap {
    * @since 0.8
    */
   public abstract MeasureMap put(MeasureLong measure, long value);
+
+  /**
+   * Associate the contextual information of an {@code Exemplar} to this {@link MeasureMap}.
+   *
+   * <p>If this method is called multiple times, only the last string map will be kept.
+   *
+   * @param attachments contextual information of an {@code Exemplar}.
+   * @return this
+   * @since 0.16
+   */
+  public abstract MeasureMap withAttachments(Map<String, String> attachments);
 
   /**
    * Records all of the measures at the same time, with the current {@link TagContext}.
--- a/api/src/main/java/io/opencensus/stats/MeasureMap.java
+++ b/api/src/main/java/io/opencensus/stats/MeasureMap.java
@@ -16,6 +16,7 @@
 
 package io.opencensus.stats;
 
+import io.opencensus.internal.Utils;
 import io.opencensus.stats.Measure.MeasureDouble;
 import io.opencensus.stats.Measure.MeasureLong;
 import io.opencensus.tags.TagContext;
@@ -55,13 +56,19 @@ public abstract class MeasureMap {
   /**
    * Associate the contextual information of an {@code Exemplar} to this {@link MeasureMap}.
    *
-   * <p>If this method is called multiple times, only the last string map will be kept.
+   * <p>If this method is called multiple times, the string maps should be merged. If there are
+   * multiple values with the same key in the maps, only the last value will be kept. I.e equivalent
+   * to {@code new HashMap<>().putAll(map1).putAll(map2)}.
    *
    * @param attachments contextual information of an {@code Exemplar}.
    * @return this
    * @since 0.16
    */
-  public abstract MeasureMap withAttachments(Map<String, String> attachments);
+  public MeasureMap putAttachments(Map<String, String> attachments) {
+    // Provides a default no-op implementation to avoid breaking other existing sub-classes.
+    Utils.checkNotNull(attachments, "attachments");
+    return this;
+  }
 
   /**
    * Records all of the measures at the same time, with the current {@link TagContext}.
--- a/api/src/main/java/io/opencensus/stats/MeasureMap.java
+++ b/api/src/main/java/io/opencensus/stats/MeasureMap.java
@@ -63,7 +63,8 @@ public abstract class MeasureMap {
    * @return this
    * @since 0.16
    */
-  public MeasureMap putAttachement(String key, String value) {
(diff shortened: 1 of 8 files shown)


api/src/main/java/io/opencensus/stats/MeasureMap.java
```diff
@@ -51,6 +52,17 @@
...
+  /**
+   * Associate the contextual information of an {@code Exemplar} to this {@link MeasureMap}.
+   *
+   * <p>If this method is called multiple times, only the last string map will be kept.
+   *
+   * @param attachments contextual information of an {@code Exemplar}.
+   * @return this
+   * @since 0.16
+   */
+  public abstract MeasureMap withAttachments(Map<String, String> attachments);
```
sebright: It other parts of the API, we use a "with" prefix for methods that manipulate the context or return a copy of an immutable object, so I don't think this method should start with "with".  Other possibilities are "setAttachments", if the method replaces existing attachments, or "putAttachements", if the method merges new and existing attachments.

api/src/main/java/io/opencensus/stats/MeasureMap.java
```diff
@@ -51,6 +52,17 @@
    */
   public abstract MeasureMap put(MeasureLong measure, long value);
 
+  /**
+   * Associate the contextual information of an {@code Exemplar} to this {@link MeasureMap}.
+   *
+   * <p>If this method is called multiple times, only the last string map will be kept.
```
sebright: Should we merge the attachments instead?  The keys allow users to add multiple types of exemplars, so I think it would be better to allow the exemplars to be added independently.  For example, a user could write helper methods:

```java
void addCurrentSpan(MeasureMap measures);
void addThrowable(MeasureMap measures, Throwable throwable);
```

Then `addCurrentSpan` could add keys such as "traceId" and "spanId", and `addThrowable` could add "errorMessage".

api/src/main/java/io/opencensus/stats/MeasureMap.java
sebright: If we use "attachments" in the method name, I think we should use that wording in opencensus-specs, to ensure consistency across APIs.  I don't mind merging this PR first, though.

sebright: This method should be non-abstract (no-op), to avoid breaking any existing subclasses.

impl_core/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
```diff
@@ -206,7 +214,11 @@ private CumulativeMutableViewData(View view, Timestamp start) {
     }
 
     @Override
-    void record(TagContext context, double value, Timestamp timestamp) {
+    void record(
+        TagContext context,
+        double value,
+        Timestamp timestamp,
+        @javax.annotation.Nullable Map<String, String> attachments) {
```
sebright: Should there be a TODO here about storing the attachments?

api/src/main/java/io/opencensus/stats/MeasureMap.java
songy23: Make sense to me, updated. And I think if a new value for an existing key is added to a `MeasureMap`, it should still overwrite the previous value (e.g putting a new TraceId should overwrite the previous one).

api/src/main/java/io/opencensus/stats/MeasureMap.java
songy23: Renamed to `putAttachements`.

songy23: Done, though I'm wondering should we rather just break other subclasses and force them to implement this method?

songy23: Yes, I'll update the method names in specs in a following PR.

impl_core/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
songy23: Done.

api/src/main/java/io/opencensus/stats/MeasureMap.java
dinooliva: Do we have any process for making this become abstract at some point?

api/src/main/java/io/opencensus/stats/MeasureMap.java
```diff
@@ -51,6 +53,23 @@
...
+   * @param attachments contextual information of an {@code Exemplar}.
+   * @return this
+   * @since 0.16
+   */
+  public MeasureMap putAttachments(Map<String, String> attachments) {
+    // Provides a default no-op implementation to avoid breaking other existing sub-classes.
+    Utils.checkNotNull(attachments, "attachments");
+    return this;
+  }
+
```
dinooliva: Consider alternatively 'put(String, String)' which accumulates an attachment (or 'putAttachement()'),

songy23: Sounds good, I updated it to `putAttachement(String, String)` since we already have other methods named `put()` for `MeasureMap`.

api/src/main/java/io/opencensus/stats/MeasureMap.java
sebright: We would need wait at least 1.5 years and increment the major version.

api/src/main/java/io/opencensus/stats/MeasureMap.java
sebright: Yes, I think it makes sense to overwrite the previous value.

api/src/main/java/io/opencensus/stats/MeasureMap.java
sebright: Actually, we could make the method abstract sooner if necessary, since we haven't released 1.0 yet, but I would rather keep it non-abstract for longer if we don't have a reason to change it.  We should at least wait one release cycle so that users are able to migrate.

songy23: Sounds good to me, I'll leave a TODO to make this method abstract in the 0.17 release.

dinooliva: sgtm

songy23:
Thanks!

PR #1287, opened 2018-06-28
Stats: Add Exemplar class to DistributionData.

songy23:
[Proto definition](https://github.com/census-instrumentation/opencensus-proto/blob/master/opencensus/proto/stats/metrics/metrics.proto#L236-L250):

```
  // Exemplars are example points that may be used to annotate aggregated
  // Distribution values. They are metadata that gives information about a
  // particular value added to a Distribution bucket.
  message Exemplar {
    // Value of the exemplar point. It determines which bucket the exemplar
    // belongs to.
    double value = 1;

    // The observation (sampling) time of the above value.
    google.protobuf.Timestamp timestamp = 2;

    // Contextual information about the example value.
    map<string, string> attachments = 3;
  }
```

Next step: record and store the `Exemplar`s in the impl (in another PR).

### DIFF
--- a/api/src/main/java/io/opencensus/stats/AggregationData.java
+++ b/api/src/main/java/io/opencensus/stats/AggregationData.java
@@ -447,7 +447,7 @@ public abstract class AggregationData {
        * Creates an {@link Exemplar}.
        *
        * @param value value of the {@link Exemplar} point.
-       * @param timestamp the observation (sampling) time of the above value.
+       * @param timestamp the time that this {@code Exemplar}'s value was recorded.
        * @param attachments the contextual information about the example value.
        * @return an {@code Exemplar}.
        * @since 0.16
--- a/api/src/main/java/io/opencensus/stats/AggregationData.java
+++ b/api/src/main/java/io/opencensus/stats/AggregationData.java
@@ -25,6 +25,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import javax.annotation.concurrent.Immutable;
 
 /**
@@ -294,10 +295,6 @@ public abstract class AggregationData {
       }
 
       Utils.checkNotNull(exemplars, "exemplar list should not be null.");
-      Utils.checkArgument(
-          exemplars.size() <= bucketCounts.size(),
-          "each histogram bucket can have up to one exemplar, "
-              + "so the length of exemplar list should be no greater than that of bucket list.");
       for (Exemplar exemplar : exemplars) {
         Utils.checkNotNull(exemplar, "exemplar should not be null.");
       }
@@ -309,7 +306,7 @@ public abstract class AggregationData {
           max,
           sumOfSquaredDeviations,
           bucketCountsCopy,
-          Collections.<Exemplar>unmodifiableList(exemplars));
+          Collections.<Exemplar>unmodifiableList(new ArrayList<Exemplar>(exemplars)));
     }
 
     /**
@@ -411,8 +408,8 @@ public abstract class AggregationData {
     }
 
     /**
-     * Example points that may be used to annotate aggregated distribution values, associated with a
-     * histogram.
+     * An example point that may be used to annotate aggregated distribution values, associated with
+     * a histogram bucket.
      *
      * @since 0.16
      */
@@ -431,9 +428,9 @@ public abstract class AggregationData {
       public abstract double getValue();
 
       /**
-       * Returns the observation (sampling) time of the above value.
+       * Returns the time that this {@link Exemplar}'s value was recorded.
        *
-       * @return the observation (sampling) time of the above value.
+       * @return the time that this {@code Exemplar}'s value was recorded.
        * @since 0.16
        */
       public abstract Timestamp getTimestamp();
@@ -458,10 +455,14 @@ public abstract class AggregationData {
       public static Exemplar create(
           double value, Timestamp timestamp, Map<String, String> attachments) {
         Utils.checkNotNull(attachments, "attachments");
+        Map<String, String> attachmentsCopy =
+            Collections.unmodifiableMap(new HashMap<String, String>(attachments));
+        for (Entry<String, String> entry : attachmentsCopy.entrySet()) {
+          Utils.checkNotNull(entry.getKey(), "key of attachments");
+          Utils.checkNotNull(entry.getValue(), "value of attachments");
+        }
         return new AutoValue_AggregationData_DistributionData_Exemplar(
-            value,
-            timestamp,
-            Collections.unmodifiableMap(new HashMap<String, String>(attachments)));
+            value, timestamp, attachmentsCopy);
       }
     }
   }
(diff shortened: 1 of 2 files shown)


api/src/main/java/io/opencensus/stats/AggregationData.java
```diff
@@ -288,8 +293,54 @@ public static DistributionData create(
...
+      Utils.checkNotNull(exemplars, "exemplar list should not be null.");
+      Utils.checkArgument(
+          exemplars.size() <= bucketCounts.size(),
+          "each histogram bucket can have up to one exemplar, "
+              + "so the length of exemplar list should be no greater than that of bucket list.");
+      for (Exemplar exemplar : exemplars) {
+        Utils.checkNotNull(exemplar, "exemplar should not be null.");
+      }
+
+      DistributionData data =
```
dinooliva: Why isn't this just a return?

songy23: Good catch, fixed.

api/src/main/java/io/opencensus/stats/AggregationData.java
```diff
@@ -288,8 +293,52 @@ public static DistributionData create(
         Utils.checkNotNull(bucket, "bucket should not be null.");
       }
 
+      Utils.checkNotNull(exemplars, "exemplar list should not be null.");
+      Utils.checkArgument(
+          exemplars.size() <= bucketCounts.size(),
+          "each histogram bucket can have up to one exemplar, "
+              + "so the length of exemplar list should be no greater than that of bucket list.");
```
sebright: I think we should either allow any number of exemplars or perform this check by comparing the exemplar values with the bucket boundaries.  The current check isn't complete, because it would allow two exemplars in one bucket if there were zero exemplars in another bucket.

api/src/main/java/io/opencensus/stats/AggregationData.java
```diff
@@ -288,8 +293,52 @@ public static DistributionData create(
...
+
       return new AutoValue_AggregationData_DistributionData(
-          mean, count, min, max, sumOfSquaredDeviations, bucketCountsCopy);
+          mean,
+          count,
+          min,
+          max,
+          sumOfSquaredDeviations,
+          bucketCountsCopy,
+          Collections.<Exemplar>unmodifiableList(exemplars));
```
sebright: The exemplar list also needs to be copied.

api/src/main/java/io/opencensus/stats/AggregationData.java
```diff
@@ -352,6 +409,61 @@ public static DistributionData create(
...
+       *
+       * @return value of the {@code Exemplar} point.
+       * @since 0.16
+       */
+      public abstract double getValue();
+
+      /**
+       * Returns the observation (sampling) time of the above value.
+       *
+       * @return the observation (sampling) time of the above value.
```
sebright: The meaning of "sampling" isn't very clear in this Javadoc (and `Exemplar.create`).

Since AggregationData isn't part of the metrics API, I think these exemplars should only be recorded by OpenCensus, so we could just say something like "the time that this Exemplar's value was recorded".

api/src/main/java/io/opencensus/stats/AggregationData.java
```diff
@@ -352,6 +409,61 @@ public static DistributionData create(
...
+       *
+       * @param value value of the {@link Exemplar} point.
+       * @param timestamp the observation (sampling) time of the above value.
+       * @param attachments the contextual information about the example value.
+       * @return an {@code Exemplar}.
+       * @since 0.16
+       */
+      public static Exemplar create(
+          double value, Timestamp timestamp, Map<String, String> attachments) {
+        Utils.checkNotNull(attachments, "attachments");
```
sebright: AutoValue should already add null checks all of the parameters that aren't nullable, though we should check that the keys and values are non-null after making the copy.

api/src/test/java/io/opencensus/stats/AggregationDataTest.java
```diff
@@ -57,6 +64,31 @@ public void testCreateDistributionData() {
...
   }
 
+  @Test
+  public void testCreateDistributionDataWithExemplar() {
+    Exemplar exemplar1 = Exemplar.create(4, TIMESTAMP_2, ATTACHMENTS);
+    Exemplar exemplar2 = Exemplar.create(1, TIMESTAMP_1, ATTACHMENTS);
+    DistributionData distributionData =
+        DistributionData.create(
+            7.7, 10, 1.1, 9.9, 32.2, Arrays.asList(4L, 1L), Arrays.asList(exemplar1, exemplar2));
+    assertThat(distributionData.getExemplars()).containsExactly(exemplar1, exemplar2).inOrder();
```
sebright: Should we sort the exemplars by value?

api/src/test/java/io/opencensus/stats/AggregationDataTest.java
```diff
@@ -57,6 +64,31 @@ public void testCreateDistributionData() {
...
+  @Test
+  public void testExemplar() {
+    Exemplar exemplar = Exemplar.create(15.0, TIMESTAMP_1, ATTACHMENTS);
+    assertThat(exemplar.getValue()).isEqualTo(15.0);
+    assertThat(exemplar.getTimestamp()).isEqualTo(TIMESTAMP_1);
+    assertThat(exemplar.getAttachments()).isEqualTo(ATTACHMENTS);
+  }
+
+  @Test
+  public void testExemplar_PreventNullAttachments() {
```
sebright: This test suite could also test a null timestamp.

api/src/main/java/io/opencensus/stats/AggregationData.java
```diff
@@ -352,6 +409,61 @@ public static DistributionData create(
         Function<? super AggregationData, T> defaultFunction) {
       return p3.apply(this);
     }
+
+    /**
+     * Example points that may be used to annotate aggregated distribution values, associated with a
```
sebright: I think point

sebright: I think "points" should be singular.

api/src/main/java/io/opencensus/stats/AggregationData.java
sebright: The Javadoc for `Exemplar.create` still says "sampling".

api/src/main/java/io/opencensus/stats/AggregationData.java
```diff
@@ -352,6 +406,65 @@ public static DistributionData create(
...
+      /**
+       * Creates an {@link Exemplar}.
+       *
+       * @param value value of the {@link Exemplar} point.
+       * @param timestamp the time that this {@code Exemplar}'s value was recorded.
+       * @param attachments the contextual information about the example value.
+       * @return an {@code Exemplar}.
+       * @since 0.16
+       */
+      public static Exemplar create(
```
dinooliva: Should this be public or package private? Either is fine, mostly curious about expected use.

songy23: I would prefer to make the factory method public. This makes it easier for any library that uses `Exemplar` (e.g exporters) to write unit tests.

Issue #1286, opened 2018-06-28
Should we mention OpenCensus-Java is currently in Beta quality level?


songy23:
Definition of [Beta](https://github.com/GoogleCloudPlatform/google-cloud-java#versioning) quality level:

> Beta: Libraries defined at a Beta quality level are expected to be mostly stable and we're working towards their release candidate. We will address issues and requests with a higher priority.

Looks like Beta matches our current quality level. I think the plan is we will officially move to GA level after the v1.0.0 major release.

From discussion: https://github.com/census-instrumentation/opencensus-java/pull/1285#discussion_r198628596.

sebright:
I think that makes sense.

PR #1290, opened 2018-06-28
Readme: Update library stage to Beta.

songy23:
Fixes https://github.com/census-instrumentation/opencensus-java/issues/1286.

Definition of [Alpha](https://github.com/GoogleCloudPlatform/google-cloud-java#versioning) quality level:

> Alpha: Libraries defined at an Alpha quality level are still a work-in-progress and are more likely to get backwards-incompatible updates. Additionally, it's possible for Alpha libraries to get deprecated and deleted before ever being promoted to Beta or GA.

Definition of [Beta](https://github.com/GoogleCloudPlatform/google-cloud-java#versioning) quality level:

> Beta: Libraries defined at a Beta quality level are expected to be mostly stable and we're working towards their release candidate. We will address issues and requests with a higher priority.

~~I think it's time to call OpenCensus-Java a Beta-stage library.~~

EDIT: I think Beta-stage matches the situation of OpenCensus-Java better.

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,13 +199,15 @@ This library follows [Semantic Versioning][semver].
   
 **GA**: Libraries defined at a GA quality level are stable, and will not introduce 
 backwards-incompatible changes in any minor or patch releases. We will address issues and requests 
-with the highest priority.
+with the highest priority. If we were to make a backwards-incompatible changes on an API, we will 
+first mark the existing API as deprecated and keep it for 18 months before removing it.
   
 **Beta**: Libraries defined at a Beta quality level are expected to be mostly stable and we're 
 working towards their release candidate. We will address issues and requests with a higher priority.
-  
-**Alpha**: Libraries defined at an Alpha quality level are still a work-in-progress and are more 
-likely to get backwards-incompatible updates.
+There may be backwards incompatible changes in a minor version release, though not in a patch 
+release. If an element is part of an API that is only meant to be used by exporters or other 
+opencensus libraries, then there is no deprecation period. Otherwise, we will deprecate it for 18 
+months before removing it, if possible.
 
 [travis-image]: https://travis-ci.org/census-instrumentation/opencensus-java.svg?branch=master
 [travis-url]: https://travis-ci.org/census-instrumentation/opencensus-java
--- a/README.md
+++ b/README.md
@@ -10,7 +10,8 @@
 OpenCensus is a toolkit for collecting application performance and behavior data. It currently
 includes 3 apis: stats, tracing and tags.
 
-The library is in alpha stage and the API is subject to change.
+The library is in [Beta][cloud-versioning] stage and APIs are expected to be mostly stable. The 
+library is expected to move to [GA][cloud-versioning] stage after v1.0.0 major release.
 
 Please join [gitter](https://gitter.im/census-instrumentation/Lobby) for help or feedback on this
 project.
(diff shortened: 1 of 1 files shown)


rghetia:
changes look good. However, when we do the next release should the version contain 'beta'?

README.md
```diff
@@ -202,6 +203,7 @@ see this [link](https://github.com/census-instrumentation/opencensus-java/tree/m
 [maven-url]: https://maven-badges.herokuapp.com/maven-central/io.opencensus/opencensus-api
 [gitter-image]: https://badges.gitter.im/census-instrumentation/lobby.svg
 [gitter-url]: https://gitter.im/census-instrumentation/lobby?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge
+[cloud-versioning]: https://github.com/GoogleCloudPlatform/google-cloud-java#versioning
```
sebright: I think these definitions are specific to google-cloud-java.  Is there a more standard definition we could use?  If not, we could just describe the level of stability that we are aiming for in a few sentences.

sebright:
@rghetia I think we are using major version 0 to indicate that it is beta.

rghetia:
@sebright  How does a user know from which version onwards it transitioned from alpha to beta?

README.md
songy23: Didn't find an official definition for the versioning. I saw that opencensus-php-exporter-stackdriver copied the definition of versioning in their [README](https://github.com/census-ecosystem/opencensus-php-exporter-stackdriver#versioning), I'll copy that to our README too.

README.md
```diff
@@ -192,6 +193,20 @@ runtime 'io.opencensus:opencensus-impl:0.15.0'
...
+This library follows [Semantic Versioning][semver].
+  
+**GA**: Libraries defined at a GA quality level are stable, and will not introduce 
+backwards-incompatible changes in any minor or patch releases. We will address issues and requests 
+with the highest priority.
+  
+**Beta**: Libraries defined at a Beta quality level are expected to be mostly stable and we're 
+working towards their release candidate. We will address issues and requests with a higher priority.
+  
+**Alpha**: Libraries defined at an Alpha quality level are still a work-in-progress and are more
```
sebright: I don't think we need the alpha section for this library now.

README.md
```diff
@@ -192,6 +193,20 @@ runtime 'io.opencensus:opencensus-impl:0.15.0'
...
+## Versioning
+  
+This library follows [Semantic Versioning][semver].
+  
+**GA**: Libraries defined at a GA quality level are stable, and will not introduce 
+backwards-incompatible changes in any minor or patch releases. We will address issues and requests 
+with the highest priority.
+  
+**Beta**: Libraries defined at a Beta quality level are expected to be mostly stable and we're 
+working towards their release candidate. We will address issues and requests with a higher priority.
```
sebright: You could say that we may introduce backwards incompatible changes in a minor version release, though not in a patch release.  You could also describe the deprecation process, though I'm not sure whether we've decided on a concrete process for versions before 1.0.  I think we've been following something like this:

If an element is part of an API that is only meant to be used by exporters or other opencensus libraries, then there is no deprecation period.  Otherwise, we will deprecate it for 1.5 years before removing it, if possible.

README.md
```diff
@@ -192,6 +193,20 @@ runtime 'io.opencensus:opencensus-impl:0.15.0'
 If the application owner wants to export in-process tracing and stats data via HTML debugging pages
 see this [link](https://github.com/census-instrumentation/opencensus-java/tree/master/contrib/zpages#quickstart).
 
+## Versioning
+  
+This library follows [Semantic Versioning][semver].
+  
+**GA**: Libraries defined at a GA quality level are stable, and will not introduce 
+backwards-incompatible changes in any minor or patch releases. We will address issues and requests 
+with the highest priority.
```
sebright: You could also describe the 1.5 year backwards compatibility window here.

sebright:
I'm not sure when this library transitioned from alpha to beta.  It has probably had beta level stability since we changed the artifact names from instrumentation to opencensus, so I think that it is probably more important to distinguish between beta and GA now.

rghetia:
@songy23 
for 0.x version it may not be an issue. But for subsequent release prior to next major release would need 'beta' indicator. I don't have any good suggestion on how to distinguish between alpha and beta. May be simply state that OC is in beta since 0.15.

rghetia:
@songy23 sounds good.

PR #1291, opened 2018-06-30
Metrics: Add MetricDescriptor.

songy23:
[Proto definition](https://github.com/census-instrumentation/opencensus-proto/blob/master/opencensus/proto/stats/metrics/metrics.proto#L54-L96):
```proto
// Defines a metric type and its schema.
message MetricDescriptor {
  // The metric type, including its DNS name prefix. It must be unique.
  string name = 1;

  // A detailed description of the metric, which can be used in documentation.
  string description = 2;

  // The unit in which the metric value is reported. Follows the format
  // described by http://unitsofmeasure.org/ucum.html.
  string unit = 3;

  // The kind of metric. It describes how the data is reported.
  //
  // A gauge is an instantaneous measurement of a value.
  //
  // A cumulative measurement is a value accumulated over a time interval. In
  // a time series, cumulative measurements should have the same start time and
  // increasing end times, until an event resets the cumulative value to zero
  // and sets a new start time for the following points.
  enum Type {
    // Do not use this default value.
    UNSPECIFIED = 0;

    // Integer gauge.
    GAUGE_INT64 = 1;

    // Floating point gauge.
    GAUGE_DOUBLE = 2;

    // Integer cumulative measurement.
    CUMULATIVE_INT64 = 3;

    // Floating point cumulative measurement.
    CUMULATIVE_DOUBLE = 4;

    // Distribution cumulative measurement.
    CUMULATIVE_DISTRIBUTION = 5;
  }
  Type type = 4;

  // The label keys associated with the metric descriptor.
  repeated LabelKey label_keys = 5;
}
```

### DIFF
--- a/metrics/src/main/java/io/opencensus/metrics/MetricDescriptor.java
+++ b/metrics/src/main/java/io/opencensus/metrics/MetricDescriptor.java
@@ -1,0 +1,155 @@
+/*
+ * Copyright 2018, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.metrics;
+
+import com.google.auto.value.AutoValue;
+import io.opencensus.common.ExperimentalApi;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import javax.annotation.concurrent.Immutable;
+
+/**
+ * {@link MetricDescriptor} defines a {@code Metric} type and its schema.
+ *
+ * @since 0.16
+ */
+@ExperimentalApi
+@Immutable
+@AutoValue
+public abstract class MetricDescriptor {
+
+  MetricDescriptor() {}
+
+  /**
+   * Creates a {@link MetricDescriptor}.
+   *
+   * @param name name of {@code MetricDescriptor}.
+   * @param description description of {@code MetricDescriptor}.
+   * @param unit the metric unit.
+   * @param type type of {@code MetricDescriptor}.
+   * @param labelKeys the label keys associated with the {@code MetricDescriptor}.
+   * @return a {@code MetricDescriptor}.
+   * @since 0.16
+   */
+  public static MetricDescriptor create(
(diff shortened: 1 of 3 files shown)


metrics/src/main/java/io/opencensus/metrics/MetricDescriptor.java
```diff
@@ -0,0 +1,155 @@
...
+   *
+   * @since 0.16
+   */
+  public enum Type {
+    /**
+     * Unspecified type. Do not use this as default value.
+     *
+     * @since 0.16
+     */
+    UNSPECIFIED,
```
sebright: Do we need UNSPECIFIED in the Java implementation?  When would it be used?

metrics/src/test/java/io/opencensus/metrics/MetricDescriptorTest.java
```diff
@@ -0,0 +1,101 @@
...
+    thrown.expect(NullPointerException.class);
+    thrown.expectMessage("labelKeys");
+    MetricDescriptor.create(METRIC_NAME_1, DESCRIPTION, UNIT, Type.GAUGE_DOUBLE, null);
+  }
+
+  @Test
+  public void preventNullLabelKey() {
+    List<LabelKey> keys = Arrays.asList(KEY_1, null);
+    thrown.expect(NullPointerException.class);
+    thrown.expectMessage("labelKey");
```
sebright: Suggestion: It might be better to test the exact error message, since this one is a substring of the error message from the test above.  For example, you could use `thrown.expectMessage(CoreMatchers.equalTo("labelKey"))`.

metrics/src/main/java/io/opencensus/metrics/MetricDescriptor.java
songy23: Short answer is I don't know.. I didn't see any specification on `Type.UNSPECIFIED` in proto or specs. I think we can ignore this value for now and add it later if needed.

metrics/src/test/java/io/opencensus/metrics/MetricDescriptorTest.java
songy23: That's true - updated. Same for other checks on null list element.

metrics/src/test/java/io/opencensus/metrics/TimeSeriesTest.java
```diff
@@ -27,6 +27,7 @@
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import org.hamcrest.CoreMatchers;
```
bogdandrutu: Why is this needed?

songy23: https://github.com/census-instrumentation/opencensus-java/pull/1291#discussion_r201159089:
> Suggestion: It might be better to test the exact error message, since this one is a substring of the error message from the test above. For example, you could use thrown.expectMessage(CoreMatchers.equalTo("labelKey")).

For example there's one exception with message "labelValue" and one with "labelValues",  CoreMatchers can be used to differentiate them.

PR #1293, opened 2018-06-30
 Metrics: Add Metric.

songy23:
~~Depend on https://github.com/census-instrumentation/opencensus-java/pull/1291.~~

[Proto definition](https://github.com/census-instrumentation/opencensus-proto/blob/master/opencensus/proto/stats/metrics/metrics.proto#L32-L45):

```proto

// Defines a Metric which has one or more timeseries.
message Metric {
  // The definition of the Metric. For now, we send the full MetricDescriptor
  // every time in order to keep the protocol stateless, but this is one of the
  // places where we can make future changes to make the protocol more
  // efficient.
  MetricDescriptor metric_descriptor = 1;

  // One or more timeseries for a single metric, where each timeseries has
  // one or more points. The type of the timeseries must match
  // metric_descriptor.type, so only one of the two should be populated.
  repeated GaugeTimeSeries gauge_timeseries = 2;
  repeated CumulativeTimeSeries cumulative_timeseries = 3;
}
```

EDIT: `MerticSet` is removed, per https://github.com/census-instrumentation/opencensus-proto/pull/66.

### DIFF
--- a/metrics/src/main/java/io/opencensus/metrics/Metric.java
+++ b/metrics/src/main/java/io/opencensus/metrics/Metric.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2018, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.metrics;
+
+import com.google.auto.value.AutoValue;
+import io.opencensus.common.ExperimentalApi;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import javax.annotation.concurrent.Immutable;
+
+/**
+ * A {@link Metric} with one or more {@link TimeSeries}.
+ *
+ * @since 0.16
+ */
+@ExperimentalApi
+@Immutable
+@AutoValue
+public abstract class Metric {
+
+  Metric() {}
+
+  /**
+   * Creates a {@link Metric}.
+   *
+   * @param metricDescriptor the {@link MetricDescriptor}.
+   * @param timeSeries the {@link TimeSeries} for this metric.
+   * @return a {@code Metric}.
+   * @since 0.16
+   */
+  public static Metric create(MetricDescriptor metricDescriptor, List<TimeSeries> timeSeries) {
+    Utils.checkNotNull(timeSeries, "timeSeriesList");
+    Utils.checkListElementNotNull(timeSeries, "timeSeries");
+    return new AutoValue_Metric(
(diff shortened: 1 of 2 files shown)


metrics/src/main/java/io/opencensus/metrics/Metric.java
```diff
@@ -0,0 +1,74 @@
...
+   * <p>Each {@link TimeSeries} has one or more {@link Point}.
+   *
+   * <p>The type of the {@link TimeSeries}s must match {@link MetricDescriptor.Type}, so the list
+   * should only contains either {@link TimeSeries.TimeSeriesCumulative} or {@link
+   * TimeSeries.TimeSeriesGauge}.
+   *
+   * @return the {@code TimeSeries}s for this metric.
+   * @since 0.16
+   */
+  public abstract List<TimeSeries> getTimeSeries();
```
sebright: This field should enforce that all of the TimeSeries in the list have the same type, as in metrics.proto.  One option is to check the consistency of the types in the factory method at runtime, though I think that adding a time series list class with two subclasses would be more similar to the Protocol Buffer.  Adding the time series list class would probably also allow us to remove the superclass for the two TimeSeries classes.

metrics/src/main/java/io/opencensus/metrics/Metric.java
```diff
@@ -0,0 +1,74 @@
...
+@AutoValue
+public abstract class Metric {
+
+  Metric() {}
+
+  /**
+   * Creates a {@link Metric}.
+   *
+   * @param metricDescriptor the {@link MetricDescriptor}.
+   * @param timeSeries the {@link TimeSeries} for this metric.
```
sebright: I think this should say `the {@link TimeSeries} list`".

metrics/src/main/java/io/opencensus/metrics/Metric.java
sebright: This field should enforce that all of the TimeSeries in the list have the same type, as in metrics.proto.  One option is to check the consistency of the types in the factory method at runtime, though I think that adding a time series list class with two subclasses would be more similar to the Protocol Buffer.  Adding the time series list class would probably also allow us to remove the superclass for the two TimeSeries classes.

metrics/src/main/java/io/opencensus/metrics/Metric.java
sebright: I think this should say "`the {@link TimeSeries} list`".

metrics/src/main/java/io/opencensus/metrics/Metric.java
```diff
@@ -0,0 +1,74 @@
...
+  /**
+   * Returns the {@link MetricDescriptor} of this metric.
+   *
+   * @return the {@code MetricDescriptor} of this metric.
+   * @since 0.16
+   */
+  public abstract MetricDescriptor getMetricDescriptor();
+
+  /**
+   * Returns the {@link TimeSeries}s for this metric.
```
sebright: Ditto

metrics/src/main/java/io/opencensus/metrics/Metric.java
songy23: > One option is to check the consistency of the types in the factory method at runtime, though I think that adding a time series list class with two subclasses would be more similar to the Protocol Buffer.

Traversing the list and check the subclass type could be expensive. The second option sounds better

> Adding the time series list class would probably also allow us to remove the superclass for the two TimeSeries classes.

I'll make another PR to make this change.

PR #1294, opened 2018-07-02
Stats: Support recording Exemplars in the impl.

songy23:
In the implementation, each histogram bucket can have up to one exemplar, and the exemplar array is guaranteed to be in ascending order. If there's no histogram, don't record exemplars.


impl_core/src/main/java/io/opencensus/implcore/stats/MutableAggregation.java
```diff
@@ -327,6 +345,13 @@ void combine(MutableAggregation other, double fraction) {
       for (int i = 0; i < bucketCounts.length; i++) {
         this.bucketCounts[i] += bucketCounts[i];
       }
+
+      for (int i = 0; i < mutableDistribution.getExemplars().length; i++) {
+        Exemplar exemplar = mutableDistribution.getExemplars()[i];
+        if (exemplar != null) {
+          this.exemplars[i] = exemplar;
```
sebright: Why does this method prefer the exemplar from `mutableDistribution`?  Is it always newer?

impl_core/src/test/java/io/opencensus/implcore/stats/MutableAggregationTest.java
```diff
@@ -135,6 +140,43 @@ public Void apply(MutableLastValue arg) {
     }
   }
 
+  @Test
+  public void testAdd_DistributionWithExemplarAttachments() {
+    MutableDistribution mutableDistribution = MutableDistribution.create(BUCKET_BOUNDARIES);
+    List<Double> values = Arrays.asList(-1.0, 1.0, -5.0, 20.0, 5.0);
+    @SuppressWarnings("unchecked")
+    List<Map<String, String>> attachmentsList =
+        Arrays.<Map<String, String>>asList(
```
sebright: I think you could use `ImmutableList.<Map<String, String>>of` to avoid the "unchecked" warning, since the list doesn't need to be modified.

impl_core/src/test/java/io/opencensus/implcore/stats/StatsRecorderImplTest.java
```diff
@@ -142,6 +156,101 @@ public void record_UnregisteredMeasure() {
...
+    viewManager.registerView(view);
+    recordWithAttachments();
+    ViewData viewData = viewManager.getView(VIEW_NAME);
+    assertThat(viewData).isNotNull();
+    CountData countData =
+        (CountData) viewData.getAggregationMap().get(Collections.singletonList(VALUE));
+    // Recording exemplar does not affect views with an aggregation other than distribution.
+    assertThat(countData.getCount()).isEqualTo(5L);
+  }
+
```
sebright: I think we should also test that a value without an exemplar doesn't overwrite an existing exemplar.

impl_core/src/main/java/io/opencensus/implcore/stats/MutableAggregation.java
```diff
@@ -236,10 +241,17 @@ long getCount() {
 
     private final BucketBoundaries bucketBoundaries;
     private final long[] bucketCounts;
+    private final Exemplar[] exemplars;
```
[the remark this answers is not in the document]
songy23: I added some comments here.

impl_core/src/main/java/io/opencensus/implcore/stats/MutableAggregation.java
songy23: > Is it always newer?

Correct, this has the same logic to `LastValue` (see https://github.com/census-instrumentation/opencensus-java/pull/1055/files#diff-21149fa86227f74f9e298e6813e88275R397 and https://github.com/census-instrumentation/opencensus-java/pull/1055/files#diff-43fa0dd191ba5d49938303943d995ee0R400). I added some explanation for it.

impl_core/src/test/java/io/opencensus/implcore/stats/MutableAggregationTest.java
songy23: I'm trying to avoid using Guava classes and methods - is it OK to use them in the impl?

impl_core/src/test/java/io/opencensus/implcore/stats/StatsRecorderImplTest.java
songy23: Sounds good, done.

impl_core/src/test/java/io/opencensus/implcore/stats/MutableAggregationTest.java
sebright: I think we should eventually remove the dependency on Guava from the impl, though I don't think we need to avoid using it in tests.

impl_core/src/main/java/io/opencensus/implcore/stats/MutableAggregation.java
```diff
@@ -241,6 +241,9 @@ long getCount() {
 
     private final BucketBoundaries bucketBoundaries;
     private final long[] bucketCounts;
+    // If there's a histogram (i.e bucket boundaries are not empty) in this MutableDistribution,
+    // exemplars will have the same size to bucketCounts; otherwise exemplars are empty.
```
sebright: Suggestion: It might be clearer to use null when there are no exemplars, since it has a different meaning from zero buckets.  It would also allow null analysis to enforce that the code handles this case.

songy23: Make sense, updated.

impl_core/src/test/java/io/opencensus/implcore/stats/MutableAggregationTest.java
songy23: SG, I updated to use `ImmutableList` instead, since we have been using Guava classes in tests anyway.

PR #1297, opened 2018-07-09
Contrib: Add an exemplar_util artifact to help record special exemplars.

songy23:
Since we represent the contextual information of `Exemplar` as a `Map<String, String>`, we need some kind of contract (i.e map key) for special contextual information, like trace id and span id. This PR adds two public string keys and helper methods for trace id and span id, so that other libraries can do:
```java
// in instrumentation code, like gRPC stats module
ExemplarUtils.putTraceIdAsAttachment(measureMap, traceId);
ExemplarUtils.putSpanIdAsAttachment(measureMap, spanId);
```
And exporters can easily get the corresponding trace id or span id:
```java
// in exporter code
String traceId = exemplar.getAttachments().get(ExemplarUtils.ATTACHMENT_KEY_TRACE_ID);
String spanId = exemplar.getAttachments().get(ExemplarUtils.ATTACHMENT_KEY_SPAN_ID);
```
These two keys serve a similar purpose as [ContextUtils.TAG_CONTEXT_KEY](https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/tags/unsafe/ContextUtils.java#L45).

/cc @bogdandrutu


api/src/main/java/io/opencensus/stats/AggregationData.java
```diff
@@ -417,6 +417,20 @@ public static DistributionData create(
     @AutoValue
     public abstract static class Exemplar {
 
+      /**
+       * Key for {@code TraceId} in the contextual information of an {@link Exemplar}.
```
bogdandrutu: Probably you want to use hex lower case for both values.

songy23: Sure, I added a note about the suggested encoding format on values.

songy23:
@sebright sounds like a better idea. I added a new artifact `exemplar_util` under `contrib` and moved attachment keys there. Also added methods on encoding and putting `TraceId`/`SpanId` to `MeasureMap`.

contrib/exemplar_util/src/main/java/io/opencensus/contrib/exemplar/util/ExemplarUtils.java
```diff
@@ -0,0 +1,89 @@
...
+
+  /**
+   * Encodes a {@link TraceId} to a string in base 16 lower case encoding, and puts the encoded
+   * string into the attachments of the given {@link MeasureMap}.
+   *
+   * @param measureMap the {@code MeasureMap}
+   * @param traceId the {@code TraceId} to be encoded and put.
+   * @since 0.16
+   */
+  public static void putTraceIdAsAttachment(MeasureMap measureMap, TraceId traceId) {
```
sebright: It might be more useful to only provide one method that takes a MeasureMap and a Span or SpanContext and adds both attachments.  We could extend it later to add other fields from the span without users needing to modify their code.

contrib/exemplar_util/README.md
```diff
@@ -0,0 +1,36 @@
+# OpenCensus Exemplar Util
+
+[![Build Status][travis-image]][travis-url]
+[![Windows Build Status][appveyor-image]][appveyor-url]
+[![Maven Central][maven-image]][maven-url]
+
+The *OpenCensus Exemplar Util for Java* is a collection of utilities for encoding and putting 
+special contextual information of `Exemplar`s.
```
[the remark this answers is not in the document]
songy23: SG, done.

contrib/exemplar_util/src/main/java/io/opencensus/contrib/exemplar/util/ExemplarUtils.java
```diff
@@ -0,0 +1,89 @@
...
+package io.opencensus.contrib.exemplar.util;
+
+import io.opencensus.stats.AggregationData.DistributionData.Exemplar;
+import io.opencensus.stats.MeasureMap;
+import io.opencensus.trace.SpanId;
+import io.opencensus.trace.TraceId;
+import javax.annotation.Nullable;
+
+/**
+ * Utils for encoding and putting special contextual information of {@link Exemplar}s.
```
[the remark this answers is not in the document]
songy23: Done.

contrib/exemplar_util/src/test/java/io/opencensus/contrib/exemplar/util/ExemplarUtilsTest.java
```diff
@@ -0,0 +1,119 @@
...
+import org.junit.rules.ExpectedException;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/** Unit tests for {@link ExemplarUtils}. */
+@RunWith(JUnit4.class)
+public class ExemplarUtilsTest {
+
+  private static final TraceId TRACE_ID = TraceId.generateRandomId(new Random());
+  private static final SpanId SPAN_ID = SpanId.generateRandomId(new Random());
```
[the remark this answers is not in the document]
songy23: Done.

contrib/exemplar_util/src/main/java/io/opencensus/contrib/exemplar/util/ExemplarUtils.java
songy23: I would prefer to keep them separate because there are cases when only `TraceId` is needed, without `SpanId`. From [Stackdriver documentation](https://cloud.google.com/monitoring/api/ref_v3/rest/v3/TimeSeries#exemplar):
> Trace ID: type.googleapis.com/google.devtools.cloudtrace.v1.Trace

contrib/exemplar_util/src/test/java/io/opencensus/contrib/exemplar/util/ExemplarUtilsTest.java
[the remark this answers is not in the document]
sebright: I think it would be safer to reuse one Random, to avoid producing duplicate values, even though the span ID and trace ID can be distinguished by their lengths.

contrib/exemplar_util/src/main/java/io/opencensus/contrib/exemplar/util/ExemplarUtils.java
sebright: I'm not sure I understand, though I'm not very familiar with exemplars.  I thought that the Stackdriver documentation just used trace ID as an example of an attachment.  What is the use case for only adding a trace ID?  Is that the most common use case?  I had thought that the main use case was adding information about the span, which could explain something like  the high latency of a request within a trace.

/cc @bogdandrutu

bogdandrutu: I would prefer one method that takes the SpanContext. we can do different optimizations in the future if needed (record only if sampled), also I think it is easier for users.

songy23: SG, done.

> What is the use case for only adding a trace ID? Is that the most common use case?

Stackdriver backend will perform check on the exemplar attachments, and trace id alone (with no span id) is one permitted use case, though I'm not sure how they make use of it in the backend. Anyway I'm fine with starting with SpanContext.

contrib/exemplar_util/src/test/java/io/opencensus/contrib/exemplar/util/ExemplarUtilsTest.java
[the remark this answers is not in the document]
songy23: Done.

contrib/exemplar_util/src/main/java/io/opencensus/contrib/exemplar/util/ExemplarUtils.java
```diff
@@ -50,31 +51,20 @@
...
+   * <p>{@link TraceId} and {@link SpanId} of the {@link SpanContext} will be encoded in base 16
+   * lower case encoding.
    *
    * @param measureMap the {@code MeasureMap}
-   * @param spanId the {@code SpanId} to be encoded and put.
+   * @param spanContext the {@code SpanContext} to be put as attachments.
    * @since 0.16
    */
-  public static void putSpanIdAsAttachment(MeasureMap measureMap, SpanId spanId) {
+  public static void putSpanContextAsAttachment(MeasureMap measureMap, SpanContext spanContext) {
```
sebright: Suggestion: I think `putSpanContextAsAttachments` (or `putSpanContextAttachments`) would be more accurate, since this method adds multiple fields.

songy23: SG, done.

songy23:
Thanks!

PR #1299, opened 2018-07-09
Mention that SampledSpanStore.getRegisteredSpanNamesForCollection is for testing.

sebright:
This commit fixes the last part of #977.  getRegisteredSpanNamesForCollection
was originally public and annotated with VisibleForTesting because it was meant
for testing code that uses OpenCensus.  However, VisibleForTesting is not ideal
because it adds a dependency on Guava and it does not appear in Javadocs.  This
commit instead describes the purpose of the method in the Javadocs.  It also
removes the temporary PublicForTesting annotation.


sebright:
Thanks!

PR #1300, opened 2018-07-10
Add the initial version of the TraceState class.

bogdandrutu:
Please focus on the public API, I will add tests and benchmarks before submitting. Any other feedback is welcome as well.

/cc @lmolkova @adriancole @songy23 @SergeyKanzhelev

### DIFF
--- a/api/src/main/java/io/opencensus/trace/TraceState.java
+++ b/api/src/main/java/io/opencensus/trace/TraceState.java
@@ -1,0 +1,173 @@
+/*
+ * Copyright 2018, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.trace;
+
+import io.opencensus.internal.Utils;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.Immutable;
+
+/**
+ * Carries tracing-system specific context in a list of key-value pairs. TraceState allows different
+ * vendors propagate additional information and inter-operate with their legacy Id formats.
+ *
+ * <p>Implementation is optimized for a small list of key-value pairs.
+ *
+ * <p>Key is opaque string up to 256 characters printable. It MUST begin with a lowercase letter,
+ * and can only contain lowercase letters a-z, digits 0-9, underscores _, dashes -, asterisks *, and
+ * forward slashes /.
+ *
+ * <p>Value is opaque string up to 256 characters printable ASCII RFC0020 characters (i.e., the
+ * range 0x20 to 0x7E) except comma , and =.
+ *
+ * @since 0.16
+ */
+@Immutable
+public final class TraceState {
+  private static final int KEY_MAX_SIZE = 256;
+  private static final int VALUE_MAX_SIZE = 256;
+  private static final int MAX_KEY_VALUE_PAIRS = 128;
+
+  // Immutable list of key-value pairs.
+  private final List<Entry> entries;
+
+  public static final TraceState EMPTY =
+      new TraceState(Collections.unmodifiableList(new ArrayList<Entry>()));
+
+  private TraceState(List<Entry> entries) {
+    this.entries = entries;
+  }
+
+  /**
+   * Returns the value to which the specified key is mapped, or null if this map contains no mapping
+   * for the key.
+   *
+   * @param key with which the specified value is to be associated
+   * @return the value to which the specified key is mapped, or null if this map contains no mapping
+   *     for the key.
+   */
+  @Nullable
+  public String get(String key) {
+    for (Entry entry : entries) {
+      if (entry.key.equals(key)) {
+        return entry.value;
+      }
+    }
+    return null;
+  }
+
+  /**
+   * Returns a {@link List} view of the mappings contained in this {@code TraceState}.
+   *
+   * @return a {@link List} view of the mappings contained in this {@code TraceState}.
+   */
+  public List<Entry> entryList() {
+    return entries;
+  }
+
+  /**
+   * Creates a TraceState by appending the extra entries to the {@code parent} in front of the
+   * key-value pairs list and removing duplicate entries.
+   *
+   * @param parent the parent {@code TraceState}.
+   * @param extraEntries the list of extra entries.
+   * @return a TraceState by appending the extra entries to the {@code parent}.
+   */
+  public static TraceState create(TraceState parent, List<Entry> extraEntries) {
+    Utils.checkNotNull(parent, "parent");
+    Utils.checkNotNull(extraEntries, "extraEntries");
+    if (extraEntries.size() == 0) {
+      return parent;
+    }
+
+    for (Entry entry : extraEntries) {
+      Utils.checkArgument(validateKey(entry.key), "Invalid key " + entry.key);
+      Utils.checkArgument(validateKey(entry.value), "Invalid value " + entry.value);
+    }
+
+    List<Entry> ret = new ArrayList<Entry>(extraEntries.size() + parent.entries.size());
+    ret.addAll(extraEntries);
+    for (Entry entry : parent.entries) {
+      boolean isPresent = false;
+      for (Entry entryExtra : extraEntries) {
+        if (entryExtra.key.equals(entry.key)) {
+          isPresent = true;
+          break;
+        }
+      }
+      if (!isPresent) {
+        ret.add(entry);
+      }
+    }
+
+    Utils.checkState(ret.size() <= MAX_KEY_VALUE_PAIRS, "Invalid size");
+
+    return new TraceState(Collections.unmodifiableList(ret));
+  }
+
+  @Immutable
+  public static final class Entry {
+    private final String key;
+    private final String value;
+
+    private Entry(String key, String value) {
+      this.key = key;
+      this.value = value;
+    }
+  }
+
+  // Key is opaque string up to 256 characters printable. It MUST begin with a lowercase letter, and
+  // can only contain lowercase letters a-z, digits 0-9, underscores _, dashes -, asterisks *, and
+  // forward slashes /.
+  private static boolean validateKey(String key) {
+    if (key.length() > KEY_MAX_SIZE
+        || key.isEmpty()
+        || key.charAt(0) < 'a'
+        || key.charAt(0) > 'z') {
+      return false;
+    }
+    for (int i = 0; i < key.length(); i++) {
+      char c = key.charAt(i);
+      if (!(c >= 'a' && c <= 'z')
+          && !(c >= '0' && c <= '9')
+          && c != '_'
+          && c != '-'
+          && c != '*'
+          && c != '/') {
+        return false;
+      }
+    }
+    return false;
+  }
+
+  // Value is opaque string up to 256 characters printable ASCII RFC0020 characters (i.e., the range
+  // 0x20 to 0x7E) except comma , and =.
+  private static boolean validateValue(String value) {
+    if (value.length() > VALUE_MAX_SIZE) {
+      return false;
+    }
+    for (int i = 0; i < value.length(); i++) {
+      char c = value.charAt(i);
+      if (c == ',' || c == '=' || c < ' ' /* '\u0020' */ || c > '~' /* '\u007E' */) {
+        return false;
+      }
+    }
+    return false;
+  }
+}
--- a/api/src/main/java/io/opencensus/trace/Tracestate.java
+++ b/api/src/main/java/io/opencensus/trace/Tracestate.java
@@ -92,25 +92,61 @@ public abstract class Tracestate {
    */
   public static final class Builder {
     private final Tracestate parent;
-    private List<Entry> toAdd;
+    private ArrayList<Entry> toAdd;
 
     private Builder(Tracestate parent) {
+      Utils.checkNotNull(parent, "parent");
       this.parent = parent;
     }
 
     /**
-     * Adds a key-value pair.
+     * Adds or updates the {@code Entry} that has the given {@code key} if it is present. The new
+     * {@code Entry} will always be added in the front of the list of entries.
      *
-     * @param key the key for the {@code Entry}.
-     * @param value the value for the {@code Entry}.
+     * @param key the key for the {@code Entry} to be added.
+     * @param value the value for the {@code Entry} to be added.
      * @return this.
+     * @since 0.16
      */
-    public Builder add(String key, String value) {
+    public Builder addOrUpdate(String key, String value) {
+      // Initially create the Entry to validate input.
+      Entry entry = Entry.create(key, value);
       if (toAdd == null) {
-        toAdd = new ArrayList<Entry>();
+        // Copy entries from the parent.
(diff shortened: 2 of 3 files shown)


api/src/main/java/io/opencensus/trace/TraceState.java
```diff
@@ -0,0 +1,173 @@
...
+   */
+  public static TraceState create(TraceState parent, List<Entry> extraEntries) {
+    Utils.checkNotNull(parent, "parent");
+    Utils.checkNotNull(extraEntries, "extraEntries");
+    if (extraEntries.size() == 0) {
+      return parent;
+    }
+
+    for (Entry entry : extraEntries) {
+      Utils.checkArgument(validateKey(entry.key), "Invalid key " + entry.key);
```
SergeyKanzhelev: so this is a typical practice to throw for invalid arguments? Need to be careful with this class

api/src/main/java/io/opencensus/trace/TraceState.java
```diff
@@ -0,0 +1,173 @@
...
+      Utils.checkArgument(validateKey(entry.value), "Invalid value " + entry.value);
+    }
+
+    List<Entry> ret = new ArrayList<Entry>(extraEntries.size() + parent.entries.size());
+    ret.addAll(extraEntries);
+    for (Entry entry : parent.entries) {
+      boolean isPresent = false;
+      for (Entry entryExtra : extraEntries) {
+        if (entryExtra.key.equals(entry.key)) {
```
SergeyKanzhelev: I'd suggest to treat `extraEntries` not only as an update for current ones, but also as a way to remove entries. Maybe value `null` may indicate `remove`.

api/src/main/java/io/opencensus/trace/TraceState.java
```diff
@@ -0,0 +1,173 @@
...
+ * and can only contain lowercase letters a-z, digits 0-9, underscores _, dashes -, asterisks *, and
+ * forward slashes /.
+ *
+ * <p>Value is opaque string up to 256 characters printable ASCII RFC0020 characters (i.e., the
+ * range 0x20 to 0x7E) except comma , and =.
+ *
+ * @since 0.16
+ */
+@Immutable
+public final class TraceState {
```
SergeyKanzhelev: Making it final indicates that you plan to parse the `tracestate` string in the begging of every request. There is an opportunity to save on parsing if nobody is modifying collection and the same protocol (http incoming and http outgoing) is being used

api/src/main/java/io/opencensus/trace/TraceState.java
```diff
@@ -0,0 +1,173 @@
...
+
+  /**
+   * Creates a TraceState by appending the extra entries to the {@code parent} in front of the
+   * key-value pairs list and removing duplicate entries.
+   *
+   * @param parent the parent {@code TraceState}.
+   * @param extraEntries the list of extra entries.
+   * @return a TraceState by appending the extra entries to the {@code parent}.
+   */
+  public static TraceState create(TraceState parent, List<Entry> extraEntries) {
```
SergeyKanzhelev: So if on incoming request I want to read and modify some properties (like read `app-id` of source component and set my own `app-id` for downstream services - will I re-create Tracestate with new entries and set it on `SpanContext`? Also I almost want to make it inside `tags` API as for hierarchical identifiers stored in tracestate you'd need to make an update before making the call to external service somehow.

I guess if you can add comment on the intended use in Javadoc - it will be helpful

api/src/main/java/io/opencensus/trace/TraceState.java
adriancole: silly but I used Tracestate iirc to make it clear that this is one word. ex when decapitalized "tracestate" not "traceState"

api/src/main/java/io/opencensus/trace/TraceState.java
```diff
@@ -0,0 +1,173 @@
...
+          && c != '/') {
+        return false;
+      }
+    }
+    return false;
+  }
+
+  // Value is opaque string up to 256 characters printable ASCII RFC0020 characters (i.e., the range
+  // 0x20 to 0x7E) except comma , and =.
+  private static boolean validateValue(String value) {
```
adriancole: maybe add href links to the w3c spec thing that describes this (to make sure for example the spec isn't wider than this, which could create a problem reading data if not)

api/src/main/java/io/opencensus/trace/TraceState.java
lmolkova: possible operations on the tracestate seems to be 
- get entry
- add entry
- remove entry
- update entry

And the flow is:
1. copy the parent state 
2. call vendor A, let it check tracestate and add/remove/update entriers
3. call vendor B, etc...

To allow such kind of scenarios (and make them more efficient), I think there should be explicit `copy state`,  `add entry`, `update entry` and `remove entry` methods.

The usage might look like

```
onNewSpan(SpanContext parentContext) {
  childState = parentContext.getTracestate().copy();
  childSpanContext = ...
  childSpanContext.setTraceState(childState);
  for (vendor : vendors) {
    vendor.onNewSpan(childState);
  }
}

...

class MicrosoftAppInsightsExtension : implements Vendor {
  public void onNewSpan(SpanContext spanContext) {
    if (spanContext.getTracestate().get("ms-app-id")) {
      spanContext.getTraceState.update("ms-app-id", "some new value");
    }
  }
}

```

SergeyKanzhelev: So you also suggesting to have a setter for the `tracestate` in `SpanContext`? In this case would it make sense to define `Tracestate` as interface or abstract class? 

(I like it too to make it immutable... so there is less problems with updating it from different child threads)

lmolkova: I don't have a strong opinion on how tracestate should be set. SpanContext constructor could create it once and builder could make it immutable. 

My point above was that it's better to have explicit add/update/delete API on a tracestate/tracestate builder than `create(parent, extra items)` with complicated behavior to update/remove entries.

SergeyKanzhelev: I think the intent of API is to keep it immutable. If `add/remove` methods will return the new object - will it be OK. Typically there is only few `tracestate` keys needs to be updated

lmolkova: there could be a builder pattern that allows tracestate entries to be changed, and after tracestate is built, it becomes immutable. There is no reason to keep it immutable while invoking vendor callbacks - they can modify it anyway

api/src/main/java/io/opencensus/trace/TraceState.java
bogdandrutu: Changed the static factory pattern to a Builder pattern. We can add support to remove if needed.

api/src/main/java/io/opencensus/trace/TraceState.java
bogdandrutu: Done.

api/src/main/java/io/opencensus/trace/TraceState.java
bogdandrutu: Changed to a builder pattern. This should kind of solve this issue.

api/src/main/java/io/opencensus/trace/Tracestate.java
```diff
@@ -0,0 +1,200 @@
...
+ * <p>Value is opaque string up to 256 characters printable ASCII RFC0020 characters (i.e., the
+ * range 0x20 to 0x7E) except comma , and =.
+ *
+ * @since 0.16
+ */
+@Immutable
+public final class Tracestate {
+  private static final int KEY_MAX_SIZE = 256;
+  private static final int VALUE_MAX_SIZE = 256;
+  private static final int MAX_KEY_VALUE_PAIRS = 128;
```
adriancole: if this is about w3c then the high amount of kv seems suspicious as total limit is 512 right? high count might encourage people to use this like tags and thoughts like that might steer library design towards things that act like tags vs relatively stable trace state entries. i noticed this in earlier comments which asked for design changes using a more tag oriented use case as example

api/src/main/java/io/opencensus/trace/Tracestate.java
```diff
@@ -0,0 +1,200 @@
...
+  public Builder toBuilder() {
+    return new Builder(this);
+  }
+
+  /**
+   * Returns a {@link List} view of the mappings contained in this {@code TraceState}.
+   *
+   * @return a {@link List} view of the mappings contained in this {@code TraceState}.
+   */
+  public List<Entry> entryList() {
```
adriancole: list suffix is not idiomatic

adriancole: also list allows repetition for a key. seems a bad idea for tracestate as it is defined. linkedhashmap is better.

api/src/main/java/io/opencensus/trace/Tracestate.java
bogdandrutu: Wrote https://github.com/w3c/distributed-tracing/pull/134/files to argue for a smaller limit (32).

api/src/main/java/io/opencensus/trace/Tracestate.java
```diff
@@ -0,0 +1,239 @@
...
+            isPresent = true;
+            break;
+          }
+        }
+        if (!isPresent) {
+          ret.add(entry);
+        }
+      }
+
+      return Tracestate.create(ret);
```
lmolkova: overall length validation should probably happen here

bogdandrutu: It happens in create to make sure that any caller will pass the check.

bogdandrutu:
@lmolkova the current add is actually an "update". Probably we should change the name, or do we want behavior like add if does not exist?

What is the use-case for remove?

api/src/main/java/io/opencensus/trace/Tracestate.java
lmolkova: I mean there is an overall header size limit of 512 bytes:

> The length of a combined header MUST be less than or equal to 512 bytes. If the length of a combined header is more than 512 bytes it SHOULD be ignored.

I think we should try to follow the protocol and check if it's too hard to implement or too inefficient, it's a good moment to fix the protocol.

lmolkova:
@bogdandrutu 

> the current add is actually an "update". Probably we should change the name, or do we want behavior like add if does not exist?

I see your point now. I was thinking about more explicit API, but I think with your API it is more efficient.
I do not have a particular scenario for delete, just a few hypothetical. Delete could also be done by inserting null value.

Changing name to something like 'addorUpdate' is a great idea

api/src/main/java/io/opencensus/trace/Tracestate.java
```diff
@@ -0,0 +1,239 @@
...
+    for (Entry entry : getEntries()) {
+      if (entry.getKey().equals(key)) {
+        return entry.getValue();
+      }
+    }
+    return null;
+  }
+
+  /**
+   * Returns a {@link List} view of the mappings contained in this {@code TraceState}.
```
songy23: Maybe also mention the order of Entries?

bogdandrutu:
PTAL. I cleaned up the code/comments and I added tests.

api/src/main/java/io/opencensus/trace/Tracestate.java
```diff
@@ -0,0 +1,255 @@
...
+ * forward slashes /.
+ *
+ * <p>Value is opaque string up to 256 characters printable ASCII RFC0020 characters (i.e., the
+ * range 0x20 to 0x7E) except comma , and =.
+ *
+ * @since 0.16
+ */
+@Immutable
+@AutoValue
+public abstract class Tracestate {
```
SergeyKanzhelev: so abstract to allow other protocols?

SergeyKanzhelev: Maybe needs to be mentioned in javadoc

api/src/main/java/io/opencensus/trace/Tracestate.java
```diff
@@ -0,0 +1,255 @@
...
+     */
+    public abstract String getValue();
+
+    Entry() {}
+  }
+
+  // Key is opaque string up to 256 characters printable. It MUST begin with a lowercase letter, and
+  // can only contain lowercase letters a-z, digits 0-9, underscores _, dashes -, asterisks *, and
+  // forward slashes /.
+  private static boolean validateKey(String key) {
```
SergeyKanzhelev: should this be public to allow other protocol's implementations to use this method?

api/src/main/java/io/opencensus/trace/Tracestate.java
```diff
@@ -0,0 +1,255 @@
...
+    /**
+     * Adds or updates the {@code Entry} that has the given {@code key} if it is present. The new
+     * {@code Entry} will always be added in the front of the list of entries.
+     *
+     * @param key the key for the {@code Entry} to be added.
+     * @param value the value for the {@code Entry} to be added.
+     * @return this.
+     * @since 0.16
+     */
+    public Builder addOrUpdate(String key, String value) {
```
SergeyKanzhelev: maybe the name `set` will be easier to understand? `set` implies `add` or `update`? Just a proposal, no strong feeling

api/src/main/java/io/opencensus/trace/Tracestate.java
```diff
@@ -0,0 +1,255 @@
...
+
+  // Value is opaque string up to 256 characters printable ASCII RFC0020 characters (i.e., the range
+  // 0x20 to 0x7E) except comma , and =.
+  private static boolean validateValue(String value) {
+    if (value.length() > VALUE_MAX_SIZE) {
+      return false;
+    }
+    for (int i = 0; i < value.length(); i++) {
+      char c = value.charAt(i);
+      if (c == ',' || c == '=' || c < ' ' /* '\u0020' */ || c > '~' /* '\u007E' */) {
```
reyang: I've updated the W3C spec by avoiding having trailing spaces ('\u0020') in value.

The previous version of ABNF is ambiguous, as the trailing space could be treated as OWS (optional white space) before the delimiter (e.g. "a=1 , b =2", with the previous syntax, we can either say the value of "a" is "1", and the space after "1" is an OWS, or we can say the value of "a" is "1 ", depending on whether we're using greedy match.)

value    = 0*255(chr) nblk-chr
nblk-chr = %x21-2B / %x2D-3C / %x3E-7E
chr      = %x20 / nblk-chr

api/src/main/java/io/opencensus/trace/Tracestate.java
adriancole: abstract is a nuance of auto value. it isn't intended to be subclasses otherwise.  usually you look for a hidden constructor which ensures this

api/src/main/java/io/opencensus/trace/Tracestate.java
adriancole: setEntry( or entry(

seems correct

api/src/main/java/io/opencensus/trace/Tracestate.java
adriancole: can you link us on protocol differences? seems important if they are more constrained not less as most constrained should win

api/src/main/java/io/opencensus/trace/Tracestate.java
bogdandrutu: @adriancole is always right. In the future if we really want we can relax that constraint because it is backwards compatible. For the moment just to allow auto-value to generate the class.

api/src/main/java/io/opencensus/trace/Tracestate.java
bogdandrutu: I would not make it public now. We can always change that when we see a need.

api/src/main/java/io/opencensus/trace/Tracestate.java
bogdandrutu: Applied the new rule.

api/src/main/java/io/opencensus/trace/Tracestate.java
bogdandrutu: Picked set because of remove. Happy to change to set/remove[Entry] if you think is better.

PR #1302, opened 2018-07-10
Add specification on Bazel build in README.

songy23:
Add specification on Bazel build so that it's easier for Bazel users to try out OpenCensus-Java.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ maven_jar(
     sha1 = "9a098392b287d7924660837f4eba0ce252013683",
 )
 ```
-Then targets can specify @io_opencensus_opencensus_api//jar as a dependency to depend on this jar:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -40,6 +40,23 @@ For Gradle add to your dependencies:
...
 
+For Bazel add the following lines to the WORKSPACE file:
+```
+maven_jar(
+    name = "io_opencensus_opencensus_api",
+    artifact = "io.opencensus:opencensus-api:0.15.0",
+    sha1 = "9a098392b287d7924660837f4eba0ce252013683",
+)
+```
+Then targets can specify @io_opencensus_opencensus_api//jar as a dependency to depend on this jar:
```
sebright: The dependency name might be clearer with backticks.

songy23:
Thanks!

PR #1305, opened 2018-07-11
Update and add imports to examples in readme.

sebright:
Three commits:

#### Update and add imports to stats example in readme.
Changes in this commit:
- Remove use of deprecated AggregationWindow.
- Make methods public to show that they could be called by code that isn't part of the example.
- Rename the class for consistency with the tracing example.
- Run google-java-format.

#### Remove use of deprecated AggregationWindow from QuickStart.java.


sebright:
Thanks!

PR #1306, opened 2018-07-11
Metrics: Move TimeSeriesCumulative and TimeSeriesGauge to top level. Hide TimeSeries.

songy23:
To make APIs more consistent with the [proto definition](https://github.com/census-instrumentation/opencensus-proto/blob/master/opencensus/proto/stats/metrics/metrics.proto#L101-L126).

### DIFF
--- a/metrics/src/main/java/io/opencensus/metrics/TimeSeries.java
+++ b/metrics/src/main/java/io/opencensus/metrics/TimeSeries.java
@@ -16,23 +16,12 @@
 
 package io.opencensus.metrics;
 
-import com.google.auto.value.AutoValue;
-import io.opencensus.common.ExperimentalApi;
-import io.opencensus.common.Function;
-import io.opencensus.common.Timestamp;
-import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 import javax.annotation.concurrent.Immutable;
 
-/**
- * A collection of data points that describes the time-varying values of a {@code Metric}.
- *
- * @since 0.16
- */
-@ExperimentalApi
+/** A collection of data points that describes the time-varying values of a {@code Metric}. */
 @Immutable
-public abstract class TimeSeries {
+abstract class TimeSeries {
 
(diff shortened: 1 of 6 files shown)


metrics/src/main/java/io/opencensus/metrics/TimeSeries.java
```diff
@@ -16,23 +16,12 @@
...
-/**
- * A collection of data points that describes the time-varying values of a {@code Metric}.
- *
- * @since 0.16
- */
-@ExperimentalApi
+/** A collection of data points that describes the time-varying values of a {@code Metric}. */
 @Immutable
-public abstract class TimeSeries {
+abstract class TimeSeries {
```
sebright: Why is this class needed, now that it isn't exposed?

metrics/src/test/java/io/opencensus/metrics/TimeSeriesTest.java
```diff
@@ -104,28 +100,6 @@ public void create_WithNullPoint() {
     TimeSeriesCumulative.create(Collections.<LabelValue>emptyList(), points, TIMESTAMP_1);
   }
```
sebright: It might be clearer to split this test suite now that the two TimeSeries classes are in separate files.

metrics/src/main/java/io/opencensus/metrics/TimeSeries.java
songy23: 1. To avoid duplicating code since `TimeSeriesGauge` and `TimeSeriesCumulative` still have some fields in common. 
2. If we later found a use case to put `TimeSeriesGauge` and `TimeSeriesCumulative` together (e.g in one list), we can easily expose this class.

metrics/src/test/java/io/opencensus/metrics/TimeSeriesTest.java
songy23: Done.

Issue #1307, opened 2018-07-12
OpenCensus representation of Histogram in Prometheus exporter differs from Prometheus representation


langecode:
Please answer these questions before submitting a bug report.

### What version of OpenCensus are you using?
0.15.0

### What JVM are you using (`java -version`)?
java version "9"
Java(TM) SE Runtime Environment (build 9+181)
Java HotSpot(TM) 64-Bit Server VM (build 9+181, mixed mode)

### What did you do?
Using the Prometheus exporter to export histogram measurements the representation of data is not following Prometheus normal practice. In Prometheus normally a sample would be represented in multiple buckets this is not the case in the data exported from OpenCensus Prometheus exporter.

If possible, provide a recipe for reproducing the error.

### What did you expect to see?
I would expect the OpenCensus Prometheus exporter to export histograms that can be used as normally in Prometheus expressions.

### What did you see instead?
I see OpenCensus only representing the sample in on bucket whereas Prometheus clients normally would represent in it multiple buckets.

Following code can reproduce the problem.

```
        PrometheusStatsCollector.createAndRegister();

        Measure.MeasureLong measure = Measure.MeasureLong.create("measurement", "to test", "ms");
        double[] buckets = {0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 8.0, 10.0, 13.0, 16.0, 20.0, 25.0, 30.0,
                40.0, 50.0, 65.0, 80.0, 100.0, 130.0, 160.0, 200.0, 250.0, 300.0, 400.0,
                500.0, 650.0, 800.0, 1000.0, 2000.0, 5000.0, 10000.0, 20000.0, 50000.0,
                100000.0};
        Stats.getViewManager()
                .registerView(
                        View.create(View.Name.create("opencensus/histogram"),
                                "Histogram exported by OpenCensus",
                                measure,
                                Aggregation.Distribution.create(
                                        BucketBoundaries.create(Arrays.stream(buckets).mapToObj(d -> Double.valueOf(d)).collect(Collectors.toList()))),
                                Collections.emptyList()
                        ));


        Stats.getStatsRecorder().newMeasureMap().put(measure, 120).record();

        Histogram histogram = Histogram.build("prometheus_histogram", "Histogram exported by Prometheus")
                .buckets(buckets)
                .register();
        histogram.observe(120);

        Collections.list(CollectorRegistry.defaultRegistry.metricFamilySamples()).stream()
                .forEach(f -> {
[...]

PR #1309, opened 2018-07-12
Export cumulative buckets for Prometheus

langecode:
Fixing #1307.

Prometheus uses cumulative values in the histogram buckets thus the buckets from OpenCensus `Distribution` needs to be summarized as they are exported to Prometheus. Otherwise the built-in functions in Prometheus, e.g., `histogram_quantile`, wont work as expected.


langecode:
I signed it!

exporters/stats/prometheus/src/main/java/io/opencensus/exporter/stats/prometheus/PrometheusExportUtils.java
```diff
@@ -190,6 +190,7 @@ public Void apply(DistributionData arg) {
             List<Double> boundaries = distribution.getBucketBoundaries().getBoundaries();
             List<String> labelNamesWithLe = new ArrayList<String>(labelNames);
             labelNamesWithLe.add(LABEL_NAME_BUCKET_BOUND);
+            long sum = 0;
```
songy23: nit: `cumulativeCount` may be a more descriptive name.

langecode: True, I will change that.

PR #1315, opened 2018-07-16
Tracing: Reduce default limit on Links to 32.

songy23:
Per Specs: https://github.com/census-instrumentation/opencensus-specs/pull/139.


songy23:
Thanks!

PR #1316, opened 2018-07-16
Tags: Add two new APIs defaultTagContext() and defaultBuilder().

songy23:
It's important to differentiate between empty and default `TagContext`, since they are not always the same in different implementations. Currently we're using `Tagger.empty()` and `Tagger.emptyBuilder()` to get the default `TagContext`s and that has caused some problems.

Adding two new APIs `defaultTagContext()` and `defaultBuilder()`. By default they are the same as `empty()` and `emptyBuilder()` but implementations can override them to have different behaviors.

/cc @zhangkun83

### DIFF
--- a/.vscode/launch.test.json
+++ b/.vscode/launch.test.json
@@ -1,0 +1,462 @@
+{
+    "run": {
+        "default": "",
+        "items": [
+            {
+                "name": "opencensus-impl-lite",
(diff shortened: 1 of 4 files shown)


songy23:
Chatted offline, we're considering an alternative approach.

PR #1321, opened 2018-07-18
Metrics: Add Exemplar to Distribution.

songy23:
Pretty much a copy-and-paste of https://github.com/census-instrumentation/opencensus-java/pull/1287.


songy23:
Thanks!

PR #1325, opened 2018-07-18
Update 'Add dependencies' section in README.

songy23:
Before there are two sections in README on adding OpenCensus dependencies to users' project. I think they are somewhat duplicated. In addition the first section doesn't mention the `impl` dependency, and that has caused confusion to users. Update README to keep one section on dependency instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,9 +16,10 @@ library is expected to move to [GA](#versioning) stage after v1.0.0 major releas
 Please join [gitter](https://gitter.im/census-instrumentation/Lobby) for help or feedback on this
 project.
 
-## OpenCensus Quickstart
+## OpenCensus Quickstart for Libraries
 
 Integrating OpenCensus with a new library means recording stats or traces and propagating context.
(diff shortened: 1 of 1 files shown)


sebright:
I think it's still useful to show the difference between the dependencies required for applications and libraries, because libraries shouldn't depend on an implementation.  Would it be clearer if we reversed the order of the two sections and  specified that the second one is for libraries, i.e., "Quickstart for Applications" and "Quickstart for Libraries"?

songy23:
> I think it's still useful to show the difference between the dependencies required for applications and libraries, because libraries shouldn't depend on an implementation. 

Yeah I forgot that - put back the `Dependencies for Libraries` part.

> Would it be clearer if we reversed the order of the two sections and specified that the second one is for libraries, i.e., "Quickstart for Applications" and "Quickstart for Libraries"?

Took a second thought, I think the current layout may make more sense. I added a pointer to application integration instead.

songy23:
Thanks!

PR #1329, opened 2018-07-19
Add opencensus-metrics to all/build.gradle.

sebright:
I didn't add opencensus-metrics to subprojects_javadoc, because I didn't want the Javadocs to be published before the artifact is released.


sebright:
Thanks!

PR #1335, opened 2018-07-23
Split findBugs annotations and jsr305 versions.

songy23:


bogdandrutu:
What is changed in 3.0.2 that we need this partial release to be used?

elharo:
Downstream and upstream dependencies including Guava are settling on 3.0.2. Using anything else leads to convergence issues.

PR #1336, opened 2018-07-23
Extract constant functions when used in match to reduce number of allocations.

bogdandrutu:


exporters/trace/zipkin/src/main/java/io/opencensus/exporter/trace/zipkin/ZipkinExporterHandler.java
```diff
@@ -167,36 +167,19 @@ private static long toEpochMicros(Timestamp timestamp) {
     return SECONDS.toMicros(timestamp.getSeconds()) + NANOSECONDS.toMicros(timestamp.getNanos());
   }
 
-  private static final Function<String, String> STRING_ATTRIBUTE_FUNCTION =
-      new Function<String, String>() {
+  // The return type needs to be nullable when this function is used as an argument to 'match' in
+  // attributeValueToString, because 'match' doesn't allow covariant return types.
+  private static final Function<Object, /*@Nullable*/ String> returnToString =
+      new Function<Object, /*@Nullable*/ String>() {
```
songy23: You need to import `import org.checkerframework.checker.nullness.qual.Nullable;`.

PR #1337, opened 2018-07-23
Metrics: Remove Range from Distribution.

songy23:
According to Specs: https://github.com/census-instrumentation/opencensus-proto/pull/67.


songy23:
Thanks!

PR #1343, opened 2018-07-25
Cleanup: Move static util methods to util class.

songy23:
Some internal clean up. None of these changes are user-visible.


songy23:
Thanks!

PR #1344, opened 2018-07-26
Add support for recording Gauges.

bogdandrutu:
Add a new concept of MetricRegistry that can create new types of Metrics like Gauges and keeps a reference to all registered metrics.

### DIFF
--- a/metrics/src/main/java/io/opencensus/metrics/MetricProducer.java
+++ b/metrics/src/main/java/io/opencensus/metrics/MetricProducer.java
@@ -14,14 +14,14 @@
  * limitations under the License.
  */
 
-package io.opencensus.metrics.export;
+package io.opencensus.metrics;
 
 import io.opencensus.common.ExperimentalApi;
-import io.opencensus.metrics.Metric;
+import io.opencensus.metrics.export.MetricProducerManager;
 import java.util.Collection;
 
 /**
- * A {@link Metric} producer that can be registered for exporting using {@link
+ * A {@link io.opencensus.metrics.Metric} producer that can be registered for exporting using {@link
  * MetricProducerManager}.
  *
  * <p>All implementation MUST be thread-safe.
@@ -30,9 +30,9 @@ import java.util.Collection;
 public abstract class MetricProducer {
 
   /**
-   * Returns a collection of produced {@link Metric}s to be exported.
+   * Returns a collection of produced {@link io.opencensus.metrics.Metric}s to be exported.
    *
-   * @return a collection of produced {@link Metric}s to be exported.
+   * @return a collection of produced {@link io.opencensus.metrics.Metric}s to be exported.
    */
   public abstract Collection<Metric> getMetrics();
 }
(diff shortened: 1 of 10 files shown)


metrics/src/main/java/io/opencensus/metrics/MetricProducer.java
```diff
@@ -14,14 +14,14 @@
  * limitations under the License.
  */
 
-package io.opencensus.metrics.export;
+package io.opencensus.metrics;
```
songy23: Why move `MetricProducer` to `metrics`, while `MetricProducerManager` is still under `metrics.export`?

metrics/src/main/java/io/opencensus/metrics/MetricRegistry.java
```diff
@@ -0,0 +1,117 @@
...
+import io.opencensus.common.ExperimentalApi;
+import io.opencensus.common.ToDoubleFunction;
+import io.opencensus.common.ToLongFunction;
+import io.opencensus.internal.Utils;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+
+/**
+ * Creates and manages your application's set of metrics. Exporters use the meter registry to
```
songy23: s/meter/metrics

metrics/src/main/java/io/opencensus/metrics/MetricRegistry.java
```diff
@@ -0,0 +1,117 @@
...
+   * <p>Must be called only once.
+   *
+   * @param name the name of the metric.
+   * @param description the description of the metric.
+   * @param unit the unit of the metric.
+   * @param obj the function argument.
+   * @param function the function to be called.
+   * @since 0.16 @ExperimentalApi
+   */
+  public abstract <T> void addLongGauge(
```
songy23: Not sure why we need this method, given that we already have the factory method on creating `MetricDescriptor` and `Metric`?

metrics/src/main/java/io/opencensus/metrics/MetricProducer.java
songy23: Chatted offline, this is to avoid circular dependency.

metrics/src/main/java/io/opencensus/metrics/MetricRegistry.java
songy23: Chatted offline, `MetricRegistry` is mainly to be used by application users to record Gauge  metrics conveniently.

metrics/src/main/java/io/opencensus/metrics/Metrics.java
```diff
@@ -40,6 +40,18 @@ public static ExportComponent getExportComponent() {
     return metricsComponent.getExportComponent();
   }
 
+  /**
+   * Returns the global {@link MetricRegistry}.
+   *
+   * <p>This {@code MetricRegistry} is already added to t
```
songy23: nit: this sentence is not complete.

PR #1346, opened 2018-07-27
Metrics: Move metrics to opencensus-api.

songy23:
Move `metrics` to `opencensus-api`, for two reasons:

- `metrics` depends on `opencensus-api` (`api-common`), and later `opencensus-api` (`api-stats`) needs to depend on `metrics`. If `metrics` continue to be an independent artifact, this will introduce circular dependency.
- Now that we have `MetricsComponent` and need to reflectively load the implementation, it's reasonable to separate the API and implementation of `metrics`.


songy23:
Holding off this PR for the moment, since we want to finish the data models first.

bogdandrutu:
Please write a big warning in the package-info that this package (classes from io.opencensus.metrics* are very experimental and users should be aware that we may change them in the future).

songy23:
> Please write a big warning in the package-info that this package (classes from io.opencensus.metrics* are very experimental and users should be aware that we may change them in the future).

SG, I updated `package-info.java` about the potential backwards-incompatible changes in `metrics`.

songy23:
Rebased, going to merge this PR once CI is green.

PR #1347, opened 2018-07-27
 Metrics: Implement ExportComponent, make ProducerManager abstract.

songy23:
~~Blocked by https://github.com/census-instrumentation/opencensus-java/pull/1346.~~

### DIFF
--- a/api/src/main/java/io/opencensus/metrics/export/MetricProducerManager.java
+++ b/api/src/main/java/io/opencensus/metrics/export/MetricProducerManager.java
@@ -20,7 +20,6 @@ import io.opencensus.common.ExperimentalApi;
 import io.opencensus.internal.Utils;
 import io.opencensus.metrics.MetricProducer;
 import java.util.Collections;
-import java.util.LinkedHashSet;
 import java.util.Set;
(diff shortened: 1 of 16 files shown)


api/src/main/java/io/opencensus/metrics/export/MetricProducerManager.java
```diff
@@ -79,11 +56,23 @@ public synchronized void remove(MetricProducer metricProducer) {
    *
    * @return all registered {@code MetricProducer}s that should be exported.
    */
-  public Set<MetricProducer> getAllMetricProducer() {
-    return metricProducers;
-  }
+  public abstract Set<MetricProducer> getAllMetricProducer();
+
+  static final class NoopMetricProducerManager extends MetricProducerManager {
```
bogdandrutu: this should be private. You should add a "public static MetricProducerManager newNoopMetricProducerManager() {}" to be consistent with the other APIs (like stats/trace).

songy23: Got it, updated.

PR #1348, opened 2018-07-27
Stats: Implement MetricProducer in Stats.

songy23:
Blocked by https://github.com/census-instrumentation/opencensus-java/pull/1347.


bogdandrutu:
Please resync.

songy23:
Rebased, PTAL.

bogdandrutu:
Let's not expose the MetricProducer publicly. I would propose the following approach:
* In implcore we have a StatsMetricProducer that uses the ViewManager to get all views and all ViewData and produces from them Metrics.
* Register the StatsMetricProducer to the global MetricProducerManager (during the initialization of the Stats in the implcore).
* Then we can change the Stats exporters to use metrics.

This way for the moment we do not expose any API in stats. What do you think?

songy23:
If the `MetricProducer` in Stats is intended to be only used by exporters/zpages (not in users' applications), I agree that it should be hidden from APIs.

> In implcore we have a StatsMetricProducer that uses the ViewManager to get all views and all ViewData and produces from them Metrics.

This is already done in https://github.com/census-instrumentation/opencensus-java/pull/1338. Whenever there's a record event, it will be turned into a `Point` in a `Metric` (or several `Metric`s, if the `Measure` maps to multiple `Metric`s).

> Register the StatsMetricProducer to the global MetricProducerManager (during the initialization of the Stats in the implcore).
> Then we can change the Stats exporters to use metrics.

Make sense to me.

PR #1349, opened 2018-07-27
Jaeger Exporter: Make the reporting duration longer in integration test.

songy23:


songy23:
Doesn't fix the problem. Closing.

PR #1354, opened 2018-08-01
Fix incorrect argument to Checker Framework, and fix nullness warnings.

sebright:
709d97aa321d5729988fd63b960bbece04cfba10 modified the -AskipDefs argument to the
Checker Framework (a regular expression) in a way that caused it to skip
checking all files.  This commit fixes the regular expression and the new
Checker Framework warnings.

### DIFF
--- a/buildscripts/checkstyle.xml
+++ b/buildscripts/checkstyle.xml
@@ -157,14 +157,15 @@
              <message key="ws.notPreceded"
              value="GenericWhitespace ''{0}'' is not preceded with whitespace."/>
         </module>
-        <module name="Indentation">
-            <property name="basicOffset" value="2"/>
-            <property name="braceAdjustment" value="0"/>
-            <property name="caseIndent" value="2"/>
-            <property name="throwsIndent" value="4"/>
-            <property name="lineWrappingIndentation" value="4"/>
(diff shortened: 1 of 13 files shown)


buildscripts/checkstyle.xml
```diff
@@ -157,14 +157,15 @@
...
         </module>
-        <module name="Indentation">
-            <property name="basicOffset" value="2"/>
-            <property name="braceAdjustment" value="0"/>
-            <property name="caseIndent" value="2"/>
-            <property name="throwsIndent" value="4"/>
-            <property name="lineWrappingIndentation" value="4"/>
-            <property name="arrayInitIndent" value="2"/>
-        </module>
+        <!-- <!-\- Checkstyle indentation rules conflict with google-java-format: -\-> -->
```
songy23: Do you want to remove these lines?

sebright: I thought these checks were from Checkstyle's google_checks.xml, so it would be better to comment them out to show what we changed.  It would also help us update the file with new changes from Checkstyle.

songy23: SGTM

sebright:
Thanks!

PR #1356, opened 2018-08-01
Update README of Stackdriver stats exporter.

songy23:
The Stackdriver Monitoring service tier is going away: https://cloud.google.com/monitoring/accounts/tiers.

Related to https://github.com/census-instrumentation/opencensus-go/issues/853.


songy23:
Thanks!

PR #1357, opened 2018-08-02
Add Checker Framework stub file for gRPC Context.Key.

sebright:
The stub file adds Nullable annotations to indicate that Context.Key.get can
return null when the key has no default or the key has a nullable value type.


sebright:
Thanks!

PR #1358, opened 2018-08-02
Minor improvements for Tracestate.

bogdandrutu:
Improve how the builder is created for empty tracestate.

### DIFF
--- a/api/src/main/java/io/opencensus/trace/Tracestate.java
+++ b/api/src/main/java/io/opencensus/trace/Tracestate.java
@@ -75,9 +75,19 @@ public abstract class Tracestate {
   public abstract List<Entry> getEntries();
 
   /**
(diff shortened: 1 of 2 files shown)


api/src/main/java/io/opencensus/trace/Tracestate.java
```diff
@@ -94,9 +104,11 @@ public Builder toBuilder() {
     private final Tracestate parent;
     @javax.annotation.Nullable private ArrayList<Entry> entries;
 
+    // Needs to be in this class to avoid initialization deadlock because super class depends on
+    // subclass (the auto-value generate class).
```
adriancole: heh

bogdandrutu: That's what IDEA told me as the error when I tried to do that.

PR #1359, opened 2018-08-03
Add Tracestate into SpanContext.

bogdandrutu:
One caveat in this PR: Tracestate name seems to not be consistent with the other class names. Maybe TraceState is better (just want to have others opinion).


adriancole:
FWIW I like the name as it matches the header. if we do otherwise tempts people to disregard that.

adriancole:
main concerns to consider is if we want this exposed as public api as it will be always available to user. for example it can be propagated package privately.. that is a choice especially as this isnt really a user api.

PR #1367, opened 2018-08-15
 Update all targets to at least java7 and android14

bogdandrutu:
gRPC updated to java7/android14 for io.context and all their artifacts see https://github.com/grpc/grpc-java/pull/4727

Currently we still cannot use try-with-resources (android 19) or ThreadLocalRandom (android 21).

### DIFF
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -6,7 +6,8 @@ dependencies {
 
     compileOnly libraries.auto_value
 
-    signature "org.codehaus.mojo.signature:java16:+@signature"
+    signature "org.codehaus.mojo.signature:java17:+@signature"
+    signature "net.sf.androidscents.signature:android-api-level-14:+@signature"
 }
 
 javadoc.exclude 'io/opencensus/internal/**'
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -6,8 +6,8 @@ dependencies {
 
     compileOnly libraries.auto_value
 
-    signature "org.codehaus.mojo.signature:java17:+@signature"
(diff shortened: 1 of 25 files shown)


bogdandrutu:
We would like to have the exporters work on Android as well. Preferably we should have the SD monitoring as well but currently uses "java.lang.management.RuntimeMXBean"

api/build.gradle
```diff
@@ -6,7 +6,8 @@ dependencies {
 
     compileOnly libraries.auto_value
 
-    signature "org.codehaus.mojo.signature:java16:+@signature"
+    signature "org.codehaus.mojo.signature:java17:+@signature"
+    signature "net.sf.androidscents.signature:android-api-level-14:+@signature"
```
adriancole: neat. I didn't know you can check for android!

adriancole:
ps to make justification clear this isnt just because of grpc, it is because we depend on grpc context which as a part of grpc moves the floor version to same as all of grpc.

is this summary right?

api/build.gradle
zhangkun83: gRPC uses specific version numbers `1.0` and `4.0_r4`, while you use `+` here. Any reason to be different?

bogdandrutu: fixed.

ejona86:
grpc/grpc-java#4727 didn't entirely drop our support of Java 6, since source and target remained 1.6. I just merged grpc/grpc-java#4799 which updated the source and target for the artifacts we were able to. A few tooling compatibility issues (grpc-java build errors) arose for some artifacts which we'll handle separately.

PR #1368, opened 2018-08-16
Fix stats to metrics implementation.

bogdandrutu:
In the metrics data model the "Point" does not represent individual points, it represents an individual point in time in the TimeSeries and in case of cumulative metrics this is the cumulative value from the start of the TimeSeries.


impl_core/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
```diff
@@ -119,6 +125,23 @@ private CumulativeMutableViewData(View view, Timestamp start, MetricMap metricMa
       }
     }
 
+    @Override
+    Metric toMetric(Timestamp now, StatsCollectionState state) {
+      if (state == StatsCollectionState.DISABLED) {
+        return null;
+      }
+      MetricDescriptor.Type type = metricDescriptor.getType();
+      List<TimeSeriesCumulative> timeSeriesCumulatives = new ArrayList<TimeSeriesCumulative>();
```
songy23: This list should be created based on `type`. If `type` is `GAUGE_DOUBLE` or `GAUGE_INT64`, this should be `TimeSeriesGauge`. See https://github.com/census-instrumentation/opencensus-java/pull/1368/files#diff-05a76803285c4ebb76d41d8822baacbfL117.

impl_core/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
```diff
@@ -119,6 +125,23 @@ private CumulativeMutableViewData(View view, Timestamp start, MetricMap metricMa
...
+      for (Entry<List<TagValue>, MutableAggregation> entry : tagValueAggregationMap.entrySet()) {
+        List<LabelValue> labelValues = MetricUtils.tagValuesToLabelValues(entry.getKey());
+        Point point = MetricUtils.mutableAggregationToPoint(entry.getValue(), now, type);
+        timeSeriesCumulatives.add(
+            TimeSeriesCumulative.create(labelValues, Collections.singletonList(point), start));
```
songy23: If each `TimeSeries` would always have only one `Point`, why do we need to define `Point`s as a list in the proto?

```proto
// A collection of data points that describes the time-varying values
// of a metric.
message TimeSeries {
  // Must be present for cumulative metrics. The time when the cumulative value
  // was reset to zero. The cumulative value is over the time interval
  // [start_timestamp, timestamp]. If not specified, the backend can use the
  // previous recorded value.
  google.protobuf.Timestamp start_timestamp = 1;

  // The set of label values that uniquely identify this timeseries. Applies to
  // all points. The order of label values must match that of label keys in the
  // metric descriptor.
  repeated LabelValue label_values = 2;

  // The data points of this timeseries. Point.value type MUST match the
  // MetricDescriptor.type.
  repeated Point points = 3;
}
```

impl_core/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
```diff
@@ -31,19 +31,26 @@
...
 import io.opencensus.stats.ViewData;
 import io.opencensus.tags.TagContext;
 import io.opencensus.tags.TagValue;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import javax.annotation.Nullable;
```
songy23: Please fix the import. You'll need to use full qualifier for `javax.annotation.Nullable`.

impl_core/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
bogdandrutu: Done.

impl_core/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
bogdandrutu: Multiple reasons why:
1) Same proto can be used for getting data. So in the storage you have more points.
2) If we produce a point each second but export every 10 seconds (not implemented this way right now).

impl_core/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
bogdandrutu: done.

impl_core/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
```diff
@@ -119,6 +127,35 @@ private CumulativeMutableViewData(View view, Timestamp start, MetricMap metricMa
...
+    @Override
+    Metric toMetric(Timestamp now, StatsCollectionState state) {
+      if (state == StatsCollectionState.DISABLED) {
+        return null;
+      }
+      // TODO(bdrutu): Refactor this after TimeSeriesGauge and TimeSeriesCumulative are combined.
+      Type type = metricDescriptor.getType();
+      if (type == Type.GAUGE_INT64 || type == Type.GAUGE_DOUBLE) {
+        List<TimeSeriesGauge> timeSeriesGauges = new ArrayList<TimeSeriesGauge>();
+        for (Entry<List<TagValue>, MutableAggregation> entry : tagValueAggregationMap.entrySet()) {
```
songy23: `TagValue` -> `/*@Nullable*/ TagValue`

impl_core/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
```diff
@@ -119,6 +127,35 @@ private CumulativeMutableViewData(View view, Timestamp start, MetricMap metricMa
...
+          List<LabelValue> labelValues = MetricUtils.tagValuesToLabelValues(entry.getKey());
+          Point point = MetricUtils.mutableAggregationToPoint(entry.getValue(), now, type);
+          timeSeriesGauges.add(
+              TimeSeriesGauge.create(labelValues, Collections.singletonList(point)));
+        }
+        return Metric.create(metricDescriptor, TimeSeriesGaugeList.create(timeSeriesGauges));
+      } else {
+        List<TimeSeriesCumulative> timeSeriesCumulatives = new ArrayList<TimeSeriesCumulative>();
+        for (Entry<List<TagValue>, MutableAggregation> entry : tagValueAggregationMap.entrySet()) {
```
songy23: ditto

impl_core/src/main/java/io/opencensus/implcore/stats/MeasureToViewMap.java
```diff
@@ -161,16 +154,24 @@ synchronized void record(TagContext tags, MeasureMapInternal stats, Timestamp ti
...
     }
   }
 
-  synchronized List<Metric> getMetrics() {
-    return metricMap.toMetrics();
+  synchronized List<Metric> getMetrics(Clock clock, StatsCollectionState state) {
+    List<Metric> metrics = new ArrayList<Metric>();
+    Timestamp now = clock.now();
+    for (Entry<String, MutableViewData> entry : mutableMap.entries()) {
+      Metric metric = entry.getValue().toMetric(now, state);
```
songy23: `Metric` -> `/*@Nullable*/ Metric`

impl_core/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
songy23: SGTM

PR #1369, opened 2018-08-16
checkstyle: 8.0 -> 8.12

sebright:
This commit also merges new changes to checkstyle.xml and fixes new checkstyle
warnings related to Javadocs.

________________________________________________________________________________

This upgrade seems to fix a bug in parsing Javadocs where a `@value` tag is split between two lines.


sebright:
Thanks!

PR #1371, opened 2018-08-17
Start adding log correlation for Log4j.

sebright:
This commit takes the approach of implementing Log4j's ContextDataInjector.  It
inserts three keys into each log event: "`openCensusTraceId`", "`openCensusSpanId`",
and "`openCensusTraceSampled`".  The tracing data can be accessed with
LogEvent.getContextData() or included in a layout.  See
https://logging.apache.org/log4j/2.x/log4j-core/apidocs/org/apache/logging/log4j/core/ContextDataInjector.html.

Here is an example of a pattern layout that includes all three tracing fields
and a log entry that could be produced by the layout:

Pattern layout:
"`%d{HH:mm:ss.SSS} [%t] traceId=%X{openCensusTraceId} spanId=%X{openCensusSpanId} sampled=%X{openCensusTraceSampled} %-5level %logger{36} - %msg%n`"

Log entry:
"`20:45:52.929 [main] traceId=0af7a7bef890695f1c5e85a8e7290164 spanId=d3f07c467ec2fbb2 sampled=true WARN  io.opencensus.contrib.logcorrelation.log4j.demo.OpenCensusLog4jLogCorrelationDemo - my log message`"

The ContextDataInjector implementation must be specified with a system property,
i.e.,
"`-Dlog4j2.contextDataInjector=io.opencensus.contrib.logcorrelation.log4j.OpenCensusTraceContextDataInjector`".

The system property
"`io.opencensus.contrib.logcorrelation.log4j.OpenCensusTraceContextDataInjector.spanSelection`"
controls whether tracing data is added to all spans, no spans, or only sampled spans.

_________________________________________________________________________

I got the initial idea from https://github.com/census-instrumentation/opencensus-specs/issues/130#issuecomment-405427214.

I opened a PR to ask for feedback on the high level approach, so it's not complete.  I'm not sure that implementing ContextDataInjector is the best solution.  The main issues that I encountered are:

- There can only be one ContextDataInjector implementation.  It would be better to implement a class that only inserts new context keys and could coexist with other classes that insert other keys.  I also looked at implementing PatternConverter, but PatternConverter.format probably doesn't run in the same thread that logs the message.
- It's difficult to wrap the default ContextDataInjector.  For example, the method that Log4j calls to get the default is private: https://github.com/apache/logging-log4j2/blob/99c40d3edfd8643cc542c25df59d11a212581076/log4j-core/src/main/java/org/apache/logging/log4j/core/impl/ContextDataInjectorFactory.java#L82
- The ContextDataInjector interface is complex, so it may be difficult to test all Log4j configurations that could affect it.  For example, it needs to add Log4j configuration properties to its result and handle different implementations of the ThreadContextMap.
- I'm not sure if this is an issue, but I noticed while testing that the ContextDataInjector can't be changed after initialization, because it is stored in a LogEventFactory that is stored in a static variable that is only set in a static initialization block.  It is not affected by configuration reloads.

/cc @g-easy @adriancole @rghetia


adriancole:
interesting. have considered this myself. it works well when there is
static accessible state backed by thread local or similar. first I saw in
practice was by @anuraaga in curiostack
https://github.com/curioswitch/curiostack/blob/73d82709350d0af40a0f02eefc1a9a8dc82c7889/common/server/framework/src/main/java/org/curioswitch/common/server/framework/logging/RequestLoggingContextInjector.java

haven't reviewed closely but the impl appears unsurprising.

g-easy:
Travis says this doesn't work on JDK7.

rghetia:
At a high level this looks good. Regarding support for multiple Injectors, ideally it would be nice to get that support from log4j but that may not be necessary right away.

sebright:
Thanks for the feedback!  @adriancole Thanks for the link!  I hadn't seen an example of a ContextDataInjector override before.

I'll continue with this approach and convert the test suite to Java 6.

contrib/log_correlation/log4j/src/main/java/io/opencensus/contrib/logcorrelation/log4j/OpenCensusTraceContextDataInjector.java
```diff
@@ -0,0 +1,219 @@
...
+ *       "true"} or {@code "false"})
+ * </ul>
+ *
+ * <p>The tracing data can be accessed with {@link LogEvent#getContextData} or included in a {@link
+ * Layout}. For example, the following patterns could be used to include the tracing data with a <a
+ * href="https://logging.apache.org/log4j/2.x/manual/layouts.html#Pattern_Layout">Pattern
+ * Layout</a>:
+ *
+ * <ul>
+ *   <li><code>%X{openCensusTraceId}</code>
```
adriancole: PS you can drop-in with some existing stuff by re-using "traceId" and "spanId" at the risk of clash. For example, I don't know how exactly it would work if someone wanted brave and census with same logging config where brave uses "traceId" and here "openCensusTraceId" yet both have literally the same value. That said, many libraries use different names ex sleuth uses the same config keys as B3 header names for legacy reasons (eventhough people can and do override that)

contrib/log_correlation/log4j/src/main/java/io/opencensus/contrib/logcorrelation/log4j/OpenCensusTraceContextDataInjector.java
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  // This method avoids getting the current span when the feature is disabled, for efficiency.
+  @Override
+  public StringMap rawContextData() {
+    switch (spanSelection) {
+      case NO_SPANS:
+        return getContextData();
+      case SAMPLED_SPANS:
+        SpanContext spanContext = tracer.getCurrentSpan().getContext();
```
adriancole: one thing to highlight about this approach is that it requires static initialization (which is not a big deal as much of census does, but to highlight).. This implies you can't do this unless you have a singleton which is the same instance for the entire lifecycle.

I know it isn't designed this way, but this is one reason why having a "context only" api can be handy as you can initialize that way ahead of usage by things like this, and multiple instances of a tracer can use the same context store.

adriancole: iotw imagine if this were reading directly from the grpc context key

sebright: I'm not sure I understand.  Do you mean that it would be better to get the tracing data directly from the gRPC context, i.e., `ContextUtils.CONTEXT_SPAN_KEY.get().getContext()`?  What is the benefit of using the context directly?  I thought that a statically accessed tracer would have the same issues as a statically accessed context.

contrib/log_correlation/log4j/src/main/java/io/opencensus/contrib/logcorrelation/log4j/OpenCensusTraceContextDataInjector.java
sebright: Thanks for the suggestion.  I was wondering which key to use, but I thought that it would be better to use longer names and avoid conflicts, since users will probably need to change the field names anyway.  Then they can choose which tracing library's field to remap.  For example, I needed to transform the trace ID field to `logging.googleapis.com/trace` to export to Stackdriver.  What are the main downsides of having multiple field names?  I could think of a few reasons:

- Shorter, more standard names could be clearer.
- Deduplicating tracing fields removes the possibility of inconsistent values.
- There would be fewer entries in the LogEvent, which could help performance.

Another reason that I prefer the longer names is that this is currently marked as an experimental feature, so I don't want it to interfere with other libraries.  Do you think it makes sense to release it with the "openCensus" field names and then agree on final key names in opencensus-specs (https://github.com/census-instrumentation/opencensus-specs/issues/123)?

contrib/log_correlation/log4j/src/main/java/io/opencensus/contrib/logcorrelation/log4j/OpenCensusTraceContextDataInjector.java
adriancole: I think we are caching a reference to the tracer, which means there's bootstrap ordering concerns. looking up just the key at runtime at least solves that right?

contrib/log_correlation/log4j/src/main/java/io/opencensus/contrib/logcorrelation/log4j/OpenCensusTraceContextDataInjector.java
adriancole: I'm just calling attention to this concern as it rose. I think if you have some cycles to see how someone can use multiple (like fallback this if not present that), or configure the key names, this would be helpful for those migrating or interoping even if not strictly our duty.

contrib/log_correlation/log4j/src/main/java/io/opencensus/contrib/logcorrelation/log4j/OpenCensusTraceContextDataInjector.java
adriancole: sorry I suppose I was unclear.. I mean to say feels like we have two choices:

* we formalize this is supported for any impl that uses CONTEXT_SPAN_KEY (including the default impl), then use this at runtime.

* we consider initialization racing, make a comment why it isn't a concern to cache Tracing.getTracer() when it is likely to be called very early in bootstrap. That or not cache it, or make it lazy

or leave it as is, and just refer to comments here if asked :)

contrib/log_correlation/log4j/src/main/java/io/opencensus/contrib/logcorrelation/log4j/OpenCensusTraceContextDataInjector.java
sebright: Now I understand.  I think that allowing users to specify the keys would be useful.  I would rather add it in a follow-up PR, though.

adriancole:
well I think you've heard my feedback which is my intent. I do think
the context key will leak and is leaking regardless of this change. Ex

https://github.com/census-instrumentation/opencensus-java/pull/1378/files#diff-4942c724b953fd881a49b94817e561b0R57

my 2p is it is better to have something stable even if it needs to
have a tweak (ex in the rare case there is a second census impl) vs
heavier and more fragile things. I would argue the most reliable way
to achieve what you are trying to do is to use the context directly.
However, I see your reasoning and glad you considered my pov.

contrib/log_correlation/log4j/src/main/java/io/opencensus/contrib/logcorrelation/log4j/OpenCensusTraceContextDataInjector.java
```diff
@@ -190,18 +188,23 @@ public StringMap rawContextData() {
...
           return getContextData();
         }
       case ALL_SPANS:
-        return getContextAndTracingData(tracer.getCurrentSpan().getContext());
+        return getContextAndTracingData(getCurrentSpanContext());
     }
     throw new AssertionError("Unknown spanSelection: " + spanSelection);
   }
 
+  private static SpanContext getCurrentSpanContext() {
```
adriancole: feel free to blame me if this turns into terrible idea :P

sebright: :)  I wanted to use the safer design if there is a possibility that Tracing.getTracer() could return the wrong object.  I also didn't realize that the Spring support used the gRPC context key directly, so I want to be consistent with that.

sebright:
Since it sounds like we agree on the high-level design and this is already a large PR, I'd like to merge it before fixing the remaining TODOs.  I fixed the build with Java 7 and temporarily removed the new library from the list of released artifacts.

sebright:
@adriancole @g-easy @rghetia Thanks for the review!

contrib/log_correlation/log4j/src/main/java/io/opencensus/contrib/logcorrelation/log4j/OpenCensusTraceContextDataInjector.java
sebright: I opened an issue: #1389

PR #1376, opened 2018-08-20
Change getState to be non-cacheable.

bogdandrutu:
This is a behavior change compared to previous comments, but it should be fine because:
1) If previously a ENABLE value is cached and the current value is DISABLE the stats will not be recorded anyway, only downside is the extra overhead on the library that caches the value to call record.
2) If previously a DISABLE value is cached and the library makes a decision of not recording stats based on this, the result will be that even if the current state is change to ENABLE the stats will not be recorded which is consistent with the previous behavior. The only difference will be that the call to setState from DISABLE -> ENABLE will not throw an exception anymore.

Not throwing a runtime exception (IllegalStateException) is backwards compatible.

This simplifies the logic of the current state and removes a global serialization point from the stats library.

Also this PR fixes an undefined behavior when setState is called. Previously if 2 threads call setStats the current library state and the view manager state can be inconsistent.

### DIFF
--- a/impl_core/src/main/java/io/opencensus/implcore/stats/StatsComponentImplBase.java
+++ b/impl_core/src/main/java/io/opencensus/implcore/stats/StatsComponentImplBase.java
@@ -70,7 +70,8 @@ public class StatsComponentImplBase extends StatsComponent {
   @SuppressWarnings("deprecation")
   public synchronized void setState(StatsCollectionState newState) {
     // Synchronized because needs to ensure that the call to resume/clear are happening atomically
-    // with the change of the state.
+    // with the change of the state. Use a different object to make sure that if this class needs
+    // synchronization in the future this logic does not contend with that.
     synchronized (setStateLock) {
       boolean stateChanged = currentState.set(newState);
       if (stateChanged) {
--- a/impl_core/src/main/java/io/opencensus/implcore/stats/StatsComponentImplBase.java
+++ b/impl_core/src/main/java/io/opencensus/implcore/stats/StatsComponentImplBase.java
@@ -16,7 +16,6 @@
 
 package io.opencensus.implcore.stats;
 
-import com.google.common.base.Preconditions;
 import io.opencensus.common.Clock;
 import io.opencensus.implcore.internal.EventQueue;
 import io.opencensus.metrics.MetricProducer;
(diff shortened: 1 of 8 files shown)


sebright:
We initially added get/setState in order to support changing the state at runtime (See issue #608).  After the API was released, we decided to postpone deciding whether to support changing the state at runtime, because allowing the state to be changed meant that client code (for example, gRPC) couldn't cache the value.  We also wanted to allow for the possibility that there would be a demand for setState functionality in the future.  We tried to leave both options open by deprecating the setState methods and only allowing the state to be set once, which the application could use for setting the state during its initialization.  In the future, we could decide to either un-deprecate the setState methods or remove setState, provide another way to disable stats at initialization, and recommend caching the result of getState.

/cc @zhangkun83 @dinooliva

impl_core/src/main/java/io/opencensus/implcore/stats/StatsComponentImplBase.java
```diff
@@ -62,18 +63,22 @@ public StatsRecorderImpl getStatsRecorder() {
...
-    boolean stateChanged = state.set(Preconditions.checkNotNull(newState, "newState"));
-    if (stateChanged) {
-      if (newState == StatsCollectionState.DISABLED) {
-        viewManager.clearStats();
-      } else {
-        viewManager.resumeStatsCollection();
+  public synchronized void setState(StatsCollectionState newState) {
+    // Synchronized because needs to ensure that the call to resume/clear are happening atomically
+    // with the change of the state.
+    synchronized (setStateLock) {
```
rghetia: Can you please explain a bit more why block synchronization is needed here on top of method synchronization?

bogdandrutu: Done

bogdandrutu:
@songy23 this PR intention is not to allow users to change the state at runtime, instead tries to simplify the guarantees that the API gives.

songy23:
Please rebase.

bogdandrutu:
Closing this PR for the moment, and try to move the discussion in https://github.com/census-instrumentation/opencensus-java/issues/1379. The global synchronization was removed from the getState in https://github.com/census-instrumentation/opencensus-java/pull/1377.

bogdandrutu:
@songy23 will do the rebase after we make the decision in the issue https://github.com/census-instrumentation/opencensus-java/issues/1379

PR #1377, opened 2018-08-20
Implement CurrentStatsState using atomic variables.

bogdandrutu:
Alternative implementation of the https://github.com/census-instrumentation/opencensus-java/pull/1376 that does not change the API behavior.

Also this PR fixes an undefined behavior when setState is called. Previously if 2 threads call setStats the current library state and the view manager state can be inconsistent.


bogdandrutu:
@songy23 fixed by ignoring that warning. See the reason in the comment.

PR #1386, opened 2018-08-27
Access current span using gRPC context key in OpenCensusTraceLoggingEnhancer.

sebright:
Avoiding accessing the OpenCensus tracing implementation early in initialization
is more robust.  See the discussion in
https://github.com/census-instrumentation/opencensus-java/pull/1371#discussion_r211440915
and
https://github.com/census-instrumentation/opencensus-java/pull/1371#discussion_r211441000.


sebright:
Thanks!

Issue #1387, opened 2018-08-27
No Trace output when running example with Logging Exporter


briantq:
Please answer these questions before submitting a bug report.

### What version of OpenCensus are you using?
0.15

### What JVM are you using (`java -version`)?
1.8.0_162

### What did you do?
If possible, provide a recipe for reproducing the error.

I am running the MultiSpansTracing.java example.  I am using the logging exporter.

`<dependency>
            <groupId>io.opencensus</groupId>
            <artifactId>opencensus-api</artifactId>
            <version>${opencensus.version}</version>
        </dependency>

        <dependency>
            <groupId>io.opencensus</groupId>
            <artifactId>opencensus-impl</artifactId>
            <version>${opencensus.version}</version>
        </dependency>

        <dependency>
            <groupId>io.opencensus</groupId>
            <artifactId>opencensus-exporter-trace-logging</artifactId>
            <version>${opencensus.version}</version>
        </dependency>`

### What did you expect to see?
When I ran the project, I expected to see some indication in System.out to denote entering/leaving/or general interaction with the span or annotations that were referenced in code.

### What did you see instead?
System.out was empty.

This could be my misunderstanding of OpenCensus but it seems that using the Logging exporter should write trace information into System.out.  If that is not the expected output location, please let me know.

songy23:
Hi @briantq , please note that in Java `Span`s are exported [every 5 seconds](https://github.com/census-instrumentation/opencensus-java/blob/660e8f375bb483a3eb817940b3aa8534f86da314/impl_core/src/main/java/io/opencensus/implcore/trace/export/ExportComponentImpl.java#L29), and by now there's no flushing mechanism for `Span` exporting yet. If you can wait for >5s before exiting the main thread (like what we did in the [helloworld example](https://github.com/census-instrumentation/opencensus-java/blob/660e8f375bb483a3eb817940b3aa8534f86da314/examples/src/main/java/io/opencensus/examples/helloworld/QuickStart.java#L106-L108)), I believe you would see some `Span`s printed to console.

FYI there's one issue about making the exporting interval configurable (https://github.com/census-instrumentation/opencensus-java/issues/1304), and one for adding a flushing for exporting (https://github.com/census-instrumentation/opencensus-specs/issues/126).

mayurkale22:
[...]

PR #1388, opened 2018-08-27
Examples : set sampler to always sample

mayurkale22:


mayurkale22:
#1387 @songy23 Please review, thanks @briantq for the report.

examples/src/main/java/io/opencensus/examples/Utils.java
```diff
@@ -0,0 +1,30 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.examples;
+
+public class Utils {
```
songy23: This class could be package-private.

examples/src/main/java/io/opencensus/examples/Utils.java
```diff
@@ -0,0 +1,30 @@
...
+
+  public static void sleep(int ms) {
```
songy23: Same here.

bogdandrutu:
One important caveat here:
* Usually people COPY-PASTE examples in their code and the "always sampler" can get to their production and cause large bills when they use a tracing service like Stackdriver.

We need to make sure this is not the case, so probably a big WARNING is good to be added.

examples/src/main/java/io/opencensus/examples/trace/MultiSpansScopedTracing.java
```diff
@@ -55,9 +59,27 @@ private static void doWork() {
...
    */
   public static void main(String[] args) {
+
+    // WARNING: Be careful before you set sampler value to always sample, especially in
+    // production environment. Trace data is often very large in size and is expensive to
+    // collect. This is why rather than collecting traces for every request(i.e. alwaysSample),
+    // downsampling is prefered.
+    //
+    // By default, OpenCensus provides a probabilistic sampler that will trace once in every
+    // 10,000 requests. If you prefer to use probabilistic sampler, you might not see trace data
```
bogdandrutu: s/". If you prefer to use probabilistic sampler"/", that's why if default probabilistic sampler is used here"

and everywhere where you have this comment.

PR #1390, opened 2018-08-27
Refactor propagation benchmarks and add b3format benchmarks.

bogdandrutu:


benchmarks/src/jmh/java/io/opencensus/benchmarks/trace/propagation/B3FormatImplBenchmark.java
```diff
@@ -0,0 +1,100 @@
...
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+
+/** Benchmarks for {@link io.opencensus.implcore.trace.propagation.B3Format}. */
+@State(Scope.Benchmark)
+public class B3FormatImplBenchmark {
+  @State(Scope.Thread)
+  public static class Data {
```
songy23: Is it possible to also extract this class and use for both `B3FormatImplBenchmark` and `BinaryFormatImplBenchmark`?

bogdandrutu: Has different formatter (binary vs text) and carrier (on uses bytes[] the other Map<String, String>).

I can create an util class that at least we share the code to generate a SpanContext for the benchmark (I will do that in the next PR where I am adding benchmarks for TraceContextFormat).

songy23: Yep, sounds good.

PR #1393, opened 2018-08-28
Remove guava dependency from the API artifact.

bogdandrutu:
Copied the BaseEncoding class for base16 from Guava (including tests).

/cc @saturnism

EDIT: fixes https://github.com/census-instrumentation/opencensus-java/issues/1081

### DIFF
--- a/api/src/main/java/io/opencensus/trace/internal/LowerCaseBase16Encoding.java
+++ b/api/src/main/java/io/opencensus/trace/internal/LowerCaseBase16Encoding.java
@@ -23,23 +23,28 @@ import java.util.Arrays;
 /** Internal copy of the Guava implementation of the {@code BaseEncoding.base16().lowerCase()}. */
 @Internal
 public final class LowerCaseBase16Encoding {
-  private static final LowerCaseBase16Encoding INSTANCE = new LowerCaseBase16Encoding();
   private static final String ALPHABET = "0123456789abcdef";
   private static final int ASCII_CHARACTERS = 128;
-  private final char[] encoding = new char[512];
-  private final byte[] decoding = new byte[ASCII_CHARACTERS];
+  private static final char[] ENCODING = buildEncodingArray();
+  private static final byte[] DECODING = buildDecodingArray();
 
-  private LowerCaseBase16Encoding() {
+  private static char[] buildEncodingArray() {
+    char[] encoding = new char[512];
     for (int i = 0; i < 256; ++i) {
       encoding[i] = ALPHABET.charAt(i >>> 4);
       encoding[i | 0x100] = ALPHABET.charAt(i & 0xF);
     }
+    return encoding;
+  }
 
+  private static byte[] buildDecodingArray() {
+    byte[] decoding = new byte[ASCII_CHARACTERS];
     Arrays.fill(decoding, (byte) -1);
     for (int i = 0; i < ALPHABET.length(); i++) {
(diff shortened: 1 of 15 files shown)


api/src/main/java/io/opencensus/trace/internal/LowerCaseBase16Encoding.java
```diff
@@ -0,0 +1,89 @@
...
+import io.opencensus.internal.Utils;
+import java.util.Arrays;
+
+/** Internal copy of the Guava implementation of the {@code BaseEncoding.base16().lowerCase()}. */
+@Internal
+public final class LowerCaseBase16Encoding {
+  private static final LowerCaseBase16Encoding INSTANCE = new LowerCaseBase16Encoding();
+  private static final String ALPHABET = "0123456789abcdef";
+  private static final int ASCII_CHARACTERS = 128;
+  private final char[] encoding = new char[512];
```
adriancole: clever impl.. I'm betting this is much quicker than things like.

```java
      result <<= 4;
      if (c >= '0' && c <= '9') {
        result |= c - '0';
      } else if (c >= 'a' && c <= 'f') {
        result |= c - 'a' + 10;
      } else {
        return 0;
      }
```

exporters/trace/zipkin/build.gradle
```diff
@@ -7,6 +7,7 @@ description = 'OpenCensus Trace Zipkin Exporter'
 
 dependencies {
     compile project(':opencensus-api'),
+            libraries.guava,
```
adriancole: odd that this is needed here..

adriancole: I think this should be scrubbed as the following isn't useful enough to claim the dep

```bash
$ ag google src/main/
src/main/java/io/opencensus/exporter/trace/zipkin/ZipkinTraceExporter.java
19:import static com.google.common.base.Preconditions.checkState;
21:import com.google.common.annotations.VisibleForTesting;

src/main/java/io/opencensus/exporter/trace/zipkin/ZipkinExporter.java
19:import com.google.common.annotations.VisibleForTesting;
```

adriancole: by scrub I mean throw IllegalStateException directly and just comment // VisibleForTesting

sebright:
Nice!  I think this can be removed now: https://github.com/census-instrumentation/opencensus-java/blob/884015cffa807a168e26ce36e8ae6d5ab426d8bb/buildscripts/import-control.xml#L57-L58

contrib/grpc_metrics/build.gradle
```diff
@@ -8,7 +8,8 @@ apply plugin: 'java'
 }
 
 dependencies {
-    compile project(':opencensus-api')
+    compile project(':opencensus-api'),
+            libraries.guava
```
songy23: Could you leave a TODO here to remove the Guava dependency? Since gRPC also needs to depend on this artifact.

exporters/trace/zipkin/build.gradle
bogdandrutu: I think this is covered here:
https://github.com/census-instrumentation/opencensus-java/issues/1113

contrib/grpc_metrics/build.gradle
bogdandrutu: I think this is covered here https://github.com/census-instrumentation/opencensus-java/issues/1113

PR #1394, opened 2018-08-28
Avoid doing string formatting when calling checkArgument.

bogdandrutu:
This improves performance significantly for our propagations:
Before (mean latency):
propagation.B3FormatImplBenchmark.extract  sample  163623       1993.963 ±   61.193  ns/op

After (mean latency):
propagation.B3FormatImplBenchmark.extract  sample  142707        174.384 ±    26.763  ns/op

### DIFF
--- a/api/src/main/java/io/opencensus/internal/Utils.java
+++ b/api/src/main/java/io/opencensus/internal/Utils.java
@@ -17,7 +17,6 @@
 package io.opencensus.internal;
 
 import java.util.List;
-import javax.annotation.Nullable;
 
 /*>>>
 import org.checkerframework.checker.nullness.qual.NonNull;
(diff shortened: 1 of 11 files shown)


api/src/main/java/io/opencensus/metrics/Metric.java
```diff
@@ -81,19 +81,16 @@ private static void checkTypeMatch(MetricDescriptor.Type type, List<TimeSeries>
           case GAUGE_INT64:
           case CUMULATIVE_INT64:
             Utils.checkArgument(
-                value instanceof ValueLong,
-                String.format("Type mismatch: %s, %s.", type, valueClassName));
+                value instanceof ValueLong, "Type mismatch: %s, %s.", type, valueClassName);
```
songy23: Do you want to split this change to another commit? (`metrics` is not present in the v0.15 release)

bogdandrutu: Done.

PR #1402, opened 2018-08-29
Add readme for opencensus-contrib-log-correlation-log4j.

sebright:
/cc @adriancole

### DIFF
--- a/contrib/log_correlation/log4j/README.md
+++ b/contrib/log_correlation/log4j/README.md
@@ -1,7 +1,7 @@
 # OpenCensus Log4j Log Correlation
 
-This subproject is currently experimental, so it doesn't have any stability guarantees. It will
-remain experimental until we have a specification for a log correlation feature in
+This subproject is currently experimental, so it may be redesigned or removed in the future.  It
+will remain experimental until we have a specification for a log correlation feature in
 [opencensus-specs](https://github.com/census-instrumentation/opencensus-specs/)
 (issue [#123](https://github.com/census-instrumentation/opencensus-specs/issues/123)).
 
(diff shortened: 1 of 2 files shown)


contrib/log_correlation/log4j/README.md
```diff
@@ -1,3 +1,88 @@
 # OpenCensus Log4j Log Correlation
 
-TODO(sebright): Add more documentation.
+This subproject is currently experimental, so it doesn't have any stability guarantees. It will
```
g-easy: API stability?

sebright: I'm not sure I understand.  What do you think I should add?

sebright: Do you mean that this line should only say that the project doesn't have any API stability guarantees?  I wanted the warning to be more general, since there could be larger redesigns once we discuss log correlation in opencensus-specs.  For example, if we decided to implement some of the more substantial logging features from https://github.com/census-instrumentation/opencensus-specs/issues/130, it might make sense to remove this library or combine it with other logging related libraries.

g-easy: To me "no stability guarantees" sounded like "it can crash" and I don't think that's what you meant. :)

How about something like:
```
This subproject is currently experimental, do not rely on names, APIs or the overall design being unchanging.
```

sebright: Yes, I only meant that it could change between versions.  I changed it to "This subproject is currently experimental, so it may be redesigned or removed in the future", because I wanted to emphasize that the functionality could be completely removed.

g-easy: Nice.

g-easy:
Still Approve of this change. Go ahead and merge.

sebright:
Thanks!

Issue #1403, opened 2018-08-30
Bring grpc to latest version 1.14


solatis:
Please answer these questions before submitting a bug report.

### What version of OpenCensus are you using?

0.15.1

### What JVM are you using (`java -version`)?

java8

### What did you do?
If possible, provide a recipe for reproducing the error.

Install opencensus in combination with one of the later Google Cloud Java API libraries (e.g. com.google.cloud/google-cloud-logging "1.41.0"). 

### What did you expect to see?

OpenCensus Stackdriver stats exporter library able to load.

### What did you see instead?

```WARNING: Exception thrown when creating MetricDescriptor.
java.lang.NoSuchMethodError: io.grpc.CallOptions$Key.createWithDefault(Ljava/lang/String;Ljava/lang/Object;)Lio/grpc/CallOptions$Key;
        at com.google.api.gax.grpc.CallOptionsUtil.<clinit>(CallOptionsUtil.java:44)
        at com.google.api.gax.grpc.GrpcMetadataHandlerInterceptor.interceptCall(GrpcMetadataHandlerInterceptor.java:58)
        at io.grpc.ClientInterceptors$InterceptorChannel.newCall(ClientInterceptors.java:156)
        at io.grpc.internal.ManagedChannelImpl.newCall(ManagedChannelImpl.java:736)
        at io.grpc.internal.ForwardingManagedChannel.newCall(ForwardingManagedChannel.java:63)
        at com.google.api.gax.grpc.GrpcClientCalls.newCall(GrpcClientCalls.java:75)
        at com.google.api.gax.grpc.GrpcDirectCallable.futureCall(GrpcDirectCallable.java:58)
        at com.google.api.gax.grpc.GrpcExceptionCallable.futureCall(GrpcExceptionCallable.java:64)
        at com.google.api.gax.rpc.UnaryCallable$1.futureCall(UnaryCallable.java:126)
        at com.google.api.gax.rpc.UnaryCallable.futureCall(UnaryCallable.java:87)
        at com.google.api.gax.rpc.UnaryCallable.call(UnaryCallable.java:112)
        at com.google.cloud.monitoring.v3.MetricServiceClient.createMetricDescriptor(MetricServiceClient.java:747)
        at io.opencensus.exporter.stats.stackdriver.StackdriverExporterWorker.registerView(StackdriverExporterWorker.java:137)
[...]

PR #1404, opened 2018-08-30
Update gRPC version to 1.14.

songy23:
Fixes https://github.com/census-instrumentation/opencensus-java/issues/1403.

### DIFF
--- a/contrib/agent/src/integration-test/java/io/opencensus/contrib/agent/instrumentation/ThreadInstrumentationIT.java
+++ b/contrib/agent/src/integration-test/java/io/opencensus/contrib/agent/instrumentation/ThreadInstrumentationIT.java
@@ -115,8 +115,10 @@ public class ThreadInstrumentationIT {
             // been wrapped in a different context (by automatic instrumentation of
             // Executor#execute), that context will be attached when executing the Runnable.
             Context context2 = Context.current().withValue(KEY, "wrong context");
-            // Work around findbugs warning
-            Preconditions.checkNotNull(context2.attach(), "context2");
+            // Work around findbugs warning. Context.attach() is marked as @CheckReturnValue so we
+            // need to check the return
+            // value here, otherwise findbugs will fail.
+            Preconditions.checkNotNull(context2.attach(), "context2.attach()");
             Thread thread = new Thread(command);
             thread.start();
             try {
--- a/contrib/agent/src/integration-test/java/io/opencensus/contrib/agent/instrumentation/ThreadInstrumentationIT.java
+++ b/contrib/agent/src/integration-test/java/io/opencensus/contrib/agent/instrumentation/ThreadInstrumentationIT.java
@@ -18,6 +18,7 @@ package io.opencensus.contrib.agent.instrumentation;
 
 import static com.google.common.truth.Truth.assertThat;
 
+import com.google.common.base.Preconditions;
 import io.grpc.Context;
 import java.util.concurrent.Executor;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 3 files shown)


songy23:
PTAL

contrib/agent/src/main/java/io/opencensus/contrib/agent/instrumentation/ContextStrategyImpl.java
```diff
@@ -67,7 +68,7 @@ public void attachContextForThread(Thread thread) {
       Context context = savedContexts.getIfPresent(thread);
       if (context != null) {
         savedContexts.invalidate(thread);
-        context.attach();
+        Preconditions.checkNotNull(context.attach(), "context"); // Work around findbugs warning
```
bogdandrutu: I think we should not do this just for findbugs (this is very annoying and code looks bad). Please rewrite this as if context == null return.

contrib/agent/src/integration-test/java/io/opencensus/contrib/agent/instrumentation/ThreadInstrumentationIT.java
```diff
@@ -114,7 +115,8 @@ public void execute(Runnable command) {
             // been wrapped in a different context (by automatic instrumentation of
             // Executor#execute), that context will be attached when executing the Runnable.
             Context context2 = Context.current().withValue(KEY, "wrong context");
-            context2.attach();
+            // Work around findbugs warning
+            Preconditions.checkNotNull(context2.attach(), "context2");
```
bogdandrutu: This is just wrong. You  call the attach method before checking for null.

songy23: Done. (This is not for checking `context2`, rather it's checking the return value of `context2.attach()`. )

contrib/agent/src/main/java/io/opencensus/contrib/agent/instrumentation/ContextStrategyImpl.java
songy23: Got it, done.

contrib/agent/src/integration-test/java/io/opencensus/contrib/agent/instrumentation/ThreadInstrumentationIT.java
```diff
@@ -114,7 +114,10 @@ public void execute(Runnable command) {
             // been wrapped in a different context (by automatic instrumentation of
             // Executor#execute), that context will be attached when executing the Runnable.
             Context context2 = Context.current().withValue(KEY, "wrong context");
-            context2.attach();
+            Context newContext = context2.attach();
+            if (newContext == null) { // work around findBugs.
```
bogdandrutu: newContext is never used here so don't understand why this is a fingBug warning.

contrib/agent/src/main/java/io/opencensus/contrib/agent/instrumentation/ContextStrategyImpl.java
```diff
@@ -67,7 +67,10 @@ public void attachContextForThread(Thread thread) {
       Context context = savedContexts.getIfPresent(thread);
       if (context != null) {
         savedContexts.invalidate(thread);
-        context.attach();
+        Context newContext = context.attach();
```
bogdandrutu: Same question here.

contrib/agent/src/integration-test/java/io/opencensus/contrib/agent/instrumentation/ThreadInstrumentationIT.java
songy23: `Context.attach()` is marked as `@CheckReturnValue` in gRPC v0.14, so if we don't check the return value of it, `findBugs` will fail.

Or do you think we should disable this check in findbugs-exclude.xml?

songy23: Sounds good, I filed https://github.com/grpc/grpc-java/issues/4817 to keep track of this.

songy23:
Thanks!

PR #1405, opened 2018-08-30
Log4j log correlation: Add unit tests for handling configuration properties.

sebright:


contrib/log_correlation/log4j/src/test/java/io/opencensus/contrib/logcorrelation/log4j/OpenCensusTraceContextDataInjectorTest.java
```diff
@@ -82,4 +89,64 @@ public void useDefaultValueForInvalidSpanSelection() {
...
+            "00000000000000000000000000000000",
+            "openCensusSpanId",
+            "0000000000000000",
+            "openCensusTraceSampled",
+            "false");
+  }
+
+  private static Map<String, String> toMap(StringMap stringMap) {
+    final ImmutableMap.Builder<String, String> builder = ImmutableMap.builder();
+    stringMap.forEach(
```
adriancole: I don't remember project settings, but if this project allows test source to be 1.8+ method reference builder::put is much nicer. In zipkin we do this by using retrolambda on main but keep source level as 1.8. That's only because intellij at least at some point didn't pay attention to source level for test tree.

contrib/log_correlation/log4j/src/test/java/io/opencensus/contrib/logcorrelation/log4j/OpenCensusTraceContextDataInjectorTest.java
```diff
@@ -82,4 +89,64 @@ public void useDefaultValueForInvalidSpanSelection() {
...
+
+  @Test
+  public void handleEmptyConfigurationProperties() {
+    assertContainsOnlyDefaultTracingEntries(
+        new OpenCensusTraceContextDataInjector()
+            .injectContextData(null, new SortedArrayStringMap()));
+  }
+
+  @Test
+  public void handleNullConfigurationProperties() {
```
g-easy: Why is this identical to the above function?

PR #1410, opened 2018-09-05
Added ability to specify a custom gRPC stub for the stackdriver exporter

steveniemitz:
This patch adds the ability to use a custom gRPC stub for the stackdriver trace exporter.  The default behavior is still the same, if unset, one will be created.


steveniemitz:
The biggest reason is that we have a pretty customized gRPC channel setup that we use in-process already (for bigtable specifically), adding the ability to pass in our own stub lets us use that instead of the default.  

Additionally, there's some experimentation I'd like to do with other transports, and being able to pass in a totally custom stub makes it much easier to do so.

songy23:
Sounds good to me, could you also add an entry in [CHANGELOG.md](https://github.com/census-instrumentation/opencensus-java/blob/master/CHANGELOG.md), since this is an API change?

steveniemitz:
Sure thing!

steveniemitz:
thanks all!

PR #1411, opened 2018-09-06
Rename opencensus-contrib-log-correlation-log4j artifact/package to use log4j2.

sebright:
This commit renames the artifact to opencensus-contrib-log-correlation-log4j2
and renames the package to io.opencensus.contrib.logcorrelation.log4j2.  The
name change emphasizes that the artifact only works with Log4j 2 and allows us
to add support for other versions of Log4j in the future.


sebright:
Thanks!

PR #1413, opened 2018-09-07
Improve implementation of OpenCensusTraceContextDataInjector.

sebright:
This commit improves the thread safety of OpenCensusTraceContextDataInjector by
following the thread safety requirements in the Javadocs of the overridden
methods from ContextDataInjector.  It also handles the possibility of
ThreadContext.getThreadContextMap() returning null by adding a Nullable
annotation in a Checker Framework stub file and adding null checks to the code.

________________________________________________________________________________

I'm not sure if there is a good way to test the thread safety, since it depends on the behavior of Log4j.

### DIFF
--- a/contrib/log_correlation/log4j2/src/main/java/io/opencensus/contrib/logcorrelation/log4j2/ContextDataUtils.java
+++ b/contrib/log_correlation/log4j2/src/main/java/io/opencensus/contrib/logcorrelation/log4j2/ContextDataUtils.java
@@ -123,10 +123,14 @@ final class ContextDataUtils {
 
   private static StringMap getShareableContextAndTracingData(SpanContext spanContext) {
     ReadOnlyThreadContextMap context = ThreadContext.getThreadContextMap();
-    SortedArrayStringMap stringMap =
-        context == null
-            ? new SortedArrayStringMap(ThreadContext.getImmutableContext())
-            : new SortedArrayStringMap(context.getReadOnlyContextData());
+    SortedArrayStringMap stringMap;
+    if (context == null) {
+      stringMap = new SortedArrayStringMap(ThreadContext.getImmutableContext());
+    } else {
+      StringMap contextData = context.getReadOnlyContextData();
+      stringMap = new SortedArrayStringMap(contextData.size() + 3);
+      stringMap.putAll(contextData);
+    }
     stringMap.putValue(
         OpenCensusTraceContextDataInjector.TRACE_ID_CONTEXT_KEY,
         new TraceIdToLowerBase16Formatter(spanContext.getTraceId()));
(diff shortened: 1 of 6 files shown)


contrib/log_correlation/log4j2/src/main/java/io/opencensus/contrib/logcorrelation/log4j2/OpenCensusTraceContextDataInjector.java
```diff
@@ -170,53 +177,126 @@ private static SpanSelection parseSpanSelection(String spanSelection) {
...
     return new SortedArrayStringMap(map);
   }
+
+  private static SpanContext getCurrentSpanContext() {
+    Span span = ContextUtils.CONTEXT_SPAN_KEY.get();
+    return span == null ? SpanContext.INVALID : span.getContext();
+  }
+
+  @Immutable
+  private static final class EmptyReadOnlyStringMap implements ReadOnlyStringMap {
```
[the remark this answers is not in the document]
sebright: I realized that this class was already using another internal Log4j class, so I removed it in #1421.

contrib/log_correlation/log4j2/src/main/java/io/opencensus/contrib/logcorrelation/log4j2/ContextDataUtils.java
```diff
@@ -0,0 +1,230 @@
...
+      Map<String, String> contextMap = ThreadContext.getImmutableContext();
+      return contextMap.isEmpty()
+          ? UnmodifiableReadOnlyStringMap.EMPTY
+          : new UnmodifiableReadOnlyStringMap(contextMap);
+    }
+  }
+
+  private static StringMap getShareableContextAndTracingData(SpanContext spanContext) {
+    ReadOnlyThreadContextMap context = ThreadContext.getThreadContextMap();
+    SortedArrayStringMap stringMap =
```
bogdandrutu: Small optimization here:
```java
SortedArrayStringMap stringMap;
if (context == null) {
  Map<String, String> threadContextMap = ThreadContext.getImmutableContext();
  stringMap = new SortedArrayStringMap(threadContextMap.size() + 3);
  stringMap.putAll(context);
else {
  StringMap threadContextMap = context.getReadOnlyContextData();
  stringMap = new SortedArrayStringMap(threadContextMap.size() + 3);
  stringMap.putAll(context);
}
```

This way you avoid a capacity increase in the SortedArrayStringMap.

contrib/log_correlation/log4j2/src/main/java/io/opencensus/contrib/logcorrelation/log4j2/ContextDataUtils.java
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    @Override
+    public String toString() {
+      return spanContext.getSpanId().toLowerBase16();
+    }
+  }
+
+  @Immutable
+  private static final class UnmodifiableReadOnlyStringMap implements ReadOnlyStringMap {
```
bogdandrutu: nit: I think you can use SortedArrayStringMap + freeze() to get the same behavior and not have to implement this. Motivation is that they to some special thing when copy from a SortedArrayStringMap (see constructor or putAll).

contrib/log_correlation/log4j2/src/main/java/io/opencensus/contrib/logcorrelation/log4j2/ContextDataUtils.java
sebright: I made the optimization for the else branch, but I didn't see a corresponding `putAll` method for adding the entries of a Map.

PR #1414, opened 2018-09-07
Log4j log correlation: Capitalize "opencensus" as one word in context key names.

sebright:
For example, this commit renames "openCensusTraceId" to "opencensusTraceId".


sebright:
Thanks!

PR #1415, opened 2018-09-07
Add opencensus-contrib-log-correlation-log4j2 to the list of released artifacts.

sebright:
This pull request depends on #1413 and #1414.


sebright:
I also updated the changelog.

PR #1420, opened 2018-09-07
Fix a memory reference issue caused by the fact that we always append a new value.

bogdandrutu:


contrib/spring_sleuth_v1x/src/main/java/io/opencensus/contrib/spring/sleuth/v1x/OpenCensusSleuthSpanContextHolder.java
```diff
@@ -52,10 +52,22 @@ static void setCurrentSpan(Span span) {
...
   // See close() for a better alternative in instrumetation
-  @SuppressWarnings("CheckReturnValue")
   static void removeCurrentSpan() {
-    CURRENT_SPAN.remove();
-    Context.current().withValue(ContextUtils.CONTEXT_SPAN_KEY, null).attach();
+    removeCurrentSpanInternal(null);
+  }
+
+  @SuppressWarnings("CheckReturnValue")
+  private static SpanContext removeCurrentSpanInternal(SpanContext toRestore) {
```
songy23: `toRestore` needs to be marked as `/*Nullable*/`.

bogdandrutu: Done.

PR #1425, opened 2018-09-10
Make jmh available for all subprojects.

bogdandrutu:
Main pros:
* Use the same version across all artifacts (for example agent was using jmh 1.19);
* Allow to easily create/add benchmarks in every artifact (allows access to package protected classes for example the propagation implementations);

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -119,6 +119,36 @@ such as
 `Disallowed import - edu.umd.cs.findbugs.annotations.SuppressFBWarnings. [ImportControl]`
 could mean that `import-control.xml` needs to be updated.
 
+## Benchmarks
+
+### Invoke all benchmarks on a sub-project
+
+```bash
+$ ./gradlew clean :opencensus-impl-core:jmh
+```
+
+### Invoke on a single benchmark class
+
+```bash
+./gradlew -PjmhIncludeSingleClass=BinaryFormatImplBenchmark clean :opencensus-impl-core:jmh
+```
+
+### Debug compilation errors
+When you make incompatible changes in the Benchmarks classes you may get compilation errors which
+are related to the old code not being compatible with the new code. Some of the reasons are:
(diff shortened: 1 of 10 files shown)


CONTRIBUTING.md
```diff
@@ -119,6 +119,35 @@ such as
...
+```bash
+./gradlew -PjmhIncludeSingleClass=BinaryFormatImplBenchmark clean :opencensus-impl-core:jmh
+```
+
+### Debug compilation errors
+When you make incompatible changes in the Benchmarks classes you may get compilation errors which
+are related to the old code not being compatible with the new code. Some of the reasons are:
+* Any plugin cannot delete the generated code (jmh generates code) because if the user configured
+the directory as the same as source code the plugin will delete users source code.
+* After you run jmh a gradle demon will stay alive which may cache the generated code in memory and
```
songy23: nit: s/demon/daemon

all/build.gradle
```diff
@@ -1,11 +1,5 @@
 description = "OpenCensus All"
 
-buildscript {
```
songy23: Why remove this?

CONTRIBUTING.md
```diff
@@ -119,6 +119,35 @@ such as
...
+generates the same code even if the files were changed.
```
rghetia: This line is copied from previous place but somehow it is not making sense. Should this line be as follows?

"After you run jmh, a gradle daemon will stay alive which may cache the generated code in memory and USE THE GENERATED ( instead of 'generate the same') code even if the files were changed"

CONTRIBUTING.md
bogdandrutu: Done.

all/build.gradle
bogdandrutu: Question is why was there? This is not needed because we don't load any plugin here, maybe I am wrong but I think this is useless here.

CONTRIBUTING.md
bogdandrutu: Tried to fix.

PR #1426, opened 2018-09-12
Temporarily move "metrics" package into impl_core/ for release.

sebright:
The "metrics" package isn't ready to be released yet, so this commit moves it
out of the API artifact.  The package can still be accessed by the stats
implementation in impl_core/.  This commit can be reverted once the package is
ready to be exposed.  The moved package names also contain "temporary" so that
there is no possibility of class name conflicts between different versions of
opencensus-api and opencensus-impl-core.  For example,
io.opencensus.metrics.export is renamed to
io.opencensus.temporary.metrics.export.

/cc @g-easy


api/src/main/java/io/opencensus/common/ToDoubleFunction.java
```diff
@@ -17,7 +17,7 @@
 package io.opencensus.common;
 
 /**
- * Represents a function that produces a double-valued result. See {@link
+ * Represents a function that produces a double-valued result. See {@code
```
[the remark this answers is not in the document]
sebright: I had to replace the link to the metrics class with the class name in code font because the class is no longer accessible from the API library.  I think it will be easy to just revert this commit because git handles moved files well.

impl/src/main/java/io/opencensus/impl/metrics/MetricsComponentImpl.java
```diff
@@ -18,7 +18,7 @@
 
 import io.opencensus.implcore.common.MillisClock;
 import io.opencensus.implcore.metrics.MetricsComponentImplBase;
-import io.opencensus.metrics.MetricsComponent;
+import io.opencensus.temporary.metrics.MetricsComponent;
```
bogdandrutu: I would prefer the namespace to be:
import io.opencensus.**implcore**.temporary.metrics.MetricsComponent;

sebright: Done

PR #1430, opened 2018-09-14
Update release instructions for new Github branch protection rules.

sebright:
Now protected branches can be specified with a pattern, so we can protect all
release branches with one rule. See
https://help.github.com/articles/about-protected-branches/.  There is no need to
enable branch protection when making a release.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -71,27 +71,10 @@ token](https://help.github.com/articles/creating-a-personal-access-token-for-the
     $ git checkout -b v$MAJOR.$MINOR.x master
     $ git push upstream v$MAJOR.$MINOR.x
     ```
+    The branch will be automatically protected by the GitHub branch protection rule for release
+    branches.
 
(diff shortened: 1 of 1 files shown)


RELEASING.md
```diff
@@ -71,27 +71,10 @@ token](https://help.github.com/articles/creating-a-personal-access-token-for-the
...
-
-    - Open the branch protection settings for the new branch, by following
-      [Github's instructions](https://help.github.com/articles/configuring-protected-branches/).
-    - Copy the settings from a previous branch, i.e., check
-      - `Protect this branch`
-      - `Require pull request reviews before merging`
-      - `Require status checks to pass before merging`
-      - `Include administrators`
-
-      Enable the following required status checks:
```
songy23: Will these CI checks be automatically enabled as well?

sebright: Yes, currently there is one branch protection rule for all release branches, which enables the same checks as the per-branch branch protection settings that we had before.

PR #1443, opened 2018-09-17
Common: Add an API Functions.returnToString().

songy23:


CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 ## Unreleased
 - Add `AttributeValueDouble` to `AttributeValue`.
-
+- Add an API `Functions.returnToString()`.
```
bogdandrutu: Don't think this really deserves a changelog change. But up to you.

api/src/main/java/io/opencensus/common/Functions.java
```diff
@@ -53,6 +53,14 @@ public Void apply(Object ignored) {
         }
       };
 
+  private static final Function<Object, /*@Nullable*/ String> RETURN_TO_STRING =
+      new Function<Object, /*@Nullable*/ String>() {
+        @Override
+        public String apply(Object input) {
+          return input.toString();
```
bogdandrutu: should this be:

return input == null ? null : input.toString()?

exporters/trace/zipkin/src/main/java/io/opencensus/exporter/trace/zipkin/ZipkinExporterHandler.java
```diff
@@ -172,21 +172,18 @@ private static long toEpochMicros(Timestamp timestamp) {
...
+  // TODO: Fix the Checker Framework warning.
+  @SuppressWarnings("nullness")
   private static String attributeValueToString(AttributeValue attributeValue) {
     return attributeValue.match(
         returnToString,
         returnToString,
         returnToString,
         returnToString,
-        Functions.<String>returnConstant(""));
+        Functions.</*@Nullable*/ String>returnConstant(""));
```
bogdandrutu: why did this change?

api/src/main/java/io/opencensus/common/Functions.java
songy23: Updated.

CHANGELOG.md
songy23: I'll prefer to leave it in the CHANGELOG since it's a public API. Later when people want to write new trace exporters, this API may be useful.

exporters/trace/zipkin/src/main/java/io/opencensus/exporter/trace/zipkin/ZipkinExporterHandler.java
songy23: It's not needed after suppressing the nullness warning. Removed.

Issue #1433, opened 2018-09-14
Allow arbitrary metric prefixes


mayurkale22:
Today OpenCensus only supports custom.googleapis.com/ and external.googleapis.com/ domain prefixes but Stackdriver supports more prefixes  (and it's expanding). OC should not enforce metric prefix validation (There will never be a way to statically know the set of prefixes), we should allow users to report metrics for their registered domain.

PR #1449, opened 2018-09-18
allow arbitrary metric prefixes (v0.16.x)

sebright:
(cherry picked from commit 69360e984884902e60c0ffafe4c73a18702896ad)


songy23:
> Why do you need this?

This commit fixes the metric prefix issue with Stackdriver Monitoring (https://github.com/census-instrumentation/opencensus-java/issues/1433). The bug is just released in v0.16.0 so we think it makes sense to include it in this patch release.

/cc @mayurkale22

bogdandrutu:
It is only in the exporter so I am not that concerned.

Issue #1408, opened 2018-09-04
Add OC-Agent TraceService Java Exporter


songy23:

bogdandrutu:
Closing this as all the PRs were merged.

songy23:
Need one more PR to mark this done. The metrics service exporter (https://github.com/census-instrumentation/opencensus-java/issues/1567) is done though.

songy23:
OcAgent Trace exporter is added to release artifacts in #1761. Remaining work is tracked in https://github.com/census-instrumentation/opencensus-java/issues/1797. Closing this for now.

PR #1471, opened 2018-09-25
Exporter/OCAgent: Add OcAgentNodeUtils.

songy23:

Add utilities for detecting and creating Node.
Equivalent to https://github.com/census-ecosystem/opencensus-go-exporter-ocagent/blob/master/nodeinfo.go.

### DIFF
--- a/exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentNodeUtils.java
+++ b/exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentNodeUtils.java
@@ -41,6 +41,10 @@ import javax.annotation.Nullable;
 /** Utilities for detecting and creating {@link Node}. */
 final class OcAgentNodeUtils {
 
+  // The current version of the OpenCensus OC-Agent Exporter.
+  @VisibleForTesting
+  static final String OC_AGENT_EXPORTER_VERSION = "0.17.0-SNAPSHOT"; // CURRENT_OPENCENSUS_VERSION
+
   @VisibleForTesting static final String RESOURCE_TYPE_ATTRIBUTE_KEY = "OPENCENSUS_SOURCE_TYPE";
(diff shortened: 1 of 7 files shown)


songy23:
Friendly ping @bogdandrutu

exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentNodeUtils.java
```diff
@@ -0,0 +1,181 @@
...
+  @Nullable
+  private static final MonitoredResource RESOURCE = MonitoredResourceUtils.getDefaultResource();
+
+  // Creates a Node with information from the OpenCensus library and environment variables.
+  static Node getNodeInfo(String serviceName) {
+    String jvmName = ManagementFactory.getRuntimeMXBean().getName();
+    Timestamp censusTimestamp = Timestamp.fromMillis(System.currentTimeMillis());
+    return Node.newBuilder()
+        .setIdentifier(getProcessIdentifier(jvmName, censusTimestamp))
+        .setLibraryInfo(getLibraryInfo(OpenCensusLibraryInformation.VERSION))
```
bogdandrutu: I think this exporter should have it's own version.

songy23: Are you suggesting using a different version for this exporter in the `LibraryInfo` message, or using a different version even for releasing to Maven?

All other exporters have the same version as `opencensus-api`, having a different version only for this exporter may be confusing.

bogdandrutu: If I compile the application with opencensus-api version 0.16 and opencensus-exporter-trace-ocagent version 0.17 I think this should be reflected in the proto that we send. So I suggest to have a VERSION string in the ocagent package which we update in the same way we update the one in the core library.

songy23: Make sense to me, updated.

songy23:
Thanks!

PR #1475, opened 2018-09-28
(Prototype) Add LongGauge and DoubleGauge Metrics APIs

mayurkale22:
This APIs will allow users to report already aggregated data into OpenCensus.


api/src/main/java/io/opencensus/metrics/DoubleGaugeMetric.java
```diff
@@ -0,0 +1,65 @@
...
+
+  public abstract DataPoint<T> addDataPoint(List<LabelValue> labelValues);
+
+  public abstract DataPoint<T> addDataPoint(
+      List<LabelValue> labelValues, T obj, ToDoubleFunction<T> function);
+
+  public abstract DataPoint<T> getDefaultDataPoint();
+
+  // TODO(mayurkale) : Change DataPoint to some meaningful name
+  public abstract static class DataPoint<T> {
```
bogdandrutu: Why generics?

api/src/main/java/io/opencensus/metrics/DoubleGaugeMetric.java
```diff
@@ -0,0 +1,65 @@
...
+ * limitations under the License.
+ */
+
+package io.opencensus.metrics;
+
+import io.opencensus.common.ToDoubleFunction;
+import java.util.List;
+
+// TODO(mayurkale) : Add comments and examples
+public abstract class DoubleGaugeMetric<T> implements Gauge {
```
bogdandrutu: Why generics?

api/src/main/java/io/opencensus/metrics/DoubleGaugeMetric.java
```diff
@@ -0,0 +1,65 @@
...
+
+  public abstract DataPoint<T> addDataPoint(List<LabelValue> labelValues);
+
+  public abstract DataPoint<T> addDataPoint(
```
bogdandrutu: Don't need to return a DataPoint here correct? user can do whatever they want (reset, inc, etc.) via the combination {obj, function}

api/src/main/java/io/opencensus/metrics/Gauge.java
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+
+package io.opencensus.metrics;
+
+import io.opencensus.common.Clock;
+
+// TODO(mayurkale) : Add comments
+
+public interface Gauge {
```
bogdandrutu: I would remove this completely. And make DataPoint not having a public function to getMetrics (this is an implementation detail).

api/src/main/java/io/opencensus/metrics/LongGaugeMetric.java
```diff
@@ -0,0 +1,121 @@
...
+ *      callbackQueueDataPoint.dec();
+ *   }
+ * }
+ * }</pre>
+ *
+ * @since 0.17
+ */
+
+// TODO(mayurkale) : Add comments
+public abstract class LongGaugeMetric<T> implements Gauge {
```
bogdandrutu: Same comments apply here for generics and addDataPoint with function.

api/src/main/java/io/opencensus/metrics/MetricRegistry.java
```diff
@@ -80,33 +68,27 @@ static MetricRegistry newNoopMetricRegistry() {
...
+    public <T> DoubleGaugeMetric<T> addDoubleGaugeMetric(
+        String name, String description, String unit, List<LabelKey> labelKeys) {
       Utils.checkNotNull(name, "name");
       Utils.checkNotNull(description, "description");
       Utils.checkNotNull(unit, "unit");
-      Utils.checkNotNull(labels, "labels");
-      Utils.checkNotNull(function, "function");
+      Utils.checkNotNull(labelKeys, "labelKeys");
+
+      // TODO(mayurkale) - Fix null return
```
bogdandrutu: This needs to be solved.

api/src/main/java/io/opencensus/metrics/DoubleGaugeMetric.java
mayurkale22: Missed during code refactor. 👍

api/src/main/java/io/opencensus/metrics/DoubleGaugeMetric.java
mayurkale22: Make sense

bogdandrutu:
@mayurkale22 you responded to some of the comments without actually pushing the changes. Please ping me when the changes are available for review.

api/src/main/java/io/opencensus/metrics/DoubleGaugeMetric.java
```diff
@@ -0,0 +1,142 @@
...
+     * @param val to assign to the gauge.
+     */
+    public abstract void set(double val);
+
+    /**
+     * Get the value of the gauge.
+     *
+     * @return The current gauge value.
+     */
+    public abstract double get();
```
bogdandrutu: Do we need to expose this public?

api/src/main/java/io/opencensus/metrics/DoubleGaugeMetric.java
```diff
@@ -0,0 +1,142 @@
...
+   * increase and decrease values as per your service requirements. Must be passed the same number
+   * of label values are were passed to {@link MetricRegistry#addDoubleGaugeMetric} label keys.
+   *
+   * @param labelValues the list of label values.
+   * @return DataPoint the value of single gauge
+   */
+  public abstract DataPoint addDataPoint(List<LabelValue> labelValues);
+
+  /**
+   * Add and returns new Data Point that reports the value of the object after the function. This is
```
bogdandrutu: Adds and returns. Comments should be descriptive ("Opens the file") rather than imperative ("Open the file"), please fix all the comments.

api/src/main/java/io/opencensus/metrics/LongGaugeMetric.java
```diff
@@ -0,0 +1,151 @@
...
+     * Set the gauge to the given value.
+     *
+     * @param val to assign to the gauge.
+     */
+    public abstract void set(long val);
+
+    /**
+     * Get the value of the gauge.
+     *
+     * @return The current gauge value.
```
bogdandrutu: for return statements start with lower-case -> "the"

api/src/main/java/io/opencensus/metrics/DoubleGaugeMetric.java
```diff
@@ -0,0 +1,142 @@
...
+   * , slightly more common form of gauge is one that monitors some non-numeric object. The last
+   * argument establishes the function that is used to determine the value of the gauge when the
+   * gauge is collected. Must be passed the same number of label values are were passed to {@link
+   * MetricRegistry#addDoubleGaugeMetric} label keys.
+   *
+   * @param labelValues the list of label values.
+   * @param obj The state object from which the function derives a measurement.
+   * @param function the function to be called.
+   * @param <T> The type of the object upon which the function derives a measurement.
+   */
```
songy23: nit: please add the `@since` tag to all the public methods.

api/src/main/java/io/opencensus/metrics/DoubleGaugeMetric.java
```diff
@@ -0,0 +1,156 @@
...
+ *      dataPoint.set(12.5);
+ *      // Your code here.
+ *      dataPoint.dec();
+ *   }
+ * }
+ * }</pre>
+ *
+ * @since 0.17
+ */
+public abstract class DoubleGaugeMetric {
```
songy23: Does it make more sense to have a parent abstract class `GaugeMetric`, and have `DoubleGaugeMetric` and `LongGaugeMetric` extended it? Similar to what we have in [Measure](https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/stats/Measure.java).

api/src/main/java/io/opencensus/metrics/DoubleGaugeMetric.java
```diff
@@ -0,0 +1,156 @@
...
+  /**
+   * Returns a Data Point for a gauge without labels.
+   *
+   * @return DataPoint the value of default gauge
+   * @since 0.17
+   */
+  public abstract DataPoint getDefaultDataPoint();
+
+  /** The value of a single Gauge. */
+  public abstract static class DataPoint {
```
songy23: Same here, maybe have one parent class `DataPoint` and two subclasses `DataPointDouble` and `DataPointLong`?

api/src/main/java/io/opencensus/metrics/DoubleGaugeMetric.java
mayurkale22: I think in future we have to add new API for counter as well (something like ```CounterMetric```). To hold everything together, we should have either parent abstract class or interface (could be a ```Meter```).

api/src/main/java/io/opencensus/metrics/DoubleGaugeMetric.java
mayurkale22: Done

api/src/main/java/io/opencensus/metrics/LongGaugeMetric.java
mayurkale22: Done

api/src/main/java/io/opencensus/metrics/DoubleGaugeMetric.java
bogdandrutu: Please remove it until we have a need.

api/src/main/java/io/opencensus/metrics/DoubleGaugeMetric.java
```diff
@@ -0,0 +1,156 @@
...
+  /**
+   * Adds and returns new Data Point. This is more convenient form when you can want to manually
+   * increase and decrease values as per your service requirements. Must be passed the same number
+   * of label values are were passed to {@link MetricRegistry#addDoubleGaugeMetric} label keys.
+   *
+   * @param labelValues the list of label values.
+   * @return DataPoint the value of single gauge
+   * @since 0.17
+   */
+  public abstract DataPoint addDataPoint(List<LabelValue> labelValues);
```
bogdandrutu: Can we call this Point now? We will rename the classes in the exporter with a suffix "Data" to be consistent with trace/stats (MetricData, etc).

api/src/main/java/io/opencensus/metrics/DoubleGaugeMetric.java
bogdandrutu: Please mark these classes with `@ThreadSafe` and make them to be thread safe. I think all the public API should be annotated MetricRegistry and these classes that you added.

api/src/main/java/io/opencensus/metrics/DoubleGaugeMetric.java
mayurkale22: Done

api/src/main/java/io/opencensus/metrics/DoubleGaugeMetric.java
mayurkale22: Done

api/src/main/java/io/opencensus/metrics/DoubleGaugeMetric.java
mayurkale22: Changed DataPoint -> ```Point```.

api/src/main/java/io/opencensus/metrics/DoubleGaugeMetric.java
```diff
@@ -0,0 +1,156 @@
...
+ *
+ * @since 0.17
+ */
+@ThreadSafe
+public abstract class DoubleGaugeMetric {
+
+  /**
+   * Adds and returns new Point. This is more convenient form when you can want to manually increase
+   * and decrease values as per your service requirements. Must be passed the same number of label
+   * values are were passed to {@link MetricRegistry#addDoubleGaugeMetric} label keys.
```
songy23: Should be "The number of label values must be the same to that of the label keys passed to ...".

api/src/main/java/io/opencensus/metrics/DoubleGaugeMetric.java
```diff
@@ -0,0 +1,156 @@
...
+   * @return Point the value of single gauge
+   * @since 0.17
+   */
+  public abstract Point addPoint(List<LabelValue> labelValues);
+
+  /**
+   * Adds and returns new Point that reports the value of the object after the function. This is ,
+   * slightly more common form of gauge is one that monitors some non-numeric object. The last
+   * argument establishes the function that is used to determine the value of the gauge when the
+   * gauge is collected. Must be passed the same number of label values are were passed to {@link
```
songy23: ditto

api/src/main/java/io/opencensus/metrics/DoubleGaugeMetric.java
```diff
@@ -0,0 +1,156 @@
...
+
+package io.opencensus.metrics;
+
+import io.opencensus.common.ToDoubleFunction;
+import java.util.List;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.ThreadSafe;
+
+/**
+ * Double Gauge metric to report instantaneous values. Gauges can go both up and down. For example,
```
songy23: You can also mention that gauge values can be negative.

api/src/main/java/io/opencensus/metrics/LongGaugeMetric.java
```diff
@@ -0,0 +1,164 @@
...
+ *
+ * @since 0.17
+ */
+@ThreadSafe
+public abstract class LongGaugeMetric {
+
+  /**
+   * Adds and returns new Point. This is more convenient form when you can want to manually increase
+   * and decrease values as per your service requirements. Must be passed the same number of label
+   * values are were passed to {@link MetricRegistry#addLongGaugeMetric} label keys.
```
songy23: ditto

api/src/main/java/io/opencensus/metrics/LongGaugeMetric.java
```diff
@@ -0,0 +1,164 @@
...
+   * @return Point the value of single gauge
+   * @since 0.17
+   */
+  public abstract Point addPoint(List<LabelValue> labelValues);
+
+  /**
+   * Adds and returns new Point that reports the value of the object after the function. This is a
+   * self sufficient gauge, slightly more common form of gauge is one that monitors some non-numeric
+   * object. The last argument establishes the function that is used to determine the value of the
+   * gauge when the gauge is collected. Must be passed the same number of label values are were
```
songy23: ditto

api/src/main/java/io/opencensus/metrics/MetricRegistry.java
```diff
@@ -78,33 +65,15 @@ static MetricRegistry newNoopMetricRegistry() {
...
-        T obj,
-        ToDoubleFunction<T> function) {
-      Utils.checkNotNull(name, "name");
-      Utils.checkNotNull(description, "description");
-      Utils.checkNotNull(unit, "unit");
-      Utils.checkNotNull(labels, "labels");
-      Utils.checkNotNull(function, "function");
+    public NoopDoubleGaugeMetric addDoubleGaugeMetric(
+        String name, String description, String unit, List<LabelKey> labelKeys) {
+      return new NoopDoubleGaugeMetric(name, description, unit, labelKeys);
```
songy23: Same here.

api/src/main/java/io/opencensus/metrics/MetricRegistry.java
```diff
@@ -78,33 +65,15 @@ static MetricRegistry newNoopMetricRegistry() {
...
-        T obj,
-        ToLongFunction<T> function) {
-      Utils.checkNotNull(name, "name");
-      Utils.checkNotNull(description, "description");
-      Utils.checkNotNull(unit, "unit");
-      Utils.checkNotNull(labels, "labels");
-      Utils.checkNotNull(function, "function");
+    public NoopLongGaugeMetric addLongGaugeMetric(
+        String name, String description, String unit, List<LabelKey> labelKeys) {
+      return new NoopLongGaugeMetric(name, description, unit, labelKeys);
```
songy23: Is it possible to create a static instance of `NoopLongGaugeMetric` and always return the same instance?

api/src/main/java/io/opencensus/metrics/NoopDoubleGaugeMetric.java
```diff
@@ -0,0 +1,72 @@
...
+    Utils.checkNotNull(unit, "unit");
+    Utils.checkNotNull(labelKeys, "labelKeys should not be null.");
+    Utils.checkListElementNotNull(labelKeys, "labelKeys element should not be null.");
+  }
+
+  @Override
+  public NoopPoint addPoint(List<LabelValue> labelValues) {
+    Utils.checkNotNull(labelValues, "labelValues should not be null.");
+    Utils.checkListElementNotNull(labelValues, "labelValues element should not be null.");
+    return new NoopPoint();
```
songy23: Same, can we create and reuse one static instance of `NoopPoint`?

api/src/main/java/io/opencensus/metrics/NoopLongGaugeMetric.java
```diff
@@ -0,0 +1,72 @@
...
+    Utils.checkNotNull(unit, "unit");
+    Utils.checkNotNull(labelKeys, "labelKeys should not be null.");
+    Utils.checkListElementNotNull(labelKeys, "labelKeys element should not be null.");
+  }
+
+  @Override
+  public NoopPoint addPoint(List<LabelValue> labelValues) {
+    Utils.checkNotNull(labelValues, "labelValues should not be null.");
+    Utils.checkListElementNotNull(labelValues, "labelValues element should not be null.");
+    return new NoopPoint();
```
songy23: Same.

buildscripts/import-control.xml
```diff
@@ -191,6 +191,7 @@ General guidelines on imports:
     <allow pkg="com.google.common"/>
     <allow pkg="io.opencensus.common"/>
     <allow pkg="io.opencensus.implcore"/>
+    <allow pkg="io.opencensus.internal"/>
```
songy23: I remember we want to avoid depending on `api_internal` from `implcore`. Consider using `com.google.common.base.Preconditions` instead.

api/src/main/java/io/opencensus/metrics/NoopDoubleGaugeMetric.java
```diff
@@ -68,5 +73,14 @@ public void dec(double amt) {}
 
     @Override
     public void set(double val) {}
+
+    /**
+     * Returns a {@code NoopPoint}.
+     *
+     * @return a {@code NoopPoint}.
+     */
+    public static NoopPoint getInstance() {
```
songy23: nit: This doesn't need to be public.

api/src/main/java/io/opencensus/metrics/NoopLongGaugeMetric.java
```diff
@@ -68,5 +74,14 @@ public void dec(long amt) {}
 
     @Override
     public void set(long val) {}
+
+    /**
+     * Returns a {@code NoopPoint}.
+     *
+     * @return a {@code NoopPoint}.
+     */
+    public static NoopPoint getInstance() {
```
songy23: ditto

mayurkale22:
@songy23 Thanks for the reviews.

api/src/main/java/io/opencensus/metrics/DoubleGaugeMetric.java
```diff
@@ -0,0 +1,156 @@
...
+   * @param obj the state object from which the function derives a measurement.
+   * @param function the function to be called.
+   * @param <T> the type of the object upon which the function derives a measurement.
+   * @since 0.17
+   */
+  public abstract <T> void addPoint(
+      List<LabelValue> labelValues, @Nullable T obj, ToDoubleFunction<T> function);
+
+  /**
+   * Returns a Point for a gauge without labels.
```
bogdandrutu: Is not "without labels" is with all labels not set, or default labels

api/src/main/java/io/opencensus/metrics/DoubleGaugeMetric.java
```diff
@@ -0,0 +1,156 @@
...
+     * Decrements the gauge value by one.
+     *
+     * @since 0.17
+     */
+    public abstract void dec();
+
+    /**
+     * Decrements the gauge by the given amount.
+     *
+     * @param amt Amount to subtract from the gauge.
```
bogdandrutu: start with lower case character: "amount" same everywhere where you comment params.

api/src/main/java/io/opencensus/metrics/NoopDoubleGaugeMetric.java
```diff
@@ -0,0 +1,86 @@
...
+ */
+
+package io.opencensus.metrics;
+
+import io.opencensus.common.ToDoubleFunction;
+import io.opencensus.internal.Utils;
+import java.util.List;
+
+/** No-op implementations of DoubleGaugeMetric class. */
+final class NoopDoubleGaugeMetric extends DoubleGaugeMetric {
```
bogdandrutu: This can be moved to the DoubleGaugeMetric, that way is consistent with other no-ops. Same for the Long version.

impl_core/src/main/java/io/opencensus/implcore/metrics/DoubleGaugeMetricImpl.java
```diff
@@ -0,0 +1,164 @@
...
+      List<LabelValue> labelValues, @Nullable T obj, ToDoubleFunction<T> function) {
+    checkNotNull(labelValues, "labelValues should not be null.");
+    checkArgument(labelKeys.size() == labelValues.size(), "Incorrect number of labels.");
+
+    addAndGetPoint(labelValues, obj, checkNotNull(function, "function"));
+  }
+
+  @Override
+  public Point getDefaultPoint() {
+    List<LabelValue> labelValues = new ArrayList<LabelValue>(labelKeys.size());
```
bogdandrutu: I think this should be pre-computed. Same in the Long.

impl_core/src/main/java/io/opencensus/implcore/metrics/DoubleGaugeMetricImpl.java
```diff
@@ -0,0 +1,164 @@
...
+    return addAndGetPoint(labelValues, null, null);
+  }
+
+  @Override
+  public <T> void addPoint(
+      List<LabelValue> labelValues, @Nullable T obj, ToDoubleFunction<T> function) {
+    checkNotNull(labelValues, "labelValues should not be null.");
+    checkArgument(labelKeys.size() == labelValues.size(), "Incorrect number of labels.");
+
+    addAndGetPoint(labelValues, obj, checkNotNull(function, "function"));
```
bogdandrutu: You need to make a copy of the list if you use the list as the key because the List is not immutable. Same in the Long.

impl_core/src/main/java/io/opencensus/implcore/metrics/DoubleGaugeMetricImpl.java
```diff
@@ -0,0 +1,164 @@
...
+  @Override
+  public Point getDefaultPoint() {
+    List<LabelValue> labelValues = new ArrayList<LabelValue>(labelKeys.size());
+
+    return addAndGetPoint(labelValues, null, null);
+  }
+
+  private <T> Point addAndGetPoint(
+      List<LabelValue> labelValues, @Nullable T obj, @Nullable ToDoubleFunction<T> function) {
+    Point currentPoint = pointConcurrentHashMap.get(labelValues);
```
bogdandrutu: Here you have a race-condition and you can end with multiple PointImpl objects created. If this implementation is created this is not how should be implemented. Maybe easier just to synchronize this method for the moment and always create a new map and you keep a volatile variable for the map. See the trick here https://github.com/census-instrumentation/opencensus-java/blob/master/impl_core/src/main/java/io/opencensus/implcore/metrics/export/MetricProducerManagerImpl.java#L31. With this you put more overhead into getting the point but lock free to retrieve the points.

impl_core/src/main/java/io/opencensus/implcore/metrics/DoubleGaugeMetricImpl.java
```diff
@@ -0,0 +1,164 @@
...
+
+  /** Implementation of {@link io.opencensus.metrics.DoubleGaugeMetric.Point}. */
+  public static final class PointImpl<T> extends Point {
+    private final AtomicDouble value = new AtomicDouble(0);
+    private final List<LabelValue> labelValues;
+    @Nullable private T obj;
+    @Nullable private ToDoubleFunction<T> function;
+
+    PointImpl(List<LabelValue> labelValues) {
+      this.labelValues = Collections.unmodifiableList(labelValues);
```
bogdandrutu: PointImpl is private (only trusted APIs call this) so no need to copy the labels here. Actually you should do this before you add the key in the map and use the same instance here.

impl_core/src/main/java/io/opencensus/implcore/metrics/DoubleGaugeMetricImpl.java
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  @Override
+  public Metric getMetric(Clock clock) {
+    return Metric.create(metricDescriptor, getTimeSeries(clock));
+  }
+
+  /** Implementation of {@link io.opencensus.metrics.DoubleGaugeMetric.Point}. */
+  public static final class PointImpl<T> extends Point {
+    private final AtomicDouble value = new AtomicDouble(0);
```
bogdandrutu: I would do a bit different here:
1. Add an interface in the impl_core for the moment called Meter:
```
interface Meter {
  Metric getMetric();
}
```
2. The map in the Gauge should be <List<values>, Meter>
3. PointImpl extends Point implements Meter
4. Add a class PointWithFunction implements Meter.

I think we complicate things here a bit. By saying that people can retrieve the Point, what do we do if first add with function then add without function? Maybe consider to not let people retrieve the same Point and return an error if already exists?

mayurkale22: I have already added same interface (```Meter```) but it is implemented by both LongGaugeMetric and DoubleGaugeMetric. With this it is convenient to register Meters in MetricRegistryImpl. If we go with your approach, does it mean we are creating Metric for each individual points? Let me know if I am missing something.

bogdandrutu: So my point here is that in this class you duplicate the logic of a point and a obj, function. You should have a common interface that returns a TimeSeries:

```java
interface XYZ {
  TimeSeries getTimeSeries();
}

// for AddPoint without function.
PointImpl extends Point implements XYZ {
}

// For when use with AddPoint with function
ABC implements XYZ {
}
```

bogdandrutu: Also one more point that I made was to throw an exception if people try to create a Point/TimeSeries once with a function then without a function.

Also reading more the code seems that the function should be "createTimeSeries" or "addTimeSeries" because they add/create a TimeSeries and they get a Point back which is part of the TimeSeries.

impl_core/src/main/java/io/opencensus/implcore/metrics/DoubleGaugeMetricImpl.java
```diff
@@ -0,0 +1,175 @@
...
+    return addAndGetPoint(labelValuesCopy, null, null);
+  }
+
+  @Override
+  public <T> void addPoint(
+      List<LabelValue> labelValues, @Nullable T obj, ToDoubleFunction<T> function) {
+    checkNotNull(labelValues, "labelValues should not be null.");
+
+    List<LabelValue> labelValuesCopy = new ArrayList<LabelValue>(labelValues); // Deep copy.
+    checkArgument(labelKeysSize == labelValuesCopy.size(), "Incorrect number of labels.");
```
bogdandrutu: Need also to check that non of the labelvalues are null correct?

impl_core/src/main/java/io/opencensus/implcore/metrics/LongGaugeMetricImpl.java
```diff
@@ -0,0 +1,176 @@
...
+    checkNotNull(labelValues, "labelValues should not be null.");
+
+    List<LabelValue> labelValuesCopy = new ArrayList<LabelValue>(labelValues); // Deep copy.
+    checkArgument(labelKeysSize == labelValuesCopy.size(), "Incorrect number of labels.");
+
+    addAndGetPoint(labelValuesCopy, obj, checkNotNull(function, "function"));
+  }
+
+  @Override
+  public Point getDefaultPoint() {
```
bogdandrutu: getDeaultPoint will throw an exception if called twice. seems bad.

impl_core/src/main/java/io/opencensus/implcore/metrics/LongGaugeMetricImpl.java
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    return addAndGetPoint(labelValues, null, null);
+  }
+
+  private synchronized <T> Point addAndGetPoint(
+      List<LabelValue> labelValues, @Nullable T obj, @Nullable ToLongFunction<T> function) {
+
+    Point existingPoint = points.get(labelValues);
+    if (existingPoint != null) {
```
bogdandrutu: See one of my previous comment. If called twice with the same addPoint API maybe we should not return an error, if we do return and error add is not a good prefix maybe register is better.

impl_core/src/main/java/io/opencensus/implcore/metrics/DoubleGaugeMetricImpl.java
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  @Override
+  public Metric getMetric(Clock clock) {
+    return Metric.create(metricDescriptor, getTimeSeries(clock));
+  }
+
+  /** Implementation of {@link io.opencensus.metrics.DoubleGaugeMetric.Point}. */
+  public static final class PointImpl<T> extends Point {
+    private final AtomicDouble value = new AtomicDouble(0);
```
bogdandrutu: Please add a comment about DoubleAdder and LongAdder to be used if we change to java8. Maybe we can add the improvement later by allowing multiple implementations but is good to document that.

mayurkale22: Added TODO.

impl_core/src/main/java/io/opencensus/implcore/metrics/DoubleGaugeMetricImpl.java
mayurkale22: Done.

impl_core/src/main/java/io/opencensus/implcore/metrics/DoubleGaugeMetricImpl.java
mayurkale22: Changed the function name from addPoint to addTimeSeries.

impl_core/src/main/java/io/opencensus/implcore/metrics/LongGaugeMetricImpl.java
mayurkale22: Fixed.

api/src/main/java/io/opencensus/metrics/DoubleGaugeMetric.java
```diff
@@ -0,0 +1,237 @@
...
+  /**
+   * Adds and returns a {@code Point}. This is more convenient form when you can want to manually
+   * increase and decrease values as per your service requirements. The number of label values must
+   * be the same to that of the label keys passed to {@link MetricRegistry#addDoubleGaugeMetric}.
+   *
+   * @param labelValues the list of label values.
+   * @return a {@code Point} the value of single gauge.
+   * @since 0.17
+   */
+  public abstract Point addPoint(List<LabelValue> labelValues);
```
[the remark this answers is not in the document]
mayurkale22: We have changed the function name from addPoint to addTimeSeries. 

> The caller has to keep several variables or a map to lookup points again based on label values. That adds a lot of code overhead and essentially just replicates what the metric has internally already.
Giving this add-or-get semantics seems preferable. If performance is a concern, the caller can still keep a point directly in a variable.

I agree with your concern, but for self sufficient gauge (with obj and function) we do not return Point object as users can do whatever they want. I think one option is to expose 
```Point getPoint(List<LabelValue> labels) ``` function which should work only with manually operating gauge, otherwise throw an exception.

File: api/src/main/java/io/opencensus/metrics/MetricRegistry.java [hunk 39]
```diff
@@ -59,17 +51,27 @@
...
-      T obj,
-      ToDoubleFunction<T> function);
+  @ExperimentalApi
+  public abstract DoubleGauge addDoubleGauge(
+      String name, String description, String unit, List<LabelKey> labelKeys);
+
+  /**
+   * Build a new double gauge to be added to the registry.
+   *
+   * <p>Must be called only once.
```
bogdandrutu: For a metric not in general.

api/src/main/java/io/opencensus/metrics/DoubleGauge.java
```diff
@@ -0,0 +1,236 @@
...
+  /**
+   * Removes the {@code TimeSeries} from gauge metric, if it is present.
+   *
+   * @param labelValues the list of label values.
+   * @since 0.17
+   */
+  public abstract void removeTimeSeries(List<LabelValue> labelValues);
+
+  /**
+   * Removes all {@code TimeSeries}s from gauge metric.
```
bogdandrutu: We should say that references to previous Point object are invalid (not part of the metric).

impl_core/src/main/java/io/opencensus/implcore/metrics/DerivedDoubleGaugeImpl.java
```diff
@@ -0,0 +1,137 @@
...
+      return;
+    }
+    registeredTimeSeries = Collections.unmodifiableMap(registeredTimeSeriesCopy);
+  }
+
+  private synchronized <T> void registerTimeSeries(
+      List<LabelValue> labelValues, @Nullable T obj, ToDoubleFunction<T> function) {
+    TimeSeriesProducer existingTimeSeries = registeredTimeSeries.get(labelValues);
+    if (existingTimeSeries != null) {
+      return;
```
bogdandrutu: Better to throw an exception or override the previous value. Just completely ignoring this will not signal the user.

impl_core/src/main/java/io/opencensus/implcore/metrics/DerivedDoubleGaugeImpl.java
```diff
@@ -0,0 +1,137 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+/** Implementation of {@link DerivedDoubleGauge}. */
+public final class DerivedDoubleGaugeImpl extends DerivedDoubleGauge implements Meter {
+  private final MetricDescriptor metricDescriptor;
+  private volatile Map<List<LabelValue>, TimeSeriesProducer> registeredTimeSeries =
+      Maps.newHashMap();
```
bogdandrutu: HashMap is not good for iteration in general. better to use LinkedHashMap here. Please benchmark this.

impl_core/src/main/java/io/opencensus/implcore/metrics/DoubleGaugeImpl.java
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    @Override
+    public void dec() {
+      dec(1);
+    }
+
+    @Override
+    public void dec(double amt) {
+      value.addAndGet(-amt);
```
bogdandrutu: This is hard to use and confusing in my opinion. What about we have inc, dec and add (which accepts negative values)? This will be more consistent with the AtomicDouble/AtomicLong APIs.

impl_core/src/main/java/io/opencensus/implcore/metrics/DoubleGaugeImpl.java
```diff
@@ -0,0 +1,165 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/** Implementation of {@link DoubleGauge}. */
+public final class DoubleGaugeImpl extends DoubleGauge implements Meter {
+  private final MetricDescriptor metricDescriptor;
+  private volatile Map<List<LabelValue>, TimeSeriesProducer> registeredTimeSeries =
```
bogdandrutu: You can have PointImpl here now that only PointImpl are added.

impl_core/src/main/java/io/opencensus/implcore/metrics/DoubleGaugeImpl.java
```diff
@@ -0,0 +1,165 @@
...
+
+  DoubleGaugeImpl(String name, String description, String unit, List<LabelKey> labelKeys) {
+    labelKeysSize = labelKeys.size();
+    defaultLabelValues = new ArrayList<LabelValue>(labelKeysSize);
+    this.metricDescriptor =
+        MetricDescriptor.create(name, description, unit, Type.GAUGE_DOUBLE, labelKeys);
+  }
+
+  @Override
+  public Point getOrCreateTimeSeries(List<LabelValue> labelValues) {
```
bogdandrutu: You can try to get the value initially and if present return it.

impl_core/src/main/java/io/opencensus/implcore/metrics/DoubleGaugeImpl.java
```diff
@@ -0,0 +1,165 @@
...
+
+    TimeSeriesProducer newTimeSeries = new PointImpl(labelValues);
+    // Updating the map of time series happens under a lock to avoid multiple add operations
+    // to happen in the same time.
+    Map<List<LabelValue>, TimeSeriesProducer> registeredTimeSeriesCopy =
+        new HashMap<List<LabelValue>, TimeSeriesProducer>(registeredTimeSeries);
+    registeredTimeSeriesCopy.put(labelValues, newTimeSeries);
+    registeredTimeSeries = Collections.unmodifiableMap(registeredTimeSeriesCopy);
+
+    return (PointImpl) newTimeSeries;
```
bogdandrutu: See my comment about avoiding the cast.

impl_core/src/main/java/io/opencensus/implcore/metrics/Meter.java
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.implcore.metrics;
+
+import io.opencensus.common.Clock;
+import io.opencensus.metrics.export.Metric;
+
+public interface Meter {
```
bogdandrutu: Please not public.

impl_core/src/main/java/io/opencensus/implcore/metrics/TimeSeriesProducer.java
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+
+package io.opencensus.implcore.metrics;
+
+import io.opencensus.common.Clock;
+import io.opencensus.metrics.export.TimeSeries;
+
+/** This is an interface that must be implemented by {@code Point}, that returns a TimeSeries. */
+public interface TimeSeriesProducer {
```
bogdandrutu: Please not public.

impl_core/src/main/java/io/opencensus/implcore/metrics/DoubleGaugeImpl.java
mayurkale22: Added lock free retrieval logic.

impl_core/src/main/java/io/opencensus/implcore/metrics/DerivedDoubleGaugeImpl.java
mayurkale22: Throwing an exception if already present.

impl_core/src/main/java/io/opencensus/implcore/metrics/DoubleGaugeImpl.java
mayurkale22: /cc @fabxc any thoughts? I think inc(), inc(amt), dec(), dec(amt), set(amt) seems clear for me.

impl_core/src/main/java/io/opencensus/implcore/metrics/DoubleGaugeImpl.java
```diff
@@ -0,0 +1,160 @@
...
+    @Override
+    public void set(double val) {
+      value.set(val);
+    }
+
+    @Override
+    public TimeSeries getTimeSeries(Clock clock) {
+      return TimeSeries.create(
+          labelValues,
+          Collections.singletonList(
```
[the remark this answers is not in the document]
bogdandrutu: Please do this in a separate PR.

impl_core/src/main/java/io/opencensus/implcore/metrics/DoubleGaugeImpl.java
```diff
@@ -58,11 +58,21 @@ public Point getOrCreateTimeSeries(List<LabelValue> labelValues) {
     checkArgument(labelKeysSize == labelValues.size(), "Incorrect number of labels.");
     Utils.checkListElementNotNull(labelValues, "labelValues element should not be null.");
 
+    // lock free point retrieval, if it is present
```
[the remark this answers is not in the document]
mayurkale22: Done

api/src/main/java/io/opencensus/metrics/DerivedDoubleGauge.java
```diff
@@ -26,17 +26,42 @@
...
-   * establishes the function that is used to determine the value of the gauge when the gauge is
-   * collected. The number of label values must be the same to that of the label keys passed to
-   * {@link MetricRegistry#addDerivedDoubleGauge}.
+   * Creates a TimeSeries, that reports the value of the object after the function. This is,
+   * slightly more common form of gauge is one that monitors some non-numeric object. The last
+   * argument establishes the function that is used to determine the value of the gauge when the
+   * gauge is collected. The number of label values must be the same to that of the label keys
+   * passed to {@link MetricRegistry#addDerivedDoubleGauge}.
    *
    * @param labelValues the list of label values.
```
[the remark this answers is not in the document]
mayurkale22: Done

api/src/main/java/io/opencensus/metrics/DoubleGauge.java
```diff
@@ -0,0 +1,212 @@
...
+ * <pre>{@code
+ * class YourClass {
+ *
+ *   private static final MetricRegistry metricRegistry = Metrics.getMetricRegistry();
+ *   List<LabelKey> labelKeys = Arrays.asList(LabelKey.create("Type", "desc"));
+ *   List<LabelValue> getLabelValues = Arrays.asList(LabelValue.create("GET"));
+ *   List<LabelValue> postLabelValues = Arrays.asList(LabelValue.create("POST"));
+ *
+ *   DoubleGauge httpDuration = metricRegistry.addDoubleGauge(
+ *       "http_request_duration", "Duration of the last http request", "second", labelKeys);
```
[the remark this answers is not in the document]
mayurkale22: Changed to "queue_size" example.

api/src/main/java/io/opencensus/metrics/DoubleGauge.java
```diff
@@ -0,0 +1,212 @@
...
+public abstract class DoubleGauge {
+
+  /**
+   * Creates a {@code TimeSeries} and returns a {@code Point}, which is part of the TimeSeries. This
+   * is more convenient form when you can want to manually increase and decrease values as per your
+   * service requirements. The number of label values must be the same to that of the label keys
+   * passed to {@link MetricRegistry#addDoubleGauge}.
+   *
+   * @param labelValues the list of label values.
+   * @return a {@code Point} the value of single gauge.
```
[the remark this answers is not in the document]
mayurkale22: Done

api/src/main/java/io/opencensus/metrics/DerivedDoubleGauge.java
```diff
@@ -0,0 +1,142 @@
...
+   * Removes the {@code TimeSeries} from the gauge metric, if it is present. i.e. references to
+   * previous {@code Point} objects are invalid (not part of the metric).
+   *
+   * @param labelValues the list of label values.
+   * @since 0.17
+   */
+  public abstract void removeTimeSeries(List<LabelValue> labelValues);
+
+  /**
+   * References to all previous {@code Point} objects are invalid (not part of the metric).
```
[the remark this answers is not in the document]
mayurkale22: Removed.

api/src/main/java/io/opencensus/metrics/DerivedDoubleGauge.java
```diff
@@ -0,0 +1,142 @@
...
+   * @param function the function to be called.
+   * @param <T> the type of the object upon which the function derives a measurement.
+   * @since 0.17
+   */
+  public abstract <T> void createTimeSeries(
+      List<LabelValue> labelValues, @Nullable T obj, ToDoubleFunction<T> function);
+
+  /**
+   * Removes the {@code TimeSeries} from the gauge metric, if it is present. i.e. references to
+   * previous {@code Point} objects are invalid (not part of the metric).
```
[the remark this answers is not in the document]
mayurkale22: Removed.

impl_core/src/main/java/io/opencensus/implcore/metrics/DerivedDoubleGaugeImpl.java
```diff
@@ -0,0 +1,141 @@
...
+    private final ToDoubleFunction<T> function;
+
+    PointWithFunction(List<LabelValue> labelValues, @Nullable T obj, ToDoubleFunction<T> function) {
+      this.labelValues = labelValues;
+      ref = obj != null ? new WeakReference<T>(obj) : null;
+      this.function = function;
+    }
+
+    @Override
+    public TimeSeries getTimeSeries(Clock clock) {
```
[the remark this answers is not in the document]
mayurkale22: This is implemented through ```TimeSeriesProducer``` interface.

impl_core/src/main/java/io/opencensus/implcore/metrics/DerivedDoubleGaugeImpl.java
```diff
@@ -0,0 +1,141 @@
...
+    // Updating the map of time series happens under a lock to avoid multiple add operations
+    // to happen in the same time.
+    Map<List<LabelValue>, TimeSeriesProducer> registeredPointsCopy =
+        new HashMap<List<LabelValue>, TimeSeriesProducer>(registeredPoints);
+    registeredPointsCopy.put(labelValues, newTimeSeries);
+    registeredPoints = Collections.unmodifiableMap(registeredPointsCopy);
+  }
+
+  @Override
+  public Metric getMetric(Clock clock) {
```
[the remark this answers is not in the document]
mayurkale22: This is implemented through ```Meter``` interface.

api/src/main/java/io/opencensus/metrics/DoubleGauge.java
```diff
@@ -0,0 +1,217 @@
...
+ *
+ *   List<LabelKey> labelKeys = Arrays.asList(LabelKey.create("Name", "desc"));
+ *   List<LabelValue> labelValues = Arrays.asList(LabelValue.create("Inbound"));
+ *
+ *   DoubleGauge gauge = metricRegistry.addDoubleGauge(
+ *       "queue_size", "Pending jobs in a queue", "1", labelKeys);
+ *
+ *   void doSomeWork() {
+ *      // Your code here.
+ *      gauge.getOrCreateTimeSeries(labelValues).set(15);
```
bogdandrutu: Please keep a reference to the returned point instead of calling all the time getOrCreateTimeSeries

api/src/main/java/io/opencensus/metrics/DoubleGauge.java
```diff
@@ -0,0 +1,217 @@
...
+   * @return the no-op implementation of the {@code DoubleGauge}.
+   * @since 0.17
+   */
+  static DoubleGauge getNoopDoubleGauge(
+      String name, String description, String unit, List<LabelKey> labelKeys) {
+    return NoopDoubleGauge.getInstance(name, description, unit, labelKeys);
+  }
+
+  /**
+   * The value of a single Gauge.
```
bogdandrutu: The value of a single point in the Gauge.TimeSeries?

api/src/main/java/io/opencensus/metrics/LongGauge.java
```diff
@@ -0,0 +1,217 @@
...
+ *
+ *   List<LabelKey> labelKeys = Arrays.asList(LabelKey.create("Name", "desc"));
+ *   List<LabelValue> labelValues = Arrays.asList(LabelValue.create("Inbound"));
+ *
+ *   LongGauge gauge = metricRegistry.addLongGauge(
+ *       "queue_size", "Pending jobs in a queue", "1", labelKeys);
+ *
+ *   void doSomeWork() {
+ *      // Your code here.
+ *      gauge.getOrCreateTimeSeries(labelValues).set(15);
```
bogdandrutu: ditto.

impl_core/src/main/java/io/opencensus/implcore/metrics/DerivedLongGaugeImpl.java
```diff
@@ -0,0 +1,148 @@
...
+    // to happen in the same time.
+    Map<List<LabelValue>, PointWithFunction> registeredPointsCopy =
+        new HashMap<List<LabelValue>, PointWithFunction>(registeredPoints);
+    registeredPointsCopy.put(labelValues, newTimeSeries);
+    registeredPoints = Collections.unmodifiableMap(registeredPointsCopy);
+  }
+
+  @Override
+  @SuppressWarnings("rawtypes")
+  public Metric getMetric(Clock clock) {
```
bogdandrutu: We should make this to return null if no TimeSeries correct? Same everywhere.

impl_core/src/main/java/io/opencensus/implcore/metrics/DerivedLongGaugeImpl.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    private TimeSeries getTimeSeries(Clock clock) {
+      final T obj = ref != null ? ref.get() : null;
+
+      @SuppressWarnings("incompatible") // if obj is null
+      long value = function.applyAsLong(obj);
+
+      return TimeSeries.createWithOnePoint(
+          labelValues,
```
bogdandrutu: Here we will re-evaluate the labelValues all the time. Consider (maybe in a separate PR) to have a helper function:

```java
class TimeSeries {
  TimeSeries setPoint(Point point) {
     return new TimeSeries(getLableValues(), Collections.singletonList(point));
  }
}
```

This way we avoid checking the labelValues all the time. Here you will have a:

```java
public static final class PointWithFunction {
  private final TimeSeries defaultTimeSeries = TimeSeries.create(labelValues);

  private TimeSeries getTimeSeries(Clock clock) {
     return defaultTimeSeries.setPoint(io.opencensus.metrics.export.Point.create(Value.longValue(value), clock.now()));
  }
}
```

Can be done separate PR. Maybe file an issue.

mayurkale22:
Closing this PR now, and splitting into 4 small parts.

PR #1479, opened 2018-10-02
Make metrics compatible with the current proto definition.

bogdandrutu:
* Add GAUGE_DISTRIBUTION and SUMMARY types.
* Modify Distribution to use sum instead of mean.

### DIFF
--- a/api/src/main/java/io/opencensus/metrics/Summary.java
+++ b/api/src/main/java/io/opencensus/metrics/Summary.java
@@ -177,7 +177,7 @@ public abstract class Summary {
     }
   }
 
-  private static void checkCountAndSum(Long count, Double sum) {
+  private static void checkCountAndSum(@Nullable Long count, @Nullable Double sum) {
     Utils.checkArgument(count == null || count >= 0, "count must be non-negative.");
     Utils.checkArgument(sum == null || sum >= 0, "sum must be non-negative.");
     if (count != null && count == 0) {
--- a/api/src/main/java/io/opencensus/metrics/Summary.java
+++ b/api/src/main/java/io/opencensus/metrics/Summary.java
@@ -48,13 +48,9 @@ public abstract class Summary {
    * @since 0.17
    */
   public static Summary create(@Nullable Long count, @Nullable Double sum, Snapshot snapshot) {
-    Utils.checkArgument(count == null || count >= 0, "count must be non-negative.");
-    Utils.checkArgument(sum == null || sum >= 0, "sum must be non-negative.");
-    if (count != null && count == 0) {
-      Utils.checkArgument(sum == null || sum == 0, "sum must be 0 if count is 0.");
-    }
+    checkCountAndSum(count, sum);
     Utils.checkNotNull(snapshot, "snapshot");
-    return new AutoValue_Summary(count, sum, copySnapshot(snapshot));
+    return new AutoValue_Summary(count, sum, snapshot);
   }
 
   /**
@@ -128,11 +124,7 @@ public abstract class Summary {
      */
     public static Snapshot create(
         @Nullable Long count, @Nullable Double sum, List<ValueAtPercentile> valueAtPercentiles) {
-      Utils.checkArgument(count == null || count >= 0, "count must be non-negative.");
-      Utils.checkArgument(sum == null || sum >= 0, "sum must be non-negative.");
-      if (count != null && count == 0) {
-        Utils.checkArgument(sum == null || sum == 0, "sum must be 0 if count is 0.");
-      }
+      checkCountAndSum(count, sum);
       Utils.checkNotNull(valueAtPercentiles, "valueAtPercentiles");
       Utils.checkListElementNotNull(valueAtPercentiles, "value in valueAtPercentiles");
       return new AutoValue_Summary_Snapshot(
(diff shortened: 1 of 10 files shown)


api/src/main/java/io/opencensus/metrics/Summary.java
```diff
@@ -0,0 +1,192 @@
...
+     *
+     * @param count the number of values in this {@code Snapshot}.
+     * @param sum the number of values in this {@code Snapshot}.
+     * @param valueAtPercentiles the list of {@code ValueAtPercentile}.
+     * @return a {@code Snapshot} with the given values.
+     * @since 0.17
+     */
+    public static Snapshot create(
+        @Nullable Long count, @Nullable Double sum, List<ValueAtPercentile> valueAtPercentiles) {
+      Utils.checkArgument(count == null || count >= 0, "count must be non-negative.");
```
songy23: nit: checks on `count` and `sum` can be extracted to a helper method to avoid duplication.

api/src/main/java/io/opencensus/metrics/Summary.java
```diff
@@ -0,0 +1,192 @@
...
+        Utils.checkArgument(
+            0 < percentile && percentile <= 100.0,
+            "percentile must be in the interval (0.0, 100.0]");
+        Utils.checkArgument(value >= 0, "value must be non-negative");
+        return new AutoValue_Summary_Snapshot_ValueAtPercentile(percentile, value);
+      }
+    }
+  }
+
+  private static Snapshot copySnapshot(Snapshot snapshot) {
```
songy23: Is this needed? I think `Snapshot` is immutable?

api/src/main/java/io/opencensus/metrics/Value.java
```diff
@@ -75,6 +75,17 @@ public static Value distributionValue(Distribution value) {
     return ValueDistribution.create(value);
   }
 
+  /**
+   * Returns a {@link Summary} {@link Value}.
+   *
+   * @param value value in {@link Distribution}.
```
songy23: nit: s/Distribution/Summary

api/src/main/java/io/opencensus/metrics/Value.java
```diff
@@ -75,6 +75,17 @@ public static Value distributionValue(Distribution value) {
...
+   * @return a {@code Distribution} {@code Value}.
```
songy23: ditto

api/src/main/java/io/opencensus/metrics/Value.java
```diff
@@ -191,6 +206,41 @@ static ValueDistribution create(Distribution value) {
     abstract Distribution getValue();
   }
 
-  // TODO(songya): Add support for Summary type.
-  // This is an aggregation that produces percentiles directly.
+  /**
+   * {@link ValueDistribution} contains summary statistics for a population of values. It optionally
```
songy23: ditto

api/src/main/java/io/opencensus/metrics/Value.java
```diff
@@ -191,6 +206,41 @@ static ValueDistribution create(Distribution value) {
...
+   * contains a histogram representing the distribution of those values across a set of buckets.
```
songy23: This should be "It contains a snapshot representing values calculated over an arbitrary time window."

songy23:
A few minor nits. LGTM overall.

api/src/main/java/io/opencensus/metrics/Summary.java
bogdandrutu: Done.

api/src/main/java/io/opencensus/metrics/Summary.java
bogdandrutu: Good catch.

api/src/main/java/io/opencensus/metrics/Value.java
bogdandrutu: done.

api/src/main/java/io/opencensus/metrics/Value.java
bogdandrutu: done.

api/src/main/java/io/opencensus/metrics/Value.java
bogdandrutu: done.

api/src/main/java/io/opencensus/metrics/Value.java
bogdandrutu: done. too much copy paste.

api/src/main/java/io/opencensus/metrics/Summary.java
```diff
@@ -185,8 +177,11 @@ public static ValueAtPercentile create(double percentile, double value) {
     }
   }
 
-  private static Snapshot copySnapshot(Snapshot snapshot) {
-    return Snapshot.create(
-        snapshot.getCount(), snapshot.getSum(), snapshot.getValueAtPercentiles());
+  private static void checkCountAndSum(Long count, Double sum) {
```
songy23: `count` and `sum` should be `@Nullable`.

bogdandrutu: done.

api/src/main/java/io/opencensus/metrics/Summary.java
```diff
@@ -0,0 +1,187 @@
...
+     * Creates a {@link Snapshot}.
+     *
+     * @param count the number of values in this {@code Snapshot}.
+     * @param sum the number of values in this {@code Snapshot}.
+     * @param valueAtPercentiles the list of {@code ValueAtPercentile}.
+     * @return a {@code Snapshot} with the given values.
+     * @since 0.17
+     */
+    public static Snapshot create(
+        @Nullable Long count, @Nullable Double sum, List<ValueAtPercentile> valueAtPercentiles) {
```
mayurkale22: You already defined ```count``` and ```sum``` for ```Summary```. Do we still need same for ```Snapshot```?

bogdandrutu: Read the proto definition. count and sum in the Summary refer to the whole population (all events recorded) and the count and sum in the Snapshot refer to the events in that sliding windows. So the once in the Summary are counters, the once in the Snapshot are gauges.

PR #1476, opened 2018-09-29
Implementation of OC-Agent Trace Service clients.

songy23:
Fixes https://github.com/census-instrumentation/opencensus-java/issues/1408.

Equivalent to https://github.com/census-ecosystem/opencensus-go-exporter-ocagent/blob/master/ocagent.go. 

This works similarly to Go Agent exporter: 
- `Export` will be handled in the main thread. OC library (`SpanExporter`) pushes `SpanData` to `ExportClient`, then `ExportClient` translates them to `SpanProto` and pushes to Agent.
- `Config` will be handled in a daemon thread, since waiting and applying incoming `Config` will the block the thread.

PR #1481, opened 2018-10-05
Testing: Add a fake implementation of TraceService.

songy23:
This fake server will be used to test against the clients in https://github.com/census-instrumentation/opencensus-java/pull/1476.

Equivalent to https://github.com/census-ecosystem/opencensus-go-exporter-ocagent/blob/master/mock_agent_test.go.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -189,6 +189,8 @@ subprojects {
                 google_cloud_monitoring: "com.google.cloud:google-cloud-monitoring:${googleCloudGaVersion}",
                 grpc_context: "io.grpc:grpc-context:${grpcVersion}",
                 grpc_core: "io.grpc:grpc-core:${grpcVersion}",
+                grpc_netty: "io.grpc:grpc-netty:${grpcVersion}",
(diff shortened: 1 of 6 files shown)


songy23:
@bogdandrutu I considered putting this fake server under io.opencensus.testing initially. A few concerns about doing so:
1. This will add `io.grpc` and `io.opencensus.proto` dependencies to `io.opencensus.testing`.
2. We have to make the fake server implementation public, while I don't think it will be useful outside the ocagent exporter package.

That being said, I agree that there's no need to include the fake server in the ocagent-exporter artifact. I've moved it to `io.opencensus.testing`.

bogdandrutu:
@songy23 my suggestion was to have it in the src/test/java instead of src/main/java. What do you think?

songy23:
> my suggestion was to have it in the src/test/java instead of src/main/java. What do you think?

SGTM, updated. (though it's a bit strange to have tests for a class under test directory.)

songy23:
> we can move it to the main later but I don't think this should be present in the artifact that gets exported to maven central for example even if this is package protected.

Make sense to me.

PR #1485, opened 2018-10-10
Metrics: Copy TimeSeries list when creating a Metric.

songy23:


api/src/main/java/io/opencensus/metrics/export/Metric.java
```diff
@@ -48,7 +50,8 @@
    */
   public static Metric create(MetricDescriptor metricDescriptor, List<TimeSeries> timeSeriesList) {
     checkTypeMatch(metricDescriptor.getType(), timeSeriesList);
```
mayurkale22: Does it make sense do ```checkNotNull``` and ```checkListElementNotNull``` on timeSeriesList?

songy23: Make sense, updated. Also added a few tests on that.

PR #1486, opened 2018-10-10
Metrics/TimeSeries: Add createWithOnePoint method to accept only one …

mayurkale22:
…point

### DIFF
--- a/api/src/main/java/io/opencensus/metrics/export/TimeSeries.java
+++ b/api/src/main/java/io/opencensus/metrics/export/TimeSeries.java
@@ -53,6 +53,7 @@ public abstract class TimeSeries {
   public static TimeSeries create(
       List<LabelValue> labelValues, List<Point> points, @Nullable Timestamp startTimestamp) {
     Utils.checkNotNull(points, "points");
+    Utils.checkListElementNotNull(points, "point");
     return createInternal(
         labelValues, Collections.unmodifiableList(new ArrayList<Point>(points)), startTimestamp);
   }
@@ -87,7 +88,6 @@ public abstract class TimeSeries {
     // Fail fast on null lists to prevent NullPointerException when copying the lists.
     Utils.checkNotNull(labelValues, "labelValues");
     Utils.checkListElementNotNull(labelValues, "labelValue");
-    Utils.checkListElementNotNull(points, "point");
     return new AutoValue_TimeSeries(
         Collections.unmodifiableList(new ArrayList<LabelValue>(labelValues)),
         points,
--- a/api/src/main/java/io/opencensus/metrics/export/TimeSeries.java
+++ b/api/src/main/java/io/opencensus/metrics/export/TimeSeries.java
@@ -52,15 +52,9 @@ public abstract class TimeSeries {
    */
   public static TimeSeries create(
       List<LabelValue> labelValues, List<Point> points, @Nullable Timestamp startTimestamp) {
-    // Fail fast on null lists to prevent NullPointerException when copying the lists.
-    Utils.checkNotNull(labelValues, "labelValues");
     Utils.checkNotNull(points, "points");
-    Utils.checkListElementNotNull(labelValues, "labelValue");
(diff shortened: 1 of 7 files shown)


api/src/main/java/io/opencensus/metrics/export/TimeSeries.java
```diff
@@ -63,6 +63,28 @@ public static TimeSeries create(
...
+   * @since 0.17
+   */
+  public static TimeSeries createWithOnePoint(
+      List<LabelValue> labelValues, Point point, @Nullable Timestamp startTimestamp) {
+    // Fail fast on null lists to prevent NullPointerException when copying the lists.
+    Utils.checkNotNull(labelValues, "labelValues");
+    Utils.checkNotNull(point, "point");
+    Utils.checkListElementNotNull(labelValues, "labelValue");
+    return new AutoValue_TimeSeries(
+        Collections.unmodifiableList(labelValues),
```
bogdandrutu: Collections.unmodifiableList(new ArrayList<LabelValue>(labelValues))

api/src/main/java/io/opencensus/metrics/export/TimeSeries.java
```diff
@@ -63,6 +63,28 @@ public static TimeSeries create(
...
+   * @param labelValues the {@code LabelValue}s that uniquely identify this {@code TimeSeries}.
+   * @param point the single data {@code Point} of this {@code TimeSeries}.
+   * @param startTimestamp the start {@code Timestamp} of this {@code TimeSeries}. Must be non-null
+   *     for cumulative {@code Point}s.
+   * @return a {@code TimeSeries}.
+   * @since 0.17
+   */
+  public static TimeSeries createWithOnePoint(
+      List<LabelValue> labelValues, Point point, @Nullable Timestamp startTimestamp) {
+    // Fail fast on null lists to prevent NullPointerException when copying the lists.
```
songy23: Consider using `create(labelValues, Collections.singletonList(point), timestamp);` instead of copying the whole code block.

impl_core/src/test/java/io/opencensus/implcore/metrics/MetricRegistryImplTest.java
```diff
@@ -263,9 +263,9 @@ public long applyAsLong(Object value) {
                     Type.GAUGE_INT64,
                     Collections.unmodifiableList(Collections.singletonList(LABEL_KEY))),
                 Collections.singletonList(
```
bogdandrutu: We may want to add the same trick in the Metric for TimeSeries. Not now but we can keep this in mind.

api/src/main/java/io/opencensus/metrics/export/TimeSeries.java
bogdandrutu: The whole idea is to avoid:
Collections.unmodifiableList(new ArrayList<Point>(Collections.singletonList(point)));

I think we should have a createInternal version probably to avoid duplicate.

songy23: SGTM

api/src/main/java/io/opencensus/metrics/export/TimeSeries.java
```diff
@@ -52,14 +52,45 @@
...
+   *     for cumulative {@code Point}s.
+   * @return a {@code TimeSeries}.
+   */
+  private static TimeSeries createInternal(
+      List<LabelValue> labelValues, List<Point> points, @Nullable Timestamp startTimestamp) {
     // Fail fast on null lists to prevent NullPointerException when copying the lists.
     Utils.checkNotNull(labelValues, "labelValues");
-    Utils.checkNotNull(points, "points");
     Utils.checkListElementNotNull(labelValues, "labelValue");
     Utils.checkListElementNotNull(points, "point");
```
bogdandrutu: This can be moved to "create" method because that is the only one that needs it.

PR #1488, opened 2018-10-10
Exporter/OcAgent: Add methods for getting and applying config.

songy23:
Split from #1476, to make that PR shorter.

### DIFF
--- a/exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/TraceProtoUtils.java
+++ b/exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/TraceProtoUtils.java
@@ -23,8 +23,6 @@ import com.google.protobuf.UInt32Value;
 import io.opencensus.common.Function;
 import io.opencensus.common.Functions;
 import io.opencensus.common.Timestamp;
-import io.opencensus.proto.agent.common.v1.Node;
-import io.opencensus.proto.agent.trace.v1.CurrentLibraryConfig;
 import io.opencensus.proto.agent.trace.v1.UpdatedLibraryConfig;
 import io.opencensus.proto.trace.v1.AttributeValue;
 import io.opencensus.proto.trace.v1.ConstantSampler;
@@ -106,7 +104,6 @@ final class TraceProtoUtils {
    * @param spanData the {@code SpanData}.
    * @return proto representation of {@code Span}.
    */
-  @SuppressWarnings("DefaultCharset")
   static Span toSpanProto(SpanData spanData) {
     SpanContext spanContext = spanData.getContext();
     TraceId traceId = spanContext.getTraceId();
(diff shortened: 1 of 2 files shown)


exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/TraceProtoUtils.java
```diff
@@ -106,10 +105,9 @@ public AttributeValue apply(Double doubleValue) {
    *
    * @param spanData the {@code SpanData}.
    * @return proto representation of {@code Span}.
-   * @since 0.17
    */
   @SuppressWarnings("DefaultCharset")
```
bogdandrutu: We should probably have reproducible builds so we need to use a charset here if that is needed.

exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/TraceProtoUtils.java
```diff
@@ -378,5 +375,26 @@ public static TraceParams fromTraceConfigProto(
     return builder.build();
   }
 
+  // Creates a CurrentLibraryConfig message with the given Node and current TraceParams.
+  static CurrentLibraryConfig getCurrentLibraryConfig(
+      @javax.annotation.Nullable Node node, io.opencensus.trace.config.TraceConfig traceConfig) {
```
bogdandrutu: Probably better to return the TraceConfig proto here and avoid adding the only first time send the node logic in this util class.

exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/TraceProtoUtils.java
songy23: Removed - this seems to be I forgot to remove this annotation. It's no longer applicable after using `ByteString.copyFrom(bytes);`.

exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/TraceProtoUtils.java
songy23: Make sense, done.

exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/TraceProtoUtils.java
```diff
@@ -378,5 +371,21 @@ public static TraceParams fromTraceConfigProto(
...
 
+  // Creates a TraceConfig proto message with current TraceParams.
+  static TraceConfig getCurrentTraceConfig(io.opencensus.trace.config.TraceConfig traceConfig) {
+    TraceParams traceParams = traceConfig.getActiveTraceParams();
+    return toTraceConfigProto(traceParams);
+  }
+
+  // Creates an updated TraceParams with the given UpdatedLibraryConfig message and current
+  // TraceParams, then applies the updated TraceParams.
+  static void applyUpdatedConfig(
```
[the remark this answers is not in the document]
songy23: Make sense, done.

PR #1492, opened 2018-10-12
Exporter/OcAgent: Add handlers for export and config RPCs.

songy23:

Also added a long long explanation on the states and how to check streams.


exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentTraceServiceStreamObservers.java
```diff
@@ -0,0 +1,206 @@
...
+/** Stream observers of OC-Agent Trace Service. */
+final class OcAgentTraceServiceStreamObservers {
+
+  private OcAgentTraceServiceStreamObservers() {}
+
+  private static final Logger logger =
+      Logger.getLogger(OcAgentTraceServiceStreamObservers.class.getName());
+
+  /*
+   * gRPC-Java has good abstractions over the transport layer, and there's no way we can know the
```
bogdandrutu: Why do we not use this:
https://grpc.io/grpc-java/javadoc/io/grpc/ManagedChannel.html#getState-boolean-?

This seems to achieve what you need to check if the agent is up and accepts connections. If this works I think we simplify a lot the design.

exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentTraceServiceStreamObservers.java
```diff
@@ -0,0 +1,206 @@
...
+   * connection status directly (nothing like *grpc.ClientConn in Go). Instead, we should verify the
+   * connection when initiating and sending streams:
+   *
+   * 0. On client side we defined exportResponseObserver and updatedLibraryConfigObserver. On server
+   * side we defined exportRequestObserver and currentConfigObserver.
```
bogdandrutu: I think it is not very good to use "on client" and "on server". every RPC is bi-directional stream so we have 4 streams two on each RPC. You can say that for export RPC we have X as StreamObserver for request stream and Y as StreamObserver for response stream.

I would use:
* exportRequestObserverRef
* configRequestObserverRef
* exportResponseObserverRef
* configResponseObserverRef

exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentTraceServiceStreamObservers.java
```diff
@@ -0,0 +1,206 @@
...
+        currentConfigObserver.onNext(currentLibraryConfig);
+      }
+    }
+
+    @Override
+    public void onError(Throwable t) {
+      // If there's an error, reset the reference to currentConfigObserver to force
+      // re-establish a connection.
+      // TODO(songya): add retries on this connection before starting a new connection.
+      logger.log(Level.WARNING, "Config stream is disconnected because of: {0}", t);
```
bogdandrutu: Better to use this:
https://docs.oracle.com/javase/7/docs/api/java/util/logging/Logger.html#log(java.util.logging.Level,%20java.lang.String,%20java.lang.Throwable)

logger.log(Level.WARNING, "Config stream is disconnected.", t);

songy23: Done.

exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentTraceServiceStreamObservers.java
songy23: Make sense to me, updated. Also replace "Client/Server" with "Exporter/Agent".

exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentTraceServiceStreamObservers.java
songy23: I see, I didn't realize we can check channel connection with this method. This will be convenient for the worker thread that tries to connect to Agent, though I think the logic for `StreamObserver`s should stay the same.

exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentTraceServiceStreamObservers.java
```diff
@@ -0,0 +1,213 @@
...
+   *
+   * Note: "Exporter" = Agent Exporter and "Agent" = OC-Agent in the sections below.
+   *
+   * STATE 1: Unconnected/Disconnected.
+   *
+   * 1. First, Exporter will try to establish the streams by calling TraceServiceStub.export and
+   * TraceServiceStub.config in a daemon thread.
+   *
+   *   1-1. If the initial attempt succeeded, exportRequestObserver and currentConfigObserver
+   *   should be returned to Exporter as callbacks, and we keep a reference of them in the
```
bogdandrutu: "should be returned to Exporter as callbacks" they are return from where?

exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentTraceServiceStreamObservers.java
```diff
@@ -0,0 +1,213 @@
...
+   *   exportRequestObserverRef and currentConfigObserverRef. At the same time,
+   *   exportRequestObserver and updatedConfigObserver will be passed to Agent.
```
bogdandrutu: How can you pass an object to the Agent (a different process)?

exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentTraceServiceStreamObservers.java
bogdandrutu: I think we have different ways to understand what is a stream. I will send more comments.

exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentTraceServiceStreamObservers.java
```diff
@@ -0,0 +1,213 @@
...
+/** Stream observers of OC-Agent Trace Service. */
+final class OcAgentTraceServiceStreamObservers {
+
+  private OcAgentTraceServiceStreamObservers() {}
+
+  private static final Logger logger =
+      Logger.getLogger(OcAgentTraceServiceStreamObservers.class.getName());
+
+  /*
+   * We can use ManagedChannel.getState() to determine the state of current connection.
```
bogdandrutu: Why do we not use that?

exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentTraceServiceStreamObservers.java
```diff
@@ -0,0 +1,213 @@
...
+   *
+   * 0. For export RPC we hold a reference "exportRequestObserverRef" to a StreamObserver for
+   * request stream, and "exportRequestObserverRef" to a StreamObserver for response stream.
+   * For config RPC we have "currentConfigObserverRef" and "updatedConfigObserverRef". Note that
```
bogdandrutu: Please don't name the streams based on the message names. Use request/response because that is easier to understand. As I commented in a previous comment.

exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentTraceServiceStreamObservers.java
songy23: We should use it, though not in this class, since `channel` is not visible to stream observers.

exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentTraceServiceStreamObservers.java
songy23: Exporter will have a "reference" to the stream observers from Agent.

On Exporter:
```java
StreamObserver<CurrentLibraryConfig> currentConfigObserver = stub.config(updatedLibraryObserver); 
```

On Agent:
```java
currentConfigObserver = ...;  // Defined and created by Agent.

@Override
public StreamObserver<CurrentLibraryConfig> config(
      StreamObserver<UpdatedLibraryConfig> updatedLibraryConfigStreamObserver) {
    return currentConfigObserver;
}
```

exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentTraceServiceStreamObservers.java
songy23: Like I mentioned in the previous comment, the reference of observers is passed to Agent as a callback.

On Exporter:
```java
updatedLibraryObserver = ..; // Defined and created by Exporter.

StreamObserver<CurrentLibraryConfig> currentConfigObserver = stub.config(updatedLibraryObserver); 
```

On Agent:
```java
updatedLibraryObserverReference;

@Override
public StreamObserver<CurrentLibraryConfig> config(
      StreamObserver<UpdatedLibraryConfig> updatedLibraryConfigStreamObserver) {
  // updatedLibraryConfigStreamObserver is from Exporter.
  updatedLibraryObserverReference.set(updatedLibraryConfigStreamObserver);  
  return ...
}
```

exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentTraceServiceStreamObservers.java
songy23: The problem is we have two streams and they are in the opposite direction: export is Exporter -> Agent, while config is Agent -> Exporter. Let me think about how to make this more clear.

songy23: Took a second look at this block, I still think it's necessary to differentiate between the two streams (export and config) since they work in completely different ways. Using generic request/response for both of them sounds confusing to me. Again this may be my personal bias, so if you feel strongly that request/response are easier for others to understand (probably everyone except me), I'm happy to re-phrase it that way.

songy23: I.e from Exporter's perspective:
- For `export`, Exporter is sending requests and receiving responses.
- For `config`, Exporter is receiving requests and sending responses.

bogdandrutu: You can name the config streams:
configRequestObserverRef
configResponseObserverRef

exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentTraceServiceStreamObservers.java
```diff
@@ -0,0 +1,245 @@
...
+  /*
+   * We should use ManagedChannel.getState() to determine the state of current connection.
+   *
+   * // TODO(songya): avoid going into too much details of each stream and use generic
+   * request/response if that's more clear.
+   *
+   * 0. For export RPC we hold a reference "exportRequestObserverRef" to a StreamObserver for
+   * request stream, and "exportResponseObserverRef" to a StreamObserver for response stream.
+   * For config RPC we have "currentConfigObserverRef" and "updatedConfigObserverRef". Note that
+   * exportRequestObserver and currentConfigObserver are defined by Agent, while
```
bogdandrutu: Just drop the "Note that ..." sentence.

exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentTraceServiceStreamObservers.java
```diff
@@ -0,0 +1,245 @@
...
+   *
+   * STATE 1: Unconnected/Disconnected.
+   *
+   * 1. First, Exporter will try to establish the streams by calling TraceServiceStub.export and
+   * TraceServiceStub.config in a daemon thread.
+   *
+   *   1-1. If the initial attempt succeeded, references of exportRequestObserver and
+   *   currentConfigObserver should be returned to Exporter as callbacks. We keep a reference of
+   *   them in exportRequestObserverRef and currentConfigObserverRef. At the same time, references
+   *   of exportRequestObserver and updatedConfigObserver will be passed to Agent as callbacks.
```
bogdandrutu: As I mentioned before you don't pass any callback. The stream object is owned by the Exporter, you can say via these streams the Exporter receives messages from the Agent. Similar for request streams they are used to send messages to the Agent.

exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentTraceServiceStreamObservers.java
```diff
@@ -0,0 +1,245 @@
...
+   *   1-3. After each attempt, we should check if the connection succeeded or not with
+   *   ManagedChannel.getState(). (This is done in the daemon thread.)
+   *   If connection succeeded and Exporter has the references to exportRequestObserver and
+   *   currentConfigObserver, we can move forward and start sending/receiving streams
+   *   (move to STATE 2). Otherwise Exporter should retry.
+   *
+   * STATE 2: Already connected.
+   *
+   * 2. Once streams are open, they should be kept alive. Exporter sends spans to Agent by calling
+   * exportRequestObserver.onNext (the callback provided by Agent). In addition, in another daemon
```
bogdandrutu: Remove "(the callback provided by Agent)"

exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentTraceServiceStreamObservers.java
```diff
@@ -0,0 +1,245 @@
...
+   * thread Exporter watches config updates from Agent, which is done by Agent calling back
```
bogdandrutu: The agent is not "calling back". Just say: "In a different demon thread Exporter watches config updates from Agent." The updates can come only via the configResponseObserverRef because that is the stream for incoming messages for config.

exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentTraceServiceStreamObservers.java
```diff
@@ -0,0 +1,245 @@
...
+   * updatedConfigObserver.onNext.
+   *
+   *   2-1. If for some reason the connection is interrupted, the errors will be caught by
```
bogdandrutu: Unify 2-1 and 2-2 by saying this:
If for some reason the connection is interrupted, the errors will be caught by the onError method of any stream observer.  In that case, we should reset references to all the stream observers.

exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentTraceServiceStreamObservers.java
```diff
@@ -0,0 +1,245 @@
...
+    @Override
+    public void onError(Throwable t) {
+      // If there's an error, reset the reference to currentConfigObserver to force
+      // re-establish a connection.
+      // TODO(songya): add retries on this connection before starting a new connection.
+      logger.log(Level.WARNING, "Config stream is disconnected.", t);
+      /*@Nullable*/ StreamObserver<CurrentLibraryConfig> currentConfigObserver =
+          currentConfigObserverRef.get();
+      if (currentConfigObserver != null) {
+        currentConfigObserver.onError(t); // Tries to notify OC-Agent about the error.
```
bogdandrutu: Why this?

bogdandrutu:
I think also it is important that this class has methods to "createXYZ" and probably get a Runnable for onError. The owners will be in a different class.

exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentTraceServiceStreamObservers.java
songy23: SG, updated.

exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentTraceServiceStreamObservers.java
songy23: Done.

exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentTraceServiceStreamObservers.java
songy23: Rephrased a bit.

exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentTraceServiceStreamObservers.java
songy23: Done.

exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentTraceServiceStreamObservers.java
songy23: Removed all occurrence of "callback".

exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentTraceServiceStreamObservers.java
songy23: Done - also onComplete and onError are a bit different so there won't always be errors when the streams are disconnected.

exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentTraceServiceStreamObservers.java
songy23: This is for cases when Exporter saw errors in the stream but Agent hadn't seen, and Exporter would try to notify Agent by calling `onError`. Anyway this is not needed since when there're errors in the stream, stream is likely to be interrupted and `onError` won't work.

songy23:
BTW it would be a lot easier if we could discuss about this on a white board.

songy23:
Updated based on our discussion this week. Also added a bunch of unit tests. PTAL.

Other changes (`Runnable`, retry) will be in https://github.com/census-instrumentation/opencensus-java/pull/1506, since this PR is already large.

songy23:
> 750 lines for a PR seems a bit large. Consider to split it.

SG, will do.

songy23:
Closing - going to split this PR into smaller ones.

PR #1498, opened 2018-10-15
Plugs-in the LongGauge into the registry

mayurkale22:


api/src/main/java/io/opencensus/metrics/MetricRegistry.java
```diff
@@ -32,44 +30,21 @@
 @ExperimentalApi
 public abstract class MetricRegistry {
   /**
-   * Build a new long gauge to be added to the registry.
-   *
-   * <p>Must be called only once.
+   * Build a new long gauge to be added to the registry. This is more convenient form when you want
```
songy23: nit: s/Build/Builds

impl_core/src/main/java/io/opencensus/implcore/metrics/MetricRegistryImpl.java
```diff
@@ -46,56 +43,38 @@
...
+      return registeredMeters;
     }
 
-    private synchronized void registerMeter(Gauge gauge) {
-      Set<Gauge> newGaguesList = new LinkedHashSet<Gauge>(registeredGauges);
-      newGaguesList.add(gauge);
-      registeredGauges = Collections.unmodifiableSet(newGaguesList);
+    private synchronized void registerMeter(String meterName, Meter meter) {
+      Meter existingMeter = registeredMeters.get(meterName);
+      if (existingMeter != null) {
```
songy23: If `existingMeter` and `meter` is same, this should be a no-op instead of throwing exception.

mayurkale22: I think ```Meter``` interface should expose ```MetricDescriptor``` to do this check OR another way is to override ```equals``` and ```hashCode``` methods in all Meter classes (like LongGauge, DoubleGauge etc.)

songy23: Could you add a TODO on this? Not a critical issue but I think we should allow users to register the same `Meter` multiple times. (Otherwise we should document the behavior that registering the same `Meter` for more than once will throw an exception.)

songy23:
Also please fix the build.

api/src/main/java/io/opencensus/metrics/DoubleGauge.java
```diff
@@ -86,12 +86,12 @@
    * method for manual operations.
    *
    * @param labelValues the list of label values. The number of label values must be the same to
-   *     that of the label keys passed to {@link MetricRegistry#addDoubleGauge}.
```
mayurkale22: I will add it back once I register this gauge in MetricRegistry (Next PR).

api/src/main/java/io/opencensus/metrics/MetricRegistry.java
```diff
@@ -78,33 +53,16 @@ static MetricRegistry newNoopMetricRegistry() {
...
-        ToDoubleFunction<T> function) {
-      Utils.checkNotNull(name, "name");
-      Utils.checkNotNull(description, "description");
-      Utils.checkNotNull(unit, "unit");
-      Utils.checkNotNull(labels, "labels");
-      Utils.checkNotNull(function, "function");
+    public LongGauge addLongGauge(
+        String name, String description, String unit, List<LabelKey> labelKeys) {
+      Utils.checkListElementNotNull(
+          Utils.checkNotNull(labelKeys, "labelKeys should not be null."),
```
bogdandrutu: We have a convention that for null check we use the message as the name of the variable. so this should be: `Utils.checkNotNull(labelKeys, "labelKeys")`

api/src/main/java/io/opencensus/metrics/MetricRegistry.java
```diff
@@ -78,33 +53,16 @@ static MetricRegistry newNoopMetricRegistry() {
...
+          "labelKeys element should not be null.");
```
bogdandrutu: consider the message `"labelKey"`

impl_core/src/main/java/io/opencensus/implcore/metrics/MetricRegistryImpl.java
```diff
@@ -46,56 +43,39 @@
...
-      LinkedHashMap<LabelKey, LabelValue> labels,
-      T obj,
-      ToLongFunction<T> function) {
-    checkNotNull(labels, "labels");
-    registeredMeters.registerMeter(
-        new LongGauge<T>(
+  public LongGauge addLongGauge(
+      String name, String description, String unit, List<LabelKey> labelKeys) {
+    Utils.checkListElementNotNull(
+        checkNotNull(labelKeys, "labelKeys should not be null."),
```
bogdandrutu: same comments for exception messages.

PR #1499, opened 2018-10-15
ignore negative bucket bounds

mayurkale22:


songy23:
For backwards-compatibility, I think we need to continue supporting negative buckets in Stats APIs. In my opinion we should drop the negative buckets when converting `MutableViewData` to `Metric`s rather than drop them when doing the aggregation.

@bogdandrutu WDYT?

bogdandrutu:
Negative values are currently not supported by any of the backends that we support and have histograms (Stackdriver and Prometheus). In Stackdriver for example sum cannot go down.

There are multiple things to consider:
1. If you drop negative from the MutableView, you should drop from count and sum as well.
2. To do a backwards compatible change we should probably have 2 count and 2 sum and always add a 0 in the bucket definition. One count/sum for >=0 and one count/sum for <0 and make sure that we always have negative buckets and positive buckets separated (one bucket cannot have negative and positive).
3. Probably we should consider a backwards incompatible change and drop negative values in the Record API, that will simplify a lot of things (I would like to hear what others think about this). @dinooliva @songy23 @mayurkale22?

songy23:
> Probably we should consider a backwards incompatible change and drop negative values in the Record API, that will simplify a lot of things (I would like to hear what others think about this). 

Given the complexity of maintaining both the positive and negative aggregated values, I think dropping negative values makes more sense. An alternative is to throw an unchecked exception for negative values so that users will be noticed that negative values are no longer supported (instead of silently skipping them).

songy23:
>  We can start with logging an error for few versions then throw an exception. We should explain this to the current users very clearly.

Agree that logging is a good starting point. We should document this behavior in the Javadoc as well as in the CHANGELOG.

impl_core/src/main/java/io/opencensus/implcore/stats/MutableAggregation.java
```diff
@@ -292,6 +296,11 @@ static MutableDistribution create(BucketBoundaries bucketBoundaries) {
 
     @Override
     void add(double value, Map<String, String> attachments, Timestamp timestamp) {
+      if (value < 0) {
+        logger.log(Level.WARNING, "Dropping negative values for histogram count and sum.");
```
songy23: nit: You can also log the `value`.

bogdandrutu:
I agree with @songy23 we should drop negative values in the Record (and drop all recorded values in that record batch).

impl_core/src/main/java/io/opencensus/implcore/stats/MeasureMapImpl.java
```diff
@@ -37,13 +41,21 @@ private MeasureMapImpl(StatsManager statsManager) {
...
+    } else {
+      builder.put(measure, value);
+    }
     return this;
   }
 
   @Override
   public MeasureMapImpl put(MeasureLong measure, long value) {
-    builder.put(measure, value);
+    if (value < 0) {
```
bogdandrutu: Can we check and record warnings in the noop implementation as well?

I would not do this in Put, instead I would do it in the record. As I mentioned is better to drop all recorded values because users may use correlation between different values recorded at the same time.

api/src/main/java/io/opencensus/stats/BucketBoundaries.java
```diff
@@ -46,14 +50,37 @@ public static final BucketBoundaries create(List<Double> bucketBoundaries) {
...
+        Utils.checkArgument(previous < next, "Bucket boundaries not sorted.");
+        previous = next;
       }
     }
-    return new AutoValue_BucketBoundaries(Collections.unmodifiableList(bucketBoundariesCopy));
+    return new AutoValue_BucketBoundaries(
+        Collections.unmodifiableList(dropNegativeBucketBounds(bucketBoundariesCopy)));
+  }
+
+  private static List<Double> dropNegativeBucketBounds(List<Double> bucketBoundaries) {
```
bogdandrutu: Metrics histogram does not accept 0 https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/metrics/export/Distribution.java#L211 so I think you need to drop the 0 as well but don't record an error for that (or maybe just a LOG.info if you really care) because that is correctly handled by metrics with the default [0, first_item) interval.

api/src/main/java/io/opencensus/stats/NoopStats.java
```diff
@@ -140,6 +150,12 @@ public void record() {}
     @Override
     public void record(TagContext tags) {
       Utils.checkNotNull(tags, "tags");
+
+      if (hasUnsupportedValues) {
+        // drop all the recorded values
+        logger.log(Level.WARNING, "Dropping values, value to record must be non-negative.");
+        hasUnsupportedValues = false;
```
bogdandrutu: Why do you reset this value? If they call record twice they get values recorded?

impl_core/src/main/java/io/opencensus/implcore/stats/MeasureMapImpl.java
```diff
@@ -61,6 +72,13 @@ public void record() {
 
   @Override
   public void record(TagContext tags) {
+    if (hasUnsupportedValues) {
+      // drop all the recorded values
+      builder.clear();
```
bogdandrutu: this is a change that we should not add it right now. I think record should not reset the values.

impl_core/src/main/java/io/opencensus/implcore/stats/MeasureMapImpl.java
```diff
@@ -61,6 +72,13 @@ public void record() {
...
+      logger.log(Level.WARNING, "Dropping values, value to record must be non-negative.");
+      hasUnsupportedValues = false;
```
bogdandrutu: same comment here that you should not reset this.

api/src/main/java/io/opencensus/stats/NoopStats.java
mayurkale22: I thought will allow users to record values again, anyways removed the value reset.

impl_core/src/main/java/io/opencensus/implcore/stats/MeasureMapImpl.java
mayurkale22: done

api/src/main/java/io/opencensus/stats/NoopStats.java
```diff
@@ -140,6 +150,11 @@ public void record() {}
     @Override
     public void record(TagContext tags) {
       Utils.checkNotNull(tags, "tags");
+
+      if (hasUnsupportedValues) {
+        // drop all the recorded values
+        logger.log(Level.WARNING, "Dropping values, value to record must be non-negative.");
```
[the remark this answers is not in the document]
mayurkale22: Agreed, removed the singleton instance. PTAL

api/src/main/java/io/opencensus/stats/NoopStats.java
```diff
@@ -116,21 +118,28 @@ public void setState(StatsCollectionState state) {
...
-      return getNoopMeasureMap();
+      return newNoopMeasureMap();
     }
   }
 
   @Immutable
   private static final class NoopMeasureMap extends MeasureMap {
-    static final MeasureMap INSTANCE = new NoopMeasureMap();
+    private static final Logger logger = Logger.getLogger(NoopMeasureMap.class.getName());
+    private volatile boolean hasUnsupportedValues;
```
bogdandrutu: don't need to make this volatile because the builder is not marked as ThreadSafe. Also this class is no longer Immutable.

mayurkale22: Done

Issue #1138, opened 2018-04-19
Find a way to share utility classes and methods across artifacts.


sebright:
There are several reasons to add shared utilities currently:

- If we remove uses of Guava from the implementation and exporters (issue #1113), we'll need to replace the Guava functionality, such as validating arguments.  It would be nice to avoid duplicating the code in every artifact.

- There are two copies of DaemonThreadFactory, in [impl_core](https://github.com/census-instrumentation/opencensus-java/blob/9e518c6743ef46d0d1e18b5927c7dfc754b6d835/impl_core/src/main/java/io/opencensus/implcore/internal/DaemonThreadFactory.java#L24) and the [Stackdriver stats exporter](https://github.com/census-instrumentation/opencensus-java/blob/9e518c6743ef46d0d1e18b5927c7dfc754b6d835/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporter.java#L331), that could be to combined into one class.

Here are a few possible solutions that would avoid duplicating code:

- Add a new artifact and upload it to Maven Central so that all OpenCensus artifacts can depend on it.
- Add a package to opencensus-api that is similar to `io.opencensus.internal`, but stable enough to be used outside of opencensus-api.
- Add a new subproject that can be used as a shaded dependency by the other artifacts.  It wouldn't need to be uploaded to Maven Central.  There is some discussion of this idea in https://github.com/census-instrumentation/opencensus-java/pull/1114#pullrequestreview-109828652, and an experimental branch (https://github.com/sebright/opencensus-java/tree/shadow-plugin).

PR #1500, opened 2018-10-17
Exporter/OcAgent: Add DaemonThreadFactory.

songy23:

Add `DaemonThreadFactory` for GAE compatibility.

### DIFF
--- a/exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/DaemonThreadFactory.java
+++ b/exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/DaemonThreadFactory.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2018, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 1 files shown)


exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/DaemonThreadFactory.java
```diff
@@ -0,0 +1,72 @@
...
+ * limitations under the License.
+ */
+
+package io.opencensus.exporter.trace.ocagent;
+
+import com.google.common.util.concurrent.MoreExecutors;
+import java.util.concurrent.ThreadFactory;
+
+/** A lightweight {@link ThreadFactory} to spawn threads in a GAE-Java7-compatible way. */
+final class DaemonThreadFactory implements ThreadFactory {
```
bogdandrutu: I saw this class at least 3 times. What is the plan for this? Also GAE-Java7 is gone.

songy23:
> I saw this class at least 3 times. What is the plan for this? 

There's a tracking issue for making it a common shaded artifact: https://github.com/census-instrumentation/opencensus-java/issues/1138.

> Also GAE-Java7 is gone.

Good to know :) Then I don't think we need the daemon factory since only on GAE-Java7 we cannot set a thread to daemon.

PR #1505, opened 2018-10-19
 Plugs-in the DerivedLongGauge and DerivedDoubleGauge into the registry

mayurkale22:


api/src/main/java/io/opencensus/metrics/MetricRegistry.java
```diff
@@ -92,5 +128,29 @@ public DoubleGauge addDoubleGauge(
...
+          Utils.checkNotNull(description, "description"),
+          Utils.checkNotNull(unit, "unit"),
+          labelKeys);
+    }
+
+    @Override
+    public DerivedDoubleGauge addDerivedDoubleGauge(
+        String name, String description, String unit, List<LabelKey> labelKeys) {
+      Utils.checkListElementNotNull(
+          Utils.checkNotNull(labelKeys, "labelKeys"), "labelKey element should not be null.");
```
bogdandrutu: s/"labelKey element should not be null."/"labelKey" 

same everywhere.

mayurkale22: I will fix this in separate PR, I see many places we did like this.

PR #1513, opened 2018-10-24
TimeSeries/Create: Avoid re-evaluate the labelValues all the time

### LINKED ISSUE #1490: Avoid re-evaluate the labelValues all the time
Consider to have a helper function inside TimeSeries class. Something like 

```
class TimeSeries {
  TimeSeries setPoint(Point point) {
     return new TimeSeries(getLableValues(), Collections.singletonList(point));
  }
}
```

mayurkale22:
Fixes #1490


api/src/main/java/io/opencensus/metrics/export/TimeSeries.java
```diff
@@ -57,6 +57,17 @@ public static TimeSeries create(
         labelValues, Collections.unmodifiableList(new ArrayList<Point>(points)), startTimestamp);
   }
 
+  /**
+   * Creates a {@link TimeSeries}.
```
songy23: nit: You can mention that this TimeSeries has no Points.

PR #1516, opened 2018-10-25
Exporter/Stats/Stackdriver: Add user agent as endpoint.

songy23:
Equivalent to https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/blob/master/stats.go#L80-L81.

In Java there's no option named "UserAgent". The closest setting I've found is "endPoint". /cc @rakyll


songy23:
Closing - just realized `UserAgent` and `Endpoint` are different things.

PR #1518, opened 2018-10-26
Exporter/Stats/Stackdriver: Add user agent as client header. 

songy23:
Equivalent to https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/blob/master/stats.go#L80-L81.

See also https://github.com/googleapis/gax-java/blob/a60bd347b69fbf725b0dd8ae18bbcf5b00b66b7b/gax-grpc/src/main/java/com/google/api/gax/grpc/GrpcHeaderInterceptor.java#L63-L67.

### DIFF
--- a/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporter.java
+++ b/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporter.java
@@ -22,6 +22,9 @@ import static com.google.common.base.Preconditions.checkState;
 
 import com.google.api.MonitoredResource;
 import com.google.api.gax.core.FixedCredentialsProvider;
+import com.google.api.gax.grpc.InstantiatingGrpcChannelProvider;
+import com.google.api.gax.rpc.FixedHeaderProvider;
+import com.google.api.gax.rpc.HeaderProvider;
 import com.google.auth.Credentials;
 import com.google.auth.oauth2.GoogleCredentials;
 import com.google.cloud.ServiceOptions;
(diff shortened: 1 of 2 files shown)


exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporter.java
```diff
@@ -317,6 +328,10 @@ private static void createInternal(
         metricServiceClient =
             MetricServiceClient.create(
                 MetricServiceSettings.newBuilder()
+                    .setTransportChannelProvider(
```
mayurkale22: can you pls add unit test for this?

songy23: SG, done.

songy23:
Thanks for the review. Before I merge this PR I'd like to double check with GAX-Java team that this is the correct way for setting headers. Filed https://github.com/googleapis/gax-java/issues/614 and waiting on their responses.

songy23:
Looks like C++ did this in a similar way to Go: https://github.com/census-instrumentation/opencensus-cpp/pull/234.

songy23:
Quotes from https://github.com/googleapis/gax-java/issues/614#issuecomment-439067683:

> As the comment says ("only merging with gRPC default value for User-Agent is permitted"), the user-agent value is merged with the grpc one.

It seems in Java the custom `user-agent` value works the same way as other languages (merged with the default gRPC value). I've also run a few examples testing the Stackdriver exporter with this change. Everything works fine, though I'm not sure if the `user-agent` value is propagated to Stackdriver backend.

Going to merge this PR first, will change to use `ClientSettings.WithUserAgent()` once GAX-Java team added support for that.

g-easy:
Please let me know if this change was sufficient for Stackdriver and also if the format is correct.

songy23:
> Please let me know if this change was sufficient for Stackdriver and also if the format is correct.

/cc @rakyll I remember you mentioned there's a way to verify the "user-agent" header in Stackdriver backend. Maybe we can chat offline to see if is can be applied to all languages.

PR #1527, opened 2018-10-31
Example: Update dependency versions.

songy23:
Otherwise there will be Guava version conflict when running the examples (https://github.com/census-instrumentation/opencensus-java/tree/master/exporters/stats/stackdriver#why-did-i-get-an-error-javalangnosuchmethoderror-comgooglecommon-like-javalangnosuchmethoderrorcomgooglecommonbasethrowablesthrowifinstanceof).


mayurkale22:
Please fix the build.

songy23:
CI is green, PTAL.

songy23:
Thanks! Also note that we don't need to do the same for Bazel because it doesn't bring in transitive dependencies.

PR #1528, opened 2018-10-31
Update opencensus-contrib-log-correlation-log4j2 to match log correlation spec.

sebright:
This commit updates the context key names to match the shorter names in
https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/LogCorrelation.md.
It also removes the warning about the library being experimental.


sebright:
Thanks!

PR #1536, opened 2018-11-02
Improve handling of null project ID in Stackdriver log correlation.

sebright:
This commit stores null instead of an empty string when the project ID can't be
found, which allows distinguishing between null and empty project IDs.

This commit also improves null safety by adding a Nullable annotation to
com.google.cloud.ServiceOptions.getDefaultProjectId(), which can return null
when no project ID is available.  I had to suppress three warnings and add TODOs
where the result was already assumed to be non-null.

_____________________________________________________________________________

I'm planning to open an issue to fix the TODOs after this PR is merged.

### DIFF
--- a/checker-framework/stubs/google-cloud-java.astub
+++ b/checker-framework/stubs/google-cloud-java.astub
@@ -1,0 +1,8 @@
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+package com.google.cloud;
+
+class ServiceOptions<ServiceT extends Service<OptionsT>, OptionsT extends ServiceOptions<ServiceT, OptionsT>> {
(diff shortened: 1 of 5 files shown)


sebright:
I thought about this some more, and I think it might be better to hide the `create` and `getProjectId` methods in the first release, since I'm not sure they're needed.  It's likely that most code that uses this class will only interact with it through the LoggingEnhancer interface.

songy23:
Should this be closed?

sebright:
I'm planning to update it so that it can be merged within a few days.

sebright:
Thanks!

sebright:
I opened #1547 for the TODOs.

PR #1538, opened 2018-11-02
Hide methods in OpenCensusTraceLoggingEnhancer that aren't necessary currently.

sebright:
Code that creates or uses OpenCensusTraceLoggingEnhancer probably won't know its
type, so it will only call the no-arg constructor and 'enhanceLogEntry'.  This
commit hides the other methods in the class, create(String) and getProjectId().
They can be re-exposed in the future if necessary.


songy23:
Superseded by https://github.com/census-instrumentation/opencensus-java/pull/1539.

PR #1539, opened 2018-11-02
 Update Javadoc "since" tags in log correlation libraries.

sebright:
0.17 is the first non-experimental release, and it has some breaking changes, so
so we should reset the "since" tags to 0.17.

This PR includes #1538.


sebright:
Thanks!

PR #1542, opened 2018-11-02
Remove min/max from Distribution.

bogdandrutu:


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 - The histogram bucket boundaries (`BucketBoundaries`) and values (`Count` and `Sum`) are no longer
   supported for negative values. The Record API drops the negative `value` and logs the warning.
   This could be a breaking change if you are recording negative value for any `measure`.
+- Remove support for min/max in the stats Distribution to make it compatible with Metrics.
```
mayurkale22: Can we also consider to change ```mean``` -> ```sum``` in stats Distribution to make it compatible with Metrics Distribution. ?

songy23: Please file an issue for that. It should be in a separate PR.

PR #1553, opened 2018-11-04
Helper utility for Http Tracing.

rghetia:
This is a split from PR https://github.com/census-instrumentation/opencensus-java/pull/1537.

### DIFF
--- a/contrib/http_util/src/main/java/io/opencensus/contrib/http/util/HttpTraceUtil.java
+++ b/contrib/http_util/src/main/java/io/opencensus/contrib/http/util/HttpTraceUtil.java
@@ -27,6 +27,34 @@ import javax.annotation.Nullable;
  */
 @ExperimentalApi
 public final class HttpTraceUtil {
+  private static final int STATUS_200 = 200;
+  private static final int STATUS_100 = 100;
+  private static final int STATUS_101 = 101;
+  private static final int STATUS_400 = 400;
+  private static final int STATUS_401 = 401;
+  private static final int STATUS_402 = 402;
+  private static final int STATUS_403 = 403;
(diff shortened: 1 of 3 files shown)


contrib/http_util/src/main/java/io/opencensus/contrib/http/util/HttpTraceUtil.java
```diff
@@ -0,0 +1,127 @@
...
+
+    // set status according to response
+    if (statusCode == 0) {
+      return Status.UNKNOWN.withDescription(message);
+    } else {
+      if (statusCode >= 200 && statusCode < 400) {
+        return Status.OK;
+      } else {
+        // error code, try parse it
+        switch (statusCode) {
```
bogdandrutu: Looks like 90% of the statuses can be constants. Can you please update this class to use constants when possible e.g.:

private static final Status STATUS_402 = Status.UNKNOWN.withDescription("Payment Required");

rghetia: done.

bogdandrutu: I think we miss-understood each other. I want you to avoid constructing a Status object when possible to avoid that not to use const codes.

rghetia: fixed it correctly.

rghetia: fixed.

PR #1554, opened 2018-11-05
Avoid one extra allocation in TraceId and SpanId.

bogdandrutu:
In a followup PR the LowerCaseBase16Encoding should be changed to work with longs to avoid more allocations (not doing that here to keep this PR simple).

### DIFF
--- a/api/src/main/java/io/opencensus/trace/TraceId.java
+++ b/api/src/main/java/io/opencensus/trace/TraceId.java
@@ -18,7 +18,6 @@ package io.opencensus.trace;
 
 import io.opencensus.common.Internal;
 import io.opencensus.internal.Utils;
-import java.util.Arrays;
 import java.util.Random;
 import javax.annotation.Nullable;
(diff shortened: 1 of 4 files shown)


api/src/main/java/io/opencensus/trace/TraceId.java
```diff
@@ -211,12 +213,13 @@ public boolean equals(@Nullable Object obj) {
...
     TraceId that = (TraceId) obj;
-    return Arrays.equals(bytes, that.bytes);
+    return idHi == that.idHi && idLo == that.idLo;
   }
 
   @Override
   public int hashCode() {
-    return Arrays.hashCode(bytes);
+    // TODO: Use Objects.hashCode when switch to java17.
+    return Arrays.hashCode(new Long[] {idHi, idLo});
```
rghetia: is it worth to avoid creating intermediate Long[]?

bogdandrutu: done

PR #1565, opened 2018-11-08
Resource: Add resource detector

mayurkale22:
This is as per specs: https://github.com/census-instrumentation/opencensus-specs/blob/master/resource/Resource.md#detectors

Also, I think ```ChainedDetector``` name is absolute and changed to ```MultiDetector```.


api/src/main/java/io/opencensus/resource/ResourceDetector.java
```diff
@@ -0,0 +1,41 @@
...
+package io.opencensus.resource;
+
+import io.opencensus.common.ExperimentalApi;
+import java.util.List;
+
+/*>>>
+import org.checkerframework.checker.nullness.qual.Nullable;
+*/
+
+/** Interface to detect resource information. */
```
songy23: nit: interface should also have `@since` tag.

api/src/main/java/io/opencensus/resource/ResourceDetector.java
```diff
@@ -0,0 +1,41 @@
...
+@ExperimentalApi
+public interface ResourceDetector {
+
+  /**
+   * Returns a {@link Resource} that runs all input resources sequentially and merges their results.
+   * In case a type of label key is already set, the first set value is takes precedence. If the
```
songy23: nit: is takes -> takes

impl_core/src/main/java/io/opencensus/implcore/resource/ResourceDetectorImpl.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+
+/*>>>
+import org.checkerframework.checker.nullness.qual.Nullable;
+*/
+
+/** Implementation of {@link ResourceDetector}. */
```
songy23: Same, `@since 0.18`.

mayurkale22: I think ```@since``` not required here, as this is internal implementation of ```ResourceDetector```.

api/src/main/java/io/opencensus/resource/ResourceDetector.java
mayurkale22: done

api/src/main/java/io/opencensus/resource/ResourceDetector.java
mayurkale22: done

mayurkale22:
Good point. I am not sure if we want to do reflective loading for resource API. Another approach(simple) I was thinking to add ```multiDetector``` method to resource class. 

Something like...
```
class Resource {
  Resource multiDetector(List</*@Nullable*/ Resource> resources) {
     // merge logic
  }
}
```

WDYT?

songy23:
> Another approach(simple) I was thinking to add multiDetector method to resource class.

Agree, this sounds simpler.

mayurkale22:
@bogdandrutu PTAL this PR.

mayurkale22:
@bogdandrutu Friendly ping...

mayurkale22:
@songy23 @bogdandrutu If no more comments, I would like to merge this PR today.

songy23:
SGTM

api/src/main/java/io/opencensus/resource/Resource.java
```diff
@@ -99,14 +104,34 @@ public static Resource create() {
...
+  /**
+   * Returns a {@link Resource} that runs all input resources sequentially and merges their results.
+   * In case a type of label key is already set, the first set value takes precedence.
+   *
+   * @param resources a list of resources.
+   * @return a {@code Resource}.
+   * @since 0.18
+   */
+  @javax.annotation.Nullable
+  public Resource multiDetector(List</*@Nullable*/ Resource> resources) {
```
bogdandrutu: Why do you need this public method?

api/src/main/java/io/opencensus/resource/Resource.java
```diff
@@ -154,14 +179,39 @@ static String parseResourceType(@Nullable String rawEnvType) {
...
+
+    String mergedType = resource.getType() != null ? resource.getType() : otherResource.getType();
+    Map<String, String> mergedLabelMap =
+        new LinkedHashMap<String, String>(otherResource.getLabels());
+
+    // Labels from resource overwrite labels from otherResource.
+    for (Entry<String, String> entry : resource.getLabels().entrySet()) {
+      mergedLabelMap.put(entry.getKey(), entry.getValue());
+    }
+    return new AutoValue_Resource(mergedType, Collections.unmodifiableMap(mergedLabelMap));
```
bogdandrutu: Why not createInternal?

api/src/main/java/io/opencensus/resource/Resource.java
```diff
@@ -116,8 +141,8 @@ private static Resource createInternal(@Nullable String type, Map<String, String
    * <p>OC_RESOURCE_TYPE: A string that describes the type of the resource prefixed by a domain
    * namespace, e.g. “kubernetes.io/container”.
    */
-  @Nullable
-  static String parseResourceType(@Nullable String rawEnvType) {
+  /*@Nullable*/
```
bogdandrutu: I am lost. When do we use the annotation and when do we use the comment?

songy23: Yes, for public APIs we should use `@javax.annotations.Nullable`. For variables and internal methods, the comment one `/*@Nullable*/` is fine.

mayurkale22:
@bogdandrutu PTAL

api/src/main/java/io/opencensus/resource/Resource.java
```diff
@@ -99,14 +104,34 @@ public static Resource create() {
    *     ASCII string or exceed {@link #MAX_LENGTH} characters.
    * @since 0.18
    */
-  public static Resource create(@Nullable String type, Map<String, String> labels) {
+  // TODO(mayurkale): Use this method while creating resource information from GCP, AWS etc.
```
bogdandrutu: This comment it is not necessary now that you explain where do you need this.

api/src/main/java/io/opencensus/resource/Resource.java
bogdandrutu: Maybe then call it "mergeResources"? or just "merge"?

api/src/main/java/io/opencensus/resource/Resource.java
bogdandrutu: What is the difference? Why are we not consistent everywhere?

api/src/main/java/io/opencensus/resource/Resource.java
bogdandrutu: Also why not have this as static method?

api/src/main/java/io/opencensus/resource/Resource.java
songy23: In the code most of the time they are the same except that `/*@Nullable*/` can be applied to generics. When building with Gradle, `@javax.annotations.Nullable` will be used by both error-prone and nullness checker, while `/*@Nullable*/` can only be used by nullness checker.

Generally speaking we can use `@javax.annotations.Nullable` whenever possible and only use `/*@Nullable*/` for generics. IMO this is beyond the scope of this PR though.

api/src/main/java/io/opencensus/resource/Resource.java
```diff
@@ -52,21 +57,21 @@
...
 
-  @Nullable
+  @javax.annotation.Nullable
   private static final String ENV_TYPE = parseResourceType(System.getenv(OC_RESOURCE_TYPE_ENV));
 
   private static final Map<String, String> ENV_LABEL_MAP =
       parseResourceLabels(System.getenv(OC_RESOURCE_LABELS_ENV));
 
-  Resource() {}
+  protected Resource() {}
```
bogdandrutu: No. Please keep it package protected.

api/src/main/java/io/opencensus/resource/Resource.java
mayurkale22: Done

api/src/main/java/io/opencensus/resource/Resource.java
mayurkale22: Done.

api/src/main/java/io/opencensus/resource/Resource.java
mayurkale22: Done

api/src/main/java/io/opencensus/resource/Resource.java
mayurkale22: Done

api/src/main/java/io/opencensus/resource/Resource.java
bogdandrutu: So here it should be the annotation not the comment.

PR #1482, opened 2018-10-06
Add message events and span kind in Jaeger.

bogdandrutu:

PR #1569, opened 2018-11-09
Build: Exclude oc-agent-exporter from nullness checker.

songy23:
OC-Agent exporter depends on grpc-core which depends on errorprone annotations, and it doesn't work with checker framework.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -26,6 +26,10 @@ def useCheckerFramework = rootProject.hasProperty('checkerFramework')
 def useErrorProne = !useCheckerFramework
 
 subprojects {
+    // OC-Agent exporter depends on grpc-core which depends on errorprone annotations, and it
+    // doesn't work with checker framework.
+    def mustUseErrorprone = project.name == "opencensus-exporter-trace-ocagent"
+
     apply plugin: "checkstyle"
     apply plugin: 'maven'
     apply plugin: 'idea'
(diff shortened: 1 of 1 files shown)


songy23:
Build failure is because of #1482 and fixed in https://github.com/census-instrumentation/opencensus-java/pull/1571.

songy23:
CI is green, PTAL. Need to merge this PR first to unblock other PRs.

build.gradle
```diff
@@ -26,6 +26,10 @@ def useCheckerFramework = rootProject.hasProperty('checkerFramework')
 def useErrorProne = !useCheckerFramework
 
 subprojects {
+    // OC-Agent exporter depends on grpc-core which depends on errorprone annotations, and it
+    // doesn't work with checker framework.
+    def mustUseErrorprone = project.name == "opencensus-exporter-trace-ocagent"
```
bogdandrutu: Why not change only the useErrorProne = !useCheckerFramework || project.name == ...?

songy23: For extensibility. Any artifacts that depend on grpc-core should use error-prone instead of nullness checker. In the future if we add project foo that depends on grpc-core, we can simply add it to this line:
```gradle
def mustUseErrorprone = project.name == "opencensus-exporter-trace-ocagent" || project.name == "any-artifact-with-grpc-core" || ...
```

songy23: Sorry ignore the previous comment. You're right there's no need to have another variable `mustUseErrorprone`.

PR #1570, opened 2018-11-09
Exporter/OcAgent: Add export RPC handler.

songy23:
Split from https://github.com/census-instrumentation/opencensus-java/pull/1492.

### DIFF
--- a/exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentTraceServiceExportRpcHandler.java
+++ b/exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentTraceServiceExportRpcHandler.java
@@ -63,7 +63,7 @@ final class OcAgentTraceServiceExportRpcHandler {
       StreamObserver<ExportTraceServiceRequest> exportRequestObserver =
           stub.export(exportResponseObserver);
       exportRpcHandler.setExportRequestObserver(exportRequestObserver);
-    } catch (Exception e) {
+    } catch (StatusRuntimeException e) {
       exportRpcHandler.onComplete(e);
     }
     return exportRpcHandler;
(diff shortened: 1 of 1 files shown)


songy23:
Depends on #1569.

exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentTraceServiceExportRpcHandler.java
```diff
@@ -0,0 +1,138 @@
...
+  // given TraceServiceStub.
+  static OcAgentTraceServiceExportRpcHandler create(TraceServiceStub stub) {
+    OcAgentTraceServiceExportRpcHandler exportRpcHandler =
+        new OcAgentTraceServiceExportRpcHandler();
+    ExportResponseObserver exportResponseObserver = new ExportResponseObserver(exportRpcHandler);
+    try {
+      StreamObserver<ExportTraceServiceRequest> exportRequestObserver =
+          stub.export(exportResponseObserver);
+      exportRpcHandler.setExportRequestObserver(exportRequestObserver);
+    } catch (Exception e) {
```
bogdandrutu: It is not good to catch all Exceptions. What exception do we catch here? Also instead of returning an object should we just throw the exception? That will imply that  "exportRequestObserver" is never null for example.

songy23: > It is not good to catch all Exceptions. What exception do we catch here?

Make sense, here we should only catch the gRPC `StatusRuntimeException`.

> Also instead of returning an object should we just throw the exception?

I don't think we need to throw the `StatusRuntimeException`, since we've stored the terminate status if we saw a `StatusRuntimeException` and the worker thread can make use of the status. Agree other runtime exceptions can be just thrown to the worker thread. 

> That will imply that "exportRequestObserver" is never null for example.

IMO that's less an issue. `exportRequestObserver` is only created when `exportHandler` is created, and when there's an exception, `exportHandler` will be GC'ed, so is the `exportRequestObserver`.

songy23:
SG, thanks!

PR #1572, opened 2018-11-09
Exporter/OcAgent: Add config RPC handler.

songy23:
Split from #1492.


songy23:
Depends on #1569.

songy23:
Thanks!

PR #1574, opened 2018-11-09
Exporter/OcAgent: Skeleton and workflow for worker thread.

songy23:
Split from https://github.com/census-instrumentation/opencensus-java/pull/1492.


songy23:
Friendly ping. I think we need to agree on the workflow before I move on to implementation.

songy23:
PTAL.

songy23:
Thanks!

PR #1577, opened 2018-11-12
Add Trace config options and extractor for Http Servlets.

rghetia:


contrib/http_servlet/src/main/java/io/opencensus/contrib/http/servlet/TraceConfigOptions.java
```diff
@@ -0,0 +1,114 @@
...
+/**
+ * This class provide configuring options for Tracing.
+ *
+ * @since 0.18
+ */
+@ExperimentalApi
+public class TraceConfigOptions {
+  private TextFormat propagator = Tracing.getPropagationComponent().getB3Format();
+  private static final TraceConfigOptions options = new TraceConfigOptions();
+  private Sampler sampler = Samplers.probabilitySampler(0.0001);
```
bogdandrutu: Probably better to make this default to the library default sampler.

contrib/http_servlet/src/main/java/io/opencensus/contrib/http/servlet/TraceConfigOptions.java
```diff
@@ -0,0 +1,114 @@
...
+import io.opencensus.trace.samplers.Samplers;
+
+/**
+ * This class provide configuring options for Tracing.
+ *
+ * @since 0.18
+ */
+@ExperimentalApi
+public class TraceConfigOptions {
+  private TextFormat propagator = Tracing.getPropagationComponent().getB3Format();
```
bogdandrutu: we should probably make this default to trace context (you can blame me on being lazy to finish that). Maybe add a TODO. We should not ship it with b3 because we will have to deal with deprecation and the whole process we do it in go

contrib/http_servlet/src/main/java/io/opencensus/contrib/http/servlet/TraceConfigOptions.java
```diff
@@ -0,0 +1,114 @@
...
+import io.opencensus.trace.propagation.TextFormat;
+import io.opencensus.trace.samplers.Samplers;
+
+/**
+ * This class provide configuring options for Tracing.
+ *
+ * @since 0.18
+ */
+@ExperimentalApi
+public class TraceConfigOptions {
```
bogdandrutu: Make this immutable and add a builder. Suggest to use autovalue for this class.

contrib/http_servlet/src/main/java/io/opencensus/contrib/http/servlet/OcHttpServletExtractor.java
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  @Override
+  public String getUrl(HttpServletRequest request) {
+    return request.getRequestURL().toString();
+  }
+
+  @Override
+  public String getRoute(HttpServletRequest request) {
+    return "";
```
bogdandrutu: how do we plan to allow users to set this?

contrib/http_servlet/src/main/java/io/opencensus/contrib/http/servlet/TraceConfigOptions.java
rghetia: as per our conversation I will remove this for now. Add the support later.

contrib/http_servlet/src/main/java/io/opencensus/contrib/http/servlet/TraceConfigOptions.java
rghetia: I am removing this file (see above) but I'll add the TODO in other relevant place(s).

contrib/http_servlet/src/main/java/io/opencensus/contrib/http/servlet/TraceConfigOptions.java
rghetia: ok.

PR #1580, opened 2018-11-14
Add Http servlet listener and filter.

rghetia:


rghetia:
refactor based on stats work.

PR #1581, opened 2018-11-14
Exporter/OcAgent: Use dedicated Resource proto instead of string map.

songy23:
See https://github.com/census-instrumentation/opencensus-proto/blob/master/src/opencensus/proto/agent/trace/v1/trace_service.proto#L81.

### DIFF
--- a/exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentNodeUtils.java
+++ b/exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentNodeUtils.java
@@ -29,13 +29,11 @@ import io.opencensus.proto.agent.common.v1.LibraryInfo.Language;
 import io.opencensus.proto.agent.common.v1.Node;
 import io.opencensus.proto.agent.common.v1.ProcessIdentifier;
 import io.opencensus.proto.agent.common.v1.ServiceInfo;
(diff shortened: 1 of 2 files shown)


exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentNodeUtils.java
```diff
@@ -112,72 +106,61 @@ static ServiceInfo getServiceInfo(String serviceName) {
...
+      return resourceProtoBuilder.build();
     }
   }
 
-  // Encodes the attributes of MonitoredResource into a comma-separated list of labels.
-  // For example "aws_account=account1,instance_id=instance1,region=us-east-2".
-  private static String getConcatenatedResourceLabels(MonitoredResource resource) {
-    StringBuilder resourceLabels = new StringBuilder();
+  // Puts the attributes of MonitoredResource to ResourceProto.
+  private static void putResourceLabels(
```
bogdandrutu: We should probably change that package the auto-discovery package to use the Resource but in a different PR.

songy23: Agree, filed https://github.com/census-instrumentation/opencensus-java/issues/1583 to keep track.

PR #1589, opened 2018-11-16
Deprecate registration of the span names for zpages.

bogdandrutu:


Anusien:
`SampledSpanStore#registerSpanNamesForCollection` says it's deprecated and we should use `EndSpanOptions#getSampleToLocalSpanStore()`. But `EndSpanOptions#getSampleToLocalSpanStore()` is marked `@Experimental` and says we should use `SampledSpanStore#registerSpanNamesForCollection`.

Which is accurate?

PR #1598, opened 2018-11-20
Use java8 for osx builds.

bogdandrutu:


bogdandrutu:
Looks like this fixes the osx build. I will submit after appveyor passes (which should not be related)

PR #1602, opened 2018-11-20
Exclude http-jetty-client and http-servlet artifact from release 0.18

rghetia:


contrib/http_jetty_client/build.gradle
```diff
@@ -9,8 +9,6 @@ description = 'OpenCensus Http Jetty Client Plugin'
     it.targetCompatibility = 1.6
 }
 
-def opencensusVersion = "0.18.0-SNAPSHOT"
```
songy23: This doesn't need to be excluded.

rghetia: it is unused.

PR #1604, opened 2018-11-21
update http default view as per the spec. fixes #1603

### LINKED ISSUE #1603: default http server view do not have correct tags
According to [spec](https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/HTTP.md#default-views-1) default view should have following tags.
http_route
http_status
http_method

Instead, [implementation](https://github.com/census-instrumentation/opencensus-java/blob/fdb5e9fb9406d0a04c7a4ccf1d8356862ba02ea3/contrib/http_util/src/main/java/io/opencensus/contrib/http/util/HttpViewConstants.java#L150) has 
http_path
http_method

rghetia:


contrib/http_util/src/main/java/io/opencensus/contrib/http/util/HttpMeasureConstants.java
```diff
@@ -171,4 +171,12 @@ private HttpMeasureConstants() {}
    * @since 0.13
    */
   public static final TagKey HTTP_SERVER_METHOD = TagKey.create("http_server_method");
+
+  /**
+   * {@link TagKey} for the server-side logical route, a pattern that matched the URL, of a handler
+   * that processed the request.
+   *
+   * @since 0.19
```
songy23: This should be 0.18 if we were to merge this PR today.

rghetia: I wasn't sure if it would be merged before the release. I want @bogdandrutu  to review this before I merge.

rghetia:
CLIENT_PATH and SERVER_PATH are still valid but just not in default views. User can still create custom view using those tags.

PR #1609, opened 2018-11-27
Fix bug: Dropwizard Meter should have non-null start Timestamp

mayurkale22:
startTimestamp: Must be non-null for cumulative points.


contrib/dropwizard/src/main/java/io/opencensus/contrib/dropwizard/DropWizardMetrics.java
```diff
@@ -162,7 +162,7 @@ private Metric collectMeter(String dropwizardName, Meter meter) {
         TimeSeries.createWithOnePoint(
             Collections.<LabelValue>emptyList(),
             Point.create(Value.longValue(meter.getCount()), clock.now()),
-            null);
+            cumulativeStartTimestamp);
```
bogdandrutu: we use these timestamp for all the metrics. We should have a cache map<string, CacheEntry> where CacheEntry is {MetricDescriptor, startTime}. Please add a TODO.

mayurkale22: Done

Issue #1599, opened 2018-11-20
100% cpu usage by OpenCensus.Disruptor-0


danielnorberg:
Please answer these questions before submitting a bug report.

### What version of OpenCensus are you using?
0.17.0

### What JVM are you using (`java -version`)?

java version "11.0.1" 2018-10-16 LTS
Java(TM) SE Runtime Environment 18.9 (build 11.0.1+13-LTS)
Java HotSpot(TM) 64-Bit Server VM 18.9 (build 11.0.1+13-LTS, mixed mode)

### What did you do?
Normal opencensus use - low volume workload.

### What did you expect to see?
Low cpu usage.

### What did you see instead?
The `OpenCensus.Disruptor-0` thread spinning and using a whole core even when the application is otherwise idle.

This manifested after we upgraded from Java 8 to Java 11.

songy23:
@danielnorberg @steveniemitz  I've made a new SNAPSHOT release which included https://github.com/census-instrumentation/opencensus-java/pull/1605. Can you try it and see if it resolved this issue?

The `SNAPSHOT` release is under https://oss.sonatype.org/content/repositories/snapshots/io/opencensus/. For example https://oss.sonatype.org/content/repositories/snapshots/io/opencensus/opencensus-api/0.18.0-SNAPSHOT/.

To access it, please add the following lines to ~/.m2/settings.xml: 
```
<profiles>
  <profile>
     <id>allow-snapshots</id>
        <activation><activeByDefault>true</activeByDefault></activation>
     <repositories>
       <repository>
         <id>snapshots-repo</id>
         <url>https://oss.sonatype.org/content/repositories/snapshots</url>
         <releases><enabled>false</enabled></releases>
         <snapshots><enabled>true</enabled></snapshots>
       </repository>
     </repositories>
   </profile>
</profiles>
```

Then you should be able to use `0.18.0-SNAPSHOT` as the OpenCensus version in your gradle/pom file.

danielnorberg:
With `opencensus-impl-0.18.0-20181121.224929-2.jar` the `OpenCensus.Disruptor-0` thread cpu usage has dropped down to around `%10` for me on java 11.

bogdandrutu:
Do you think for the moment until we do a better fix 4% is acceptable? If yes I can do a PR or you can send me a PR. Thanks

steveniemitz:
[...]

PR #1618, opened 2018-11-28
lock-free single thread wait strategy

danielnorberg:
An attempt to implement a disruptor wait strategy that mitigates https://github.com/census-instrumentation/opencensus-java/issues/1599 and incurs less synchronization overhead than the `BlockingWaitStrategy`.

### DIFF
--- a/impl/src/main/java/io/opencensus/impl/internal/LockFreeSingleConsumerBlockingWaitStrategy.java
+++ b/impl/src/main/java/io/opencensus/impl/internal/LockFreeSingleConsumerBlockingWaitStrategy.java
@@ -1,0 +1,63 @@
+package io.opencensus.impl.internal;
+
+import com.lmax.disruptor.AlertException;
+import com.lmax.disruptor.Sequence;
+import com.lmax.disruptor.SequenceBarrier;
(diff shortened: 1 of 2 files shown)


impl/src/main/java/io/opencensus/impl/internal/LockFreeSingleConsumerBlockingWaitStrategy.java
```diff
@@ -0,0 +1,63 @@
...
+        barrier.checkAlert();
+        LockSupport.park(this);
+        if (Thread.interrupted()) {
+          throw new InterruptedException();
+        }
+        unparked = false;
+      }
+    }
+
+    while ((availableSequence = dependentSequence.get()) < sequence) {
```
felixbarny: This busy-spins on dependent sequences, which could be quite expensive, right? But if there is only one event handler there are probably no dependant sequences 🤔

felixbarny:
I like the idea very much :) Did you consider contributing this to disruptor itself? Maybe, you'll get some more feedback there as well.

songy23:
Stale. Will be fixed in OpenTelemetry-Java.

PR #1629, opened 2018-12-04
Exporter/Metrics/OcAgent: Add MetricsProtoUtils.

songy23:
~~Blocked by https://github.com/census-instrumentation/opencensus-java/pull/1627.~~

Updates https://github.com/census-instrumentation/opencensus-java/issues/1567.
Metrics APIs and Metrics protos are one-to-one mapping.

PR #1637, opened 2018-12-11
Exporter/Metrics/OcAgent: Optimization for already-sent metrics.

songy23:
As a follow-up of https://github.com/census-instrumentation/opencensus-java/pull/1629.

For already-sent metrics, set name instead of metric descriptor and don't need to set bucket options.

Updates #1567.


songy23:
I see, though I think this is why we add the `metric.name` field in the proto:
```proto
// Defines a Metric which has one or more timeseries.
message Metric {
  // The descriptor of the Metric. This is an optimization for network wire
  // size, from data-model perspective a Metric contains always a
  // MetricDescriptor.
  oneof descriptor {
    // In case of a streaming RPC can be sent only the first time a metric is
    // reported to save network traffic.
    MetricDescriptor metric_descriptor = 1;

    // In case of a streaming RPC this can be sent for metrics that already
    // sent the MetricDescriptor once.
    string name = 2;
  }
...
}
```

songy23:
We haven't made a decision on https://github.com/census-instrumentation/opencensus-proto/issues/152 yet. Closing this PR for now.

PR #1639, opened 2018-12-13
Add Stats for generic http server.

rghetia:


contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpServerHandler.java
```diff
@@ -118,20 +137,58 @@ public Span handleStart(C carrier, Q request) {
...
-    spanEnd(span, response, error);
+  public void handleEnd(
+      HttpRequestContext context, Q request, @Nullable P response, @Nullable Throwable error) {
+    recordStats(context, request, response, error);
+    spanEnd(context.span, response, error);
+  }
+
+  private void recordStats(
+      HttpRequestContext context, Q request, @Nullable P response, @Nullable Throwable error) {
+    checkNotNull(context, "context");
```
bogdandrutu: This is not a public API, why do we do the checks here? I would suggest to move all the checks in the public API and all private APIs like recordStats and spanEnd should accept non-null values. This way we avoid duplicate checks and also we make the code cleaner.

rghetia: done.

rghetia:
Client side is fine.

Issue #1643, opened 2018-12-19
opencensus-contrib-http-servlet missing in Maven Central


dansiviter:
### What version of OpenCensus are you using?
0.18.0

### What JVM are you using (`java -version`)?
Any

### What did you do?
If possible, provide a recipe for reproducing the error.
Add `io.opencensus:opencensus-contrib-http-servlet:0.18.0` as a dependency.

### What did you expect to see?
The above resource downloaded from maven central

### What did you see instead?
Dependency doesn't exist and appears to have not been uploaded to Maven Central.

bogdandrutu:
This is cause by this:
https://github.com/census-instrumentation/opencensus-java/blob/master/build.gradle#L386

@rghetia any update when the package will be available?

rghetia:
I'll post the PR today.

songy23:
opencensus-contrib-http-servlet is included in 0.19.0 Release: https://github.com/census-instrumentation/opencensus-java/releases/tag/v0.19.0.

Artifacts will be available on Maven in a minute.

PR #1645, opened 2018-12-20
Add Http servlet listener and filter.

rghetia:


rghetia:
fixes #1643

contrib/http_servlet/build.gradle
```diff
@@ -10,12 +10,14 @@ description = 'OpenCensus Http Servlet Plugin'
 }
 
 def opencensusVersion = "0.18.0-SNAPSHOT"
```
bogdandrutu: shouldn't this be 0.19.0-SNAPSHOT?

bogdandrutu: I think you should do the same as you did for http util and change every 0.18.0 to 0.19.0 (including since)

rghetia: the above version is not used. I'll open another PR and remove it. There are no other since 0.18.0

rghetia: there is however reference to 0.18.0 in readme. I'll fix it.

PR #1650, opened 2019-01-04
add support for opencensus http servlet plugin in spring cloud framework

rghetia:


rghetia:
check the build failure.

contrib/spring_cloud/core/src/main/java/io/opencensus/contrib/springcloud/instrument/web/HttpServletFilter.java
```diff
@@ -0,0 +1,28 @@
...
+package io.opencensus.contrib.springcloud.instrument.web;
+
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import io.opencensus.contrib.http.servlet.OcHttpServletFilter;
+import javax.servlet.Filter;
+import org.springframework.core.annotation.Order;
+import org.springframework.stereotype.Component;
+
+@Component
+@Order(1)
```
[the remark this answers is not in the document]
rghetia: I used +10 to leave some room for other important filter (may be security related) to precede tracing/monitoring. +10 may be too much. do you have a recommendation here?

contrib/spring_cloud/core/src/main/java/io/opencensus/contrib/springcloud/instrument/web/HttpServletFilter.java
```diff
@@ -0,0 +1,29 @@
...
+import io.opencensus.contrib.http.servlet.OcHttpServletFilter;
+import io.opencensus.contrib.springcloud.autoconfig.OpenCensusAutoConfiguration;
+import javax.servlet.Filter;
+import org.springframework.core.annotation.Order;
+import org.springframework.stereotype.Component;
+
+@Component
+@Order(OpenCensusAutoConfiguration.TRACE_FILTER_ORDER)
+@SuppressFBWarnings("RI_REDUNDANT_INTERFACES")
+public class HttpServletFilter extends OcHttpServletFilter implements Filter {}
```
meltsufin: An alternative to using the Servet filter would be to use a Spring MVC interceptor and the `WebMvcConfigurer`. We did this for including the trace id in the logs. The same concept applies to this PR. Take a look in [our code here](https://github.com/spring-cloud/spring-cloud-gcp/blob/master/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/logging/TraceIdLoggingWebMvcInterceptor.java).

rghetia: thanks for the review. Is Servlet filter approach still ok? The reason I asked is the most of the work for servlet filter is done and usable outside of the spring frame work as well. But if there are advantages of doing it using MVC interceptor i am can certainly change it.

meltsufin: Sure, Servlet Filter is more generic.

contrib/spring_cloud/core/src/main/java/io/opencensus/contrib/springcloud/instrument/web/HttpServletFilter.java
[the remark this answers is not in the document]
dhaval24: @rghetia we had similar problem in Application insights and there was a discussion around it.

You may want to refer https://github.com/Microsoft/ApplicationInsights-Java/pull/766#issuecomment-433195169 for more relevant details.

in short there are some default springboot filters which are responsible for setting error codes, and get the correct timing. It would be very hard to get that details if the filter lies below that in the chain. Therefore HIGHEST_PRECEDENCE is needed imo. But, I am open if you find better alternatives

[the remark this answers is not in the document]
devinsba: I would recommend the exact same value that Springs Tracing uses for their filter, unless you expect them both to be there, then maybe an additional `+ 1`

```
static final int TRACING_FILTER_ORDER = Ordered.HIGHEST_PRECEDENCE + 5;
```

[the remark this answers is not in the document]
rghetia: SG. I'll make the change.

contrib/spring_cloud/core/src/main/java/io/opencensus/contrib/springcloud/autoconfig/OpenCensusProperties.java
```diff
@@ -0,0 +1,38 @@
...
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * Opencensus Tracing settings.
+ *
+ * @since 0.19.0
+ */
+@ConfigurationProperties("spring.opencensus")
+public class OpenCensusProperties {
```
dinooliva: Should we have separate properties for Tracing and Stats?

all/build.gradle
```diff
@@ -19,6 +19,7 @@ def subprojects = [
         project(':opencensus-contrib-log-correlation-stackdriver'),
         project(':opencensus-contrib-monitored-resource-util'),
         project(':opencensus-contrib-spring'),
+        project(':opencensus-contrib-spring-cloud-core'),
```
bogdandrutu: Can this be part of the opencensus-contrib-spring?

contrib/spring_cloud/core/src/main/java/io/opencensus/contrib/springcloud/autoconfig/OpenCensusAutoConfiguration.java
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * @since 0.19.0
+ */
+@Configuration
+@ComponentScan(basePackages = "io.opencensus")
+@ConditionalOnProperty(value = "spring.opencensus.enabled", matchIfMissing = true)
+@EnableConfigurationProperties(OpenCensusProperties.class)
+@ExperimentalApi
+public class OpenCensusAutoConfiguration {
+  public static final int TRACE_FILTER_ORDER = Ordered.HIGHEST_PRECEDENCE + 10;
```
bogdandrutu: Comment about what this field represents?

contrib/spring_cloud/core/src/main/java/io/opencensus/contrib/springcloud/instrument/web/HttpServletFilter.java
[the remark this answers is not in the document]
dhaval24: @devinsba yes your suggestion looks more apt.

contrib/spring_cloud/core/src/main/java/io/opencensus/contrib/springcloud/autoconfig/OpenCensusProperties.java
dinooliva: Any thoughts here?

contrib/spring/src/main/java/io/opencensus/contrib/spring/autoconfig/OpenCensusAutoConfiguration.java
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * {@link org.springframework.boot.autoconfigure.EnableAutoConfiguration Auto-configuration} to
+ * enable tracing using OpenCensus.
+ *
+ * @since 0.19.0
+ */
+@Configuration
+@ComponentScan(basePackages = "io.opencensus")
+@ConditionalOnProperty(value = "spring.opencensus.enabled", matchIfMissing = true)
```
devinsba: You should be careful to use the same property namespaces as spring projects. The spring teams did a bunch of work to consolidate their use of properties to mostly be under `spring.` so that the community could have other areas. I would suggest you make `opencensus` the top level namespace

rghetia: I'll change the name.

bogdandrutu:
What do we need to ship this?

rghetia:
> What do we need to ship this?

Need to add ability to specify propagator, extractor and publicEndpoint. 
Also needs ability to enable just one tracing or metrics but not both. In the current implementation it does both. This can probably wait.

dinooliva:
As discussed irl, agree that we don't need to separate controls for tracing and metrics.

rghetia:
@dinooliva and @bogdandrutu  PTAL.

PR #1651, opened 2019-01-07
Changing compareTo to compareToIgnoreCase.

Vizerai:
Changing compareTo to compareToIgnoreCase so that keys are sorted ignoring case.  Capitalized tag keys should not go before lower case tag keys, it should be based on the lexicographic order of the letter itself.


Vizerai:
Please test this.

PR #1652, opened 2019-01-07
Add trace exporter for Datadog.

hexedpackets:
Similar to the Instana exporter, this sends to an API that is intended to be run locally through the Datadog agent. The agent handles batching, authentication, sampling, and so on before communicating with the Datadog servers. 

AFAIK, the expected span format matches up pretty closely with OpenCensus's format. There are some fields with extra meaning which would otherwise just be attributes, which I've tried to map appropriately.

### DIFF
--- a/exporters/trace/datadog/src/main/java/io/opencensus/exporter/trace/datadog/DatadogExporterHandler.java
+++ b/exporters/trace/datadog/src/main/java/io/opencensus/exporter/trace/datadog/DatadogExporterHandler.java
@@ -1,0 +1,191 @@
+/*
+ * Copyright 2019, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.exporter.trace.datadog;
+
+import com.google.gson.FieldNamingPolicy;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import io.opencensus.common.Functions;
+import io.opencensus.common.Scope;
+import io.opencensus.common.Timestamp;
+import io.opencensus.trace.AttributeValue;
+import io.opencensus.trace.Sampler;
+import io.opencensus.trace.SpanContext;
+import io.opencensus.trace.SpanId;
+import io.opencensus.trace.Status;
+import io.opencensus.trace.Tracer;
+import io.opencensus.trace.Tracing;
+import io.opencensus.trace.export.SpanData;
+import io.opencensus.trace.export.SpanExporter;
+import io.opencensus.trace.samplers.Samplers;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.nio.charset.Charset;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+
+final class DatadogExporterHandler extends SpanExporter.Handler {
+
+  private static final Tracer tracer = Tracing.getTracer();
+  private static final Sampler probabilitySpampler = Samplers.probabilitySampler(0.0001);
+  private static Gson gson =
+      new GsonBuilder()
+          .setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES)
+          .create();
+
+  private final URL agentEndpoint;
+  private final String service;
+  private final String type;
+
+  DatadogExporterHandler(final URL agentEndpoint, final String service, final String type) {
+    this.agentEndpoint = agentEndpoint;
+    this.service = service;
+    this.type = type;
+  }
+
+  @javax.annotation.Nullable
+  private static String attributeValueToString(AttributeValue attributeValue) {
+    return attributeValue.match(
+        Functions.returnToString(),
+        Functions.returnToString(),
+        Functions.returnToString(),
+        Functions.returnToString(),
+        Functions.returnNull());
+  }
+
+  private static Map<String, String> attributesToMeta(
+      final Map<String, AttributeValue> attributes) {
+    final HashMap<String, String> result = new HashMap<>();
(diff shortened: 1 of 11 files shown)


exporters/trace/datadog/src/main/java/io/opencensus/exporter/trace/datadog/DatadogExporterHandler.java
```diff
@@ -0,0 +1,191 @@
...
+              sc.getTraceId().getLowerLong(),
+              convertSpanId(sc.getSpanId()),
+              sd.getName(),
+              resource,
+              this.service,
+              this.type,
+              startTime,
+              endTime - startTime,
+              parentId,
+              errorCode(sd.getStatus()),
```
songy23: ditto

exporters/trace/datadog/src/main/java/io/opencensus/exporter/trace/datadog/DatadogSpan.java
```diff
@@ -0,0 +1,79 @@
...
+  // Required. The service you are tracing. The service name must not be longer than 100 characters.
+  private String service;
+  // Required. The type of request.
+  private String type;
+  // Required. The start time of the request in nanoseconds from the unix epoch.
+  private long start;
+  // Required. The duration of the request in nanoseconds.
+  private long duration;
+  // Optional. The span integer ID of the parent span.
+  private Long parentId;
```
songy23: Is this `Nullable`?

exporters/trace/datadog/src/main/java/io/opencensus/exporter/trace/datadog/DatadogTraceExporter.java
```diff
@@ -0,0 +1,86 @@
...
+   * Creates and registers the Datadog Trace exporter to the OpenCensus library. Only one Datadog
+   * exporter can be registered at any point.
+   *
+   * @param agentEndpoint URL for the Datadog agent.
+   * @param service Name of the service being traced.
+   * @param type Type of service being traced, eg web or db.
+   * @since 0.19
+   */
+  public static void createAndRegister(
+      final String agentEndpoint, final String service, final String type)
```
songy23: Consider using an `Option` class to construct the exporter. Similar to [StackdriverTraceConfiguration](https://github.com/census-instrumentation/opencensus-java/blob/master/exporters/trace/stackdriver/src/main/java/io/opencensus/exporter/trace/stackdriver/StackdriverTraceConfiguration.java).

exporters/trace/datadog/src/main/java/io/opencensus/exporter/trace/datadog/DatadogExporterHandler.java
```diff
@@ -0,0 +1,191 @@
...
+          new DatadogSpan(
+              sc.getTraceId().getLowerLong(),
+              convertSpanId(sc.getSpanId()),
+              sd.getName(),
+              resource,
+              this.service,
+              this.type,
+              startTime,
+              endTime - startTime,
+              parentId,
```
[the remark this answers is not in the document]
hexedpackets: I marked the optional JSON fields as `Nullable`. The current behavior is that null values will be excluded from the generated JSON.

exporters/trace/datadog/src/main/java/io/opencensus/exporter/trace/datadog/DatadogExporterHandler.java
```diff
@@ -0,0 +1,191 @@
...
+    return 1;
+  }
+
+  String convertToJson(Collection<SpanData> spanDataList) {
+    final ArrayList<DatadogSpan> datadogSpans = new ArrayList<>();
+    for (SpanData sd : spanDataList) {
+      SpanContext sc = sd.getContext();
+
+      final long startTime = timestampToNanos(sd.getStartTimestamp());
+      final long endTime = timestampToNanos(sd.getEndTimestamp());
```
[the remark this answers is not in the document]
hexedpackets: This is sort of tricky, the API requires a duration to be sent as part of the span. For now a null endTimestamp is mapped to `0L`, which produces a nonsense value for duration. 

Other options I can think of are to drop those spans or use the current system time, but I don't fully understand when a span would be exported without an endtime set. Do you have an opinion about the right way to handle those?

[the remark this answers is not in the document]
songy23: We do export sampled running spans. Since those spans are active, they don't have an end timestamp. For now I'd suggest to use current time as the end timestamp for running spans.

exporters/trace/datadog/src/main/java/io/opencensus/exporter/trace/datadog/DatadogTraceExporter.java
```diff
@@ -0,0 +1,87 @@
...
+import javax.annotation.concurrent.GuardedBy;
+
+/**
+ * An OpenCensus span exporter implementation which exports data to Datadog.
+ *
+ * <p>Example of usage:
+ *
+ * <pre>{@code
+ * public static void main(String[] args) {
+ *   DatadogTraceExporter.createAndRegister("http://localhost:8126/v0.3/traces", "myService", "web");
```
songy23: nit: please also update the code sample here.

exporters/trace/datadog/src/main/java/io/opencensus/exporter/trace/datadog/DatadogTraceExporter.java
```diff
@@ -0,0 +1,87 @@
...
+      throws MalformedURLException {
+    synchronized (monitor) {
+      checkState(handler == null, "Datadog exporter is already registered.");
+
+      String agentEndpoint = configuration.getAgentEndpoint();
+      String service = configuration.getService();
+      String type = configuration.getType();
+
+      handler = new DatadogExporterHandler(agentEndpoint, service, type);
+      Tracing.getExportComponent().getSpanExporter().registerHandler(REGISTER_NAME, handler);
```
songy23: Looks like the checker framework is still complaining about nullness here. Could you try adding a null check here? e.g if (handler != null) { ...registerHandler... }

hexedpackets: Doing a null check caused my IDE to complain that its always true. I copied what [Instana did](https://github.com/census-instrumentation/opencensus-java/blob/master/exporters/trace/instana/src/main/java/io/opencensus/exporter/trace/instana/InstanaTraceExporter.java#L67) with using an extra variable.

settings.gradle
```diff
@@ -65,6 +66,8 @@ project(':opencensus-exporter-stats-stackdriver').projectDir =
         "$rootDir/exporters/stats/stackdriver" as File
 project(':opencensus-exporter-stats-prometheus').projectDir =
         "$rootDir/exporters/stats/prometheus" as File
+project(':opencensus-exporter-trace-datadog').projectDir =
```
songy23: Please move this to "Java8Compatible" projects (https://github.com/census-instrumentation/opencensus-java/pull/1652/files#diff-88b7c47e47b8ee65263b6784b86fa0a7R84).

tylerbenson:
FYI, You have a couple of places where you still reference instana and "stan" in here...

Issue #1654, opened 2019-01-14
HttpServerHandler uses canonical status for http_status tag


langecode:
### What version of OpenCensus are you using?
0.19.0-SNAPSHOT

### What JVM are you using (`java -version`)?
```
java version "9"
Java(TM) SE Runtime Environment (build 9+181)
Java HotSpot(TM) 64-Bit Server VM (build 9+181, mixed mode)
```

### What did you do?
Testing out the functionality provided by `HttpServerHandler` it seems that the tags added when recording statistics is not following the [HTTP Stats](https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/HTTP.md) documentation. Specifically the value of the `http_server_status` is generated from the `toString()` method of `Status` class which is the canonical status.

The [HTTP Stats](https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/HTTP.md) specification says 

> http_server_status | HTTP server status code returned, as an integer e.g. 200, 404, 500.

Looking at the output from the Prometheus exporter the current implementation yields, e.g.:
```
opencensus_io_http_server_received_bytes_bucket{http_server_method="GET",http_server_route="healthz",http_server_status="Status{canonicalCode=OK, description=null}",le="1024.0",} 1.0
```

Which I do not expect is the intention.

PR #1655, opened 2019-01-14
http_server_status tag should hold http status code

langecode:
Changing `HttpServerHandler` to follow [HTTP Stats](https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/HTTP.md) such that the `http_server_status` will contain the http status code - not the OpenCensus canonical status.

See issue #1654


rghetia:
It should correspondingly change for client side as well.

langecode:
Did not look at the client handler but I guess that would be the same thing.

PR #1670, opened 2019-01-17
Update versions for guava, grpc, google cloud

bogdandrutu:


build.gradle
```diff
@@ -171,11 +171,11 @@ subprojects {
         findBugsAnnotationsVersion = '3.0.1'
         findBugsJsr305Version = '3.0.2'
         errorProneVersion = '2.3.2'
-        grpcVersion = '1.16.1'
-        guavaVersion = '20.0'
+        grpcVersion = '1.18.0'
+        guavaVersion = '25.1-android'
```
songy23: Why android?

bogdandrutu: https://github.com/google/guava

Guava comes in two flavors.

The JRE flavor requires JDK 1.8 or higher.
If you need support for JDK 1.7 or Android, use the Android flavor. You can find the Android Guava source in the android directory.

PR #1687, opened 2019-01-25
Update deps and exclude opencensus artifacts from dependencies.

bogdandrutu:


contrib/log_correlation/stackdriver/build.gradle
```diff
@@ -3,8 +3,12 @@ description = 'OpenCensus Stackdriver Log Correlation'
 apply plugin: 'java'
 
 dependencies {
-    compile project(':opencensus-api'),
-            libraries.google_cloud_logging
+    compile project(':opencensus-api')
+
+    compile (libraries.google_cloud_logging) {
+        // We will always be more up to date.
+        exclude group: 'io.opencensus', module: 'opencensus-api'
```
dinooliva: So, iiuc, this will replace the 'opencensus-api' dependency in libraries.google_cloud_logging with the locally compiled version (':opencensus-api'), correct?

bogdandrutu: Yes.

PR #1690, opened 2019-01-25
Remove build for java7.

bogdandrutu:
This is also what gRPC does for their tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,10 +5,6 @@ language: java
 matrix:
   fast_finish: true
   include:
-  - jdk: openjdk7
-    env: TASK=BUILD
-    os: linux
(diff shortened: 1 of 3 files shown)


bogdandrutu:
@dinooliva the problem is that more and more tools like errorprone/auto-value they do not work on java7 and that's why we have complicated logic in the build scripts or we avoid upgrading the dependency for auto-value. The generated code is still java7 compatible ensured by the animalsniffer but we generate the code using the java8. If we want to be more paranoid (grpc does not do this) we can build on java8 and run tests on java7 if really needed (I don't think this is needed).

dinooliva:
Understood, thanks.

PR #1691, opened 2019-01-25
Remove format and licence tasks for tests and move to gradle.

bogdandrutu:


rghetia:
> @rghetia how can I get rid of the kokoro "Check Example License" and "Check Example Format"?

There is a config in inside google that needs to be updated. Do you want me remove that first?

rghetia:
@bogdandrutu I sent a PR to disable the build from inside google.

bogdandrutu:
I need an approval for this before we disable the tasks.

Issue #1666, opened 2019-01-17
Tags: Add support for TagMetadata


songy23:
Spec is https://github.com/census-instrumentation/opencensus-specs/blob/master/tags/TagMap.md#tagscope.

PR #1696, opened 2019-01-28
[DO NOT MERGE] Tags: Add support for TagScope.

songy23:
Fixes https://github.com/census-instrumentation/opencensus-java/issues/1666.

Spec is https://github.com/census-instrumentation/opencensus-specs/blob/master/tags/TagMap.md#tagscope.

Note that this is a breaking change because from now on `TagContext` will be non-propagating by default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ## Unreleased
+- Add support for `TagScope`. This is a breaking change because from now on `TagKey`s will be non-propagating by default.
 - Add OpenCensus Java OC-Agent Trace Exporter.
 - Add OpenCensus Java OC-Agent Metrics Exporter.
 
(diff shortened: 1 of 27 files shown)


bogdandrutu:
TagScope belongs to a Tag entry not to the entire TagMap(TagContext). I think this PR needs more work. Let's not wait for this PR and move forward with the release.

songy23:
> Let's not wait for this PR and move forward with the release.

SGTM. Mark as do not merge for now.

songy23:
Updated to apply tag scope per tag key. PTAL.

songy23:
Chatted offline, holding off this PR now. We need to make it clear whether the scope applies to `Tag` or `TagKey`.

songy23:
Going to submit a new PR with `TagMetadata` as per https://github.com/census-instrumentation/opencensus-specs/pull/233/. Closing this PR for now, since there will be a lot of changes.

PR #1699, opened 2019-01-28
create scope for servlet request.

rghetia:


bogdandrutu:
From multiple perspectives this could be very wrong, starting a scope on a thread and then trying to close on a different thread will cause undefined behaviors (or more explicitly bad behaviors) for the context. Unless we can guarantee that the scope is created and closed on the same thread this PR needs to be rolled back.

Issue #1684, opened 2019-01-24
creating a span with an explicit trace id and no parent


dmichel1:
Hi,

I'm currently trying to add tracing to an internal java framework. The framework has a request decorator that allows me to inject a trace header based on span context I am creating myself...

```
    final SpanContext spanContext = SpanContext.create(
      TraceId.generateRandomId(random),
      SpanId.generateRandomId(random),
      TraceOptions.DEFAULT,
      Tracestate.builder().build()
    );

    Map<String, String> traceHeaders = new HashMap<>();
    HttpPropagationUtil.getCloudTraceFormat()
      .inject(
        spanContext,
        traceHeaders,
        new TextFormat.Setter<Map<String, String>>() {
          @Override
          public void put(final Map<String, String> map, final String key, final String value) {
            map.put(key, value);
          }
        });

    traceHeaders.forEach((k, v) -> messageBuilder.setUserHeader(k, ByteString.copyFromUtf8(v)));
```

This part works fine but I run into an issue when it comes to starting the span based on this context within a handler.

Using ` tracer.spanBuilderWithRemoteParent("service-A", spanContext).startSpan()` implies that the context is from a parent span.

The current API doesn't seem to have any method of creating a span with an explicit traceId/spanContext that was created but never started **and** with no parent. 

Is this something we could add to the API?

Thanks

dmichel1:
I found an alternative way to do this in my code.

PR #1702, opened 2019-01-29
WIP - create a span from explicit span context

dmichel1:
Took a shot at implementing https://github.com/census-instrumentation/opencensus-java/issues/1684 - before I refactor `SpanBuilderImpl` too much what do you think of the general idea?

Thanks

### DIFF
--- a/api/src/main/java/io/opencensus/trace/SpanBuilder.java
+++ b/api/src/main/java/io/opencensus/trace/SpanBuilder.java
@@ -324,6 +324,10 @@ public abstract class SpanBuilder {
       return new NoopSpanBuilder(spanName);
     }
 
+    static NoopSpanBuilder createFromContext(String spanName, @Nullable SpanContext context) {
+      return new NoopSpanBuilder(spanName);
+    }
+
     @Override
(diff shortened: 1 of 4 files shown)


bogdandrutu:
Hi @dmichel1,

I have a different opinion on this, what about we change the spanBuilderWithRemoteParent and if the SpanContext contains a valid TraceId we preserve that?

dmichel1:
hi @bogdandrutu -

That would work but currently the 2nd argument for spanBuilderWithRemoteParent is a Span. I havn't been able to find a way with the current API to build a span with an explicit context. Or where you thinking of changing the method signature of spanBuilderWithRemoteParent to something else?

bogdandrutu:
public abstract SpanBuilder spanBuilderWithRemoteParent(
      String spanName, @Nullable SpanContext remoteParentSpanContext);

It is a SpanContext

dmichel1:
I was getting mixed up with `spanBuilderWithExplicitParent` my mistake. I just got something working locally. This is a better approach than this PR. Closing and will open a different one.

Issue #148, opened 2017-03-18
Remove Bazel build files, and only build with Maven.


sebright:
Maintaining two builds is time-consuming.  Since Bazel currently doesn't support deploying to Maven Central or running different static analysis tools, we should use Maven for now.  We will need to fix #147 and possibly add Error Prone to the Maven build first.

sebright:
#170 removed Bazel support.

PR #1703, opened 2019-01-29
Proposal: Remove Bazel support for examples.

songy23:
Similar rationale to https://github.com/census-instrumentation/opencensus-java/issues/148.

Reasons:
- Bazel is not good at resolving transitive dependencies, nor is it popular in the open source world. [The example BUILD file for Bazel](https://github.com/census-instrumentation/opencensus-java/blob/master/examples/BUILD.bazel) is extremely complicated and fragile.
- For each new release, we have to do [a lot of additional manual work to update the Bazel build files](https://github.com/census-instrumentation/opencensus-java/blob/master/RELEASING.md#update-release-versions-in-documentations-and-build-files), and even the migration tool we used to use is now [deprecated](https://github.com/bazelbuild/migration-tooling#deprecated). 
- In the past we have run into several issues with building with Bazel on CI (e.g https://github.com/census-instrumentation/opencensus-java/issues/1228), and most of them have nothing to do with our code changes. 
- In addition, on https://opencensus.io/ all the Java examples are built with either Gradle or Maven - none of the examples we put there use Bazel.

Given the much effort every time when we update the Bazel files and it's not even recommended to our users, I think we're at a point that it's not worth keeping the Bazel build around. Supporting Gradle and Maven for examples should be enough.

WDYT?


songy23:
Thanks - am going to submit this PR by EOD today if there're no further concerns/comments.

Note that this PR only removes the Bazel build for the examples. Examples is kind of an independent artifact and has little to do with the OC-Java library itself. If users built with Bazel and wanted to depend on OC-Java, they're still able to do so (https://github.com/census-instrumentation/opencensus-java#add-the-dependencies-to-your-project-1).

PR #1714, opened 2019-01-30
Provide creating scope for async request http-servlet.

rghetia:


bogdandrutu:
Consider to add a comment in the readme.

rghetia:
@bogdandrutu PTAL again.
Appveyor build failure is not related to the change.

bogdandrutu:
Unfortunately this cannot be in 0.19.1 because it adds a new API, I think we should make 0.19.1 now with the rollback PR and this will get to 0.20.0

rghetia:
> Unfortunately this cannot be in 0.19.1 because it adds a new API, I think we should make 0.19.1 now with the rollback PR and this will get to 0.20.0

SG.

Issue #1590, opened 2018-11-16
Deprecate monitoredresource.util, use resource.util instead.


mayurkale22:
The "monitored" prefix is the Stackdriver specific name.

PR #1718, opened 2019-01-30
Deprecate MonitoredResource and change to use Resource.

bogdandrutu:


mayurkale22:
Can we combine this with https://github.com/census-instrumentation/opencensus-java/issues/1590?

bogdandrutu:
@mayurkale22 feel free to push a PR after this 👍 (seriously already large changes, I was actually trying to understand how to split, put I think is simple enough)

songy23:
Please fix the import.

bogdandrutu:
@songy23 finally working.

contrib/monitored_resource_util/src/main/java/io/opencensus/contrib/monitoredresource/util/MonitoredResource.java
```diff
@@ -87,23 +69,29 @@ public ResourceType getResourceType() {
      * @return the AWS account ID.
      * @since 0.13
      */
-    public abstract String getAccount();
+    public String getAccount() {
+      return firstNonNull(awsInstanceLabels.get(AwsEc2InstanceResource.ACCOUNT_ID_KEY), "");
```
songy23: `firstNonNull here is not needed because it's already called: https://github.com/census-instrumentation/opencensus-java/pull/1718/files#diff-e0f6d1e015f871479d844cca09da8bc7R34. Same for all others in this class.

contrib/monitored_resource_util/src/main/java/io/opencensus/contrib/monitoredresource/util/K8sContainerResource.java
```diff
@@ -0,0 +1,100 @@
...
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+/**
+ * Helper class for K8S container {@code Resource}.
+ *
+ * @since 0.20
+ */
+public class K8sContainerResource {
```
songy23: So `project`, `instance-id` and `zone` do not apply to the k8s resource?

contrib/monitored_resource_util/src/main/java/io/opencensus/contrib/monitoredresource/util/MonitoredResource.java
songy23: Ignore this comment - just saw this is the key not the value.

contrib/monitored_resource_util/src/main/java/io/opencensus/contrib/monitoredresource/util/K8sContainerResource.java
bogdandrutu: They are not related to k8s resource, but when we do autodetection if the pod/container runs on a GCE instance they will be added in the resource merging.

songy23: OK, make sense.

PR #1723, opened 2019-01-31
Use explicit context with spanBuilderWithRemoteParent

dmichel1:
#1684 

@bogdandrutu what do you think of something like this... I'm not totally happy with the additional conditional.


rghetia:
@dmichel1 what was the reason you closed this? Do you still need the feature in #1684?

dmichel1:
@rghetia sorry my github notifications were going to my junk folder for some reason. I found an alternate approach to getting the span.

PR #1731, opened 2019-02-01
Build: Upgrade AutoValue version and resolve a TODO.

### LINKED ISSUE #1017: Upgrade AutoValue to 1.5.3 or later.

songy23:
Fixes #1017.

We removed Java 7 build in https://github.com/census-instrumentation/opencensus-java/pull/1690. Now it's safe to upgrade to a newer version of AutoValue that works better with Java 9.

### DIFF
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -13,3 +13,10 @@ javadoc {
     exclude 'io/opencensus/internal/**'
     exclude 'io/opencensus/trace/internal/**'
 }
+
+animalsniffer {
+    sourceSets = [
+            sourceSets.main,
+            sourceSets.test
+    ]
+}
--- a/build.gradle
+++ b/build.gradle
@@ -118,8 +118,7 @@ subprojects {
         // Protobuf-generated code produces some warnings.
         // https://github.com/google/protobuf/issues/2718
         it.options.compilerArgs += ["-Xlint:-cast"]
-        if (!JavaVersion.current().isJava9() && !useErrorProne) {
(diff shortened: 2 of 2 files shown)


bogdandrutu:
Not 100% sure if this is android level 14 compatible. How can we test?

api/build.gradle
```diff
@@ -13,3 +13,10 @@ javadoc {
     exclude 'io/opencensus/internal/**'
     exclude 'io/opencensus/trace/internal/**'
 }
+
+animalsniffer {
```
bogdandrutu: Not sure that this will include the auto-value classes.

songy23: Copied this from gRPC-Java: https://github.com/grpc/grpc-java/blob/master/core/build.gradle#L49-L55.  /cc @zhangkun83 any suggestion on how to enforce `animalsniffer` on generated classes?

songy23: /cc @ericgribkoff too - looks like `animalsniffer` to Android 14 is added to gRPC in https://github.com/grpc/grpc-java/pull/4727. 

Any help is high appreciated!

songy23:
Closing, needs to have a better understanding on how this would affect gRPC before we can move forward.

PR #1738, opened 2019-02-06
 Deprecated measures use instance of new measures when 1to1 mapping.

bogdandrutu:
Intentionally split this into 2 commits to make it easy to see the changes, first commit just moves the code of deprecated measures after the new measures, second commit actually does the main change.

This fixes an issue with registerAllGrpcBasicViews because we had 2 measures (rountrip_latency) with different descriptions and this will fail here https://github.com/census-instrumentation/opencensus-java/blob/master/impl_core/src/main/java/io/opencensus/implcore/stats/MeasureToViewMap.java#L115

This is a breaking change in regard to the measure name/description for the old measures, but nobody should depend on that.

### DIFF
--- a/contrib/grpc_metrics/src/main/java/io/opencensus/contrib/grpc/metrics/RpcViews.java
+++ b/contrib/grpc_metrics/src/main/java/io/opencensus/contrib/grpc/metrics/RpcViews.java
@@ -127,7 +127,7 @@ public final class RpcViews {
           RpcViewConstants.RPC_SERVER_FINISHED_COUNT_HOUR_VIEW);
 
   @VisibleForTesting
-  static final ImmutableSet<View> RPC_REAL_TIME_METRICS_VIEWS_SET =
+  static final ImmutableSet<View> GRPC_REAL_TIME_METRICS_VIEWS_SET =
       ImmutableSet.of(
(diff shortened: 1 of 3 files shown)


bogdandrutu:
To see what changed look at this https://github.com/census-instrumentation/opencensus-java/pull/1738/commits/babf3d38df19696017d54507fabb5f5e4d9fb87c or the second commit.

bogdandrutu:
One more change - removed the old views from the grpc basic views. This method never worked because of the duplicated measure name so this is not a problem.

contrib/grpc_metrics/src/main/java/io/opencensus/contrib/grpc/metrics/RpcViews.java
```diff
@@ -127,7 +127,7 @@
           RpcViewConstants.RPC_SERVER_FINISHED_COUNT_HOUR_VIEW);
 
   @VisibleForTesting
-  static final ImmutableSet<View> RPC_REAL_TIME_METRICS_VIEWS_SET =
+  static final ImmutableSet<View> GRPC_REAL_TIME_METRICS_VIEWS_SET =
```
songy23: Please fix the tests.

PR #1741, opened 2019-02-06
Clarify that realtime views are only for streaming rpcs

bogdandrutu:


songy23:
Do you want to include this commit in the patch release?

bogdandrutu:
It will be good.

PR #1747, opened 2019-02-07
Remove bazel instructions since we don't update them

bogdandrutu:


bogdandrutu:
I don't think we do that consistently, also bazel updated their ways to access maven jars, etc. and without an example where we run this config (which example was removed) I don't feel comfortable to keep this.

songy23:
SGTM

PR #1754, opened 2019-02-08
add option for extractor and propagator for jetty client.

rghetia:


contrib/http_jetty_client/src/main/java/io/opencensus/contrib/http/jetty/client/OcJettyHttpClient.java
```diff
@@ -45,19 +48,41 @@ public void put(Request carrier, String key, String value) {
       };
 
   private static final Tracer tracer = Tracing.getTracer();
-  private final HttpClientHandler<Request, Response, Request> handler;
+  final HttpClientHandler<Request, Response, Request> handler;
```
dinooliva: So you've made this package-private for testing - is there an @VisibleForTesting annotation that we use in OC?

rghetia: added the annotation.

Issue #1758, opened 2019-02-13
Stats: Add API that supports recording with additional tags.


songy23:
Currently if users want to record against additional tags, they need to create a new `TagContext` like:
```java
TagContext ctx = tagger.currentBuilder().put(key1, value1).build();
measureMap.record(ctx);
```

Consider providing a convenient API to record with additional tags without creating a `TagContext` e.g
```java
measureMap.recordWithTags(Arrays.asList(tag1, tag2));
```

[Counterpart in Go](https://github.com/census-instrumentation/opencensus-go/blob/fa1e35e3ffe868399138d03aaf2e949d52884f1b/stats/record.go#L57-L69):
```go
// RecordWithTags records one or multiple measurements at once.
//
// Measurements will be tagged with the tags in the context mutated by the mutators.
// RecordWithTags is useful if you want to record with tag mutations but don't want
// to propagate the mutations in the context.
func RecordWithTags(ctx context.Context, mutators []tag.Mutator, ms ...Measurement) error {
	ctx, err := tag.New(ctx, mutators...)
	if err != nil {
		return err
	}
	Record(ctx, ms...)
	return nil
}
```

PR #1759, opened 2019-02-13
Stats: Add MeasureMap.recordWithTags.

songy23:
Fixes https://github.com/census-instrumentation/opencensus-java/issues/1758.

### DIFF
--- a/api/src/main/java/io/opencensus/stats/MeasureMap.java
+++ b/api/src/main/java/io/opencensus/stats/MeasureMap.java
@@ -19,7 +19,9 @@ package io.opencensus.stats;
 import io.opencensus.internal.Utils;
 import io.opencensus.stats.Measure.MeasureDouble;
 import io.opencensus.stats.Measure.MeasureLong;
+import io.opencensus.tags.Tag;
 import io.opencensus.tags.TagContext;
+import java.util.List;
 import javax.annotation.concurrent.NotThreadSafe;
 
 /**
@@ -89,4 +91,26 @@ public abstract class MeasureMap {
    * @since 0.8
    */
   public abstract void record(TagContext tags);
+
+  /**
+   * Records all of the measures at the same time, with the current {@link TagContext} plus the
+   * additional {@link Tag}s.
+   *
+   * <p>This method records all of the stats in the {@code MeasureMap} every time it is called.
+   *
+   * <p>This is equivalent to calling:
+   *
+   * <pre>{@code
+   * TagContext context = tagger.currentBuilder().put(tagKey, tagValue).build();
+   * measureMap.record(context);
+   * }</pre>
+   *
+   * @param tags additional {@link Tag}s to be recorded against.
+   * @since 0.20
+   */
+  public void recordWithTags(List<Tag> tags) {
+    // Provides a default no-op implementation to avoid breaking other existing sub-classes.
+    Utils.checkNotNull(tags, "tags");
+    Utils.checkListElementNotNull(tags, "tag");
+  }
 }
--- a/api/src/main/java/io/opencensus/stats/MeasureMap.java
+++ b/api/src/main/java/io/opencensus/stats/MeasureMap.java
@@ -93,8 +93,8 @@ public abstract class MeasureMap {
   public abstract void record(TagContext tags);
 
   /**
-   * Records all of the measures at the same time, with the current {@link TagContext} plus the
-   * additional {@link Tag}s.
+   * Records all of the measures at the same time, with the implicit current {@link TagContext} plus
+   * the additional {@link Tag}s.
    *
    * <p>This method records all of the stats in the {@code MeasureMap} every time it is called.
    *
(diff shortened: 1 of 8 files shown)


impl_core/src/main/java/io/opencensus/implcore/stats/MeasureMapImpl.java
```diff
@@ -72,6 +78,20 @@ public void record() {
 
   @Override
   public void record(TagContext tags) {
+    recordInternal(RecordUtils.getTagMap(tags));
```
dinooliva: This adds the overhead of copying the TagContext into a Map every time we call record.

impl_core/src/main/java/io/opencensus/implcore/stats/MeasureMapImpl.java
```diff
@@ -72,6 +78,20 @@ public void record() {
...
+  }
+
+  @Override
+  public void recordWithTags(List<Tag> tags) {
```
dinooliva: There should be a version that takes an explicit TagContext as well that this one should be written in terms of.

impl_core/src/main/java/io/opencensus/implcore/stats/MeasureMapImpl.java
songy23: Not really. `RecordUtils.getTagMap` is already on the critical path of `record`: https://github.com/census-instrumentation/opencensus-java/blob/9c4df3dbe085a70850cc466849d33d90ffa846f4/impl_core/src/main/java/io/opencensus/implcore/stats/MutableViewData.java#L151
This PR only moves it from later in the path to an earlier location.

In addition `RecordUtils.getTagMap` won't copy the TagContext into a Map for the existing APIs. It just extracts the internal Map from the TagContext: https://github.com/census-instrumentation/opencensus-java/blob/9c4df3dbe085a70850cc466849d33d90ffa846f4/impl_core/src/main/java/io/opencensus/implcore/stats/RecordUtils.java#L66
For the new API (`recordWithTags`) we'll have to copy into a Map anyway, since the internal map in TagContext is immutable:
https://github.com/census-instrumentation/opencensus-java/blob/9c4df3dbe085a70850cc466849d33d90ffa846f4/impl_core/src/main/java/io/opencensus/implcore/tags/TagContextImpl.java#L41

impl_core/src/main/java/io/opencensus/implcore/stats/MeasureMapImpl.java
songy23: Make sense, added.

songy23:
Chatted with @dinooliva offline, we're wondering whether (and how much) we could benefit from the new API in terms of 1. performance and 2. readability, given that the new API pretty much does the same thing as the current record API with TagContext.

To give a comparison between the two:

Suppose we have a static `TagKey`:
```java
TagKey FRONT_END_KEY = TagKey.create("my.org/front_end");
StatsRecorder recorder = Stats.getStatsRecorder();
Tagger tagger = Tags.getTagger();
```

Record with the current API:
```java
TagValue value = TagValue.create("ios");
TagContext context = tagger.currentBuilder().put(KEY, value).build(); 
recorder.newMeasureMap().put(MEASURE, 1.0).record(context);
```

Record with the new API:
```java
TagValue value = TagValue.create("ios");
List<Tag> tags = Arrays.asList(Tag.create(KEY, value)); 
recorder.newMeasureMap().put(MEASURE, 1.0).recordWithTags(tags);
```

The two code snippets above are exactly equivalent. The first one creates a new `TagContext` from current, while the second creates an extra `Tag` and a `List`.

dinooliva:
My concerns with this feature are 1. I don't see a clear benefit in terms of performance and readability and 2. it introduces another way to define tags, which potentially adds confusion to our APIs.

api/src/main/java/io/opencensus/stats/MeasureMap.java
```diff
@@ -89,4 +91,50 @@ public MeasureMap putAttachment(String key, String value) {
...
+   *
+   * <pre>{@code
+   * TagContext context = tagger.currentBuilder().put(tagKey, tagValue).build();
+   * measureMap.record(context);
+   * }</pre>
+   *
+   * @param tags additional {@link Tag}s to be recorded against.
+   * @since 0.20
+   */
+  public void recordWithTags(List<Tag> tags) {
```
bogdandrutu: I think here we currently support only "upsert" (replace the previous tag) but we may want to support other operations as well. So probably something like what golang did to have a notion of "TagMutator" is useful.

Also not sure if we need the metadata (the new Tag component), but happy to have it here, maybe just comment that metadata is ignored.

Before coding this idea let's make sure we agree on what to do. Thanks.

songy23:
Closing. As previously discussed, in Java particularly, adding a `recordWithTags` API doesn't offer us much benefits over using the existing Tag APIs. Will reconsider this if we saw a real need.

PR #1768, opened 2019-02-22
Tags: Add TagMetadata.

songy23:
Fixes https://github.com/census-instrumentation/opencensus-java/issues/1666.

~~DO NOT MERGE before https://github.com/census-instrumentation/opencensus-specs/pull/233 is submitted.~~

### DIFF
--- a/api/src/main/java/io/opencensus/tags/Tag.java
+++ b/api/src/main/java/io/opencensus/tags/Tag.java
@@ -17,6 +17,7 @@
 package io.opencensus.tags;
 
 import com.google.auto.value.AutoValue;
+import io.opencensus.tags.TagMetadata.TagTtl;
 import javax.annotation.concurrent.Immutable;
 
 /**
@@ -28,18 +29,39 @@ import javax.annotation.concurrent.Immutable;
 @AutoValue
 public abstract class Tag {
 
+  private static final TagMetadata METADATA_UNLIMITED_PROPAGATION =
+      TagMetadata.create(TagTtl.UNLIMITED_PROPAGATION);
+
   Tag() {}
 
   /**
    * Creates a {@code Tag} from the given key and value.
    *
+   * <p>This is equivalent to calling {@code create(key, value,
+   * TagMetadata.create(TagTtl.UNLIMITED_PROPAGATION))}.
+   *
    * @param key the tag key.
    * @param value the tag value.
    * @return a {@code Tag} with the given key and value.
    * @since 0.8
+   * @deprecated in favor of {@link #create(TagKey, TagValue, TagMetadata)}.
    */
+  @Deprecated
   public static Tag create(TagKey key, TagValue value) {
-    return new AutoValue_Tag(key, value);
+    return create(key, value, METADATA_UNLIMITED_PROPAGATION);
+  }
+
+  /**
+   * Creates a {@code Tag} from the given key, value and metadata.
+   *
+   * @param key the tag key.
+   * @param value the tag value.
+   * @param tagMetadata the tag metadata.
+   * @return a {@code Tag}.
+   * @since 0.20
+   */
+  public static Tag create(TagKey key, TagValue value, TagMetadata tagMetadata) {
+    return new AutoValue_Tag(key, value, tagMetadata);
   }
 
   /**
@@ -57,4 +79,12 @@ public abstract class Tag {
    * @since 0.8
    */
   public abstract TagValue getValue();
+
+  /**
+   * Returns the {@link TagMetadata} associated with this {@link Tag}.
+   *
+   * @return the {@code TagMetadata}.
+   * @since 0.20
+   */
+  public abstract TagMetadata getTagMetadata();
 }
--- a/api/src/main/java/io/opencensus/tags/Tag.java
+++ b/api/src/main/java/io/opencensus/tags/Tag.java
@@ -37,6 +37,8 @@ public abstract class Tag {
   /**
    * Creates a {@code Tag} from the given key and value.
    *
+   * <p>For backwards-compatibility this method still produces propagating {@link Tag}s.
+   *
    * <p>This is equivalent to calling {@code create(key, value,
    * TagMetadata.create(TagTtl.UNLIMITED_PROPAGATION))}.
    *
--- a/api/src/main/java/io/opencensus/tags/TagContextBuilder.java
+++ b/api/src/main/java/io/opencensus/tags/TagContextBuilder.java
@@ -28,6 +28,8 @@ public abstract class TagContextBuilder {
   /**
    * Adds the key/value pair regardless of whether the key is present.
    *
+   * <p>For backwards-compatibility this method still produces propagating {@link Tag}s.
+   *
    * <p>Equivalent to calling {@code put(key, value,
    * TagMetadata.create(TagTtl.UNLIMITED_PROPAGATION))}.
    *
--- a/api/src/main/java/io/opencensus/tags/TagContextBuilder.java
+++ b/api/src/main/java/io/opencensus/tags/TagContextBuilder.java
@@ -28,12 +28,28 @@ public abstract class TagContextBuilder {
   /**
    * Adds the key/value pair regardless of whether the key is present.
    *
+   * <p>Equivalent to calling {@code put(key, value,
+   * TagMetadata.create(TagTtl.UNLIMITED_PROPAGATION))}.
+   *
    * @param key the {@code TagKey} which will be set.
    * @param value the {@code TagValue} to set for the given key.
    * @return this
    * @since 0.8
+   * @deprecated in favor of {@link #put(TagKey, TagValue, TagMetadata)}.
    */
+  @Deprecated
   public abstract TagContextBuilder put(TagKey key, TagValue value);
+
+  /**
+   * Adds the key/value pair and metadata regardless of whether the key is present.
+   *
+   * @param key the {@code TagKey} which will be set.
+   * @param value the {@code TagValue} to set for the given key.
+   * @param tagMetadata the {@code TagMetadata} associated with this {@link Tag}.
+   * @return this
+   * @since 0.20
+   */
+  public abstract TagContextBuilder put(TagKey key, TagValue value, TagMetadata tagMetadata);
 
   /**
    * Removes the key if it exists.
--- a/api/src/main/java/io/opencensus/tags/TagMetadata.java
+++ b/api/src/main/java/io/opencensus/tags/TagMetadata.java
@@ -59,7 +59,7 @@ public abstract class TagMetadata {
    * then the tag is considered to have travelled one hop.
    *
    * <p>There could be one or more proxy(ies) between sender and receiver. Proxies are treated as
-   * transparent entities and they do not create additional hops.
+   * transparent entities and they are not counted as hops.
    *
    * <p>For now, only special values of {@link TagTtl} are supported.
    *
@@ -69,7 +69,7 @@ public abstract class TagMetadata {
 
     /**
      * A {@link Tag} with {@link TagTtl#NO_PROPAGATION} is considered to have local scope and is
-     * used within the process it created.
+     * used within the process where it's created.
      *
      * @since 0.20
      */
@@ -80,8 +80,8 @@ public abstract class TagMetadata {
      *
      * <p>However, it is still subject to outgoing and incoming (on remote side) filter criteria.
      *
-     * <p>{@link TagTtl#UNLIMITED_PROPAGATION} is typical used to represent a request, processing of
-     * which may span multiple entities.
+     * <p>{@link TagTtl#UNLIMITED_PROPAGATION} is typical used to track a request, which may be
+     * processed across multiple entities.
      *
      * @since 0.20
      */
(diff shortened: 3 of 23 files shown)


api/src/main/java/io/opencensus/tags/Tag.java
```diff
@@ -28,18 +29,39 @@
 @AutoValue
 public abstract class Tag {
 
+  private static final TagMetadata METADATA_UNLIMITED_PROPAGATION =
```
songy23: Should we expose this in the API? Otherwise users need to re-create this object in their code.

api/src/main/java/io/opencensus/tags/Tag.java
```diff
@@ -28,18 +29,39 @@
...
    * @param key the tag key.
    * @param value the tag value.
    * @return a {@code Tag} with the given key and value.
    * @since 0.8
+   * @deprecated in favor of {@link #create(TagKey, TagValue, TagMetadata)}.
    */
+  @Deprecated
   public static Tag create(TagKey key, TagValue value) {
-    return new AutoValue_Tag(key, value);
+    return create(key, value, METADATA_UNLIMITED_PROPAGATION);
```
songy23: For backwards-compatibility, `Tag`s created with this method will still be propagating, though this method is marked as deprecated so there will be warnings when compiling.

api/src/main/java/io/opencensus/tags/TagContextBuilder.java
```diff
@@ -28,13 +28,29 @@
   /**
    * Adds the key/value pair regardless of whether the key is present.
    *
+   * <p>Equivalent to calling {@code put(key, value,
+   * TagMetadata.create(TagTtl.UNLIMITED_PROPAGATION))}.
```
songy23: Same, to keep backwards-compatibility this (deprecated) method still produces propagating `Tag`s.

impl_core/src/main/java/io/opencensus/implcore/tags/TagValueWithMetadata.java
```diff
@@ -0,0 +1,47 @@
...
+import io.opencensus.common.Internal;
+import io.opencensus.tags.TagMetadata;
+import io.opencensus.tags.TagValue;
+import javax.annotation.concurrent.Immutable;
+
+/** Internal helper class that holds a TagValue and a TagMetadata. */
+@Immutable
+@AutoValue
+@Internal
+public abstract class TagValueWithMetadata {
```
songy23: This has to be public because it's used in other packages (`tags.propagation`).

songy23:
Ready to review since https://github.com/census-instrumentation/opencensus-specs/pull/233 is merged.

api/src/main/java/io/opencensus/tags/Tag.java
rghetia: agree on exposing it.

impl_core/src/test/java/io/opencensus/implcore/tags/ScopedTagMapTest.java
```diff
@@ -109,4 +119,35 @@ public void addToCurrentTagsWithBuilder() {
       scope1.close();
     }
   }
+
+  @Test
+  public void multiScopeTagMapWithMetadata() {
```
songy23: This test matches https://github.com/census-instrumentation/opencensus-specs/pull/233/files#diff-7ee77560805f6772a65a8add67f0d692R95.

api/src/main/java/io/opencensus/tags/Tag.java
songy23: SG, updated.

api/src/main/java/io/opencensus/tags/TagMetadata.java
```diff
@@ -0,0 +1,113 @@
...
+public abstract class TagMetadata {
+
+  TagMetadata() {}
+
+  /**
+   * {@link Tag}s associated with this metadata can propagate unlimited number of hops.
+   *
+   * @since 0.20
+   */
+  public static final TagMetadata METADATA_UNLIMITED_PROPAGATION =
```
bogdandrutu: Not sure we want these 2 fields public, because if we add other things in the metadata will be hard to scale to all possibilities. But I see why you want to expose this to avoid object creation, but user may be able to do this as well, not sure what is the best option.

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpClientHandler.java
```diff
@@ -150,6 +150,7 @@ public void handleEnd(
     spanEnd(context.span, response, error);
   }
 
+  @SuppressWarnings("deprecation")
```
bogdandrutu: Why not changing this to the new API? Same everywhere.

impl_core/src/main/java/io/opencensus/implcore/stats/RecordUtils.java
```diff
@@ -62,16 +61,12 @@
...
-    if (ctx instanceof TagMapImpl) {
-      return ((TagMapImpl) ctx).getTags();
-    } else {
-      Map<TagKey, TagValue> tags = Maps.newHashMap();
-      for (Iterator<Tag> i = InternalUtils.getTags(ctx); i.hasNext(); ) {
-        Tag tag = i.next();
-        tags.put(tag.getKey(), tag.getValue());
-      }
-      return tags;
+    Map<TagKey, TagValue> tags = Maps.newHashMap();
```
bogdandrutu: You force a new map creation every time this is sub-optimal because the case when the TagContext is not our implementation is very rare and probably we should move this conversion TagContext to TagMapImpl sooner and make our implementation work with TagMapImpl everywhere.

Then you use everywhere the Map<TagKey, ValueAndMetadata> to avoid to always create a new map.

So something like this:
```java
static TagMapImpl ToTagMapImpl(TagContext ctx) {
  if (!(ctx instanceof TagMapImpl)) {
    ...
    return  new TagMapImpl(...)
  }
  return (TagMapImpl) ctx
}

// In apis that we accept a TagContext like record we call this as a first thing
void record(TagContext ctx, MeasureMap) {
  TagMapImpl tagMap = ToTagMapImpl(ctx)
  // from here we only use TagMapImpl in our implementation.
}
```

Maybe this is an something we should do in a separate PR.

songy23: Agreed non-native implementation should be rare. I reverted and modified this method to return `Map<TagKey, TagValueWithMetadata>` instead and avoid re-creating a map for native implementation. Though `TagMetadata` is not used in stats impl.

songy23:
PTAL

api/src/main/java/io/opencensus/tags/TagMetadata.java
```diff
@@ -0,0 +1,96 @@
...
+     * <p>{@link TagTtl#UNLIMITED_PROPAGATION} is typical used to represent a request, processing of
+     * which may span multiple entities.
+     *
+     * @since 0.20
+     */
+    UNLIMITED_PROPAGATION(-1);
+
+    final int hops;
+
+    TagTtl(int hops) {
```
bogdandrutu: private?

api/src/main/java/io/opencensus/tags/TagMetadata.java
```diff
@@ -0,0 +1,96 @@
...
+     * <p>However, it is still subject to outgoing and incoming (on remote side) filter criteria.
+     *
+     * <p>{@link TagTtl#UNLIMITED_PROPAGATION} is typical used to represent a request, processing of
+     * which may span multiple entities.
+     *
+     * @since 0.20
+     */
+    UNLIMITED_PROPAGATION(-1);
+
+    final int hops;
```
bogdandrutu: private?

api/src/main/java/io/opencensus/tags/Tag.java
bogdandrutu: I think we should not because once you have another property in the metadata you need to expose a lot of constants correct?

api/src/main/java/io/opencensus/tags/TagContextBuilder.java
bogdandrutu: This should be in the method comment.

api/src/main/java/io/opencensus/tags/Tag.java
songy23: Removed from public APIs. If they're really needed by users we can always expose them later.

api/src/main/java/io/opencensus/tags/TagContextBuilder.java
songy23: SG, added to Javadoc.

api/src/main/java/io/opencensus/tags/TagMetadata.java
songy23: Done.

api/src/main/java/io/opencensus/tags/TagMetadata.java
songy23: Done.

api/src/main/java/io/opencensus/tags/TagMetadata.java
```diff
@@ -0,0 +1,96 @@
...
+
+  /**
+   * {@link TagTtl} is an integer that represents number of hops a tag can propagate.
+   *
+   * <p>Anytime a sender serializes a tag, sends it over the wire and receiver deserializes the tag
+   * then the tag is considered to have travelled one hop.
+   *
+   * <p>There could be one or more proxy(ies) between sender and receiver. Proxies are treated as
+   * transparent entities and they do not create additional hops.
+   *
```
dinooliva: they do not create additional hops => are not counted as hops

api/src/main/java/io/opencensus/tags/TagMetadata.java
```diff
@@ -0,0 +1,96 @@
...
+   * <p>For now, only special values of {@link TagTtl} are supported.
+   *
+   * @since 0.20
+   */
+  public enum TagTtl {
+
+    /**
+     * A {@link Tag} with {@link TagTtl#NO_PROPAGATION} is considered to have local scope and is
+     * used within the process it created.
```
dinooliva: nit: process where it's created

api/src/main/java/io/opencensus/tags/TagMetadata.java
```diff
@@ -0,0 +1,96 @@
...
+    NO_PROPAGATION(0),
+
+    /**
+     * A {@link Tag} with {@link TagTtl#UNLIMITED_PROPAGATION} can propagate unlimited hops.
+     *
+     * <p>However, it is still subject to outgoing and incoming (on remote side) filter criteria.
+     *
+     * <p>{@link TagTtl#UNLIMITED_PROPAGATION} is typical used to represent a request, processing of
+     * which may span multiple entities.
+     *
```
dinooliva: fine as-is, could consider: is typically used to track a request, which may be processed across multiple entities.

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpServerHandler.java
```diff
@@ -176,11 +177,18 @@ private void recordStats(
     TagContext startCtx =
         tagger
             .toBuilder(context.tagContext)
-            .put(HTTP_SERVER_METHOD, TagValue.create(methodStr == null ? "" : methodStr))
-            .put(HTTP_SERVER_ROUTE, TagValue.create(routeStr == null ? "" : routeStr))
+            .put(
+                HTTP_SERVER_METHOD,
```
dinooliva: I would think we wouldn't want to propagate some of these tags but instead have the endpoints set them explicitly during message processing.

api/src/main/java/io/opencensus/tags/TagMetadata.java
songy23: SG, updated.

api/src/main/java/io/opencensus/tags/TagMetadata.java
songy23: Updated.

api/src/main/java/io/opencensus/tags/TagMetadata.java
songy23: Updated.

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpServerHandler.java
songy23: Changing these tags to non-propagating by default would be a breaking change, though the impact could be minimal since HTTP handlers are just released in v0.19. WDYT? @rghetia

songy23:
Chatted offline, we don't propagate tags over HTTP before so it would be fine to just change them to non-propagating. Updated in https://github.com/census-instrumentation/opencensus-java/pull/1768/commits/9e5f0fd127f8cfe1c9740e2beb411859d94a4c86, PTAL @dinooliva.

contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpServerHandler.java
rghetia: as discussed. HTTP handlers were not propagating tags. So no breaking change.

songy23:
Thanks!

PR #1769, opened 2019-02-23
SpanKind is missing in HttpServerHandler and HttpClientHandler

rghetia:


contrib/http_util/src/main/java/io/opencensus/contrib/http/HttpClientHandler.java
```diff
@@ -112,7 +113,7 @@ public HttpRequestContext handleStart(@Nullable Span parent, C carrier, Q reques
     }
     String spanName = getSpanName(request, extractor);
     SpanBuilder builder = tracer.spanBuilderWithExplicitParent(spanName, parent);
-    Span span = builder.startSpan();
+    Span span = builder.setSpanKind(Kind.CLIENT).startSpan();
```
bogdandrutu: Do we also prefix the SpanName with "Recv." and "Sent." ?

bogdandrutu: Asking because we should not do that.

rghetia: no we don't. I'll add that.

bogdandrutu: Did you see my previous comment. We SHOULD NOT do that.

rghetia: for some reason it was showing only the first comment. Later I saw your comment and I edited my response. Anyway, I am not going to add the prefix.

PR #1771, opened 2019-02-26
Exporter/Stats/Stackdriver: Support Exemplar.

songy23:
~~Depending on https://github.com/census-instrumentation/opencensus-java/pull/1770.~~

Stackdriver Monitoring has added support for `Exemplar` in the latest client libraries. Modify our Stackdriver Stats exporter to support it. 

~~Note that for now Stackdriver Monitoring backend only accepts string attachments. If we were to send other attachments (e.g `type.googleapis.com/google.devtools.cloudtrace.v1.Trace`), Stackdriver would return an error response saying that type is not supported.~~

### DIFF
--- a/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExportUtils.java
+++ b/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExportUtils.java
@@ -19,6 +19,7 @@ package io.opencensus.exporter.stats.stackdriver;
 import com.google.api.Distribution;
 import com.google.api.Distribution.BucketOptions;
 import com.google.api.Distribution.BucketOptions.Explicit;
+import com.google.api.Distribution.Exemplar;
 import com.google.api.LabelDescriptor;
 import com.google.api.LabelDescriptor.ValueType;
 import com.google.api.Metric;
@@ -34,6 +35,8 @@ import com.google.monitoring.v3.Point;
 import com.google.monitoring.v3.TimeInterval;
 import com.google.monitoring.v3.TimeSeries;
 import com.google.monitoring.v3.TypedValue;
+import com.google.protobuf.Any;
+import com.google.protobuf.ByteString;
 import com.google.protobuf.Timestamp;
 import io.opencensus.common.Function;
 import io.opencensus.common.Functions;
@@ -105,6 +108,15 @@ final class StackdriverExportUtils {
 
   @VisibleForTesting static final String SUMMARY_SUFFIX_COUNT = "_summary_count";
   @VisibleForTesting static final String SUMMARY_SUFFIX_SUM = "_summary_sum";
+
+  @VisibleForTesting
+  static final String EXEMPLAR_ATTACHMENT_TYPE_STRING =
+      "type.googleapis.com/google.protobuf.StringValue";
+  // TODO: currently Stackdriver Monitoring only accepts string attachment.
+  // private static final String EXEMPLAR_ATTACHMENT_TYPE_TRACE_ID =
+  //     "type.googleapis.com/google.devtools.cloudtrace.v1.Trace";
+  // private static final String EXEMPLAR_ATTACHMENT_TYPE_DROPPED_LABELS =
+  //     "type.googleapis.com/google.monitoring.v3.DroppedLabels";
 
   // Constant functions for TypedValue.
   private static final Function<Double, TypedValue> typedValueDoubleFunction =
@@ -337,13 +349,15 @@ final class StackdriverExportUtils {
   // Convert a OpenCensus Distribution to a StackDriver Distribution
   @VisibleForTesting
   static Distribution createDistribution(io.opencensus.metrics.export.Distribution distribution) {
-    return Distribution.newBuilder()
-        .setBucketOptions(createBucketOptions(distribution.getBucketOptions()))
-        .addAllBucketCounts(createBucketCounts(distribution.getBuckets()))
-        .setCount(distribution.getCount())
-        .setMean(distribution.getCount() == 0 ? 0 : distribution.getSum() / distribution.getCount())
-        .setSumOfSquaredDeviation(distribution.getSumOfSquaredDeviations())
-        .build();
+    Distribution.Builder builder =
+        Distribution.newBuilder()
+            .setBucketOptions(createBucketOptions(distribution.getBucketOptions()))
+            .setCount(distribution.getCount())
+            .setMean(
+                distribution.getCount() == 0 ? 0 : distribution.getSum() / distribution.getCount())
+            .setSumOfSquaredDeviation(distribution.getSumOfSquaredDeviations());
+    setBucketCountsAndExemplars(distribution.getBuckets(), builder);
+    return builder.build();
   }
 
   // Convert a OpenCensus BucketOptions to a StackDriver BucketOptions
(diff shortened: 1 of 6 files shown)


exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExportUtils.java
```diff
@@ -106,6 +109,15 @@
   @VisibleForTesting static final String SUMMARY_SUFFIX_COUNT = "_summary_count";
   @VisibleForTesting static final String SUMMARY_SUFFIX_SUM = "_summary_sum";
 
+  @VisibleForTesting
+  static final String EXEMPLAR_ATTACHMENT_TYPE_STRING =
+      "type.googleapis.com/google.protobuf.StringValue";
+  // TODO: currently Stackdriver Monitoring only accepts string attachment.
```
bogdandrutu: Not sure I understand this. I made a prototype some time ago and I was able to push the "SpanContext"

songy23: Which type are you using?

I got this error with type `google.devtools.cloudtrace.v1.Trace`:
```json
{
  "code": 400,
  "message": "Invalid value at 'time_series[0].points[0].value.distribution_value.exemplars[0].attachments[0]' (Any), Invalid type URL, unknown type: google.devtools.cloudtrace.v1.Trace",
  "status": "INVALID_ARGUMENT",
}
```

bogdandrutu: Use the proto that you added :) https://github.com/googleapis/googleapis/blob/master/google/monitoring/v3/span_context.proto

songy23: I see - I tested the previous one because it's on Stackdriver documentation (https://cloud.google.com/monitoring/api/ref_v3/rest/v3/TimeSeries#exemplar):

> Contextual information about the example value. Examples are:
Trace ID: type.googleapis.com/google.devtools.cloudtrace.v1.Trace

Though it didn't work...

songy23: Updated to support `SpanContext`, PTAL.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExportUtils.java
```diff
@@ -397,17 +412,42 @@ private static Exemplar toProtoExemplar(
         Exemplar.newBuilder()
             .setValue(exemplar.getValue())
             .setTimestamp(convertTimestamp(exemplar.getTimestamp()));
+    @javax.annotation.Nullable String traceId = null;
+    @javax.annotation.Nullable String spanId = null;
     for (Map.Entry<String, String> attachment : exemplar.getAttachments().entrySet()) {
-      // TODO: add support for special attachments once Stackdriver Monitoring accepts them.
+      String key = attachment.getKey();
+      String value = attachment.getValue();
+      if (ExemplarUtils.ATTACHMENT_KEY_TRACE_ID.equals(key)) {
```
bogdandrutu: I know that currently this package is not used, and I am curious to know if it is better to extend the Exemplar to support SpanContext directly to avoid these 2 keys, what do you think?

Something like https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/stats/MeasureMap.java#L67
```java
public MeasureMap putAttachment(SpanContext spanContext);
```

When we export in metrics we can export using these headers probably, what do you think? I think I am mostly worried about recording path, but in our Metrics if we can keep it as <string, string>.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExportUtils.java
```diff
@@ -397,17 +412,42 @@ private static Exemplar toProtoExemplar(
...
+        traceId = value;
+      } else if (ExemplarUtils.ATTACHMENT_KEY_SPAN_ID.equals(key)) {
+        spanId = value;
+      } else { // Everything else will be treated as plain strings.
+        builder.addAttachments(
```
bogdandrutu: Should we use "DroppedLabels"? not sure what is that about.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExportUtils.java
songy23: I remember we discussed this approach before. That would introduce dependency on Tracing APIs to Stats APIs, which is not ideal. That's also the main reason that we chose <String, String> as exemplar attachments.

An alternative is to add the API `putAttachment(SpanContext spanContext)` in a contrib util package.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExportUtils.java
songy23: > not sure what is that about.

Me either, but I think that's a special attachment and probably not everything fits into it. I left a [TODO](https://github.com/census-instrumentation/opencensus-java/pull/1771/files#diff-d2886f5b11a5e3be7a00a8f5fff5a1f0R126) to support it once I have a better understanding on it.

exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverExportUtils.java
bogdandrutu: Other option is to have:

abstract class LazyAttachmentValue {
  public String getValue();
}

putAttachment(String key,  LazyAttachmentValue value)

And then in the util we create the LazyAttachmentValue for SpanContext, also we make the SpanContext attachment to be one line, which I think we should do even in Logs if you ask me. Maybe that sail is shipped already.

songy23: I file an issue https://github.com/census-instrumentation/opencensus-java/issues/1775 to keep track of this, since it involves new API support. This PR is mainly on exporter changes.

bogdandrutu: Yeah, but before we add this support we can guarantee that nobody uses the keys defined in the exemplar util so I think we should do that first, this way we can simply do a breaking change in the exemplar util.

songy23: OK that makes sense to me, putting this PR on hold for now.

songy23:
Updated to use the refactored APIs in https://github.com/census-instrumentation/opencensus-java/pull/1771/commits/fd2003029560cdbb93540e65283874e419ca6f50. PTAL.

PR #1773, opened 2019-02-27
Exporter/Trace/OcAgent: Refactor Export RPC.

songy23:
Export Handler is called every 5 seconds in a different thread (where [SpanExporter](https://github.com/census-instrumentation/opencensus-java/blob/master/impl_core/src/main/java/io/opencensus/implcore/trace/export/SpanExporterImpl.java) is running).
- If no connection is established with Agent, try to reconnect every time when a new batch of spans comes.
- If connection is already established, use the connected `exportRpcHandler` to send spans.

`Config` is not included in the PR. (It's not yet supported on Agent either.)

Tested locally against the [FakeOcAgentTraceServiceGrpcImpl](https://github.com/census-instrumentation/opencensus-java/blob/master/exporters/trace/ocagent/src/test/java/io/opencensus/exporter/trace/ocagent/FakeOcAgentTraceServiceGrpcImpl.java) and it worked fine.

### DIFF
--- a/exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentTraceExporterHandler.java
+++ b/exporters/trace/ocagent/src/main/java/io/opencensus/exporter/trace/ocagent/OcAgentTraceExporterHandler.java
@@ -16,19 +16,38 @@
 
 package io.opencensus.exporter.trace.ocagent;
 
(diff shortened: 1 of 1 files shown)


bogdandrutu:
Hope to get some tests :)

songy23:
Will add in a follow-up PR.

Issue #1780, opened 2019-03-04
http-servlet make default value for publicEndpoint to false.


rghetia:
Make default value for publicEndpoint to false. In more cases services will not be serving publicEndpoint than those that would.

PR #1781, opened 2019-03-04
Modify default value to false for publicEndpoint in Http-Servlet.

rghetia:
fixes #1780

PR #1784, opened 2019-03-04
Changed publicEndpoint to be an init parameter, default to false

alexamies:
For HTTP integration with Jetty, changed publicEndpoint to be an init parameter to be easier to configure via web.xml web application file and changed default to false to addresses

https://github.com/census-instrumentation/opencensus-java/issues/1780

### DIFF
--- a/contrib/http_servlet/src/main/java/io/opencensus/contrib/http/servlet/OcHttpServletFilter.java
+++ b/contrib/http_servlet/src/main/java/io/opencensus/contrib/http/servlet/OcHttpServletFilter.java
@@ -145,13 +145,9 @@ public class OcHttpServletFilter implements Filter {
       extractor = new OcHttpServletExtractor();
     }
 
-    obj = context.getAttribute(OC_PUBLIC_ENDPOINT);
+    String publicEndVal = context.getInitParameter(OC_PUBLIC_ENDPOINT);
     if (obj != null) {
-      if (obj instanceof Boolean) {
-        publicEndpoint = (Boolean) obj;
-      } else {
-        throw new ServletException(EXCEPTION_MESSAGE + OC_PUBLIC_ENDPOINT);
-      }
+        publicEndpoint = Boolean.parseBoolean(publicEndVal);
     } else {
       publicEndpoint = false;
     }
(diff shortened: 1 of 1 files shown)


contrib/http_servlet/src/main/java/io/opencensus/contrib/http/servlet/OcHttpServletFilter.java
```diff
@@ -145,15 +145,11 @@ public void init(FilterConfig filterConfig) throws ServletException {
...
     if (obj != null) {
-      if (obj instanceof Boolean) {
-        publicEndpoint = (Boolean) obj;
-      } else {
-        throw new ServletException(EXCEPTION_MESSAGE + OC_PUBLIC_ENDPOINT);
-      }
+        publicEndpoint = Boolean.parseBoolean(publicEndVal);
     } else {
-      publicEndpoint = true;
+      publicEndpoint = false;
```
rghetia: This is already fixed in #1781. Please rebase your change

contrib/http_servlet/src/main/java/io/opencensus/contrib/http/servlet/OcHttpServletFilter.java
```diff
@@ -145,15 +145,11 @@ public void init(FilterConfig filterConfig) throws ServletException {
       extractor = new OcHttpServletExtractor();
     }
 
-    obj = context.getAttribute(OC_PUBLIC_ENDPOINT);
+    String publicEndVal = context.getInitParameter(OC_PUBLIC_ENDPOINT);
```
rghetia: Does this also mean that getting Propagator and Extractor should also be changed?

contrib/http_servlet/src/main/java/io/opencensus/contrib/http/servlet/OcHttpServletFilter.java
alexamies: Rebased in latest push

contrib/http_servlet/src/main/java/io/opencensus/contrib/http/servlet/OcHttpServletFilter.java
alexamies: Maybe but I am not sure why an application developer would want to change that. Those are more purely within the realm of settings for tool configuration. I have not done testing related to those, like I have with publicEndVal.

PR #1791, opened 2019-03-07
Exemplar: Refactor to be under metrics.data.

songy23:
~~Blocked by https://github.com/census-instrumentation/opencensus-java/pull/1779.~~


songy23:
Rebased and updated. The reasons for making this change:
1. By its nature `Exemplar` is a correlation between multiple signals (histogram and trace in our case). IMO it doesn't belong to an existing signal. We already made `Exemplar` in [Go](https://github.com/census-instrumentation/opencensus-go/tree/master/exemplar).
2. `Exemplar` and its `Attachment`s are required at both recording time and exporting time. For recording we have `stats` package and for exporting we have `metrics`. We don't want `stats` and `metrics` APIs to have dependency on each other to avoid circular dependency. Therefore previously we have two identical implementation of `Exemplar`: `stats.AggregationData.Exemplar` and `metrics.export.Distribution.Exemplar`. This doesn't work if we were to use one consistent `Exemplar` for both recording and exporting.

Thus I proposed to extract `Exemplar` to common package so that both `stats` and `metrics` packages can depend on it. Note this is a breaking change to those who use the current `Exemplar` APIs.

bogdandrutu:
I think we are moving to the model where stats is just for recording and not for exporting (currently all our exporters are now using Metric no more ViewData). Probably we should remove the ViewData, hence we will remove the AggregationData.

So I think we should work to remove the ViewData and AggregationData, so that will remove the duplicated code.

songy23:
> So I think we should work to remove the ViewData and AggregationData, so that will remove the duplicated code.

But still we need `AttachmentValue` in both recording APIs and exporting APIs:
```java
statsRecorder.newMeasureMap().put(m, v).putAttachment(key, attachmentValue).record();
...
metricProducer.getMetrics().get(0)...getDistribution().getExemplars().get(0).getAttachments();
```
so the only viable option I saw here is still to extract it to a common artifact.

bogdandrutu:
For the common types that we need during recording as well as exporting I suggest we do what Go does for metrics and add a "data" sub-package. stats, metrics and metrics.exporter can depend on metrics.data.

songy23:
> I suggest we do what Go does for metrics and add a "data" sub-package. stats, metrics and metrics.exporter can depend on metrics.data.

OK that makes sense to me. I'll post a separate PR for adding that package first.

songy23:
https://github.com/census-instrumentation/opencensus-java/pull/1791/commits/4eea6a2ec29dc787ccd3df9e683213a8ae34a47b moved `Exemplar` and `AttachmentValue` to `metrics.data` so that both stats and metrics packages can depend on it. PTAL.

Issue #1667, opened 2019-01-17
Tags: Support HTTP propagation.


songy23:
1. Agree on the text format for tag HTTP propagation.
2. Add HTTP plugin for tags.

dinooliva:
Shouldn't 1 be cross-language (e.g. an issue on opencensus-specs)?

songy23:
The spec issue is https://github.com/census-instrumentation/opencensus-specs/issues/65.

This issue is particularly for tracking the Java implementation.

PR #1800, opened 2019-03-15
Tags: Add HTTP text serializer API.

songy23:
Updates https://github.com/census-instrumentation/opencensus-java/issues/1667.

This PR only contains API changes. Implementation will be in another PR.


api/src/main/java/io/opencensus/tags/propagation/TagContextTextSerializer.java
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Object for serializing and deserializing {@link TagContext}s with HTTP text format.
+ *
+ * <p>OpenCensus uses W3C Correlation Context as the HTTP text format. For more details, see <a
+ * href="https://github.com/w3c/correlation-context">correlation-context</a>.
+ *
+ * @since 0.21
+ */
+public abstract class TagContextTextSerializer {
```
bogdandrutu: In this approach you force tags to be propagated in one string. What if we want to propagate them in multiple strings? Take a look at the trace part:
https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/trace/propagation/TextFormat.java

songy23: OK make sense to me, added support for encoding in multiple strings in https://github.com/census-instrumentation/opencensus-java/pull/1800/commits/56573558e5c861280d6e984d412f8d6c80b97933.

PR #1807, opened 2019-03-21
Set kind on the outgoing HTTP spans in HttpClientHandler

lmolkova:
[Http spec](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md) suggests setting Span kind to Client on outgoing calls. 

`Outgoing requests should be a span kind of CLIENT and incoming requests should be a span kind of SERVER.`
Also HTTP instrumentation tests validate it. https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/http-out-test-cases.json

It allows backends that have special logic for incoming/outgoing calls properly identify them.  Presence or absence of the remote/local parent is not really reliable for this purpose.


lmolkova:
Sorry, I need to pull master more frequently :)

PR #1811, opened 2019-03-22
Tags: Make the new API non-abstract to avoid breaking other subclasses.

songy23:


api/src/main/java/io/opencensus/tags/TagContextBuilder.java
```diff
@@ -51,7 +50,11 @@
    * @return this
    * @since 0.20
    */
-  public abstract TagContextBuilder put(TagKey key, TagValue value, TagMetadata tagMetadata);
+  public TagContextBuilder put(TagKey key, TagValue value, TagMetadata tagMetadata) {
```
bogdandrutu: No need to do this, we try to be backwards compatible only with user public functions not with the implementation.

songy23: This broke gRPC: https://github.com/grpc/grpc-java/blob/master/testing/src/main/java/io/grpc/internal/testing/StatsTestUtils.java#L287.

PR #1833, opened 2019-04-03
Trace Benchmarks

dinooliva:
Adds microbenchmarks for OpenCensus Trace.


benchmarks/src/jmh/java/io/opencensus/benchmarks/trace/BasicOperationsBenchmark.java
```diff
@@ -0,0 +1,305 @@
...
+  @OutputTimeUnit(TimeUnit.NANOSECONDS)
+  public Span getCurrentSpan(Data data) {
+    return data.tracer.getCurrentSpan();
+  }
+
+  /** Encode a span using binary format. */
+  @Benchmark
+  @BenchmarkMode(Mode.AverageTime)
+  @OutputTimeUnit(TimeUnit.NANOSECONDS)
+  public byte[] encodeSpanBinary(Data data) {
```
rghetia: encode and decode will vary based on number of attributes, links and annotations. Will that be covered later?

dinooliva: Encoding/Decoding only apply to the span context - tracestate will get added eventually but we don't support that at the moment.

dinooliva:
Thanks!

Issue #1809, opened 2019-03-22
Consider changing DisruptorEventQueue.enqueue() to avoid possibility of blocking 


jmullins:
Does it make sense to change DisruptorEventQueue.enqueue() to use RingBuffer.tryNext() instead of RingBuffer.next() to avoid possibly blocking in the case the producers overrun the consumer due to load or a future bug in the consumer?

In all of our use-cases our preference would be to drop the metrics rather than block.

https://github.com/census-instrumentation/opencensus-java/blob/aa8ee6a526e11e5763d2f508bd6952a1b87e3a51/impl/src/main/java/io/opencensus/impl/internal/DisruptorEventQueue.java#L130-L142

saiya:
I created PR for this issue! : https://github.com/census-instrumentation/opencensus-java/pull/1837

Issue #1767, opened 2019-02-21
Memory settings


bputt:
When exporting trace information:

1. What's the high/low watermark for heap usage, if any?
2. If the heap usage hits a certain threshold, does the exporter become blocking at that point?
3. Is there a way to configure high watermarks for the client?

We were able to verify heap usage increased steadily as we added backpressure to our ingest service. This can become a big issue if we had actual latency issues or if our ingest service was down.

dmichel1:
I can confirm this behavior when the ingest service becomes overloaded/slow the heap steadily increases and eventually causes memory pressure leading to an OOM event in the application.

I havn't found the same behavior when the ingest service is completely down but perhaps I need to test that case more thoroughly. 

I'm attaching a few screenshots from a heap dump I was able to capture...

![Screen Shot 2019-05-02 at 10 36 09 AM](https://user-images.githubusercontent.com/677960/57083554-be16cb00-6cc6-11e9-985b-4b016729ad27.png)
![Screen Shot 2019-05-02 at 10 36 21 AM](https://user-images.githubusercontent.com/677960/57083561-c111bb80-6cc6-11e9-8750-ae750de9e2c8.png)


This issue seems different than https://github.com/census-instrumentation/opencensus-java/issues/1813 where spans that aren't closed properly will cause an OOM event (?)

dmichel1:
It seems https://github.com/census-instrumentation/opencensus-specs/issues/262 and https://github.com/census-instrumentation/opencensus-java/pull/1837 are related to this issue.

bogdandrutu:
@dmichel1 by any chance can you take dump the thread stacktrace as well?

bogdandrutu:
Based on the picture the attach the memory is hold by the SpanExporterImpl thread. 

The current design is this:
ThreadExecutesRequest -> DisruptorThread -> SpanExporterImplThread -> Exporter (usually no thread here, executed from the span exporter impl thread)

Keep in mind that threads are passed to the SpanExporterImplThread only when they end, so clearly not related to this https://github.com/census-instrumentation/opencensus-java/issues/1813.

For me this does not correlate with the  blocking queue issue. May still be a correlation if the Disruptor thread bursts the Spans into the SpanExporterImplThread but probably that is not the issue.

[...]

PR #1837, opened 2019-04-09
Prevent blocking on queue overflow (#1809)

saiya:
https://github.com/census-instrumentation/opencensus-java/issues/1809

I implemented queue overflow handling to prevent blocking foreground (application) thread.

It drops event and show `WARN` log when queue get full.

### DIFF
--- a/impl/src/main/java/io/opencensus/impl/internal/DisruptorEventQueue.java
+++ b/impl/src/main/java/io/opencensus/impl/internal/DisruptorEventQueue.java
@@ -30,6 +30,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import javax.annotation.CheckReturnValue;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.ThreadSafe;
 
@@ -135,7 +136,7 @@ public final class DisruptorEventQueue implements EventQueue {
     DisruptorEnqueuer enqueuer =
         new DisruptorEnqueuer() {
           @Override
-          public void enqueue(Entry entry) {
+          public boolean enqueue(Entry entry) {
             long sequence;
             try {
               sequence = ringBuffer.tryNext();
@@ -147,7 +148,7 @@ public final class DisruptorEventQueue implements EventQueue {
                     "Dropping some events due to queue overflow."
                         + " Consider to reduce sampling rate.");
               }
-              return;
+              return false;
             }
             try {
               DisruptorEvent event = ringBuffer.get(sequence);
@@ -155,6 +156,7 @@ public final class DisruptorEventQueue implements EventQueue {
             } finally {
               ringBuffer.publish(sequence);
             }
+            return true;
           }
         };
     return new DisruptorEventQueue(disruptor, enqueuer);
@@ -176,7 +178,9 @@ public final class DisruptorEventQueue implements EventQueue {
    */
   @Override
   public void enqueue(Entry entry) {
-    enqueuer.enqueue(entry);
+    if (!enqueuer.enqueue(entry)) {
+      entry.rejected();
+    }
   }
 
   /** Shuts down the underlying disruptor. */
@@ -187,10 +191,11 @@ public final class DisruptorEventQueue implements EventQueue {
           final AtomicBoolean logged = new AtomicBoolean(false);
 
           @Override
-          public void enqueue(Entry entry) {
+          public boolean enqueue(Entry entry) {
             if (!logged.getAndSet(true)) {
               logger.log(Level.INFO, "Attempted to enqueue entry after Disruptor shutdown.");
             }
+            return false;
           }
         };
 
@@ -200,7 +205,14 @@ public final class DisruptorEventQueue implements EventQueue {
   // Allows this event queue to safely shutdown by not enqueuing events on the ring buffer
   private abstract static class DisruptorEnqueuer {
 
-    public abstract void enqueue(Entry entry);
+    /**
+     * Try to enqueue Entry into the queue.
+     *
+     * @param entry Entry to enqueue
+     * @return If enqueue failed, returns false.
+     */
+    @CheckReturnValue
+    public abstract boolean enqueue(Entry entry);
   }
 
   // An event in the {@link EventQueue}. Just holds a reference to an EventQueue.Entry.
--- a/impl/src/main/java/io/opencensus/impl/internal/DisruptorEventQueue.java
+++ b/impl/src/main/java/io/opencensus/impl/internal/DisruptorEventQueue.java
@@ -16,8 +16,10 @@
 
 package io.opencensus.impl.internal;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.lmax.disruptor.EventFactory;
 import com.lmax.disruptor.EventHandler;
+import com.lmax.disruptor.InsufficientCapacityException;
 import com.lmax.disruptor.RingBuffer;
 import com.lmax.disruptor.SleepingWaitStrategy;
 import com.lmax.disruptor.dsl.Disruptor;
@@ -25,6 +27,7 @@ import com.lmax.disruptor.dsl.ProducerType;
 import io.opencensus.implcore.internal.DaemonThreadFactory;
 import io.opencensus.implcore.internal.EventQueue;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.annotation.Nullable;
@@ -96,7 +99,7 @@ public final class DisruptorEventQueue implements EventQueue {
   // TODO(aveitch): consider making this a parameter to the constructor, so the queue can be
   // configured to a size appropriate to the system (smaller/less busy systems will not need as
   // large a queue.
-  private static final int DISRUPTOR_BUFFER_SIZE = 8192;
+  @VisibleForTesting static final int DISRUPTOR_BUFFER_SIZE = 8192;
   // The single instance of the class.
   private static final DisruptorEventQueue eventQueue = create();
 
@@ -104,6 +107,8 @@ public final class DisruptorEventQueue implements EventQueue {
   private final Disruptor<DisruptorEvent> disruptor;
 
   private volatile DisruptorEnqueuer enqueuer;
+
+  @VisibleForTesting static final AtomicLong overflowCount = new AtomicLong();
 
   // Creates a new EventQueue. Private to prevent creation of non-singleton instance.
   private DisruptorEventQueue(Disruptor<DisruptorEvent> disruptor, DisruptorEnqueuer enqueuer) {
@@ -131,7 +136,19 @@ public final class DisruptorEventQueue implements EventQueue {
         new DisruptorEnqueuer() {
           @Override
           public void enqueue(Entry entry) {
-            long sequence = ringBuffer.next();
+            long sequence;
+            try {
+              sequence = ringBuffer.tryNext();
+            } catch (InsufficientCapacityException e) {
+              // Show warning log only once to prevent blocking caused by logger.
+              if (overflowCount.getAndIncrement() == 0) {
+                logger.log(
+                    Level.WARNING,
+                    "Dropping some events due to queue overflow."
+                        + " Consider to reduce sampling rate.");
+              }
+              return;
+            }
             try {
               DisruptorEvent event = ringBuffer.get(sequence);
               event.setEntry(entry);
--- a/buildscripts/import-control.xml
+++ b/buildscripts/import-control.xml
@@ -270,6 +270,7 @@ General guidelines on imports:
     <allow pkg="io.opencensus.trace"/>
   </subpackage>
   <subpackage name="impl">
+    <allow pkg="com.google.common"/>
     <allow pkg="com.lmax.disruptor"/>
     <allow pkg="io.opencensus.common"/>
     <allow pkg="io.opencensus.impl"/>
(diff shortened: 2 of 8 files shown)


saiya:
I signed it! > CLA

impl/src/main/java/io/opencensus/impl/internal/DisruptorEventQueue.java
```diff
@@ -131,7 +136,19 @@ private static DisruptorEventQueue create() {
         new DisruptorEnqueuer() {
           @Override
           public void enqueue(Entry entry) {
```
songy23: We may consider tracing this method too (not suggesting doing it in this PR though).

songy23:
We need to evaluate the potential memory leak issue with this PR. Marked as DO NOT MERGE for now.

saiya:
Thank you for checking this PR and quick response!

I organized thoughts about your concern and implemented some improvements to mitigate the issue in 6ff8ccd03de1307905acb53c1be0ec754cca28a1 & 8b886b0540779ca164e6474c81b4fa63a75c04ca. Could you check this? @songy23 

# potential memory leak issue

I checked subclasses of `EventQueue.Entry`. Problem is a `SpanEndEvent#process` because it calls `runningSpanStore.onEnd` to remove span from `runningSpans`. If we skip it, causes memory leak due to runningSpan. In my understanding, it is a cause for your concern.

My idea to mitigate the issue of `runningSpans` potential leak:

- Idea 1-A) Call `EventQueue.Entry#process()` in foreground thread when queue get full. This idea seems bad because it may cause lock contention and severe performance issue.
- Idea 1-B) Add `EventQueue.Entry#rejected()` method to implement minimum cleanup code. Call it in foreground thread when queue get full. Call `runningSpanStore.onEnd` from `SpanEndEvent#rejected()` to prevent memory leak.
- Idea 2) Change `InProcessRunningSpanStoreImpl` to use `WeakReference` to prevent memory leak caused by it. This may cause additional memory usage due to allocating additional objects to hold reference to spans.

I feel Idea 1-B is feasible and effective way, so that implemented Idea 1-B in 6ff8ccd03de1307905acb53c1be0ec754cca28a1 . It also prevent memory leak even if events come after queue shutdown (current code has such potential issue).

( And also `InProcessSampledSpanStoreImpl.UnregisterSpanNameEvent` seems to have similar potential issue. But the `Map<String, PerSpanNameSamples>` inside it should not be huge because its key is name of span, not span itself. So I feel it is not serious problem. )


# Skipping SpanStartEvent

When we drop `SpanStartEvent`, it skips `RunningSpanStoreImpl#onStart(span)`. So that `RunningSpanStoreImpl` might receive `onEnd(span)` without `onStart(span)`. 

Current only one implementation `InProcessRunningSpanStoreImpl` raises `IllegalArgumentExeption` in `runningSpans.removeElement`. In such case, do nothing is correct behavior in my understanding. So that I implemented exception handling in 8b886b0540779ca164e6474c81b4fa63a75c04ca .

saiya:
Friendly ping @songy23 . Let me know if I can help in any way with getting this PR merged.

songy23:
Apologies for getting back late to this PR and thanks for investigating the issue.

> Idea 1-B) Add EventQueue.Entry#rejected() method to implement minimum cleanup code. Call it in foreground thread when queue get full. Call runningSpanStore.onEnd from SpanEndEvent#rejected() to prevent memory leak.

This sounds a reasonable approach to me. We may also extend `EventQueue.Entry#rejected()` to support measurement recording in Stats.  @bogdandrutu @dinooliva @rghetia WDYT?

songy23:
Hi @saiya, we would like to have a better understanding about the background of the issue. Could you describe a bit about how the thread blocking impacted your application? 

(Depending on the impact we may want to take a different approach, e.g do some optimization to reduce the events put onto the queue.)

saiya:
> we would like to have a better understanding about the background of the issue. Could you describe a bit about how the thread blocking impacted your application?

Hi @songy23, here is the situation I am facing:
1. There are so many products/microservices in my company
    - Over 100 systems
    - There are complex dependency between API subsystems
    - Estimated as ten billion HTTP API calls/month at least
2. Some product has traffic spikes
    - It causes unexpectable spike for related API subsystems
3. Some important & huge services are running in on-premise environment
    - Difficult for elastic scaling / frequent failover, so that we are running large instances to process traffic
4. All of our service are implemented with concurrent way (multi-thread or event driven)
    - __Blocking is not welcome (even if it is "possible" blocking)__
    - Blocking cause service down. It hurts profit & customer loyalty
5. We had used Zipkin, but maintaining Zipkin infrastructure is hard work so that considering OpenCensus + Stackdriver Trace

There are some alternative idea I can imagine for this issue, but not suitable for OpenTracing:

1. Use low sampling ratio (e.g. trace only 10%)
    - Cannot capture random performance degrade
    - Does not help debugging for complex bug due to missing trace
2. Use [RateLimiting sampler](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/Sampling.md#ratelimiting-sampler-implementation-details)
    - I found the document today, it seems not supported by opencensus-java yet
    - Because I want to avoid blocking as possible, I need to set conservative RateLimit. 
        - It means low sampling ratio as same as idea 1.
3. Implement dynamic sampling to sample slow request only
    - Does not provide uniform sampling, so that may cause problem for statistics
    - Does not solve queue overflow because we need to start tracing before we know the request is slow

This is why I am interested with this queue overflow issue.

> (Depending on the impact we may want to take a different approach, e.g do some optimization to reduce the events put onto the queue.)

Because large system (large scale microservice system) is chaotic, it is hard to determine "safe" sampling ratio / rate limit. If there are risk of blocking, I need to set very low sampling ratio. Even if we reduce events on the queue, the issue still alive.

In my working experience, many large service provider (not only current company) has similar needs. It is hard to introduce additional blocking risk only for tracing/instrumentation. If the tracing SDK does not have such risk, it is awesome. This is why I made this PR.

But reducing event is also wonderful because it prevent from dropping events.


Thank you for talking with this PR, feel free to ask me for anything I can share more.

songy23:
Thank you @saiya for adding the detailed background, now we have a much better understanding about your use case. There're some concerns from my teammates, and also we want to make sure this behavior is consistent across all OC languages, so I filed https://github.com/census-instrumentation/opencensus-specs/issues/262 for a broader discussion.

saiya:
Hi @songy23 !

Currently this PR drops events if queue get full. But it calls `runningSpanStore.onEnd(span);` even on `rejected()` so it does not cause memory leak. I added corner-case handling to prevent memory leak in rare case also in a55ccbc.

And also I rebased this PR based on latest master branch (there are many refactoring to related classes :-) ).

I know there are discussion in https://github.com/census-instrumentation/opencensus-specs/issues/262. Can such improvements be done in another branch/PR? Or should we improve this PR? (it might make this PR huge.)  This PR prevents memory leak and shows warning log on queue overflow. I think this is minimal viable solution.

songy23:
Hi @saiya, did you get a chance to try the 0.22.0 release? The new release included the fix proposed by @bogdandrutu (https://github.com/census-instrumentation/opencensus-specs/issues/262#issuecomment-487792069) and should reduce the possibility of queue blocking. https://github.com/census-instrumentation/opencensus-java/pull/1893 implemented a similar approach on dropping spans.

@bogdandrutu had some concerns about silently dropping events. Maybe the fact that event queue got full and blocking is an indicator that sampling rate is too high. In this case having users explicitly reduce the sampling rate may be a better option. (IMO dropping queue events is actually equivalent to having a lower sampling rate.) Some quotes from the [Dapper paper](https://static.googleusercontent.com/media/research.google.com/en//archive/papers/dapper-2010-1.pdf):

> In practice, we have found that there is still an adequate amount of trace data for high-volume services when using a sampling rate as low as 1/1024. 

> The first production version of Dapper used a uniform sampling probability for all processes at Google, averaging one sampled trace for every 1024 candidates. This simple scheme was effective for our high-throughput online services since the vast majority of events of interest were still very likely to appear often enough to be captured.

Personally I think adaptive sampling (a.k.a [rate-limited sampler](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/Sampling.md#ratelimiting-sampler-implementation-details)) may be the silver bullet here. But before that happened, I think it's reasonable to make queue non-blocking and log a warning to let users know they may need to use a lower sampling rate.

saiya:
Hi @songy23, thank you for kind response.

> IMO dropping queue events is actually equivalent to having a lower sampling rate

Yes, I think so. And also I know that dropping events causes non-uniform sampling ratio (it sometimes drops, sometimes not drop). If user knows adequate sampling ratio, user should set it.

But the key point I want to emphasize is that it is hard to know proper sampling ratio for some kind of environment. In our case, some microservice amplify request (generate many outgoing API request for each one incoming request). And there are random user activity spikes. In such situation, hard to know good sampling ratio. If I set very pessimistic sampling ratio (such as 0.001%) to avoid blocking, it can easily ignore issues of some percent of users.

> Personally I think adaptive sampling (a.k.a rate-limited sampler) may be the silver bullet here.

Yes, adaptive sampling is an ideal solution for this matter. But, IMHO, I feel it takes some days to implement it.

> But before that happened, I think it's reasonable to make queue non-blocking and log a warning to let users know they may need to use a lower sampling rate.

Absolutely agree. It helps us to search good sampling ratio (rathe than very pessimistic one such as 0.001%). And we really need non blocking one to avoid stopping our service in production.

I can try 0.22.0 in non-important service in our company. But to try it in most hot service, non blocking is important.

songy23:
> But the key point I want to emphasize is that it is hard to know proper sampling ratio for some kind of environment. In our case, some microservice amplify request (generate many outgoing API request for each one incoming request). And there are random user activity spikes. In such situation, hard to know good sampling ratio. If I set very pessimistic sampling ratio (such as 0.001%) to avoid blocking, it can easily ignore issues of some percent of users.

Yes I was thinking about the same reason, and I agree in this scenario it's better to have drop-and-log rather than slow down users' whole application.

@bogdandrutu WDYT? The only alternative I see here is to always recommend a low sampling rate, but like @saiya said it has its own problem.

songy23:
Hi @saiya, I chatted with @bogdandrutu about this PR today. Bogdan mentioned we'll provide the non-blocking option in the new [OpenTelemetry client libraries](https://github.com/open-telemetry/opentelemetry-java). (As you may be aware of, we're currently migrating the OpenCensus project to the merged [OpenTelemetry project](https://github.com/open-telemetry), and OpenCensus will be put to maintenance mode afterwards.) Given that,

1. Do you need this feature urgently in OpenCensus, or can you wait for us implement it in OpenTelemetry?
2. If you need this feature now in OpenCensus, can you keep blocking as the default, and make non-blocking configurable through manifest/env vars/flags? Changing the default from blocking to non-blocking is kind of a breaking change, and this is something we want to avoid.
3. Have you got a chance to try out [the v0.22.1 release](https://search.maven.org/search?q=g:io.opencensus)? We have another customer facing a similar issue, and the latest release fixed it (https://github.com/census-instrumentation/opencensus-java/issues/1767).

bputt:
@songy23 when do you expect there'll be a beta release for OpenTelemetry or a stable release

songy23:
> when do you expect there'll be a beta release for OpenTelemetry or a stable release

The current plan is to have API ready and published by the end of this month. SDK (a.k.a implementation) will be ready early next quarter.

saiya:
Hi @songy23 , thank you for sharing current status.

>  Bogdan mentioned we'll provide the non-blocking option in the new OpenTelemetry client libraries. 

Very good news! Are there any specification / code / PR / issue in OpenTelemetry repositories (e.g. documentation in https://github.com/open-telemetry/opentelemetry-specification) that clarify non-blocking option? Or can I send PR to any repository of OpenTelemetry?

> Do you need this feature urgently in OpenCensus, or can you wait for us implement it in OpenTelemetry?

Not urgent. I can wait for some months (but not want to wait for years). In my case, I currently enabled OpenCensus in non-heavy-traffic systems only. 

In my understanding, OpenTelemetry will provide OpenCensus shim. So that I can reuse my OpenCensus integration to use OpenTelemetry.

> The current plan is to have API ready and published by the end of this month. SDK (a.k.a implementation) will be ready early next quarter.

Good news to hear it! I hope OpenTelemetry will provide SDK for Stackdriver Tracing in next quarter.

> Have you got a chance to try out the v0.22.1 release? We have another customer facing a similar issue, and the latest release fixed it (#1767)

I am using 0.22.1 in not heavy system only. It is working well. But not enabled OpenCensus in heavy systems.

songy23:
> Are there any specification / code / PR / issue in OpenTelemetry repositories (e.g. documentation in https://github.com/open-telemetry/opentelemetry-specification) that clarify non-blocking option? Or can I send PR to any repository of OpenTelemetry?

Not yet since for now we're focusing on the API/data models. Please feel free to open an issue under https://github.com/open-telemetry/opentelemetry-specification.

> I am using 0.22.1 in not heavy system only. It is working well. But not enabled OpenCensus in heavy systems.

Thanks for the info!

> In my understanding, OpenTelemetry will provide OpenCensus shim. So that I can reuse my OpenCensus integration to use OpenTelemetry.

Exactly. The goal is OpenTelemetry will be 100% backwards-compatible with OpenCensus so the existing integration will continue to work.

saiya:
I found a fresh issue in https://github.com/open-telemetry/opentelemetry-specification/issues/94 (`Add recommendations about blocking / queuing / resource consumption for language libraries`), I hope OpenTelemetry explicitly supports non-blocking behavior.

@songy23 Thank you for supporting this PR. I think we can continue discussion in the https://github.com/open-telemetry/opentelemetry-specification/issues/94, please comment anything in the issue if you have thoughts. I feel it is okay to close this PR.

songy23:
Thanks for all the contributions and discussions - your time and efforts are highly appreciated! Let's continue the discussion at open-telemetry/opentelemetry-specification#94 to make sure we have the consistent behavior in all OpenTelemetry languages.

> I hope OpenTelemetry explicitly supports non-blocking behavior.

Yes that's on our radar. As of today we've finished the initial version of APIs in OpenTelemetry-Java. SDK implementation is in progress: https://github.com/open-telemetry/opentelemetry-java/labels/sdk.

PR #1842, opened 2019-04-11
Metrics: Allow to set constant labels for Gauge APIs.

songy23:


songy23:
> it would be better if it could be abstracted somehow

Make sense to me, I filed https://github.com/census-instrumentation/opencensus-java/issues/1843 to keep track on improving these classes.

Issue #1547, opened 2018-11-03
Handle null default project ID in Stackdriver exporters


sebright:
When I added a Nullable annotation to [com.google.cloud.ServiceOptions.getDefaultProjectId()](https://github.com/googleapis/google-cloud-java/blob/9c111f436fb95e0409df446406241a38804c38bb/google-cloud-clients/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java#L320-L332) in #1536, I noticed that the Stackdriver stats and tracing exporters assume that the value is non-null.  I had to suppress several Checker Framework errors:

- https://github.com/census-instrumentation/opencensus-java/blob/2727a58f5220c15e54b210b49d128abaed992a0f/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporter.java#L310-L311

- https://github.com/census-instrumentation/opencensus-java/blob/2727a58f5220c15e54b210b49d128abaed992a0f/exporters/trace/stackdriver/src/main/java/io/opencensus/exporter/trace/stackdriver/StackdriverExporter.java#L113-L114

- https://github.com/census-instrumentation/opencensus-java/blob/2727a58f5220c15e54b210b49d128abaed992a0f/exporters/trace/stackdriver/src/main/java/io/opencensus/exporter/trace/stackdriver/StackdriverTraceExporter.java#L82-L83

PR #1846, opened 2019-04-17
Exporter/Stackdriver: Fail fast on invalid project ID.

songy23:
Also require a few configurations to be non-null.

Related to https://github.com/census-instrumentation/opencensus-java/issues/1547. /cc @sebright


songy23:
https://github.com/census-instrumentation/opencensus-java/pull/1846/commits/b53a2927dbbd7aeb83ffff2c2fba10f489e09179 fixes the same issue with Stackdriver Trace exporter.

PR #1847, opened 2019-04-18
 update jetty client and server example to use updated api.

rghetia:
- use client api that supports ssl context.
- use default value for publicEndpoint on server.
- Add comment on how to use B3 format on both sides, client and server.

### DIFF
--- a/examples/src/main/java/io/opencensus/examples/http/jetty/client/HelloWorldClient.java
+++ b/examples/src/main/java/io/opencensus/examples/http/jetty/client/HelloWorldClient.java
@@ -30,11 +30,16 @@ import org.apache.log4j.BasicConfigurator;
 import org.apache.log4j.Level;
 import org.apache.log4j.Logger;
 import org.eclipse.jetty.client.HttpRequest;
+import org.eclipse.jetty.client.http.HttpClientTransportOverHTTP;
 import org.eclipse.jetty.client.util.StringContentProvider;
 import org.eclipse.jetty.http.HttpMethod;
+import org.eclipse.jetty.util.ssl.SslContextFactory;
 
-/** Sample application that shows how to instrument jetty client. */
+/**
+ * Sample application that shows how to instrument jetty client.
+ */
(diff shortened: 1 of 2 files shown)


examples/src/main/java/io/opencensus/examples/http/jetty/server/HelloWorldServer.java
```diff
@@ -179,18 +183,35 @@ private static void initTracing() {
...
+    contextHandler.addServlet(sh, "/request/*");
+
+    // Enable tracing by adding OcHttpServleFilter for all path
+    contextHandler.addFilter(OcHttpServletFilter.class, "/*", EnumSet.of(DispatcherType.REQUEST));
+
+    // Uncomment following line to use B3Format for trace context propagation.
+    // contextHandler.setAttribute(
+    //    OC_TRACE_PROPAGATOR, Tracing.getPropagationComponent().getB3Format());
+
+    // By default publicEndpoint parameter is set to false and incoming trace context is added as
```
dinooliva: I may be misreading this but it sounds like the incoming trace context is added as a parent link in both the true and false cases.

rghetia: you are right. the first one should be parent.

examples/src/main/java/io/opencensus/examples/http/jetty/client/HelloWorldClient.java
```diff
@@ -68,17 +71,37 @@ public static void main(String[] args) throws Exception {
     initTracing();
     initStatsExporter();
 
-    OcJettyHttpClient httpClient = new OcJettyHttpClient();
+    // Create http client that will trace requests. By default trace context is propagated using
+    // using w3c TraceContext propagator.
```
songy23: Remove the duplicated "using"

rghetia: done.

Issue #1849, opened 2019-04-18
Undeprecate TagContextBuilder and Tag's `put`  without TagMetadata


paivagustavo:
### What version of OpenCensus are you using?
0.20.0

### What did you do?
When upgrading the lib to 0.20.0 some methods got deprecated without no good reasoning. `TagContextBuilder` method `put(TagKey key, TagValue value)` was deprecated in favor for `put(TagKey key, TagValue value, TagMetadata tagMetadata)` but I'm having hard time understanding why `TagMetadata` should be obligatory for everyone. 

### What did you expect to see?
Remove the deprecation for `put(TagKey key, TagValue value)`. We should aim for simplicity while designing an API and this changes makes the current one unnecessarily complicated.

I would like to be able to simply do:
```java
put(someTag, TagValue.create(someValue));
```

### What did you see instead?
I'm forced to either do:
```java
put(someTag, TagValue.create(someValue), null)
```
or 
```java
put(someTag, TagValue.create(someValue), tagMetadata.create(TagTtl.UNLIMITED_PROPAGATION))
```
Which is overly verbose.

songy23:
> TagContextBuilder method put(TagKey key, TagValue value) was deprecated in favor for put(TagKey key, TagValue value, TagMetadata tagMetadata) but I'm having hard time understanding why TagMetadata should be obligatory for everyone.

The reason is `TagContextBuilder.put(TagKey key, TagValue value)` used to be the only API that generates tags, and it produced propagating tags by default. After the discussion in https://github.com/census-instrumentation/opencensus-specs/pull/233, we agreed that most of time tags are only used to group metrics and most tags should only be within the process (non-propagating). To avoid breaking the existing behavior but also bringing this to users' attention, we decided to deprecate the old API. With the new API 
`put(TagKey key, TagValue value, TagMetadata tagMetadata)` users should be very aware of which tags they created are propagating and which are not.

> We should aim for simplicity while designing an API and this changes makes the current one unnecessarily complicated.

Agree, but still we want to discourage users (or at least give a warning) on producing propagating tags unintentionally. An alternative approach is to add a new convenient API like `putLocal(TagKey key, TagValue value)` which produces non-propagating tags while doesn't need to take a metadata every time.

paivagustavo:
I understand the problem and the reasons why now. I think the `putLocal` solution is more than fine for me.

songy23:
[...]

PR #1851, opened 2019-04-19
Tags: Add TagContextBuilder.putLocal().

songy23:
Fixes https://github.com/census-instrumentation/opencensus-java/issues/1849.

Add a convenient API so that users don't need to put a tag metadata over and over again when putting tags.


songy23:
> wouldn't it make sense to add putPropagating also?

Make sense to me, will add in a separate PR.

Issue #1764, opened 2019-02-19
Track progress of gRPC switching to the new Measures and Tags.


bogdandrutu:
Currently we are in a bad state where gRPC is half switched to the new measures, views, tags. We need to track progress here.

See https://github.com/grpc/grpc-java/issues/5291

dinooliva:
Need to release new version of OpenCensus and update gRPC.

PR #1854, opened 2019-04-19
Stats: Map deprecated RPC tags when record against new tags.

songy23:
gRPC may migrate to use the new RPC tags (e.g "grpc_client_method") in https://github.com/grpc/grpc-java/pull/5601 and stop recording against the deprecated tags (e.g "method"). However probably there're users who still use the deprecated RPC constants. To avoid breaking their metrics, we need to do some mappings when recording stats against the new tags.

Note the deprecated RPC measures are already mapped to the new ones in https://github.com/census-instrumentation/opencensus-java/pull/1738.

Updates https://github.com/census-instrumentation/opencensus-java/issues/1764.

### DIFF
--- a/impl_core/src/main/java/io/opencensus/implcore/stats/RecordUtils.java
+++ b/impl_core/src/main/java/io/opencensus/implcore/stats/RecordUtils.java
@@ -68,8 +68,8 @@ final class RecordUtils {
   @VisibleForTesting static final TagKey RPC_METHOD = TagKey.create("method");
   @VisibleForTesting static final TagKey GRPC_CLIENT_STATUS = TagKey.create("grpc_client_status");
   @VisibleForTesting static final TagKey GRPC_CLIENT_METHOD = TagKey.create("grpc_client_method");
-  private static final TagKey GRPC_SERVER_STATUS = TagKey.create("grpc_server_status");
-  private static final TagKey GRPC_SERVER_METHOD = TagKey.create("grpc_server_method");
+  @VisibleForTesting static final TagKey GRPC_SERVER_STATUS = TagKey.create("grpc_server_status");
+  @VisibleForTesting static final TagKey GRPC_SERVER_METHOD = TagKey.create("grpc_server_method");
   private static final Map<TagKey, TagKey[]> RPC_TAG_MAPPINGS =
       ImmutableMap.<TagKey, TagKey[]>builder()
           .put(RPC_STATUS, new TagKey[] {GRPC_CLIENT_STATUS, GRPC_SERVER_STATUS})
--- a/impl_core/src/main/java/io/opencensus/implcore/stats/RecordUtils.java
+++ b/impl_core/src/main/java/io/opencensus/implcore/stats/RecordUtils.java
@@ -17,9 +17,8 @@
 package io.opencensus.implcore.stats;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.collect.ImmutableMultimap;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
-import com.google.common.collect.Multimap;
 import io.opencensus.common.Function;
 import io.opencensus.common.Functions;
 import io.opencensus.implcore.stats.MutableAggregation.MutableCount;
@@ -71,12 +70,10 @@ final class RecordUtils {
   @VisibleForTesting static final TagKey GRPC_CLIENT_METHOD = TagKey.create("grpc_client_method");
   private static final TagKey GRPC_SERVER_STATUS = TagKey.create("grpc_server_status");
   private static final TagKey GRPC_SERVER_METHOD = TagKey.create("grpc_server_method");
-  private static final Multimap<TagKey, TagKey> RPC_TAG_MAPPINGS =
-      ImmutableMultimap.<TagKey, TagKey>builder()
-          .put(RPC_STATUS, GRPC_CLIENT_STATUS)
-          .put(RPC_STATUS, GRPC_SERVER_STATUS)
-          .put(RPC_METHOD, GRPC_CLIENT_METHOD)
-          .put(RPC_METHOD, GRPC_SERVER_METHOD)
+  private static final Map<TagKey, TagKey[]> RPC_TAG_MAPPINGS =
+      ImmutableMap.<TagKey, TagKey[]>builder()
+          .put(RPC_STATUS, new TagKey[] {GRPC_CLIENT_STATUS, GRPC_SERVER_STATUS})
+          .put(RPC_METHOD, new TagKey[] {GRPC_CLIENT_METHOD, GRPC_SERVER_METHOD})
           .build();
 
   static Map<TagKey, TagValueWithMetadata> getTagMap(TagContext ctx) {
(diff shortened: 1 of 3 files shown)


impl_core/src/main/java/io/opencensus/implcore/stats/RecordUtils.java
```diff
@@ -83,15 +100,29 @@
...
         tagValues.add(tags.get(tagKey).getTagValue());
       }
     }
     return tagValues;
   }
 
+  // TODO(songy23): remove the mapping once we completely remove the deprecated RPC constants.
+  private static TagValue getTagValueForDeprecatedRpcTag(
+      Map<? extends TagKey, TagValueWithMetadata> tags, TagKey oldKey) {
+    for (TagKey newKey : RPC_TAG_MAPPINGS.get(oldKey)) {
```
zhangkun83: @carl-mastrangelo, I think the for-each construct will create an Iterator, which you may not want to see. Maybe `RPC_TAG_MAPPINGS` should be defined as a `Map<TagKey, TagKey[]>` because for-each on an array presumably doesn't create an Iterator?

impl_core/src/test/java/io/opencensus/implcore/stats/RecordUtilsTest.java
```diff
@@ -83,6 +86,19 @@ public void testGetTagValues() {
         .inOrder();
   }
 
+  @Test
+  public void testGetTagValues_MapDeprecatedRpcTag() {
+    List<TagKey> columns = Arrays.asList(RecordUtils.RPC_STATUS, RecordUtils.RPC_METHOD);
+    Map<TagKey, TagValueWithMetadata> tags =
+        ImmutableMap.of(
+            RecordUtils.GRPC_CLIENT_METHOD, METHOD_V_WITH_MD,
+            RecordUtils.GRPC_CLIENT_STATUS, STATUS_V_WITH_MD);
```
dinooliva: What happens when we have both a client method and a server method?

songy23: Depending on what tag keys we have in the view:
1. If we have the new "client_method" or "server_method" key (or even both), everything works fine;
2. If we have the old "method" key, a random value will be picked. While without this mapping the value will always be null.

IMO this should be less an issue because eventually users should migrate to use the new keys. WDYT?

impl_core/src/main/java/io/opencensus/implcore/stats/RecordUtils.java
```diff
@@ -83,15 +97,34 @@
     for (int i = 0; i < columns.size(); ++i) {
       TagKey tagKey = columns.get(i);
       if (!tags.containsKey(tagKey)) {
-        // replace not found key values by null.
-        tagValues.add(UNKNOWN_TAG_VALUE);
+        @javax.annotation.Nullable TagValue tagValue = UNKNOWN_TAG_VALUE;
+        if (RPC_TAG_MAPPINGS.containsKey(tagKey)) {
```
carl-mastrangelo: nit: this is a double lookup of the key.  It would be better to just get() it, and check for Null

impl_core/src/main/java/io/opencensus/implcore/stats/RecordUtils.java
```diff
@@ -83,15 +97,34 @@
...
+  @javax.annotation.Nullable
+  private static TagValue getTagValueForDeprecatedRpcTag(
+      Map<? extends TagKey, TagValueWithMetadata> tags, TagKey oldKey) {
+    TagKey[] newKeys = RPC_TAG_MAPPINGS.get(oldKey);
+    if (newKeys == null) { // fix checker framework
+      return UNKNOWN_TAG_VALUE;
+    }
+    for (TagKey newKey : newKeys) {
+      if (tags.containsKey(newKey)) {
+        return tags.get(newKey).getTagValue();
```
zhangkun83: nit: this is also a double lookup.

impl_core/src/test/java/io/opencensus/implcore/stats/RecordUtilsTest.java
zhangkun83: By looking at `getTagValueForDeprecatedRpcTag()`, it seems the underlying data (columns?) is still in the old format, thus `getTagValueForDeprecatedRpcTag()` tries to get the values from the new tags in case the old tags are not set. If both "client_method" or "server_method" are set, "client_method" will be used. Is it going to be the case even after the old tags are removed?

dinooliva: So, iiuc, if we have a view with and old key (e.g. "rpc_method") and tags 

1. contains tag rpc_method:v1 then v1 is used
2. otherwise, contains tag grpc_client_method:v2 then v2 is used
3. otherwise contains tag grpc_server_method:v3 then v3 is used
4. otherwise UNKNOWN_TAG_VALUE is used

Is that correct? If so, it may be worth enumerating in the unit tests. Also, when we have both grpc_client_method and grpc_server_method, which one matches closer to the original notion of rpc_method?

impl_core/src/main/java/io/opencensus/implcore/stats/RecordUtils.java
songy23: Done, good catch.

impl_core/src/main/java/io/opencensus/implcore/stats/RecordUtils.java
songy23: Fixed.

impl_core/src/test/java/io/opencensus/implcore/stats/RecordUtilsTest.java
songy23: > If both "client_method" or "server_method" are set, "client_method" will be used.

Apologies for the confusion, what Kun said is correct. If both are present and a view is using the old column, client method value will be picked.

> if we have a view with and old key (e.g. "rpc_method") and tags

Yes the order is correct. I added a few tests on this.

songy23: > If both "client_method" or "server_method" are set, "client_method" will be used. Is it going to be the case even after the old tags are removed?

No, the new "client_method" and "server_method" tags will be matched properly regardless of the old tag. This change is only trying to get the values from the new tags for old tag if it's still in use. Added one more test in https://github.com/census-instrumentation/opencensus-java/pull/1854/commits/cec1f5ac0c74309f7bfc14907b610b9a560a4247 to verify.

impl_core/src/test/java/io/opencensus/implcore/stats/RecordUtilsTest.java
```diff
@@ -83,6 +92,61 @@ public void testGetTagValues() {
...
+
+  @Test
+  public void testGetTagValues_MapDeprecatedRpcTag_PreferClientTag() {
+    List<TagKey> columns = Arrays.asList(RecordUtils.RPC_STATUS, RecordUtils.RPC_METHOD);
+    Map<TagKey, TagValueWithMetadata> tags =
+        ImmutableMap.of(
+            RecordUtils.GRPC_SERVER_METHOD, METHOD_V_WITH_MD,
+            RecordUtils.GRPC_SERVER_STATUS, STATUS_V_WITH_MD,
+            RecordUtils.GRPC_CLIENT_METHOD, METHOD_V_2_WITH_MD,
+            RecordUtils.GRPC_CLIENT_STATUS, STATUS_V_2_WITH_MD);
```
dinooliva: Yes - these tests clarify what's going on, thanks@ - you may want to add a test that contains GRPC_SERVER_METHOD, GRPC_CLIENT_METHOD, and RPC_METHOD to verify that the RPC_METHOD tag value does not unexpectedly get overridden by the other tags but I'll leave that up to you.

songy23: Good idea, added one in https://github.com/census-instrumentation/opencensus-java/pull/1854/commits/9e371bafa42be0b43b7f0a9f111de0c3c9ea89db.

songy23:
Thanks for the review!

Issue #1863, opened 2019-04-25
Don't expose gRPC context key directly. Use APIs instead.


songy23:
Otherwise whenever we migrated to a new context key, it will break context propagation. See https://github.com/bogdandrutu/openconsensus/pull/211.

bogdandrutu:
Can we get this in the new release please :)

PR #1864, opened 2019-04-25
Restrict access to context key, add helper methods to interact with Context.

songy23:
Fixes https://github.com/census-instrumentation/opencensus-java/issues/1863.

Similar to https://github.com/bogdandrutu/openconsensus/pull/211.

/cc @zhangkun83

### DIFF
--- a/api/src/main/java/io/opencensus/tags/unsafe/ContextUtils.java
+++ b/api/src/main/java/io/opencensus/tags/unsafe/ContextUtils.java
@@ -43,7 +43,7 @@ public final class ContextUtils {
    *
    * @since 0.8
    * @deprecated from API since 0.21. Use {@link #withValue(Context, TagContext)} and {@link
-   * #getValue(Context)} instead.
+   *     #getValue(Context)} instead.
    */
   // TODO(songy23): make this private once gRPC migrates to use the alternative APIs.
   @Deprecated
--- a/api/src/main/java/io/opencensus/tags/unsafe/ContextUtils.java
+++ b/api/src/main/java/io/opencensus/tags/unsafe/ContextUtils.java
@@ -17,6 +17,7 @@
 package io.opencensus.tags.unsafe;
 
 import io.grpc.Context;
+import io.opencensus.internal.Utils;
 import io.opencensus.tags.Tag;
 import io.opencensus.tags.TagContext;
 import java.util.Collections;
@@ -41,9 +42,36 @@ public final class ContextUtils {
    * {@link io.grpc.Context}.
    *
    * @since 0.8
+   * @deprecated from API since 0.21. Use {@link #withValue(Context, TagContext)} and {@link
+   * #getValue(Context)} instead.
    */
+  // TODO(songy23): make this private once gRPC migrates to use the alternative APIs.
+  @Deprecated
   public static final Context.Key<TagContext> TAG_CONTEXT_KEY =
       Context.keyWithDefault("opencensus-tag-context-key", EMPTY_TAG_CONTEXT);
+
+  /**
+   * Creates a new {@code Context} with the given value set.
+   *
+   * @param context the parent {@code Context}.
+   * @param tagContext the value to be set.
+   * @return a new context with the given value set.
+   * @since 0.21
+   */
+  public static Context withValue(Context context, TagContext tagContext) {
+    return Utils.checkNotNull(context, "context").withValue(TAG_CONTEXT_KEY, tagContext);
+  }
+
+  /**
+   * Returns the value from the specified {@code Context}.
+   *
+   * @param context the specified {@code Context}.
+   * @return the value from the specified {@code Context}.
+   * @since 0.21
+   */
+  public static TagContext getValue(Context context) {
+    return TAG_CONTEXT_KEY.get(context);
+  }
 
   @Immutable
   private static final class EmptyTagContext extends TagContext {
--- a/api/src/main/java/io/opencensus/tags/unsafe/ContextUtils.java
+++ b/api/src/main/java/io/opencensus/tags/unsafe/ContextUtils.java
@@ -24,6 +24,10 @@ import java.util.Collections;
 import java.util.Iterator;
 import javax.annotation.concurrent.Immutable;
 
+/*>>>
+import org.checkerframework.checker.nullness.qual.Nullable;
+*/
+
 /**
  * Utility methods for accessing the {@link TagContext} contained in the {@link io.grpc.Context}.
  *
@@ -47,7 +51,7 @@ public final class ContextUtils {
    */
   // TODO(songy23): make this private once gRPC migrates to use the alternative APIs.
   @Deprecated
-  public static final Context.Key<TagContext> TAG_CONTEXT_KEY =
+  public static final Context.Key</*@Nullable*/ TagContext> TAG_CONTEXT_KEY =
       Context.keyWithDefault("opencensus-tag-context-key", EMPTY_TAG_CONTEXT);
 
   /**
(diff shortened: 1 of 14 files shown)


api/src/main/java/io/opencensus/trace/unsafe/ContextUtils.java
```diff
@@ -39,7 +40,35 @@ private ContextUtils() {}
...
    *
    * @since 0.5
+   * @deprecated from API since 0.21. Use {@link #withValue(Context, Span)} and {@link
+   * #getValue(Context)} instead.
    */
+  // TODO(songy23): make this private once gRPC migrates to use the alternative APIs.
+  @Deprecated
   public static final Context.Key</*@Nullable*/ Span> CONTEXT_SPAN_KEY =
-      Context.key("opencensus-trace-span-key");
+      Context.<Span>key("opencensus-trace-span-key");
```
songy23: Default value for context span key is still `null` for backwards-compatibility.

bogdandrutu: Who needs it? gRPC does not need to be null.

songy23: gRPC expected this behavior in their tests, e.g https://github.com/grpc/grpc-java/blob/master/core/src/test/java/io/grpc/internal/CensusModulesTest.java#L306. I'm not sure if this is used anywhere else. /cc @zhangkun83

api/src/main/java/io/opencensus/tags/unsafe/ContextUtils.java
```diff
@@ -41,10 +42,37 @@ private ContextUtils() {}
...
 
+  /**
+   * Creates a new {@code Context} with the given value set.
+   *
+   * @param context the parent {@code Context}.
+   * @param tagContext the value to be set.
+   * @return a new context with the given value set.
+   * @since 0.21
+   */
+  public static Context withValue(Context context, TagContext tagContext) {
```
dinooliva: Shouldn't you check whether tagContext is null? Or annotate as nullable?

api/src/main/java/io/opencensus/tags/unsafe/ContextUtils.java
```diff
@@ -41,10 +42,37 @@ private ContextUtils() {}
...
+  }
+
+  /**
+   * Returns the value from the specified {@code Context}.
+   *
+   * @param context the specified {@code Context}.
+   * @return the value from the specified {@code Context}.
+   * @since 0.21
+   */
+  public static TagContext getValue(Context context) {
```
dinooliva: Maybe these names should be less generic since they are doing specific work - e.g. withTagContext and getTagContext

api/src/main/java/io/opencensus/tags/unsafe/ContextUtils.java
bogdandrutu: No, because the key is initialized with a default value.

api/src/main/java/io/opencensus/tags/unsafe/ContextUtils.java
bogdandrutu: I would keep them as they are to not make confusion with the main APIs.

api/src/main/java/io/opencensus/trace/unsafe/ContextUtils.java
bogdandrutu: But we don't change the behavior for the key, only for the new API which we can implement whatever we want.

api/src/main/java/io/opencensus/tags/unsafe/ContextUtils.java
```diff
@@ -41,10 +42,37 @@ private ContextUtils() {}
...
+  /**
+   * Creates a new {@code Context} with the given value set.
+   *
+   * @param context the parent {@code Context}.
+   * @param tagContext the value to be set.
+   * @return a new context with the given value set.
+   * @since 0.21
+   */
+  public static Context withValue(Context context, TagContext tagContext) {
+    return Utils.checkNotNull(context, "context").withValue(TAG_CONTEXT_KEY, Utils.checkNotNull(tagContext, "tagContext"));
```
bogdandrutu: tagContext may be null. no need to check.

api/src/main/java/io/opencensus/trace/unsafe/ContextUtils.java
```diff
@@ -39,7 +37,34 @@ private ContextUtils() {}
    * The {@link io.grpc.Context.Key} used to interact with {@link io.grpc.Context}.
    *
    * @since 0.5
+   * @deprecated from API since 0.21. Use {@link #withValue(Context, Span)} and {@link
+   *     #getValue(Context)} instead.
+   */
+  // TODO(songy23): make this private once gRPC migrates to use the alternative APIs.
+  @Deprecated
+  public static final Context.Key<Span> CONTEXT_SPAN_KEY =
+      Context.<Span>keyWithDefault("opencensus-trace-span-key", BlankSpan.INSTANCE);
```
bogdandrutu: we cannot change the default for the key.

api/src/main/java/io/opencensus/trace/unsafe/ContextUtils.java
```diff
@@ -39,7 +37,34 @@ private ContextUtils() {}
...
+   * Returns the value from the specified {@code Context}.
+   *
+   * @param context the specified {@code Context}.
+   * @return the value from the specified {@code Context}.
+   * @since 0.21
    */
-  public static final Context.Key</*@Nullable*/ Span> CONTEXT_SPAN_KEY =
-      Context.key("opencensus-trace-span-key");
+  public static Span getValue(Context context) {
+    return CONTEXT_SPAN_KEY.get(Utils.checkNotNull(context, "context"));
```
bogdandrutu: here we implement the default logic.

api/src/main/java/io/opencensus/trace/unsafe/ContextUtils.java
```diff
@@ -39,7 +37,34 @@ private ContextUtils() {}
...
+  /**
+   * Creates a new {@code Context} with the given value set.
+   *
+   * @param context the parent {@code Context}.
+   * @param span the value to be set.
+   * @return a new context with the given value set.
+   * @since 0.21
+   */
+  public static Context withValue(Context context, Span span) {
+    return Utils.checkNotNull(context, "context").withValue(CONTEXT_SPAN_KEY, Utils.checkNotNull(span, "span"));
```
bogdandrutu: no need to check for not null for the span, because we can support null Span correct?

api/src/main/java/io/opencensus/tags/unsafe/ContextUtils.java
songy23: Reverted, though I'm wondering when we would want to set `TagContext` or `Span` to null in a context.

api/src/main/java/io/opencensus/tags/unsafe/ContextUtils.java
songy23: Reverted.

api/src/main/java/io/opencensus/trace/unsafe/ContextUtils.java
songy23: Makes sense, fixed.

api/src/main/java/io/opencensus/trace/unsafe/ContextUtils.java
songy23: Fixed.

api/src/main/java/io/opencensus/trace/unsafe/ContextUtils.java
songy23: Updated.

api/src/main/java/io/opencensus/tags/unsafe/ContextUtils.java
dinooliva: Agreed that the default value does mean the tagContext can be null but still not clear to me that there is a good use-case for this.

If there is, shouldn't it be annotated Nullable?

songy23: I added the Nullable annotation to the context value. The new API still returns non-null value though.

Issue #1841, opened 2019-04-11
Stackdriver metric units missing


mwuertinger:
### What version of OpenCensus are you using?
0.20.0

### What JVM are you using (`java -version`)?
openjdk version "11.0.2" 2019-01-15
OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.2+9)
OpenJDK 64-Bit Server VM AdoptOpenJDK (build 11.0.2+9, mixed mode)

### What did you do?
I'm exporting the default gRPC metrics to Stackdriver:
```
StackdriverStatsConfiguration.Builder builder = StackdriverStatsConfiguration.builder();
StackdriverStatsExporter.createAndRegister(builder.build());
RpcViews.registerAllViews();
```

### What did you expect to see?
I expect the units of the metrics to be set accordingly (eg. `By`, `ms`, `1`).

### What did you see instead?
There are no units set for the metrics.

### Workaround
Downgrading to OpenCensus 0.19.1 resolves the issue.

songy23:
Spent some time looking into this issue today. It turns out in v0.20.0 the Stackdriver `MetricDescriptor` is created from time series instead of metric descriptors. 

A bit more background: in Stackdriver there're two ways on creating a `MetricDescriptor`:
1. Explicitly call the `createMetricDescriptor` API (https://cloud.google.com/monitoring/custom-metrics/creating-metrics#call_the_create_method). With this approach `unit` and `description` can be specified for a `MetricDescriptor`. The OpenCensus Stackdriver exporter should have created `MetricDescriptor` with this approach, but somehow in v0.20.0 it didn't work and fall back to the second approach below.
2. Implicitly auto-creating it with uploaded time series (https://cloud.google.com/monitoring/custom-metrics/creating-metrics#auto-creation). `MetricDescriptor`s created in this way won't have units.

For example, when I tried to export the same metric with exporters in 0.19.2 and 0.20.0:

With v0.19.2 I got a `MetricDescriptor`:
```json
{
   "name": "xxx",
   "labels": ["xxx"],
   "metricKind": "CUMULATIVE",
   "valueType": "DISTRIBUTION",
   "description": "test Stackdriver metric",
   "type": "xxx",
   "unit": "ms",
}
```

With v0.20.0 I got a `MetricDescriptor`:
```json
{
 "name": "xxx",
 "labels": ["xxx"],
 "metricKind": "CUMULATIVE",
 "valueType": "DISTRIBUTION",
[...]

PR #1866, opened 2019-04-29
Exporter/Stackdriver: Do create metric descriptor for custom metrics.

songy23:
Fixes https://github.com/census-instrumentation/opencensus-java/issues/1841.

Previously metric names like "my.org/metric" are treated as Stackdriver built-in metrics by accident. This PR fixes this bug.

### DIFF
--- a/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/CreateMetricDescriptorExporter.java
+++ b/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/CreateMetricDescriptorExporter.java
@@ -44,7 +44,7 @@ final class CreateMetricDescriptorExporter extends MetricExporter {
   private static final Logger logger =
       Logger.getLogger(CreateMetricDescriptorExporter.class.getName());
   private static final ImmutableSet<String> SUPPORTED_EXTERNAL_DOMAINS =
-      ImmutableSet.<String>of("custom.googleapis.com/", "external.googleapis.com/");
+      ImmutableSet.<String>of("custom.googleapis.com", "external.googleapis.com");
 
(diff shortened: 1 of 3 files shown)


exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/CreateMetricDescriptorExporter.java
```diff
@@ -153,12 +153,17 @@ public void export(Collection<Metric> metrics) {
...
-      if (metricName.startsWith(domain)) {
-        return true;
-      }
+  private static boolean isBuiltInMetric(String metricName) {
+    int domainIndex = metricName.indexOf('/');
+    if (domainIndex < 0) {
+      return false;
+    }
+    String metricDomain = metricName.substring(0, domainIndex);
+    if (!metricDomain.contains("googleapis.com")) {
```
bogdandrutu: Why not just startsWith?

songy23: Built-in metrics are something like "bigquery.googleapis.com", "agent.googleapis.com", thus using `startsWith` "googleapis.com" won't work since the first few characters (indicate the product) can be different.

bogdandrutu: You are correct, but just for this we can probably use endsWith.

songy23: SGTM, updated.

Issue #1813, opened 2019-03-25
Configure a max number spans in RunningSpanStoreImpl


dmichel1:
Hi,

I recently was involved with an incident where a service was not properly closely a span when an error was raised. This caused the JVM to go OOM after a few hours.

Do you have any thoughts or suggestions on how to prevent something similar from happening again? Some configurable limit of how many spans can be stored in the RunningSpanStoreImpl?

![image](https://user-images.githubusercontent.com/677960/54950197-1663eb80-4f17-11e9-8fc0-73de5e2a2bf5.png)

songy23:
Unfortunately for now we don't have an API for configuring the buffer size of in-process running spans. In the implementation running span store is an unbounded list:
https://github.com/census-instrumentation/opencensus-java/blob/f0e78ba5a10165cce2b63809bfdfe3e34b1eb456/impl_core/src/main/java/io/opencensus/implcore/trace/export/InProcessRunningSpanStoreImpl.java#L36

We could add an API in `TraceConfig` to allow users to configure the limit of running spans. Before that I would suggest 1) don't keep too many spans running concurrently and 2) don't keep the lifetime of each span too long.

/cc @bogdandrutu WDYT?

bogdandrutu:
Couple of things we should do to help here:
1. RunningSpanStoreImpl should be a no-op if no z-pages installed.
1. Add a limit probably when the z-pages are configured that is passed to the RunningSpanStoreImpl.

These things may imply a bit of refactoring of the code, we should start talking about this. I think the first item should be easier to fix.

bogdandrutu:
@dmichel1 can you please tell me if you enable z-pages https://github.com/census-instrumentation/opencensus-java/tree/master/contrib/zpages or if you use the RunningSpanStore for other reasons?

I was thinking to make that disabled by default and if enabled then will keep a reference to all the Spans.

dmichel1:
@bogdandrutu we don't have zpages enabled.

Some services might pull in the `opencensus-contrib-zpage` dependencies for other reasons but only having the dependencies listed won't enable anything unless `PageHandlers.startHttpServerAndRegisterAll(8080);` is specifically called, right? 

I'm also not aware of any `RunningSpanStore` usage outside of the zpages.

songy23:
> only having the dependencies listed won't enable anything unless PageHandlers.startHttpServerAndRegisterAll(8080); is specifically called, right?

[...]

PR #1878, opened 2019-05-06
zPages: Enable RunningSpanStore only when /tracez is in use.

songy23:
A follow-up of https://github.com/census-instrumentation/opencensus-java/issues/1813#issuecomment-489685886.

Enable RunningSpanStore only when /tracez is in use, instead of always enabling it statically.

### DIFF
--- a/contrib/zpages/src/main/java/io/opencensus/contrib/zpages/ZPageHandlers.java
+++ b/contrib/zpages/src/main/java/io/opencensus/contrib/zpages/ZPageHandlers.java
@@ -25,6 +25,7 @@ import io.opencensus.stats.View;
 import io.opencensus.trace.Tracing;
 import java.io.IOException;
 import java.net.InetSocketAddress;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Logger;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.GuardedBy;
@@ -79,6 +80,7 @@ public final class ZPageHandlers {
       StatszZPageHandler.create(Stats.getViewManager());
 
   private static final Object monitor = new Object();
+  private static final AtomicBoolean isRunningSpanStoreInitialized = new AtomicBoolean(false);
 
   @GuardedBy("monitor")
   @Nullable
@@ -94,8 +96,9 @@ public final class ZPageHandlers {
    * <p>If no sampled spans based on latency and error codes are available for a given name, make
    * sure that the span name is registered to the {@code SampledSpanStore}.
    *
-   * <p>When this method is called, {@link io.opencensus.trace.export.RunningSpanStore} will be
-   * enabled automatically.
+   * <p>When this method is called for the first time, {@link
+   * io.opencensus.trace.export.RunningSpanStore} will be enabled automatically. Subsequent calls
+   * won't update {@link io.opencensus.trace.export.RunningSpanStore} again.
    *
    * @return a {@code ZPageHandler} for tracing debug.
    * @since 0.6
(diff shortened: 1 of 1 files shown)


contrib/zpages/src/main/java/io/opencensus/contrib/zpages/ZPageHandlers.java
```diff
@@ -148,6 +147,7 @@ public static ZPageHandler getStatszZPageHandler() {
    * @since 0.6
    */
   public static void registerAllToHttpServer(HttpServer server) {
+    enableRunningSpanStore();
```
bogdandrutu: Don't need to call it from here because getTracezPageHandler will be called and will enable this.

contrib/zpages/src/main/java/io/opencensus/contrib/zpages/ZPageHandlers.java
```diff
@@ -99,10 +94,14 @@
...
    * sure that the span name is registered to the {@code SampledSpanStore}.
    *
+   * <p>When this method is called, {@link io.opencensus.trace.export.RunningSpanStore} will be
+   * enabled automatically.
+   *
    * @return a {@code ZPageHandler} for tracing debug.
    * @since 0.6
    */
   public static ZPageHandler getTracezZPageHandler() {
+    enableRunningSpanStore();
```
bogdandrutu: Currently setting the max number of spans resets the store. Probably we should call enable only once.

dinooliva: Could this end up running multiple times (i.e. could/would an application call this method multiple times)?

songy23: Make sense, updated in https://github.com/census-instrumentation/opencensus-java/pull/1878/commits/5d347352a5b7280acabaca1701609c30198b978d.

songy23: It should only be executed once, updated in https://github.com/census-instrumentation/opencensus-java/pull/1878/commits/5d347352a5b7280acabaca1701609c30198b978d.

contrib/zpages/src/main/java/io/opencensus/contrib/zpages/ZPageHandlers.java
songy23: Removed.

contrib/zpages/src/main/java/io/opencensus/contrib/zpages/ZPageHandlers.java
```diff
@@ -201,5 +203,13 @@ private static void stop() {
     }
   }
 
+  // Sets the maximum number of elements as Integer.MAX_VALUE to enable RunningSpanStore.
+  // This method will only execute once even if called multiple times.
+  private static void enableRunningSpanStore() {
+    if (!(isRunningSpanStoreInitialized.getAndSet(true))) {
```
bogdandrutu:
```java
private volatile boolean isRunningSpanStoreInitialized = false;

private static void enableRunningSpanStore() {
  if (!isRunningSpanStoreInitialized) {
    synchronize(this) {
      if (isRunningSpanStoreInitialized) {
        return;  // Already initialized, small race
      }
      Tracing.getExportComponent().getRunningSpanStore().setMaxNumberOfSpans(Integer.MAX_VALUE);
      isRunningSpanStoreInitialized = true;
    }
  }
}

```

dinooliva: lgtm

songy23: Updated. (Note the variables and locks are static.)

Issue #1880, opened 2019-05-06
Exporters: update all exporters to take a Deadline option and don't retry.


songy23:
- [x] Stackdriver Trace
- [x] Zipkin
- [x] Jaeger
- [x] Instana
- [x] Datadog
- [x] Elasticsearch
- [x] Stackdriver Monitoring
- [ ] OC-Agent

bogdandrutu:
Prometheus is pull so no need to update that.

bogdandrutu:
@songy23 I think this should be P0 for the moment.

songy23:
Marked as P0 for now, will downgrade to P1 after SD is fixed.

rghetia:
> Marked as P0 for now, will downgrade to P1 after SD is fixed.

@songy23 is this only required for Java?

bogdandrutu:
@rghetia we have a clear problem with memory in Java, after we fix this we will see how we should modify all the languages.

songy23:
To add to that, according to https://github.com/census-instrumentation/opencensus-java/issues/1767#issuecomment-489372100 the memory issue is only with exporting spans. Metrics export is less a problem because only the exporters have references to the metrics. Once metrics exporters are blocked, no new metrics will be generated so memory usage would stay the same (while for span export, undelivered spans can grow unbounded).

PR #1881, opened 2019-05-06
Exporter/Trace/Stackdriver: Provide Deadline option and don't do retry by default.

songy23:
Stackdriver Trace part of https://github.com/census-instrumentation/opencensus-java/issues/1880.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased
 - Disable RunningSpanStore by default unless the z-pages contrib is linked and initialized. This may
 break behaviors for applications that rely on this to be always enabled.
+- Provide a `Deadline` option to Stackdriver Trace exporter.
 
 ## 0.21.0 - 2019-04-30
 - Add HTTP text format serializer to Tag propagation component.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## Unreleased
 - Disable RunningSpanStore by default unless the z-pages contrib is linked and initialized. This may
 break behaviors for applications that rely on this to be always enabled.
-- Provide a `Deadline` option to Stackdriver Trace exporter.
+- Provide a `Deadline` option to Stackdriver Trace exporter. Default value is 10 seconds if it's not set.
 
 ## 0.21.0 - 2019-04-30
 - Add HTTP text format serializer to Tag propagation component.
--- a/exporters/trace/stackdriver/src/main/java/io/opencensus/exporter/trace/stackdriver/StackdriverV2ExporterHandler.java
+++ b/exporters/trace/stackdriver/src/main/java/io/opencensus/exporter/trace/stackdriver/StackdriverV2ExporterHandler.java
@@ -19,6 +19,7 @@ package io.opencensus.exporter.trace.stackdriver;
 import static com.google.api.client.util.Preconditions.checkNotNull;
 
 import com.google.api.gax.core.FixedCredentialsProvider;
+import com.google.api.gax.retrying.RetrySettings;
 import com.google.auth.Credentials;
 import com.google.cloud.trace.v2.TraceServiceClient;
 import com.google.cloud.trace.v2.TraceServiceSettings;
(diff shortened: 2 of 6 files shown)


exporters/trace/stackdriver/src/main/java/io/opencensus/exporter/trace/stackdriver/StackdriverV2ExporterHandler.java
```diff
@@ -171,15 +173,42 @@ static StackdriverV2ExporterHandler createWithStub(
...
+       *       .setMaxRetryDelay(Duration.ofMillis(1000L))
+       *       .setInitialRpcTimeout(Duration.ofMillis(30000L))
+       *       .setRpcTimeoutMultiplier(1.5)
+       *       .setMaxRpcTimeout(Duration.ofMillis(60000L))
+       *       .setTotalTimeout(Duration.ofMillis(120000L))
+       *       .build();
+       *
+       * Override the default settings with settings that don't retry.
+       */
+      builder
```
dinooliva: So, iiuc, after this change there will be no retry for exporting traces and they will be dropped unless users set an explicit deadline, at which point they will get the default retry settings - is that about right?

songy23: Partially correct, when users set an explicit deadline still no retries will be performed (`setSimpleTimeoutNoRetries`). It's the expected behavior that not to do retry by default, since by doing retries we hold references to undelivered spans longer, thus it makes memory usage worse. Note users always have an option to use a custom Stackdriver Trace client: https://github.com/census-instrumentation/opencensus-java/blob/4475dd9ec123e0378bcb4726fc261e6fe233daf4/exporters/trace/stackdriver/src/main/java/io/opencensus/exporter/trace/stackdriver/StackdriverTraceConfiguration.java#L127, which they can configure to use other retry settings.

Maybe we can also consider adding a top-level `retrySettings` option but I'd prefer to wait and see if it's really needed.

dinooliva: Got it - I think that's a good approach.

exporters/trace/stackdriver/src/main/java/io/opencensus/exporter/trace/stackdriver/StackdriverV2ExporterHandler.java
```diff
@@ -171,15 +173,42 @@ static StackdriverV2ExporterHandler createWithStub(
...
+          .batchWriteSpansSettings()
+          .setRetryableCodes() // Sets to empty so no retryable code will be used.
+          // RetrySettings.newBuilder().build() returns settings with no retries.
+          .setRetrySettings(RetrySettings.newBuilder().build());
```
bogdandrutu: What is the default Deadline if non provided by the user?

songy23: Total timeout is 120000 ms (2 min). See https://github.com/census-instrumentation/opencensus-java/pull/1881/files#diff-2f7e8b7487a68c644aef32903d35ca13R199.

bogdandrutu: So the deadline will be setInitialRpcTimeout(60sec) because no retries? Can we explicitly set that? Can we also make it 10sec?

songy23: > So the deadline will be setInitialRpcTimeout(60sec) because no retries? 

Apologies I forgot to mention 2min is the default deadline **before** this PR. After this PR if users don't set deadline, the default settings will be no retries and no timeout (see https://github.com/googleapis/gax-java/blob/master/gax/src/main/java/com/google/api/gax/retrying/RetrySettings.java#L137-L148). So if the initial RPC failed, all spans are immediately dropped.

> Can we explicitly set that? Can we also make it 10sec?

Yes and yes. Made the change in https://github.com/census-instrumentation/opencensus-java/pull/1881/commits/62aa27a4e47a9fe14f4cdfffa9b91b7655b24de9.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## Unreleased
 - Disable RunningSpanStore by default unless the z-pages contrib is linked and initialized. This may
 break behaviors for applications that rely on this to be always enabled.
+- Provide a `Deadline` option to Stackdriver Trace exporter.
```
bogdandrutu: Update this to reflect the default.

songy23: Done.

PR #1882, opened 2019-05-06
Export SpanData in batches of maximum bufferSize.

bogdandrutu:


impl_core/src/main/java/io/opencensus/implcore/trace/export/SpanExporterImpl.java
```diff
@@ -187,27 +177,43 @@ public void run() {
...
+        spansCopy = new ArrayList<>(spans);
         spans.clear();
       }
+      // Execute the batch export outside the synchronized to not block all producers.
+      exportBatches(spansCopy);
+    }
 
-      final List<SpanData> spanDataList = fromSpanImplToSpanData(spansCopy);
+    private void exportBatches(ArrayList<RecordEventsSpanImpl> spanList) {
+      ArrayList<SpanData> spanDataList = new ArrayList<>(bufferSize);
```
songy23: Why not just use `Lists.partition`?

songy23: Never mind, forgot we don't want to depend on Guava here.

impl_core/src/main/java/io/opencensus/implcore/trace/export/SpanExporterImpl.java
```diff
@@ -187,27 +177,43 @@ public void run() {
...
-        spansCopy = new ArrayList<RecordEventsSpanImpl>(spans);
+        spansCopy = new ArrayList<>(spans);
         spans.clear();
       }
+      // Execute the batch export outside the synchronized to not block all producers.
+      exportBatches(spansCopy);
+    }
 
-      final List<SpanData> spanDataList = fromSpanImplToSpanData(spansCopy);
+    private void exportBatches(ArrayList<RecordEventsSpanImpl> spanList) {
```
dinooliva: It might be worth noting that spanList is assumed not to be shared across other threads.

impl_core/src/main/java/io/opencensus/implcore/trace/export/SpanExporterImpl.java
bogdandrutu: Not only that, but we want to remove the reference to the RecordEventsSpanImpl asap, with Lists.partitions they use List.subList which is implemented as a wrapper on top of the original list also immutable so it is harder to remove the reference to that object.

PR #1883, opened 2019-05-07
Exporter/Stats/Stackdriver: Provide Deadline option and don't do retry by default.

songy23:
Stackdriver Monitoring part of #1880.

### DIFF
--- a/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporter.java
+++ b/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporter.java
@@ -36,6 +36,7 @@ import com.google.auth.oauth2.GoogleCredentials;
 import com.google.cloud.ServiceOptions;
 import com.google.cloud.monitoring.v3.MetricServiceClient;
 import com.google.cloud.monitoring.v3.MetricServiceSettings;
+import com.google.cloud.monitoring.v3.stub.MetricServiceStub;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.monitoring.v3.CreateMetricDescriptorRequest;
 import com.google.monitoring.v3.CreateTimeSeriesRequest;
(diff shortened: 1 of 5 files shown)


exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/StackdriverStatsExporter.java
```diff
@@ -365,6 +391,38 @@ static MetricServiceClient createMetricServiceClient(@Nullable Credentials crede
...
+        createMetricDescriptorSettings = settingsBuilder.createMetricDescriptorSettings();
+    UnaryCallSettings.Builder<CreateTimeSeriesRequest, Empty> createTimeSeriesSettings =
+        settingsBuilder.createTimeSeriesSettings();
+    if (deadline != null) {
+      org.threeten.bp.Duration stackdriverDuration =
+          org.threeten.bp.Duration.ofMillis(deadline.toMillis());
+      createMetricDescriptorSettings.setSimpleTimeoutNoRetries(stackdriverDuration);
+      createTimeSeriesSettings.setSimpleTimeoutNoRetries(stackdriverDuration);
+    } else {
+      /*
```
dinooliva: Similarly with tracing - iiuc, the only way to get the default retry settings is to explicitly set the deadline - will that negatively impact any customers, i.e. are there customers that have been relying on retries?

songy23: Similar comment to https://github.com/census-instrumentation/opencensus-java/pull/1881#discussion_r281708276. I'll add a commit to allow users to use a custom `MetricClient` so that they can use custom settings.

dinooliva: Sounds good.

songy23: Added the new option in https://github.com/census-instrumentation/opencensus-java/pull/1883/commits/730e37ae91e4c4c054c025c9b85947691015e200, PTAL.

PR #1884, opened 2019-05-08
Exporter/Jaeger: Use configurations and add deadline.

songy23:
Jaeger part of https://github.com/census-instrumentation/opencensus-java/issues/1880.

PR #1885, opened 2019-05-08
Exporter/Zipkin: Use configurations and add deadline.

songy23:
Zipkin part of https://github.com/census-instrumentation/opencensus-java/issues/1880.

Almost identical to #1884.

### DIFF
--- a/exporters/trace/zipkin/src/main/java/io/opencensus/exporter/trace/zipkin/ZipkinExporterConfiguration.java
+++ b/exporters/trace/zipkin/src/main/java/io/opencensus/exporter/trace/zipkin/ZipkinExporterConfiguration.java
@@ -1,0 +1,185 @@
+/*
+ * Copyright 2019, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


exporters/trace/zipkin/src/main/java/io/opencensus/exporter/trace/zipkin/ZipkinExporterConfiguration.java
```diff
@@ -0,0 +1,185 @@
...
+
+  /**
+   * Returns the {@link SpanBytesEncoder}.
+   *
+   * <p>Default is {@link SpanBytesEncoder#JSON_V2}.
+   *
+   * @return the {@code SpanBytesEncoder}
+   * @since 0.22
+   */
+  public abstract SpanBytesEncoder getEncoder();
```
dinooliva: This is a little surprising to me - why is it necessary here?

songy23: The encoder has been a required field before:
https://github.com/census-instrumentation/opencensus-java/blob/1dc78a87748ed5642ff560ab0ed35c2f1021170d/exporters/trace/zipkin/src/main/java/io/opencensus/exporter/trace/zipkin/ZipkinExporterHandler.java#L68

It's used here:
https://github.com/census-instrumentation/opencensus-java/blob/1dc78a87748ed5642ff560ab0ed35c2f1021170d/exporters/trace/zipkin/src/main/java/io/opencensus/exporter/trace/zipkin/ZipkinExporterHandler.java#L199

PR #1886, opened 2019-05-08
Exporter/Datadog, Elasticsearch: Add deadline option.

songy23:
Another two of https://github.com/census-instrumentation/opencensus-java/issues/1880.

I'm consider adding a common `SpanExportWithDeadline` interface to reduce the duplicated codes, similar to https://github.com/census-instrumentation/opencensus-java/tree/master/exporters/metrics/util. WDYT?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Monitoring client to make RPCs.
 with default value 10 seconds.
 - Use `ZipkinExporterConfiguration` for creating `ZipkinTraceExporter`. Provide a `Deadline` option
 with default value 10 seconds.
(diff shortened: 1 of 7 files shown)


dinooliva:
> Another two of #1880.
> 
> I'm consider adding a common `SpanExportWithDeadline` interface to reduce the duplicated codes, similar to https://github.com/census-instrumentation/opencensus-java/tree/master/exporters/metrics/util. WDYT?

Agreed that there is a lot of duplication of logic/code related to deadlines, it would be better if there was a common way of handling them.

songy23:
Sounds good, I'll send a follow-up PR.

PR #1887, opened 2019-05-09
Exporter/Trace: Extract timeout logic to a common util artifact.

songy23:
Updates https://github.com/census-instrumentation/opencensus-java/issues/1880.

The goal of this change is to reduce the code duplication added in recent exporter PRs (e.g #1884 #1885 #1886). This is similar to PR https://github.com/census-instrumentation/opencensus-java/pull/1647.

If this change looks good I'll go ahead and refactor other Trace exporters to use this util.

PR #1890, opened 2019-05-10
Exporter/Trace: Refactor to use TimeLimitedHandler.

songy23:
Updates #1880.

The goal of this change is to reduce the code duplication added in recent exporter PRs (e.g #1884 #1885 #1886), as a follow up of #1887.

### DIFF
--- a/buildscripts/import-control.xml
+++ b/buildscripts/import-control.xml
@@ -224,6 +224,7 @@ General guidelines on imports:
     </subpackage>
     <subpackage name="trace">
       <allow pkg="io.opencensus.trace"/>
+      <allow pkg="io.opencensus.exporter.trace.util"/>
       <subpackage name="instana">
         <allow pkg="io.opencensus.exporter.trace.instana"/>
       </subpackage>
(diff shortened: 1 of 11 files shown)


dmichel1:
hey @songy23 i'm curious about the deadlines you are adding to all the trace exporters. Is this to avoid blocking the `SpanExporterImplThread` since some of the exporters are running in that thread?

Can the Stackdriver exporter still block the `SpanExporterImplThread` even though the work is being delegated to the grpc thread pool?

songy23:
> Is this to avoid blocking the SpanExporterImplThread since some of the exporters are running in that thread?

Correct, see https://github.com/census-instrumentation/opencensus-java/issues/1767#issuecomment-489372100.

> Can the Stackdriver exporter still block the SpanExporterImplThread even though the work is being delegated to the grpc thread pool?

Stackdriver exporter still blocks the SpanExporterImplThread before deadline is exceeded. After that the exporter just drops any undelivered spans. This is the first step described in https://github.com/census-instrumentation/opencensus-java/issues/1767#issuecomment-489372100. The second step is to make exporters async (non-blocking).

dmichel1:
@songy23 makes sense thanks!

PR #1893, opened 2019-05-12
Start dropping spans when the span exporter thread holds more than 10K spans references.

bogdandrutu:
Fixes https://github.com/census-instrumentation/opencensus-java/issues/1767

### DIFF
--- a/impl_core/src/main/java/io/opencensus/implcore/trace/export/SpanExporterImpl.java
+++ b/impl_core/src/main/java/io/opencensus/implcore/trace/export/SpanExporterImpl.java
@@ -38,6 +38,10 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.annotation.concurrent.GuardedBy;
 
+/*>>>
+import org.checkerframework.checker.nullness.qual.Nullable;
+*/
+
 /** Implementation of the {@link SpanExporter}. */
 public final class SpanExporterImpl extends SpanExporter {
   private static final Logger logger = Logger.getLogger(ExportComponent.class.getName());
@@ -82,34 +86,6 @@ public final class SpanExporterImpl extends SpanExporter {
   static SpanExporterImpl create(int bufferSize, Duration scheduleDelay) {
     // TODO(bdrutu): Consider to add a shutdown hook to not avoid dropping data.
     final Worker worker = new Worker(bufferSize, scheduleDelay);
-    droppedSpans.createTimeSeries(
-        Collections.<LabelValue>emptyList(),
-        worker,
-        new ToLongFunction<Worker>() {
-          @Override
-          public long applyAsLong(Worker worker) {
-            return worker.getDroppedSpans();
-          }
-        });
-    referencedSpans.createTimeSeries(
-        Collections.<LabelValue>emptyList(),
-        worker,
-        new ToLongFunction<Worker>() {
-          @Override
-          public long applyAsLong(Worker worker) {
-            return worker.getReferencedSpans();
-          }
-        });
-    pushedSpans.createTimeSeries(
-        Collections.<LabelValue>emptyList(),
-        worker,
-        new ToLongFunction<Worker>() {
-          @Override
-          public long applyAsLong(Worker value) {
-            return worker.getPushedSpans();
-          }
-        });
-
     return new SpanExporterImpl(worker);
   }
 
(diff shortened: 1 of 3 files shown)


impl_core/src/main/java/io/opencensus/implcore/trace/export/SpanExporterImpl.java
```diff
@@ -154,6 +254,7 @@ private void onBatchExport(List<SpanData> spanDataList) {
     private Worker(int bufferSize, Duration scheduleDelay) {
       spans = new ArrayList<>(bufferSize);
       this.bufferSize = bufferSize;
+      this.maxReferencedSpans = 10L * bufferSize;
```
songy23: Any specific reason we chose this value? Do we want to make it configurable?

rghetia: I was wondering the same.

impl_core/src/main/java/io/opencensus/implcore/trace/export/SpanExporterImpl.java
```diff
@@ -51,7 +85,7 @@
    */
   static SpanExporterImpl create(int bufferSize, Duration scheduleDelay) {
     // TODO(bdrutu): Consider to add a shutdown hook to not avoid dropping data.
-    Worker worker = new Worker(bufferSize, scheduleDelay);
+    final Worker worker = new Worker(bufferSize, scheduleDelay);
```
dinooliva: Nit: why 'final'?

impl_core/src/main/java/io/opencensus/implcore/trace/export/SpanExporterImpl.java
```diff
@@ -32,9 +38,37 @@
...
 
+/*>>>
+import org.checkerframework.checker.nullness.qual.Nullable;
+*/
+
 /** Implementation of the {@link SpanExporter}. */
 public final class SpanExporterImpl extends SpanExporter {
   private static final Logger logger = Logger.getLogger(ExportComponent.class.getName());
+  private static final DerivedLongCumulative droppedSpans =
+      Metrics.getMetricRegistry()
```
dinooliva: Tracing impl now depends on Metrics - that seems fine - I prefer not to have cyclic dependencies but, iiuc, this doesn't introduce them. Interesting issue, tho - if we split the artifacts for tracing and metrics, this won't work.

bogdandrutu: It depends only on the API artifact not the impl artifact.

dinooliva: My second point was that a service that links in a tracing impl but not a metrics impl will get unexpected results - not an issue with the current way we package artifacts but could be a problem if we ever decide to split the implementation artifacts.

impl_core/src/main/java/io/opencensus/implcore/trace/export/SpanExporterImpl.java
bogdandrutu: Done.

impl_core/src/main/java/io/opencensus/implcore/trace/export/SpanExporterImpl.java
```diff
@@ -110,14 +195,29 @@ Thread getServiceExporterThread() {
...
+
+    private final Map<String, Handler> serviceHandlers = new ConcurrentHashMap<>();
     private final int bufferSize;
+    private final long maxReferencedSpans;
     private final long scheduleDelayMillis;
 
     // See SpanExporterImpl#addSpan.
     private void addSpan(RecordEventsSpanImpl span) {
       synchronized (monitor) {
+        if (referencedSpans + 1L > maxReferencedSpans) {
```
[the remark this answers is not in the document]
bogdandrutu: done.

impl_core/src/main/java/io/opencensus/implcore/trace/export/SpanExporterImpl.java
bogdandrutu: Added some explanation about why I decided to go with 4 * bufferSize in the end.

PR #1895, opened 2019-05-13
Exporter/OcAgent: Make required fields non-null.

songy23:


songy23:
> would it make sense to document (javadoc) the defaults?

Good idea, added.

Issue #1897, opened 2019-05-13
OcAgentExporter does not work (on windows?)


lmolkova:
Please answer these questions before submitting a bug report.

### What version of OpenCensus are you using?
0.21.0

### What JVM are you using (`java -version`)?
1.8.0_201

### What did you do?
If possible, provide a recipe for reproducing the error.

```java
  public static void main(String[] args) throws Exception {
    TraceConfig traceConfig = Tracing.getTraceConfig();
    traceConfig.updateActiveTraceParams(
        traceConfig.getActiveTraceParams().toBuilder().setSampler(Samplers.alwaysSample()).build());

    Tracer tracer = Tracing.getTracer();
    OcAgentTraceExporter.createAndRegister();
    try (Scope _ = tracer.spanBuilder("tracingExample1").startScopedSpan()) {
    }

    Tracing.getExportComponent().shutdown();
  }
```

```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0"
  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
  <modelVersion>4.0.0</modelVersion>

  <groupId>deleteme</groupId>
  <artifactId>deleteme</artifactId>
  <version>1.0-SNAPSHOT</version>
  <build>
    <plugins>
      <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-compiler-plugin</artifactId>
        <configuration>
          <source>7</source>
          <target>7</target>
        </configuration>
      </plugin>
    </plugins>
  </build>
  <dependencies>
    <dependency>
      <groupId>io.opencensus</groupId>
      <artifactId>opencensus-api</artifactId>
      <version>0.21.0</version>
    </dependency>
    <dependency>
      <groupId>io.opencensus</groupId>
      <artifactId>opencensus-exporter-trace-ocagent</artifactId>
      <version>0.21.0</version>
    </dependency>
    <dependency>
      <groupId>io.opencensus</groupId>
      <artifactId>opencensus-impl</artifactId>
      <version>0.21.0</version>
      <scope>runtime</scope>
    </dependency>
  </dependencies>

</project>
```

### What did you expect to see?
Spans are exported

### What did you see instead?

Bunch of exceptions in the log, no data is exported
[...]

PR #1898, opened 2019-05-14
Exporter/OcAgent: Add a note about netty-tcnative-boringssl dependency.

songy23:
Fixes https://github.com/census-instrumentation/opencensus-java/issues/1897.

### DIFF
--- a/exporters/trace/ocagent/README.md
+++ b/exporters/trace/ocagent/README.md
@@ -53,3 +53,11 @@ public class MyMainClass {
   }
 }
 ```
+
+## FAQ
+
+### Why is `netty-tcnative-boringssl-static` needed? Which version should I use?
+
+It's because this artifact depends on `grpc-netty` which requires a runtime dependency on `netty-tcnative-boringssl-static`
+for it to work. For more details and a table of known working version combinations, see
+https://github.com/grpc/grpc-java/blob/master/SECURITY.md#netty.
--- a/exporters/trace/ocagent/README.md
+++ b/exporters/trace/ocagent/README.md
@@ -26,6 +26,12 @@ For Maven add to your `pom.xml`:
     <version>0.22.0</version>
     <scope>runtime</scope>
   </dependency>
+  <dependency>
+    <groupId>io.netty</groupId>
+    <artifactId>netty-tcnative-boringssl-static</artifactId>
+    <version>2.0.20.Final</version>
+    <scope>runtime</scope>
+  </dependency>
 </dependencies>
 ```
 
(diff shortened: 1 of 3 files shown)


exporters/trace/ocagent/README.md
```diff
@@ -26,6 +26,12 @@ For Maven add to your `pom.xml`:
     <version>0.21.0</version>
     <scope>runtime</scope>
   </dependency>
+  <dependency>
+    <groupId>io.netty</groupId>
+    <artifactId>netty-tcnative-boringssl-static</artifactId>
```
bogdandrutu: Very confused about this:
https://github.com/census-instrumentation/opencensus-java/blob/master/examples/pom.xml#L102
https://github.com/census-instrumentation/opencensus-java/blob/master/examples/pom.xml#L128
and here with 2.0.20.Final version??

Can we clarify what exactly it is needed? @zhangkun83 can you help us with this?

songy23: IIRC, previously `io.grpc:grpc-netty-shaded` shaded the `netty-tcnative-boringssl` artifact, but at some point `netty-tcnative-boringssl` was removed from `grpc-netty-shaded`, and as a result we had to add it explicitly as a runtime dependency. See https://github.com/census-instrumentation/opencensus-java/pull/1247. I believe this runtime dependency is still needed but I'll defer to @zhangkun83 for clarification.

zhangkun83: @ejona86 should be able to help.

ejona86: `netty-tcnative-boringssl-static` is not necessary, and specifying it should be useless. The shaded netty can't use it. grpc-netty-shaded _must_ include tcnative for it to work at all.

#1247 doesn't really provide any information for me to correct or verify. So I can only say, "no, that's not right." Is there any more information as to why this was believed? I'm surprised a bug wasn't opened and that we weren't notified, especially since there was no mention of dependency changes in the v1.10 the release notes.

FYI: I'm not part of the census-instrumentation github org, so I can't route github email for notifications like this. They go to my person email address and tend to be lost.

ejona86: Taking a look at #1897, the error message implies you aren't using `grpc-netty-shaded`. It looks like [`io.opencensus:opencensus-exporter-trace-ocagent` depends on grpc-netty](https://search.maven.org/artifact/io.opencensus/opencensus-exporter-trace-ocagent/0.22.0/jar). grpc-netty-shaded is nowhere to be found. Looking at #1247 again, it is clear it was using grpc-netty and not grpc-netty-shaded. So now the mystery is solved.

netty-tcnative-boringssl-static is necessary when using grpc-netty instead of grpc-netty-shaded. To choose the correct version of tcnative, you can either follow the release notes or look at the table under "Below are known to work version combinations:" at https://github.com/grpc/grpc-java/blob/master/SECURITY.md#netty

songy23: Thanks for providing the context, it's very useful. Yes we need to depend on `grpc-netty` rather than `grpc-netty-shaded` in these two artifacts because they used `io.netty.handler.ssl.SslContext`. 

Currently we depend on grpc 1.19.0, from the table I think `tcnative:2.0.20.Final` should work.

songy23:
PTAL.

Issue #1905, opened 2019-05-15
Allow setting span start/end times explicitly


tigrannajaryan:
We have a use case where the start/end times of traces is coming from an external system.
We need to be able to explicitly set the times instead of the times automatically 
retrieved from the local clock.

The suggested change to the Span API is the following:

1. Add `void setStartTime(Timestamp startTimestamp)` to Span.
2. Add `void end(EndSpanOptions options, Timestamp endTimestamp)` to Span.

Here is the draft of proposed solution that I can turn into Pull Request if this issue is accepted:

https://github.com/census-instrumentation/opencensus-java/compare/master...tigrannajaryan:feature/tigran/span-times?expand=1

bogdandrutu:
Do you really need to construct a Span or you can create directly the whole SpanData?

tigrannajaryan:
I am constructing Spans, I didn't consider creating SpanData directly. I can check that to see if it works for my use case.

For reference here is how I use this feature in my code currently:
```java
Span span = tracer.spanBuilderWithExplicitParent(operationName, parentSpan).startSpan();
span.setStartTime(Timestamp.fromMicros(startTimeMicros));
span.end(EndSpanOptions.DEFAULT, Timestamp.fromMicros(endTimeMicros));
```

tigrannajaryan:
@bogdandrutu I cannot see how to achieve the same end result using SpanData. I need to create spans and have them exported using OcAgentTraceExporter (which I register earlier). Can you give me some hints on how it can be done with SpanData?
I see that SpanData accepts start/end timestamps directly, so that's good but I don't see how to initiate the delivery of SpanData.

bogdandrutu:
@tigrannajaryan you are correct, we lack that but my comment was mostly to confirm that you have everything you need if we add a RecordSpanData on the tracer similar to OpenTelemetry.

I will give a try soon.

tigrannajaryan:
@bogdandrutu sounds good.I will use our fork for now.

dietervdw-spotify:
Just want to add here that my team is also very interested in this.

Our use-case is that we are tracing through a pipeline system (where each step is a different system / JVM). So for each step in the pipeline, we have a span that starts after its root span has ended.
We'd like to construct a span that covers the whole of the pipeline: the start is in the first step of the pipeline and the end being the end-time of processing the last step in the pipeline.

adriancole:
[...]

PR #1906, opened 2019-05-16
Allow setting span start/end times explicitly

tigrannajaryan:
We have a use case where the start/end times of traces is coming from an external system.
We need to be able to explicitly set the times instead of the times automatically
retrieved from the local clock.

The change to the Span API is the following:

1. Add `void setStartTime(Timestamp startTimestamp)` to Span.
2. Add `void end(EndSpanOptions options, Timestamp endTimestamp)` to Span.

### DIFF
--- a/api/src/main/java/io/opencensus/trace/Span.java
+++ b/api/src/main/java/io/opencensus/trace/Span.java
@@ -16,6 +16,7 @@
 
 package io.opencensus.trace;
 
+import io.opencensus.common.Timestamp;
 import io.opencensus.internal.Utils;
 import io.opencensus.trace.internal.BaseMessageEventUtils;
 import java.util.Collections;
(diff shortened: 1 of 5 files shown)


api/src/main/java/io/opencensus/trace/Span.java
```diff
@@ -244,6 +257,22 @@ public final void end() {
...
+   * <p>Only the timing of the first end call for a given {@code Span} will be recorded, and
+   * implementations are free to ignore all further calls. The default implementation ignores
+   * endTimestamp parameter and calls {@link end} method. Implementations that keep track of end
+   * time are expected to override this method.
+   *
+   * @param options the options to be used for the end of the {@code Span}.
+   * @param endTimestamp the {@link Timestamp} to set end time of the span to.
+   * @since 0.23
+   */
+  public void end(EndSpanOptions options, Timestamp endTimestamp) {
```
rghetia: can endTimestamp be part of EndSpanOptions?

bogdandrutu:
Please don't merge this. As discussed in https://github.com/census-instrumentation/opencensus-java/issues/1905 we want to allow users to record an entire SpanData, not to set timestamps for different events.

api/src/main/java/io/opencensus/trace/Span.java
tigrannajaryan: Interesting, I like the suggestion. I can try that, but since Bogdan has a different idea I will wait for now.

MahatmaFatalError:
@bogdandrutu Why does #1905 block this merge? Is there a PR to record an entire SpanData?

PR #1908, opened 2019-05-16
Stats: Don't put stats event onto the queue if no measures are registered.

songy23:
Small improvement on #1809.

This matches Go: https://github.com/census-instrumentation/opencensus-go/blob/0ac3701b0da6461885aa3ae14c555221d967c2c8/stats/record.go#L99-L108.

### DIFF
--- a/impl_core/src/main/java/io/opencensus/implcore/stats/MeasureToViewMap.java
+++ b/impl_core/src/main/java/io/opencensus/implcore/stats/MeasureToViewMap.java
@@ -175,6 +175,15 @@ final class MeasureToViewMap {
     return metrics;
   }
 
+  synchronized boolean isAnyMeasureRegistered(MeasureMapInternal measurementValues) {
+    for (Iterator<Measurement> iterator = measurementValues.iterator(); iterator.hasNext(); ) {
+      if (registeredMeasures.containsKey(iterator.next().getMeasure().getName())) {
+        return true;
+      }
+    }
+    return false;
+  }
+
   // Clear stats for all the current MutableViewData
   synchronized void clearStats() {
     for (Entry<String, Collection<MutableViewData>> entry : mutableMap.asMap().entrySet()) {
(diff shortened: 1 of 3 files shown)


impl_core/src/main/java/io/opencensus/implcore/stats/MeasureToViewMap.java
```diff
@@ -175,6 +175,15 @@ synchronized void record(TagContext tags, MeasureMapInternal stats, Timestamp ti
     return metrics;
   }
 
+  synchronized boolean isAnyMeasureRegistered(MeasureMapInternal measurementValues) {
```
dinooliva: Any way to cache a true result so that this won't add much overhead in the common case?

songy23: Unfortunately I cannot think of any way, because `MeasureMapInternal` is basically a list of `Measurement`s. It doesn't make much sense to cache results for different combinations of `Measurement`s.

bogdandrutu: The way I can think of is to have an internal method and field on the Measure (volatile Boolean) that we update when a view is registered for that metric.

This is solved on the OpenTelemetry by having the measure ctor on the Meter(StatsRecorder)

dinooliva: I may be misunderstanding this change completely but it seems like we are adding overhead to the common case to optimize an uncommon special case.

Is it the case that applications generally have a lot of measures defined w/o registered views?

songy23: > The way I can think of is to have an internal method and field on the Measure (volatile Boolean) that we update when a view is registered for that metric.

Make sense. This is more consistent with what Go did.

> I may be misunderstanding this change completely but it seems like we are adding overhead to the common case to optimize an uncommon special case.
> Is it the case that applications generally have a lot of measures defined w/o registered views?

That's a valid concern. The only common case I can think of is gRPC measures. gRPC records against those measures on every RPC, but I wonder not every user would register the RPC views. In that case it may improve the performance a bit by not putting measurements onto the event queue.

PR #1909, opened 2019-05-17
CI: Add JDK 10 and 11 build on Travis.

songy23:


.travis.yml
```diff
@@ -13,6 +13,14 @@ matrix:
     env: TASK=BUILD
     os: linux
 
+  - jdk: openjdk10
```
songy23: Note `oraclejdk10` is not supported on Travis.

songy23:
> Will this add more waiting time to submit?

Hopefully not. All sub tasks will be executed in parallel.

songy23:
Seems JDK 11 build has some problem: https://travis-ci.org/census-instrumentation/opencensus-java/jobs/533909313. Disable it for now.

Issue #1917, opened 2019-05-22
StackdriverStatsExporter#unregister


dansiviter:
### Is your feature request related to a problem? If so, please describe it.
I'd like to be able to gracefully shutdown the application, but the lack of `StackdriverStatsExporter#unregister` prevents this. It's also not possible to use in a container environment as this would fail on repeat deploy.

### Describe the solution you'd like.
Similar to `StackdriverTraceExporter#unregister` allow the ability to unregister the singleton.

### Describe alternatives you've considered.
N/A

### Additional context.
N/A

PR #1918, opened 2019-05-22
Exporter/Stackdriver: Expose unregister API.

songy23:
Fixes https://github.com/census-instrumentation/opencensus-java/issues/1917.


songy23:
> Why do you need this API?

See #1917 for background:

> I'd like to be able to gracefully shutdown the application, but the lack of StackdriverStatsExporter#unregister prevents this. It's also not possible to use in a container environment as this would fail on repeat deploy.

dinooliva:
Ahh - apologies, missed that, thanks!

PR #1926, opened 2019-06-07
Recording http metrics at host level

iamgr0ot:
Host name level opencensus metrics are insanely useful while troubleshooting production issues and we would like to give it back to the community.

- Recording opencensus client http metrics at hostname level based on configuration.
- Introducing `HttpClientMetricGranularity` to define, at what level, one would like to collect metrics.
- The change is backward compatible and metrics will be collected at http method level by default.


bogdandrutu:
Can you please sign the CLA?

contrib/http_util/src/main/java/io/opencensus/contrib/http/util/HttpClientMetricGranularity.java
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.contrib.http.util;
+
+/** Enum class that determines the granularity at which opencensus http metrics are emitted. */
+public enum HttpClientMetricGranularity {
```
bogdandrutu: Not sure why do you need this? What you can do is to always add the HOST tag, but don't change the default views. So you can define your own custom views that include the host :)

iamgr0ot:
I signed it!

contrib/http_util/src/main/java/io/opencensus/contrib/http/util/HttpClientMetricGranularity.java
iamgr0ot: what if some applications have a high cardinality hostname space, for example  web crawlers? Metric backends may run into cardinality issues in that case. What do you think?

bogdandrutu: If the defined views do not include the HOST we just ignore the tag so no cardinality issue.

iamgr0ot: I don't quite get your comment on custom views.

iamgr0ot:
@bogdandrutu - You are right. There is no need to change default views. I could define my own view with the exposed measure. Now, I am just tagging host name and not changing any default views.

iamgr0ot:
@songy23 - I would love to get this change with a release. Could you please help me with that?

songy23:
@langecode We can do a new release after finishing the remaining HTTP work - @rghetia do you have an estimate on it?

rghetia:
I have two PRs pending. I will post them today. we can release later today or tomorrow.

PR #1928, opened 2019-06-10
Exporter/Jaeger: Set span status as tags.

### LINKED ISSUE #1927: OC Status info is lost when exporting to Jaeger

`io.opencensus:opencensus-api:0.18.0`

1.8

### What did you do?
Set status of span exported it to Jaeger.

### What did you expect to see?
Similar to the Zipkin exporter I expected status (code and message) to be preserved in some Jaeger tag.

### What did you see instead?
Nothing related to the status was set on the span.

### Additional context
It will be good to have the specs covering what attributes will be mapping OC `Status` so exporters in different languages can be consistent. I will create an issue on specs to track this.

songy23:
Fixes #1927.

Tags conversion matches Go Jaeger exporter: 
https://github.com/census-ecosystem/opencensus-go-exporter-jaeger/blob/e8b55949d948652e47aae4378212f933ecee856b/jaeger.go#L210-L219

### DIFF
--- a/exporters/trace/jaeger/src/test/java/io/opencensus/exporter/trace/jaeger/JaegerExporterHandlerTest.java
+++ b/exporters/trace/jaeger/src/test/java/io/opencensus/exporter/trace/jaeger/JaegerExporterHandlerTest.java
@@ -48,6 +48,8 @@ import io.opencensus.trace.TraceId;
 import io.opencensus.trace.TraceOptions;
 import io.opencensus.trace.Tracestate;
 import io.opencensus.trace.export.SpanData;
+import io.opencensus.trace.export.SpanData.TimedEvent;
+import java.util.Collections;
 import java.util.List;
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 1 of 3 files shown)


exporters/trace/jaeger/src/test/java/io/opencensus/exporter/trace/jaeger/JaegerExporterHandlerTest.java
```diff
@@ -142,6 +144,44 @@ public void exportShouldConvertFromSpanDataToJaegerThriftSpan() throws SenderExc
...
+    assertThat(spans.size()).isEqualTo(1);
+    Span span = spans.get(0);
+
+    assertThat(span.tags.size()).isEqualTo(4);
+    assertThat(span.tags)
+        .containsExactly(
+            new Tag("span.kind", TagType.STRING).setVStr("server"),
+            new Tag("status.code", TagType.LONG).setVLong(4),
+            new Tag("status.message", TagType.STRING).setVStr(statusMessage),
+            new Tag("error", TagType.BOOL).setVBool(true));
```
rghetia: nit:s/status.code/STATUS_CODE/ 
same with other strings.

songy23: SG, fixed in https://github.com/census-instrumentation/opencensus-java/pull/1928/commits/c7816a7e16802ab835f5eeb6ce471ae863316bc1.

pjanotti:
Thanks for putting out a fix so quickly @songy23!

I like mapping Status to `status.code` and `status.message`. 

I don't agree with automatically setting `error` to true as explained in the conversation at https://github.com/census-instrumentation/opencensus-go/issues/1041#issuecomment-479269830 I still think that this is a decision that should be made by the application itself. Automatically setting `error` to true will tend to mark most spans as of trace as "error" making it redundant. This comment here defends the same position https://github.com/opentracing/specification/issues/96#issuecomment-355795243

pjanotti:
The Zipkin exporter also should use the same OT tags (`status.code` and `status.message`) instead of the custom one (`census.statu_code`).

songy23:
@pjanotti Thanks for the context. If it's the convention to set the `error` tag at application level, I'm good with not having it automatically set. Removed in https://github.com/census-instrumentation/opencensus-java/pull/1928/commits/652f7fffef3aa45aab2996673bb58ef80e631a79.

Updates to Zipkin will be in another PR.

pjanotti:
Thanks @songy23! LGTM.

PR #1931, opened 2019-06-11
Exporter/Zipkin: Add a note about status tags.

songy23:
Suggested by https://github.com/census-instrumentation/opencensus-java/pull/1928#issuecomment-500671005. Seems this is the convention from OpenTracing.

/cc @adriancole

### DIFF
--- a/exporters/trace/zipkin/src/main/java/io/opencensus/exporter/trace/zipkin/ZipkinExporterHandler.java
+++ b/exporters/trace/zipkin/src/main/java/io/opencensus/exporter/trace/zipkin/ZipkinExporterHandler.java
@@ -59,6 +59,7 @@ final class ZipkinExporterHandler extends TimeLimitedHandler {
   // https://github.com/apache/incubator-zipkin-brave/blob/643b7245c462dc14d47afcdb076b2603fd421497/instrumentation/grpc/src/main/java/brave/grpc/GrpcParser.java#L67-L73
   @VisibleForTesting static final String STATUS_CODE = "census.status_code";
   @VisibleForTesting static final String STATUS_DESCRIPTION = "census.status_description";
+  @VisibleForTesting static final String STATUS_ERROR = "error";
 
   private final SpanBytesEncoder encoder;
   private final Sender sender;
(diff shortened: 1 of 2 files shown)


adriancole:
opentracing !=zipkin so OT is not a valid rationale for zipkin data policy
especially when prior art exists.

since this is about zipkin I think we should use conventions present with a
comment saying where it was inherited from.

here is the default way we tag status for grpc. this has been the case
before opentracing did anything with grpc.

maybe correct and make a line note about why.

https://github.com/apache/incubator-zipkin-brave/blob/master/instrumentation/grpc/src/main/java/brave/grpc/GrpcParser.java#L67

thx

songy23:
@adriancole Thanks for the pointers. I reverted the changes to these tags and added a reference link.

/cc @pjanotti

exporters/trace/zipkin/src/main/java/io/opencensus/exporter/trace/zipkin/ZipkinExporterHandler.java
```diff
@@ -53,8 +54,12 @@
 final class ZipkinExporterHandler extends TimeLimitedHandler {
   private static final Logger logger = Logger.getLogger(ZipkinExporterHandler.class.getName());
   private static final String EXPORT_SPAN_NAME = "SendZipkinSpans";
-  private static final String STATUS_CODE = "census.status_code";
-  private static final String STATUS_DESCRIPTION = "census.status_description";
+
+  // The naming follows Zipkin convention. As an example see:
+  // https://github.com/apache/incubator-zipkin-brave/blob/643b7245c462dc14d47afcdb076b2603fd421497/instrumentation/grpc/src/main/java/brave/grpc/GrpcParser.java#L67-L73
+  @VisibleForTesting static final String STATUS_CODE = "census.status_code";
```
adriancole: one important thing is that we do use non-ok as "error" message. this controls analysis etc, so I would recommend including this logic here.

songy23: Got it, added in https://github.com/census-instrumentation/opencensus-java/pull/1931/commits/c1f588a81d5c9bb94b981969a72adb77a550a63e.

PR #1939, opened 2019-06-13
spring servlet example.

rghetia:


examples/spring/servlet/src/main/java/io/opencensus/examples/spring/servlet/Application.java
```diff
@@ -0,0 +1,59 @@
...
+    HttpViews.registerAllServerViews();
+    HttpViews.registerAllClientViews();
+
+    // Register Prometheus exporters and export metrics to a Prometheus HTTPServer.
+    // Refer to https://prometheus.io/ to run Prometheus Server.
+    PrometheusStatsCollector.createAndRegister();
+    HTTPServer prometheusServer = new HTTPServer(9090, true);
+  }
+
+  private static void initTracing() {
```
dinooliva: nit - consider naming something analogous to stats init - e.g. initTracingAndLoggingExporter

examples/spring/servlet/src/main/java/io/opencensus/examples/spring/servlet/HelloController.java
```diff
@@ -0,0 +1,91 @@
...
+  /**
+   * Serves index page.
+   *
+   * @return String
+   */
+  @RequestMapping("/")
+  public String index() {
+    String str = "Hello from servlet instrumented with opencensus-spring";
+    String resp = restTemplate.getForObject("http://localhost:8080/loopback", String.class);
+
```
dinooliva: nit - should the port be configurable?

rghetia: done.

rghetia:
> LGTM overall. One nit: it would be great to exclude the gradle wrappers under examples if there's a way to do so.

I tried but have not found a good way to do. Earlier I had created sub-project under example/settings.gradle. That did not work because CreateStartScripts did not for work for subproject. May be there is a way.

PR #1943, opened 2019-06-19
Resource: Add deployment label for k8s resource.

songy23:
Followed https://github.com/census-instrumentation/opencensus-specs/pull/267, https://github.com/census-instrumentation/opencensus-go/pull/1148 and https://github.com/census-ecosystem/kubernetes-operator/pull/13.

### DIFF
--- a/contrib/resource_util/src/main/java/io/opencensus/contrib/resource/util/K8sResource.java
+++ b/contrib/resource_util/src/main/java/io/opencensus/contrib/resource/util/K8sResource.java
@@ -19,8 +19,10 @@ package io.opencensus.contrib.resource.util;
 import static com.google.common.base.MoreObjects.firstNonNull;
 import static com.google.common.base.Preconditions.checkNotNull;
 
+import com.google.common.base.Splitter;
 import io.opencensus.resource.Resource;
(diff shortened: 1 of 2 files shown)


contrib/resource_util/src/main/java/io/opencensus/contrib/resource/util/K8sResource.java
```diff
@@ -65,20 +76,47 @@
...
     return Resource.create(TYPE, labels);
   }
 
   static Resource detect() {
+    String podName = firstNonNull(System.getenv("HOSTNAME"), "");
+    String deploymentName = "";
+    // Extract deployment name from the pod name. Pod name is created using
+    // format: [deployment-name]-[Random-String-For-ReplicaSet]-[Random-String-For-Pod]
+    List<String> parts = splitter.splitToList(podName);
+    if (parts.size() == 3) {
```
rghetia: if the deployment name has '-' then this will not work. so condition should be at least 3 and strip of last two parts to make it deployment name.

songy23: Make sense, upated.

PR #1944, opened 2019-06-20
Removing AppEngine restriction on Java version.

Vizerai:
AppEngine uses Java 8 now.

### DIFF
--- a/impl_core/src/main/java/io/opencensus/implcore/internal/DaemonThreadFactory.java
+++ b/impl_core/src/main/java/io/opencensus/implcore/internal/DaemonThreadFactory.java
@@ -22,11 +22,6 @@ import java.util.concurrent.atomic.AtomicInteger;
 
 /** A {@link ThreadFactory} implementation that starts all {@link Thread} as daemons. */
 public final class DaemonThreadFactory implements ThreadFactory {
-  // AppEngine runtimes have constraints on threading and socket handling
-  // that need to be accommodated.
-  public static final boolean IS_RESTRICTED_APPENGINE =
-      System.getProperty("com.google.appengine.runtime.environment") != null
-          && "1.8".equals(System.getProperty("java.specification.version"));
   private static final String DELIMITER = "-";
   private static final ThreadFactory threadFactory = MoreExecutors.platformThreadFactory();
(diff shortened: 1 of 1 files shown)


impl_core/src/main/java/io/opencensus/implcore/internal/DaemonThreadFactory.java
```diff
@@ -26,7 +26,7 @@
   // that need to be accommodated.
   public static final boolean IS_RESTRICTED_APPENGINE =
       System.getProperty("com.google.appengine.runtime.environment") != null
-          && "1.7".equals(System.getProperty("java.specification.version"));
+          && "1.8".equals(System.getProperty("java.specification.version"));
```
bogdandrutu: I think for Java 8 we don't need to do this. In the Java 8 appengine they removed this requirement. Can you please check that?

Vizerai: It looks like AppEngine still does not support renaming of threads.  thread.setName() isn't allowed.

bogdandrutu: ok if that is the case do this:

```
  @Override
  public Thread newThread(Runnable r) {
    Thread thread = threadFactory.newThread(r);
    try {
      thread.setName(name);
      thread.setDaemon(true);
    } catch (SecurityException e) {
      // OK if we can't set the name or daemon in this environment.
    }
    return thread;
  }
```

This is also how guava implements this in `MoreExecutors.newThread()` which unfortunately is not public.

This way we remove the IS_RESTRICTED_APPENGINE and we are ok. What do you think?

Vizerai: Yes, we can do that.

bogdandrutu:
Please update the description and name of the PR.

Vizerai:
AppEngine restriction on Java version has been removed and replaced with an SecurityException check when trying to rename the thread or set it as a daemon thread.

rghetia:
thanks @songy23 . This issue doesn't fix #1954.

PR #1956, opened 2019-07-31
Use Zipkin quickstart script in download instructions

shakuzen:
The group id of the Zipkin Server maven artifact has changed in recent versions, so the previous instructions would not download the latest version of the server. The quickstart script provided by Zipkin is the recommended way to download the Zipkin Server jar and will download the latest version.

### DIFF
--- a/exporters/trace/zipkin/README.md
+++ b/exporters/trace/zipkin/README.md
@@ -18,7 +18,7 @@ applications instrumented with Census. The easiest way to start a zipkin
 server is to paste the below:
 
 ```bash
-wget -O zipkin.jar 'https://search.maven.org/remote_content?g=io.zipkin.java&a=zipkin-server&v=LATEST&c=exec'
+curl -sSL https://zipkin.io/quickstart.sh | bash -s
 java -jar zipkin.jar
 ```
 


shakuzen:
@googlebot I signed it!

exporters/trace/zipkin/README.md
```diff
@@ -18,7 +18,7 @@ applications instrumented with Census. The easiest way to start a zipkin
 server is to paste the below:
 
 ```bash
-wget -O zipkin.jar 'https://search.maven.org/remote_content?g=io.zipkin.java&a=zipkin-server&v=LATEST&c=exec'
+curl -sSL https://zipkin.io/quickstart.sh | bash -s
```
songy23:
```suggestion
curl -ssl https://zipkin.io/quickstart.sh | bash -s
```

shakuzen: The `-s` and `-S` flags are different. It is not one flag with three letters (perhaps the ordering of them is unfortunate). From `curl --help`:

```
 -S, --show-error    Show error. With -s, make curl show errors when they occur
 -s, --silent        Silent mode (don't output anything)
```

The point being to only output errors from the `curl` command, I believe. Note that this command was copied from the [Zipkin quickstart page](https://zipkin.io/pages/quickstart.html).

songy23: Sounds good, thanks for the explanation.

PR #1965, opened 2019-08-23
Clear handler reference on OcAgentTraceExporter unregister

c2nes:
This allows the exporter to be re-registered without triggering an `IllegalStateException` in `createAndRegister`.


c2nes:
@songy23 done

PR #1409, opened 2018-09-04
Contrib/agent: detach() should take the context returned from attach().

songy23:
Part of https://github.com/census-instrumentation/opencensus-java/issues/1407.

PR #1984, opened 2019-10-02
Add Wrapper for WriteListener.

### LINKED ISSUE #1983: Wrapper for WriteListener to execute asynchronous write in parent context
[ServletOutputStream](https://docs.oracle.com/javaee/5/api/javax/servlet/ServletOutputStream.html)  allows to execute write asynchronously via [WriteListener.onWritePossible()](https://docs.oracle.com/javaee/7/api/javax/servlet/WriteListener.html#onWritePossible--) at later time. In such case, the context at the time of execution is different than at the time setting listener. In order to have correct span associated with the write operation, a wrapper for WriteListener is required. The wrapper would save context at the time of creation and use this context to execute onWritePossible() method of the object being wrapped.

/cc @blt-gw @bogdandrutu

rghetia:
fixes #1983 

/cc @blt-gw

### DIFF
--- a/contrib/http_servlet/src/test/java/io/opencensus/contrib/http/servlet/WriteListenerWrapperTest.java
+++ b/contrib/http_servlet/src/test/java/io/opencensus/contrib/http/servlet/WriteListenerWrapperTest.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2019, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.contrib.http.servlet;
+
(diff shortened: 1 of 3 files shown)


contrib/http_servlet/src/test/java/io/opencensus/contrib/http/servlet/WriteListenerWrapperTest.java
```diff
@@ -0,0 +1,62 @@
...
+@RunWith(JUnit4.class)
+public class WriteListenerWrapperTest {
+
+  @Test
+  public void testWriteOnPossibleWithContext() throws IOException, ServletException {
+
+    Key<String> key = Context.<String>key("test-key");
+    Context curr = Context.current();
+    assertThat(curr).isNotNull();
+    assertThat(curr.withValue(key, "parent").attach()).isNotNull();
```
songy23: https://github.com/census-instrumentation/opencensus-java/pull/1409#discussion_r215103654:

> attach() should always be immediately followed by a try-finally that calls detach() in finally.

contrib/http_servlet/src/test/java/io/opencensus/contrib/http/servlet/WriteListenerWrapperTest.java
```diff
@@ -0,0 +1,62 @@
...
+                Context curr = Context.current();
+                assertThat(curr).isNotNull();
+                assertThat(curr).isEqualTo(parentContext);
+              }
+
+              @Override
+              public void onError(Throwable t) {}
+            });
+
+    assertThat(parentContext.withValue(key, "child").attach()).isNotNull();
```
songy23: Same here.

Issue #1400, opened 2018-08-29
implement "b3 single" header format


adriancole:
As discussed on https://github.com/openzipkin/b3-propagation/issues/21 and first implemented here: https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/propagation/B3SingleFormat.java https://github.com/openzipkin/brave/blob/master/brave/src/test/java/brave/propagation/B3SingleFormatTest.java

Let's support at least reading "b3" header from a single string, most commonly traceid-spanid-1
It would also be nice to support optionally writing this, especially in message providers or others with constrained environments.

Brave currently has a property like this, but its name could change with feedback:
```java
    /**
     * When true, only writes a single {@link B3SingleFormat b3 header} for outbound propagation.
     *
     * <p>Use this to reduce overhead. Note: normal {@link Tracing#propagation()} is used to parse
     * incoming headers. The implementation must be able to read "b3" headers.
     */
    public Builder b3SingleFormat(boolean b3SingleFormat) {
      this.b3SingleFormat = b3SingleFormat;
      return this;
}
```

l0s:
Is the intention to introduce a new `io.opencensus.trace.propagation.TextFormat` implementation that extracts the `b3` header (or `tracestate` header with "b3" prefix) and injects the `b3` header?

adriancole:
the b3 single header was indeed to be designed as both a top-level header and also a tracestate entry for when it is necessary to carry data present in b3, but no longer in the traceparent spec (such as before-the-fact sampling instructions b3=0, or the parent id)

l0s:
@adriancole I provided two potential implementations. #1989 introduces a new `TextFormat` implementation and #1990 modifies the existing `B3Format`. The former extracts b3 from the single format as well as injects it in the same format. The latter extracts b3 from either the single or multi formats but only injects data in the multi format. Can you comment on which one most closely matches the intention of this issue?

adriancole:
@IOs you are awesome btw. thank you for the help

adriancole:
not sure if github is case-sensitive @l0s ^^

adriancole:
[...]

PR #1990, opened 2019-10-13
Extract "b3 single" header format

l0s:
This is an alternative implementation of #1400 that modifies the existing `B3Format` class to read and prioritise the single `b3` header or `tracestate` header with "b3=" prefix over the multi-header format. It only affects extraction, no changes are made to injection.

Addresses: #1400

### DIFF
--- a/api/src/main/java/io/opencensus/trace/propagation/PropagationComponent.java
+++ b/api/src/main/java/io/opencensus/trace/propagation/PropagationComponent.java
@@ -50,6 +50,19 @@ public abstract class PropagationComponent {
   public abstract TextFormat getB3Format();
 
   /**
+   * Returns the B3 {@link TextFormat} with the provided implementations. See <a href=
+   * "https://github.com/openzipkin/b3-propagation">b3-propagation</a> for more information. If no
+   * implementation is provided then no-op implementation will be used. Extraction is equivalent to
+   * {@link #getB3Format()}, however when injecting fields, the compact B3 single format is used
+   * instead.
+   *
+   * @since 0.25.0
+   * @return the B3 {@code TextFormat} implementation.
+   */
+  @ExperimentalApi
+  public abstract TextFormat getB3SingleFormat();
+
+  /**
    * Returns the TraceContext {@link TextFormat} with the provided implementations. See <a
    * href="https://github.com/w3c/distributed-tracing">w3c/distributed-tracing</a> for more
    * information. If no implementation is provided then no-op implementation will be used.
@@ -78,6 +91,11 @@ public abstract class PropagationComponent {
 
     @Override
     public TextFormat getB3Format() {
+      return TextFormat.getNoopTextFormat();
+    }
+
+    @Override
+    public TextFormat getB3SingleFormat() {
       return TextFormat.getNoopTextFormat();
     }
 
--- a/api/src/main/java/io/opencensus/trace/propagation/PropagationComponent.java
+++ b/api/src/main/java/io/opencensus/trace/propagation/PropagationComponent.java
@@ -47,20 +47,22 @@ public abstract class PropagationComponent {
    * @return the B3 {@code TextFormat} implementation.
    */
   @ExperimentalApi
-  public abstract TextFormat getB3Format();
+  public TextFormat getB3Format() {
+    return getB3Format(B3InjectionFormat.MULTI);
+  }
 
   /**
    * Returns the B3 {@link TextFormat} with the provided implementations. See <a
    * href="https://github.com/openzipkin/b3-propagation">b3-propagation</a> for more information. If
    * no implementation is provided then no-op implementation will be used.
    *
+   * @param formats the injection formats to use. At least one must be specified.
    * @since 0.25.0
-   * @param singleOutput true if and only if the compact "b3 single" should be used when emitting
-   *     trace data, otherwise the multiple header format will be used
    * @return the B3 {@code TextFormat} implementation.
+   * @see B3InjectionFormat
    */
   @ExperimentalApi
-  public abstract TextFormat getB3Format(boolean singleOutput);
+  public abstract TextFormat getB3Format(B3InjectionFormat... formats);
 
   /**
    * Returns the TraceContext {@link TextFormat} with the provided implementations. See <a
@@ -90,12 +92,7 @@ public abstract class PropagationComponent {
     }
 
     @Override
-    public TextFormat getB3Format() {
-      return TextFormat.getNoopTextFormat();
-    }
-
-    @Override
-    public TextFormat getB3Format(final boolean singleOutput) {
+    public TextFormat getB3Format(final B3InjectionFormat... formats) {
       return TextFormat.getNoopTextFormat();
     }
 
--- a/api/src/main/java/io/opencensus/trace/propagation/PropagationComponent.java
+++ b/api/src/main/java/io/opencensus/trace/propagation/PropagationComponent.java
@@ -50,17 +50,17 @@ public abstract class PropagationComponent {
   public abstract TextFormat getB3Format();
 
   /**
-   * Returns the B3 {@link TextFormat} with the provided implementations. See <a href=
-   * "https://github.com/openzipkin/b3-propagation">b3-propagation</a> for more information. If no
-   * implementation is provided then no-op implementation will be used. Extraction is equivalent to
-   * {@link #getB3Format()}, however when injecting fields, the compact B3 single format is used
-   * instead.
+   * Returns the B3 {@link TextFormat} with the provided implementations. See <a
+   * href="https://github.com/openzipkin/b3-propagation">b3-propagation</a> for more information. If
+   * no implementation is provided then no-op implementation will be used.
    *
    * @since 0.25.0
+   * @param singleOutput true if and only if the compact "b3 single" should be used when emitting
+   *     trace data, otherwise the multiple header format will be used
    * @return the B3 {@code TextFormat} implementation.
    */
   @ExperimentalApi
-  public abstract TextFormat getB3SingleFormat();
+  public abstract TextFormat getB3Format(boolean singleOutput);
 
   /**
    * Returns the TraceContext {@link TextFormat} with the provided implementations. See <a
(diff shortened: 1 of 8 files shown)


impl_core/src/main/java/io/opencensus/implcore/trace/propagation/B3Format.java
```diff
@@ -81,15 +83,27 @@
...
+
     try {
+      // check for single header format
+      final String b3Header = getter.get(carrier, B3);
+      if (b3Header != null) {
+        return parseB3Value(b3Header);
+      }
+      final String traceStateHeader = getter.get(carrier, TRACESTATE);
+      if (traceStateHeader != null) {
+        final String[] components = traceStateHeader.split("=", 2);
```
adriancole: if there are multiple state entries.. would limit of 2 work?

l0s: Ah, it looks like I should split on commas first before looking for the b3 entry ([spec for my reference](https://www.w3.org/TR/trace-context/#tracestate-header)). However, based on the discussion in #1400, it seems I should remove the tracestate handling logic altogether.

api/src/main/java/io/opencensus/trace/propagation/PropagationComponent.java
```diff
@@ -49,6 +49,19 @@
   @ExperimentalApi
   public abstract TextFormat getB3Format();
 
+  /**
+   * Returns the B3 {@link TextFormat} with the provided implementations. See <a href=
```
adriancole: maybe

```
 Returns the {@link TextFormat B3 single text format}.
```

api/src/main/java/io/opencensus/trace/propagation/PropagationComponent.java
```diff
@@ -49,6 +49,19 @@
...
+   * "https://github.com/openzipkin/b3-propagation">b3-propagation</a> for more information. If no
+   * implementation is provided then no-op implementation will be used. Extraction is equivalent to
```
adriancole: in general I think the fallback stuff and extract in general probably will end up needing to be decoupled. It sounds ok to always fall back to b3 multi at the moment, but you might imagine case in the future where someone will want "only b3 single" without having to wait for parse failures.. granted you only need to wait for 1 failure (failure to extract `X-B3-TraceId`

iotw in the future, with a way to register extract rules independently, we won't have to overload impls like this. I know it is doing this as the design is mostly copied from brave, but at any rate something to consider improving regardless even if not now.

api/src/main/java/io/opencensus/trace/propagation/PropagationComponent.java
l0s: As I started modifying the Spring contrib classes to support "single" as a configuration parameter, I realised it's probably better to expose a `boolean` parameter here that controls whether or not to emit trace data in the compact format. What do you think?

api/src/main/java/io/opencensus/trace/propagation/PropagationComponent.java
l0s: Would it make sense to add a `strict` flag? When it's set, fallbacks would be disabled.

adriancole:
I dont quite understand why in spring you cant choose the format
explicitly? it would be better to not overload too many things here which
can be done in the config layer.

KimbeLu79:
Leave me alone and stop messing with my device and accounts!!! I've been
heavy chested and light-headed today so much damn stress that I'm almost
sure that a stroke is coming. So unless you want to be a responsible party
for that leave me alone.

Kimberly  Moncibaiz

adriancole:
for example the only thing muddy right now is if you can fallback on
extract. in brave we only currently do that in the b3 legacy (not even our
impl of b3 single). for years no request for fine tuning and we have
support directly for spring 2.5 and also through sleuth boot for the same
amount of time.

rather than dig holes into the inject side where that is now also
conditional.. which hints that b3 single is a different format which it is.
I would leave it alone as separate. that way when inevitably you end up
with request specific injection it wont be muddied with flags. if you
choose the b3 single one it is in fact that and not overridden due to
pushing config directly into the format class.

adriancole:
can someone with access try to unsubscribe the person who just mentioned on this thread? maybe accidentally subscribed..

KimbeLu79:
I apologize you all is that had a really rough the only day but week also.
Does my tablet and my phone have been hacked down to the root I don't know
what to do fix the issue they are both new devices and they are connected
with my cell/internet  provider.

Kimberly  Moncibaiz

l0s:
@KimbeLu79 I'm sorry for the disturbances. I don't know how you got subscribed to this thread, but I reached out to GitHub support to see if someone can remove you. I am not a maintainer on this project and I do not have access to do that myself. Alternatively, if you can see an "unsubscribe" link in the notification emails, click it and it should stop the notifications.

@adriancole I really appreciate your feedback. I do have a commit forthcoming that will modify the Spring configuration so clients do not need to mess with `PropagationComponent` directly. I tried as much as possible to follow the existing pattern. It appears that `Tracing.getPropagationComponent().get...` is the established way to get a `TextFormat` implementation. The `B3Format` class currently only has package visibility.

Regarding the fallback logic, do I understand you correctly that it would be more appropriate to choose between "B3 multi" only and "B3 single" only with no fallbacks in either scenario? If so, I'm happy to implement it that way.

adriancole:
I think on reads it makes sense to fallback (tolerant read pattern). Someone can request to add a flag about strict or not, on reads. We've not had one since. It would seem better time to spend the time used to explain a flag on something more hard hitting like a general purpose fallback component. IOTW, I like tolerant reads (b3 single first), but in hind sight would be better to make a composite extractor https://github.com/openzipkin/zipkin-gcp/blob/master/propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/CompositeExtractor.java

in any case the main idea is to make strict writes. composite injector could also be made to output multiple formats, or a request-specific one could be made for conditional injection. At any rate, being able to strictly write allows this.

l0s:
@adriancole I just pushed a commit that includes the Spring configuration changes. In addition, the B3 Format follows the robustness principle in that it accepts either B3 format (prioritises single) but only produces one format as the configuration dictates.

I like the idea of a composite extractor, but I think that should be outside the scope of this issue. It could accept more, non B3-related formats.

api/src/main/java/io/opencensus/trace/propagation/PropagationComponent.java
```diff
@@ -49,6 +49,19 @@
...
+   * href="https://github.com/openzipkin/b3-propagation">b3-propagation</a> for more information. If
+   * no implementation is provided then no-op implementation will be used.
+   *
+   * @since 0.25.0
+   * @param singleOutput true if and only if the compact "b3 single" should be used when emitting
+   *     trace data, otherwise the multiple header format will be used
+   * @return the B3 {@code TextFormat} implementation.
+   */
+  @ExperimentalApi
+  public abstract TextFormat getB3Format(boolean singleOutput);
```
adriancole: I'm not a fan of this, but I understand why you decided to do this.

l0s:
@adriancole right now, the property `opencensus.spring.trace.propagation` controls whether to use B3 ( `TRACE_PROPAGATION_B3` ) or TraceContext ( `TRACE_PROPAGATION_TRACE_CONTEXT ). I can add a new propagation value (e.g. `TRACE_PROPAGATION_B3_SINGLE` ). Both of the B3 propagation formats can have the same extraction logic (look for B3 single first and fall back on multi). However, each would inject a specific format only. This removes the need for the "singleOutput" flag. What do you think?

adriancole:
I think controlling injection is the primary goal, yeah. in fact we need similar in messaging as we don't want to use a format that has incompatibility with JMS. However, in Brave this likely will end up a carrier-specific rule.

KimbeLu79:
What is all this crap????

KimbeLu79:
What's going on and what does it have to do with me????

adriancole:
something subscribed you. either you have a github account, or someone
signed you up for one. The email at the bottom shows how to unsubscribe. no
one here can unsubscribe for you, sadly.

KimbeLu79:
I realize I'm receiving this because I was mentioned. I also realize that
people use GitHub a tool to hack people's devices. Which I am totally in
belief that my devices laptop, multiple phones and 2 tablet has been
hacked.

KimbeLu79:
I'm pretty sure I know who the main person is. He has also used pocket and
XDA

adriancole:
my account was compromised once and it was used to star projects I never
heard of. Check your account to see if you are following or starred things,
as it is typically these days that bots are used to fake social stats.

adriancole:
@l0s fyi, we are recently tuning our B3 injector logic in support of a messaging abstraction and it might be helpful for you to see the option https://github.com/openzipkin/brave/pull/1020 It should be config friendly for the common case of controlling client outbound choices.

I'm not going to be following up on this change.. no offense or anything, just I've said all to say and think you've got what you need to make a decision.

l0s:
@adriancole thanks for the feedback. That's helpful. I'll work on these changes.

impl_core/src/main/java/io/opencensus/implcore/trace/propagation/B3Format.java
```diff
@@ -69,10 +95,21 @@
     checkNotNull(spanContext, "spanContext");
     checkNotNull(setter, "setter");
     checkNotNull(carrier, "carrier");
-    setter.put(carrier, X_B3_TRACE_ID, spanContext.getTraceId().toLowerBase16());
-    setter.put(carrier, X_B3_SPAN_ID, spanContext.getSpanId().toLowerBase16());
-    if (spanContext.getTraceOptions().isSampled()) {
-      setter.put(carrier, X_B3_SAMPLED, SAMPLED_VALUE);
+    if (getInjectionFormats().contains(B3InjectionFormat.SINGLE)) {
```
l0s: An alternative approach would have been to put the injection logic in the enum. That would provide a compile-time guarantee that every case was implemented. This code would simplify to:

    for( final B3InjectionFormat format : getInjectionFormats() ) {
        format.inject( setter, carrier, spanContext );
    }

However, I opted not to go with that approach as `B3Injectionformat` is in the api module by necessity and it did not seem appropriate to put business logic there.

KimbeLu79:
Tell Brandon May that I will love him for the rest of my life. But I'm done
playing games I want better things for my life and if he can't add to it
and be willing to grow in all areas of life with me, then I'm saying
goodbye. And yes this is a struggle I want f****** relapse right now but
instead I am going to go to a N/A meeting.

PR #1991, opened 2019-10-16
Set Zipkin status values to OpenCensus.

### LINKED ISSUE #904: Improve Zipkin status tag names
https://github.com/census-instrumentation/opencensus-java/blob/817f27b7f96df5b5c10db16e01899bf2ee7f095b/exporters/trace/zipkin/src/main/java/io/opencensus/exporter/trace/zipkin/ZipkinExporterHandler.java#L63

"error" would be better for STATUS_CODE because Zipkin treats that tag specially if it is set.

We used opencensus.status_description instead of census.status_description for STATUS_DESCRIPTION in Go, it would be good to make that the same too.

esukram:
The previous Zipkin values used to be plain `census` but those changed to be
`opencensus`.

This change was proposed in census-instrumentation/opencensus-go#242

Closes #904


esukram:
@googlebot I signed it!

adriancole:
consistency is good. thanks! :shipit:

PR #2017, opened 2020-03-09
feat: Allow users to register the same Meter multiple times without exception

mayurkale22:


impl_core/src/test/java/io/opencensus/implcore/metrics/MetricRegistryImplTest.java
```diff
@@ -244,10 +244,26 @@ public void getMetrics() {
   }
 
   @Test
-  public void registerDifferentMetricSameName() {
+  public void shouldNotThrowWhenRegisterSameMetric() {
+    metricRegistry.addLongGauge(NAME, DESCRIPTION, UNIT, LABEL_KEYS);
+    metricRegistry.addLongGauge(NAME, DESCRIPTION, UNIT, LABEL_KEYS);
+    metricRegistry.addLongGauge(NAME, DESCRIPTION, UNIT, LABEL_KEYS);
+    metricRegistry.addLongGauge(NAME, DESCRIPTION, UNIT, LABEL_KEYS);
```
rghetia: only need two times.

rghetia:
As discussed, please add test for 
```
m1 = register.AddGauge("g1")
m2 = register.AddGauge("g1")

expect m1 and m2 to be the same.

```

impl_core/src/test/java/io/opencensus/implcore/metrics/MetricRegistryImplTest.java
```diff
@@ -247,8 +247,35 @@ public void getMetrics() {
...
+                LONG_METRIC_DESCRIPTOR,
+                Arrays.asList(
+                    TimeSeries.createWithOnePoint(
+                        Arrays.asList(LABEL_VALUE, LABEL_VALUE_2),
+                        Point.create(Value.longValue(200), TEST_TIME),
+                        null),
+                    TimeSeries.createWithOnePoint(
+                        Arrays.asList(LABEL_VALUE_2, LABEL_VALUE_2),
+                        Point.create(Value.longValue(300), TEST_TIME),
+                        null))));
```
rghetia: This is fine but isn't comparing longGauge and longGauge1 sufficient?
With this test shouldNotThrowWhenRegisterSameMetric is not necessary.

mayurkale22: Removed `shouldNotThrowWhenRegisterSameMetric`  test.

mayurkale22:
@rghetia please review and approve when you get a chance, thank you :)

impl_core/src/main/java/io/opencensus/implcore/metrics/MetricRegistryImpl.java
```diff
@@ -165,17 +157,21 @@ public DerivedDoubleCumulative addDerivedDoubleCumulative(String name, MetricOpt
...
 
-    private synchronized void registerMeter(String meterName, Meter meter) {
+    private synchronized Meter registerMeter(String meterName, Meter meter) {
       Meter existingMeter = registeredMeters.get(meterName);
       if (existingMeter != null) {
-        // TODO(mayurkale): Allow users to register the same Meter multiple times without exception.
-        throw new IllegalArgumentException(
-            "A different metric with the same name already registered.");
+        if (!existingMeter.getMetricDescriptor().equals(meter.getMetricDescriptor())) {
+          throw new IllegalArgumentException(
```
olavloite: (Note, this comment is not relevant for this PR, and is only intended for discussing whether this fix would also fix [this specific issue](https://github.com/googleapis/java-spanner/issues/106) for the Spanner client library).

@mayurkale22 Can you confirm that registering multiple metrics with different label _values_ (but equal label _keys_) will not go down this code path?

mayurkale22: > Can you confirm that registering multiple metrics with different label values (but equal label keys) will not go down this code path?

Yes, it is normal to have different label values - those will be represented as individual TimeSeries in the metric. i.e:
```
Metric:
     {name, label_keys, description, unit, type, ...}
TimeSeries
    [{labels_values1, value1}, {labels_values2, value2}, ...]
```
See: https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/metrics/export/Metric.java#L51

The exception will be raised only when you use the same metric name but with a different type or unit or label_keys.

olavloite: That sounds good :+1: Thanks for looking into this.

I think this fix is better than what I added to the Spanner client library, so I'll close the [PR](https://github.com/googleapis/java-spanner/pull/107).

The only concern that I still have is what it looks like in the dashboards when there are multiple different instances of a `SessionPool` with the same label values. I understand that it won't cause any errors in the client library, but how is it presented to the user? This situation would occur when an application is opening two connections to the same database with different `SessionPoolOptions`. This would also mean different values for constant metrics like `MaxSessions`.

PR #2035, opened 2020-05-20
Rate limiting sampler

avshenuk:
Implement Rate limiting sampler as per https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/Sampling.md#ratelimiting-sampler-implementation-details

### DIFF
--- a/api/src/main/java/io/opencensus/trace/samplers/RateLimitingSampler.java
+++ b/api/src/main/java/io/opencensus/trace/samplers/RateLimitingSampler.java
@@ -1,0 +1,120 @@
+/*
+ * Copyright 2017, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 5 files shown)


api/src/main/java/io/opencensus/trace/samplers/RateLimitingSampler.java
```diff
@@ -0,0 +1,120 @@
...
+          oneValidLinkFound = true;
+          if (parentLink.getContext().getTraceOptions().isSampled()) {
+            return true;
+          }
+        }
+      }
+      if (oneValidLinkFound) {
+        return false;
+      }
+    }
```
avshenuk, line 95: Question is: why don't we propagate the negative decision from the parent in `ProbabilitySampler` the same way? According to `SpanBuilderImpl` it's possible only if the span has a remote parent or the sampler is set on the span level explicitly. Nevertheless, the positive decision is propagated always...

As far as I understand, if you decided to blacklist a certain span, you most likely want all child spans to be blacklisted too. Same as with whitelisting.

avshenuk:
Not working as expected...

Issue #1876, opened 2019-05-03
Log error if Spans are gc-ed without being closed.


bogdandrutu:
With the RunningSpanStore disable by default, log error if Spans are gc-ed without being ended.

We should override the finalize method and log an error.
https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#finalize()

PR #1877, opened 2019-05-04
Log error if Spans are gc-ed without being closed.

songy23:
Fixes https://github.com/census-instrumentation/opencensus-java/issues/1876.

PR #2043, opened 2020-07-16
Remove finalize from RecordEventsSpanImpl

### LINKED ISSUE #2045: Performance problem with defining finalizer
This issue asked for logging of spans that were GC:ed without being closed first.
https://github.com/census-instrumentation/opencensus-java/issues/1876
And this PR implemented support for that by using finalizers:
https://github.com/census-instrumentation/opencensus-java/pull/1877

I made the following simple PR to revert that behaviour, but it does not seem like it will be merged:
https://github.com/census-instrumentation/opencensus-java/pull/2043

The reason for wanting to revert this behavior is that we saw significant performance problems that could be traced back to a large increase of finalization code in GC and also creation of opencensus spans - there is a static synchronization object that is used for that.

Avoiding finalizers is a well established practice - see Effective Java. It's also mentioned in the Checkstyle rule that was disabled: https://checkstyle.sourceforge.io/apidocs/com/puppycrawl/tools/checkstyle/checks/coding/NoFinalizerCheck.html

So instead of going directly for a PR, perhaps it would be better to start a discussion as an issue instead.
Let's try to align on a good solution before making the code changes.

An alternative solution to simply removing the finalization could be to split the class into a base class and two subclasses, where only only of the subclasses overrides finalize and then make the behaviour configurable.
Either you would want to avoid finalizers for performance reasons or you would want to enable finalizers to get logging of misuse of spans. You could even have a mix of it - use sampling to only apply finalizers to a small subset of spans. If there's a systematic problem of not closing spans it would show up anyway.

We could even take it a step further - We could start by not using finalizers but instead keep an atomic counter of currently open spans. If that is continously growing that would be an indicator that we are leaking spans and then we should enable finalizers to assist identifying the problem.

spkrka:
Fixes #2045

This will reduce GC pressure since the VM will not call
Finalizer.register() when creating Span objects.

Since this method is synchronized on a static lock
it can be a source of contention.

It also avoids contention on the same lock when running
runFinalizer() as part of GC

### DIFF
--- a/impl_core/src/main/java/io/opencensus/implcore/trace/RecordEventsSpanImpl.java
+++ b/impl_core/src/main/java/io/opencensus/implcore/trace/RecordEventsSpanImpl.java
@@ -586,15 +586,4 @@ public final class RecordEventsSpanImpl extends Span implements Element<RecordEv
         timestampConverter != null ? timestampConverter : TimestampConverter.now(clock);
     startNanoTime = clock.nowNanos();
   }
-
-  @SuppressWarnings("NoFinalizer")
-  @Override
-  protected void finalize() throws Throwable {
-    synchronized (this) {
-      if (!hasBeenEnded) {
-        logger.log(Level.SEVERE, "Span " + name + " is GC'ed without being ended.");
-      }
-    }
-    super.finalize();
-  }
 }
--- a/buildscripts/checkstyle.xml
+++ b/buildscripts/checkstyle.xml
@@ -173,7 +173,6 @@
             <message key="name.invalidPattern"
              value="Interface type name ''{0}'' must match pattern ''{1}''."/>
         </module>
-        <module name="NoFinalizer"/>
         <module name="GenericWhitespace">
             <message key="ws.followed"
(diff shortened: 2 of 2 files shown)


impl_core/src/main/java/io/opencensus/implcore/trace/RecordEventsSpanImpl.java
```diff
@@ -586,15 +586,4 @@ private RecordEventsSpanImpl(
         timestampConverter != null ? timestampConverter : TimestampConverter.now(clock);
     startNanoTime = clock.nowNanos();
   }
-
-  @SuppressWarnings("NoFinalizer")
```
nilebox, line 590: Please see original PR #1877 which fixed the issue #1876 by adding this code.
As you can see, this code change was deliberate, so removing it would probably require addressing the issue differently.

nilebox: Given the severity of the issue this code is causing, we decided to revisit the decision and proceed with this change. See #2045 for more context.

nilebox (approved):

PR #2046, opened 2020-07-20
Avoid using finalizers when not leaking open spans

spkrka:


spkrka:
See #2045

nilebox:
As described in https://github.com/census-instrumentation/opencensus-java/issues/2045#issuecomment-696507809, I suggest to revisit #2043 and merge it instead.

nilebox:
Since #2043 was merged and fixed the issue #2045, this change is no longer relevant and can be closed I believe.

PR #2059, opened 2020-10-14
Add Support for Opencensus to OpenTelemetry migration

zoercai:
Branched off @nilebox  changes here: https://github.com/nilebox/opencensus-java/pull/1

> Declare the ContextManager interface which allows for overriding the context storage in OpenCensus.
> Default implementation keeps backward compatible behaviour using gRPC context with OpenCensus-specific key.

Added reflection for OpenTelemetry trace component which will be in the migration library.


api/src/main/java/io/opencensus/trace/Tracing.java
```diff
@@ -89,6 +90,19 @@ public static TraceConfig getTraceConfig() {
   // Any provider that may be used for TraceComponent can be added here.
   @DefaultVisibilityForTesting
   static TraceComponent loadTraceComponent(@Nullable ClassLoader classLoader) {
+    try {
+      // Call Class.forName with literal string name of the class to help shading tools.
+      return Provider.createInstance(
+          Class.forName(
+              "migration.OpenTelemetryTraceComponentImpl", /*initialize=*/ true, classLoader),
```
nilebox, line 97: Supposedly, the actual package will be different? Something like `com.google.cloud.opentelemetry.trace.opencensus.OpenTelemetryTraceComponent` (as part of https://github.com/GoogleCloudPlatform/opentelemetry-operations-java)?

api/src/main/java/io/opencensus/trace/Ctx.java
```diff
@@ -0,0 +1,6 @@
+package io.opencensus.trace;
+
+public interface Ctx {
```
nilebox, line 3: I know I came up with this naming (sorry), but it might better to call this interface `Context` or `ContextHandle`?
Then we can also rename `CtxUtils -> ContextHandleUtils` or similar.

Keeping `Ctx` is also fine - whatever you feel like is better :)

api/src/main/java/io/opencensus/trace/Tracing.java
zoercai: yep I'll change this once I put up the migration library PR :)

zoercai: Changed this here as @james-bebbington suggested moving the migration library to https://github.com/open-telemetry/opentelemetry-java

nilebox: Did you change it though?

zoercai: Do you mean the class name here, or the migration library in `opentelemetry-java`? I've changed the class name here to `io.opentelemetry.opencensusshim.OpenTelemetryTraceComponentImpl`, I have the changes to `opentelemetry-java` locally, will make a PR when they're ready, but for now it should be okay to set it here first right?

api/src/main/java/io/opencensus/trace/unsafe/ContextHandleUtils.java
```diff
@@ -0,0 +1,79 @@
...
+          ContextManager.class);
+    } catch (ClassNotFoundException e) {
+      LOGGER.log(
+          Level.FINE,
+          "Couldn't load full implementation for OpenTelemetry context manager, now loading "
+              + "original implementation.",
+          e);
+    }
+    return new ContextManagerImpl();
+  }
```
zoercai, line 52: Loading the migration library context manager by reflection here so applications migrating only have to add the new library dependency instead of adding new code.
More info here: https://github.com/zoercai/opentelemetry-operations-java/pull/3/files/7029c849e2f0e90c5178e044043e4e9aa0c68bda#r505253778

api/src/main/java/io/opencensus/trace/Tracing.java
nilebox: Ah yep, sorry.
I meant the package name, and this file is outdated now, so I've been looking in the wrong place.
Apologies for confusion.

api/src/main/java/io/opencensus/trace/unsafe/ContextHandleUtils.java
nilebox: LGTM.

api/src/main/java/io/opencensus/trace/unsafe/ContextHandleImpl.java
```diff
@@ -0,0 +1,45 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opencensus.trace.unsafe;
+
+import io.grpc.Context;
+import io.opencensus.trace.ContextHandle;
+
+/** {@code Ctx} implementation using {@see io.grpc.Context}. */
```
nilebox, line 22:
```suggestion
/** {@code ContextHandle} implementation using {@see io.grpc.Context}. */
```

api/src/main/java/io/opencensus/trace/unsafe/ContextHandleUtils.java
```diff
@@ -0,0 +1,79 @@
...
+
+  public static ContextHandle currentContext() {
+    return CONTEXT_MANAGER.currentContext();
+  }
+
+  /**
+   * Creates a new {@code Ctx} with the given value set.
```
nilebox, line 59:
```suggestion
   * Creates a new {@code ContextHandle} with the given value set.
```

api/src/main/java/io/opencensus/trace/unsafe/ContextHandleUtils.java
```diff
@@ -0,0 +1,79 @@
...
+   *
+   * @param context the parent {@code Ctx}.
```
nilebox, line 61:
```suggestion
   * @param context the parent {@code ContextHandle}.
```

api/src/main/java/io/opencensus/trace/unsafe/ContextHandleUtils.java
```diff
@@ -0,0 +1,79 @@
...
+   * @param span the value to be set.
+   * @return a new context with the given value set.
+   */
+  public static ContextHandle withValue(
+      ContextHandle context, @javax.annotation.Nullable Span span) {
+    return CONTEXT_MANAGER.withValue(context, span);
+  }
+
+  /**
+   * Returns the value from the specified {@code Ctx}.
```
nilebox, line 71: ContextHandle here and below

nilebox (approved):

PR #2063, opened 2020-10-21
Update release versions for all readme and build files.

zoercai:


README.md
```diff
@@ -32,14 +32,14 @@ For Maven add to your `pom.xml`:
   <dependency>
     <groupId>io.opencensus</groupId>
     <artifactId>opencensus-api</artifactId>
-    <version>0.27.0</version>
+    <version>0.28.1</version>
```
nilebox, line 35: Out of curiosity: why did we skip 0.28.0?

nilebox (approved):

README.md
zoercai: I did release 0.28.0, but only realised afterwards that I released my own fork's master, which didn't have the new changes 😓  So I did another 0.28.1 release with the new changes

nilebox: Makes sense, thanks!

PR #2071, opened 2021-01-07
Fix time skew exceptions crashing disruptor thread

### LINKED ISSUE #2068: Clock Skew(?) can cause the distruptor thread to crash

0.24.0, but looks to be present in all versions

1.8.0

Occasionally, at process start, we'll see this error in our logs:

```
Exception in thread "OpenCensus.Disruptor-0" java.lang.RuntimeException: java.lang.IllegalArgumentException: Current time must be within or after the last bucket.
	at com.lmax.disruptor.FatalExceptionHandler.handleEventException(FatalExceptionHandler.java:45)
	at com.lmax.disruptor.dsl.ExceptionHandlerWrapper.handleEventException(ExceptionHandlerWrapper.java:18)
	at com.lmax.disruptor.BatchEventProcessor.processEvents(BatchEventProcessor.java:187)
	at com.lmax.disruptor.BatchEventProcessor.run(BatchEventProcessor.java:125)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.IllegalArgumentException: Current time must be within or after the last bucket.
	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:141)
	at io.opencensus.implcore.stats.MutableViewData$IntervalMutableViewData.refreshBucketList(MutableViewData.java:308)
	at io.opencensus.implcore.stats.MutableViewData$IntervalMutableViewData.record(MutableViewData.java:262)
	at io.opencensus.implcore.stats.MeasureToViewMap.record(MeasureToViewMap.java:160)
	at io.opencensus.implcore.stats.StatsManager$StatsEvent.process(StatsManager.java:101)
	at io.opencensus.impl.internal.DisruptorEventQueue$DisruptorEventHandler.onEvent(DisruptorEventQueue.java:229)
	at io.opencensus.impl.internal.DisruptorEventQueue$DisruptorEventHandler.onEvent(DisruptorEventQueue.java:222)
	at com.lmax.disruptor.BatchEventProcessor.processEvents(BatchEventProcessor.java:168)
	... 2 more
```

If this happens, all publisher threads eventually will hang forever, since there is no longer a consumer of the LMAX Disruptor queue.  It seems like the cause of this is clock skew causing measurements to occur before the bucket start, and there is even a TODO in the code to handle this:
https://github.com/census-instrumentation/opencensus-java/blob/master/impl_core/src/main/java/io/opencensus/implcore/stats/MutableViewData.java#L307

I'm not sure what the correct behavior should be when the measurement timestamp is before the bucket start though, drop the event possibly?

jsuereth:
For now, we fix all known issues with time skew in Metric Views.

- In the event of a detected time skew, we just throw away all our data.  We really don't know if it's safe to push anything we have, and we cannot safely do any aggregation / report correct timestamps.
- We add (the first?) tests for `MutableViewData` to ensure it does not throw when time rewinds, but instead drops recorded data points and resets itself.

Fixes #2068 (deemed to not have a workaround)

### DIFF
--- a/impl_core/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
+++ b/impl_core/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
@@ -132,6 +132,7 @@ abstract class MutableViewData {
     @javax.annotation.Nullable
     @Override
     Metric toMetric(Timestamp now, State state) {
+      handleTimeRewinds(now);
       if (state == State.DISABLED) {
         return null;
       }
@@ -166,6 +167,7 @@ abstract class MutableViewData {
 
     @Override
     ViewData toViewData(Timestamp now, State state) {
+      handleTimeRewinds(now);
       if (state == State.ENABLED) {
         return ViewData.create(
             super.view,
@@ -177,6 +179,18 @@ abstract class MutableViewData {
             super.view,
             Collections.<List</*@Nullable*/ TagValue>, AggregationData>emptyMap(),
             ViewData.AggregationWindowData.CumulativeData.create(ZERO_TIMESTAMP, ZERO_TIMESTAMP));
+      }
+    }
+
+    /**
+     * This method attemps to migrate this view into a reasonable state in the event of time going
+     * backwards.
+     */
+    private void handleTimeRewinds(Timestamp now) {
+      if (now.compareTo(start) < 0) {
+        // Time went backwards, physics is broken, forget what we know.
+        tagValueAggregationMap.clear();
+        start = now;
       }
     }
 
@@ -311,6 +325,7 @@ abstract class MutableViewData {
         // - Drop events in the future, keep others within a duration.
         // - Drop all events on skew
         // - Guess at time-skew and "fix" events
+        // - Reset our "start" time to now if necessary.
         buckets.clear();
         shiftBucketList(N + 1, now);
         return;
--- a/impl_core/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
+++ b/impl_core/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
@@ -189,7 +189,7 @@ abstract class MutableViewData {
     private void handleTimeRewinds(Timestamp now) {
       if (now.compareTo(start) < 0) {
         // Time went backwards, physics is broken, forget what we know.
-        tagValueAggregationMap.clear();
+        clearStats();
         start = now;
       }
     }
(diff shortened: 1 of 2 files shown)


impl_core/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
```diff
@@ -180,6 +182,18 @@ ViewData toViewData(Timestamp now, State state) {
...
     }
 
+    /**
+     * This method attemps to migrate this view into a reasonable state in the event of time going
+     * backwards.
+     */
+    private void handleTimeRewinds(Timestamp now) {
+      if (now.compareTo(start) < 0) {
+        // Time went backwards, physics is broken, forget what we know.
+        tagValueAggregationMap.clear();
```
aabmass, line 192: I also see `clearStats()` in this class, might be more clear

impl_core/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
```diff
@@ -180,6 +182,18 @@ ViewData toViewData(Timestamp now, State state) {
       }
     }
 
+    /**
+     * This method attemps to migrate this view into a reasonable state in the event of time going
+     * backwards.
+     */
+    private void handleTimeRewinds(Timestamp now) {
+      if (now.compareTo(start) < 0) {
```
aabmass, line 190: Should this compare to the last recorded timestamp from L156 instead of `start`?

aabmass (commented):
LGTM as a fix. I can approve if no one with more context on this reviews in the next fews days.

Is the `IntervalMutableViewData` only used for distributions?

impl_core/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
jsuereth: This class isn't recording timestamps.  See L146 where it just sets each data point its collected to `now`.  While it *looks* like the `MutableAggregation` is receiving timestamps, it's actually dropping them and just storing `Sum` or `Count` in most cases.

impl_core/src/main/java/io/opencensus/implcore/stats/MutableViewData.java
jsuereth: done.

punya (approved):

aabmass (approved):

PR #2072, opened 2021-01-07
Mark ContextUtils as public (but deprecated) 

### LINKED ISSUE #2069: opencensus 0.28.1 breaks gRPC
0.28.1 hides `io.opencensus.trace.unsafe.ContextUtils` which is used directly by gRPC. This was broken by #2059. It seems ContextUtils should be marked deprecated for the moment, instead of hidden entirely. This breakage is especially surprising given the change was in a point release.

Migrating to ContextHandleUtils will need to take some careful thought, as there are multiple ways it would break with the current gRPC integration. Even if we fix this today in gRPC, it'd still be a month before the next release. In addition, a flag-day change like this requires upgrading an entire application all-at-once and so may _prevent_ people from upgrading grpc and getting on the new opencensus APIs; keeping the old class available allows upgrading a piece at a time without breaking the world.

CC @zoercai 

```
grpc-java/census/src/main/java/io/grpc/census/CensusTracingModule.java:42: error: ContextUtils is not public in io.opencensus.trace.unsafe; cannot be accessed from outside package
import io.opencensus.trace.unsafe.ContextUtils;
                                 ^
grpc-java/census/src/main/java/io/grpc/census/CensusTracingModule.java:344: error: ContextUtils is not public in io.opencensus.trace.unsafe; cannot be accessed from outside package
      return ContextUtils.withValue(context, span);
             ^
grpc-java/census/src/main/java/io/grpc/census/CensusTracingModule.java:385: error: ContextUtils is not public in io.opencensus.trace.unsafe; cannot be accessed from outside package
          newClientCallTracer(ContextUtils.getValue(Context.current()), method);
                              ^
```

jsuereth:
Fixes #2069  - bincompat issues with 0.18.0->0.18.1

### DIFF
--- a/api/src/main/java/io/opencensus/trace/unsafe/ContextHandleUtils.java
+++ b/api/src/main/java/io/opencensus/trace/unsafe/ContextHandleUtils.java
@@ -79,7 +79,7 @@ public class ContextHandleUtils {
   }
 
   /**
-   * Attempts to pull the {@see io.grpc.Context} out of an OpenCensus {@code ContextHandle}.
+   * Attempts to pull the {@link io.grpc.Context} out of an OpenCensus {@code ContextHandle}.
    *
    * @return The context, or null if not a GRPC backed context handle.
    */
--- a/api/src/main/java/io/opencensus/trace/unsafe/ContextHandleUtils.java
+++ b/api/src/main/java/io/opencensus/trace/unsafe/ContextHandleUtils.java
@@ -83,6 +83,7 @@ public class ContextHandleUtils {
    *
    * @return The context, or null if not a GRPC backed context handle.
    */
+  @Nullable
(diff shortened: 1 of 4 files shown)


aabmass (approved):

gfelbing:
Thanks @jsuereth. To ease the migration of grpc-java to use the ContextHandle, would it be possible to add the `io.opencensus.trace.unsafe.ContextHandleImpl#getContext` to the `io.opencensus.trace.ContextHandle` interface?

jsuereth:
@gfelbing I'm not going to expose an unsafe API in the old stable `ContextHandle`, but I can give you a helper method from `ContextUtils` which will give you access to it.  Fixing some tests, will post shortly.

PR #2078, opened 2021-07-20
Add github action PR build

jsuereth:
- Migrate our existing travis CI workflow to github actions.
- ONLY runs Java8 build (no Java9/Java10 etc.)   Java 11 build is not working right now
- This is a precursor to getting some of the security fixes in.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,26 +14,23 @@ jobs:
         os:
           - macos-latest
           - ubuntu-18.04
-        java-version:
+        java:
           - 8
           - 11
         include:
           - os: ubuntu-18.04
-            test-java-version: 8
+            java: 8
             coverage: true
     steps:
       - uses: actions/checkout@v2.3.4
         with:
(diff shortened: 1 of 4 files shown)


tbarker25 (approved):

.github/workflows/build.yml
```diff
@@ -0,0 +1,36 @@
...
+        os:
+          - macos-latest
+          - ubuntu-18.04
+        java:
+          - 8
+          # TODO: Java 11 build doesn't work due to Java7 target.
+        include:
+          - os: ubuntu-18.04
+            java: 8
+            coverage: true
```
aabmass, line 23: This variable is unused, but im guessing you'll use it later

aabmass (approved):
LGTM

.github/workflows/build.yml
jsuereth: Right, I had to drop jacoco for now, we can look into adding it back in.   Tests are more more immediate concern than coverage right now.

PR #2082, opened 2021-11-05
Fix flaky LinkTest

Quincunx271:
LinkTest has recently been having flaky failures. It turns out that the `link_ToString()` test relied on hash ordering. The failing assertion only passed when a copy of the `HashMap` had the same `toString()` as the original map. If copying the map changed the iteration order, the test would fail.

This entire test function is questionable, however, as it is testing the `toString()` method on an `AutoValue`. This is problematic, because `AutoValue` specifies its `toString()` as "returning a useful (but unspecified) string representation of the instance" (see https://github.com/google/auto/blob/master/value/userguide/index.md#whats-going-on-here). As such, it could make sense to remove this test altogether. This change does not remove the test, because it's operating under the assumption that the test is wanted.

This change fixes the test by checking that every entry in the map is contained in the `link.toString()`. This means that the order of the entries in the `link.toString()` is no longer important. Note that if `Link` is changed to use a different map implementation not based on `AbstractMap` that does not similarly guarantee its `toString()`, this test may break, although the previous assert would also break in such a situation.

Tested:
 - I tested this change by running LinkTest 100 times to verify that the flaky failure went away.

### DIFF
--- a/api/src/test/java/io/opencensus/trace/LinkTest.java
+++ b/api/src/test/java/io/opencensus/trace/LinkTest.java
@@ -102,11 +102,18 @@ public class LinkTest {
     assertThat(link.toString()).contains(spanContext.getTraceId().toString());
     assertThat(link.toString()).contains(spanContext.getSpanId().toString());
     assertThat(link.toString()).contains("CHILD_LINKED_SPAN");
-    assertThat(link.toString()).contains(attributesMap.toString());
+    for (Map.Entry<String, AttributeValue> entry : attributesMap.entrySet()) {
+      // This depends on HashMap#toString(), via AbstractMap#toString(), having a specified format.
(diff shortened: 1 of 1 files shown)


Quincunx271:
There is no log showing why the build failed. At any rate, I don't see why this would affect the build unless pre Java 5 is supported (before the for each loop).

punya (approved):

Quincunx271:
Any update on this?

songy23 (approved):

songy23:
CI is broken, can you rebase and do a force push?

Quincunx271:
> CI is broken, can you rebase and do a force push?

I'm not sure I understand. It's already at the latest HEAD of the master branch.

songy23:
That is meant to rerun CI builds.

Just found Github allows manual CI rerun now. Restarted to see if it helps.

PR #2095, opened 2022-01-27
Update release versions for all readme and build files

punya:


aabmass (approved):
@punya is this still needed? LGTM

PR #2096, opened 2022-02-01
Use ImmutableMap.Builder.buildOrThrow() instead of deprecated build()

dashpole:


dashpole:
cc @punya @jsuereth

punya (changes requested):
`buildOrThrow` was added in Guava 31, but we currently depend on Guava 29. I'm not sure that fixing this deprecation warning merits a bump in our Guava dependency given that it could be a breaking change for users.

@jsuereth curious to hear your thoughts.

PR #2109, opened 2023-02-22
POC: Allow limit the number of samples in InProcessSampledSpanStoreImpl

vkostyukov:
Some of our production systems were OOME-imng because `InProcessSampledSpanStoreImpl` is unbounded. This PR adds an API to limit the number of span names for which samples will be collected.

### DIFF
--- a/api/src/main/java/io/opencensus/trace/export/SampledSpanStore.java
+++ b/api/src/main/java/io/opencensus/trace/export/SampledSpanStore.java
@@ -74,6 +74,18 @@ public abstract class SampledSpanStore {
   public abstract Summary getSummary();
 
   /**
+   * Sets the maximum number of Spans in th {@code SampledSpanStore}.
+   *
+   * Should be a positive number. The {@code SampledSpanStore} can only grow in size.
+   *
+   * @param maxNumberOfSpans the maximum number of Spans in th {@code SampledSpanStore}.
+   * @throws IllegalArgumentException if {@code maxNumberOfSpans} is negative or zero.
+   */
+  public void setMaxNumberOfSpans(int maxNumberOfSpans) {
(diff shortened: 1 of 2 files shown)


vkostyukov:
@punya @dashpole Do you think there is a chance to get this PR merged? It's a very rough POC but I'm willing to work on it and polish some rough edges if there is an interested from the community in taking this in?

dashpole:
I'm reluctant to add this, as the project has been in maintenance mode for a few years now.

Are you able to migrate to OpenTelemetry?  OpenTelemetry has a stable java API + SDK, and has an OpenCensus bridge to get traces and metrics from your OpenCensus instrumentation.  If this is an issue in OTel, they would be more likely to accept fixes.

vkostyukov:
Thank you for you reply, @dashpole!

I took a brief look at OT <> OC shim and it seems like it's hardcoded [to drop the local span store entirely](https://github.com/open-telemetry/opentelemetry-java/blob/main/opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/OpenTelemetryTraceComponentImpl.java#L55). I'm worried it's going to be quick an uphill battle to get the community to enable and reimplement (!) it in OT.

dashpole:
The shim sends traces recorded using OpenCensus to your configured OpenTelemetry exporters, which is why it drops the OpenCensus ones.  See https://opentelemetry.io/docs/reference/specification/compatibility/opencensus/ for more details on how the migration is expected to work.

> I'm worried it's going to be quick an uphill battle to get the community to enable and reimplement (!) it in OT.

That is where the bridges can help.  You don't need to reimplement everything at once.  Once you've installed the shim and an OpenTelemetry exporter, you can migrate each piece of OpenCensus instrumentation to OpenTelemetry one-at-a-time without breaking your traces.

PR #2111, opened 2023-05-02
Update README with OpenCensus sunset notice

aabmass:


punya (approved):

aabmass:
I believe the worklows are stuck in Queued state because they use `ubuntu-18.04` runner which is "removed" according to https://github.blog/changelog/2022-08-09-github-actions-the-ubuntu-18-04-actions-runner-image-is-being-deprecated-and-will-be-removed-by-12-1-22/

I could upgrade the workflows to a newer version but that seems unnecessary since it's just a README update.

